PR #1, opened 2019-05-29
CR zone replacements

amaisano:
#### Summary of changes
**Asana Ticket:** [Add CR Zones to Liquid Fare Replacement](https://app.asana.com/0/555089885850811/794989287876740)

Adds `zone:x` and `interzone:x` tokens to replacement system. Prices will use `cash` amount unless otherwise specified.

image

#### Checklist
  - [x] `npm run dialyzer`

### DIFF
--- a/apps/site/lib/site/content_rewriters/liquid_objects/fare.ex
+++ b/apps/site/lib/site/content_rewriters/liquid_objects/fare.ex
@@ -49,8 +49,8 @@ defmodule Site.ContentRewriters.LiquidObjects.Fare do
 
   @type request_tuple ::
           {:ok, [repo_arg]}
-          | {summary_mode, [repo_arg]}
-          | {zone_type, String.t(), [repo_arg]}
+          | {:ok, {summary_mode, [repo_arg]}}
+          | {:ok, {zone_type, String.t(), [repo_arg]}}
 
   @default_args [reduced: nil, duration: :single_trip]
   @summary_atoms [:commuter_rail, :bus_subway, :ferry]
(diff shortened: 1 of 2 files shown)


apps/site/lib/site/content_rewriters/liquid_objects/fare.ex
```diff
@@ -185,18 +220,33 @@ defmodule Site.ContentRewriters.LiquidObjects.Fare do
...
     |> get_fares()
     |> Format.summarize(mode)
   end
 
-  defp request_fares({:ok, args}), do: get_fares(args)
-  defp request_fares(error), do: error
+  # If the request is for a zone or interzone fare, merge placeholder keys into tuple value for :name
+  defp request_fares({:ok, {zone_type, zone_id, args}}) when zone_id in @zone_id do
+    args
+    |> Keyword.put(:name, {zone_type, zone_id})
```
katehedgpeth: why does this need to happen here, instead of inside of `compose_args`?

apps/site/lib/site/content_rewriters/liquid_objects/fare.ex
```diff
@@ -236,4 +279,13 @@ defmodule Site.ContentRewriters.LiquidObjects.Fare do
       Keyword.put(args, k, String.to_existing_atom(v))
     end)
   end
+
+  # Fill in any missing/required arguments with the default,
```
katehedgpeth: why did this move?

amaisano: I typically like to have these secondary functions in the order they are called in the pipeline. filter_insert occurs before get_fares.

apps/site/lib/site/content_rewriters/liquid_objects/fare.ex
amaisano: You're right that it should be in `compose_args`. I had it separate but it really didn't need to be.

PR #4, opened 2019-05-29
add interactions to trip planner leaflet map

katehedgpeth:
#### Summary of changes
**Asana Ticket:** [🔍 OpenStreetMaps | Trip Planner - handle interactions](https://app.asana.com/0/555089885850811/1120178496978622)

Enables trip planner Leaflet map to listen for input events and show/hide markers.

#### Checklist
  - [x] `npm run dialyzer`

<br>
Assigned to: @meagonqz


apps/site/assets/ts/trip-planner/trip-planner-loader.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import "whatwg-fetch";
```
meagonqz: I think this isn't needed.

apps/site/assets/ts/trip-planner/components/TripPlannerPage.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const onRemoveCb = (ev: Event): void =>
+      onRemove(ev as RemoveEvent, dispatch);
+    const onUpdateCb = (ev: Event): void =>
+      onUpdate(ev as UpdateEvent, dispatch);
+
+    document.removeEventListener("trip-plan:remove-marker", onRemoveCb);
+    document.addEventListener("trip-plan:remove-marker", onRemoveCb);
+
+    document.removeEventListener("trip-plan:update-marker", onUpdateCb);
+    document.addEventListener("trip-plan:update-marker", onUpdateCb);
```
meagonqz: Are you able to move the `document.removeEventListener`s to the returned cleanup function from useEffect?

as in the last line of useEffect being
```
return () => { 
document.removeEventListener("trip-plan:remove-marker", onRemoveCb); 
document.removeEventListener("trip-plan:update-marker", onUpdateCb);
};
```

apps/site/assets/ts/trip-planner/__tests__/TripPlannerPageTest.tsx
```diff
@@ -0,0 +1,201 @@
...
+  it("registers event listeners", () => {
+    const cbHash: { [event: string]: Function } = {};
+    document.addEventListener = (name: string, cb: EventListener) => {
+      cbHash[name] = cb;
+    };
+    renderer
+      .create(<TripPlannerPage mapData={mapData} />)
+      // update triggers useEffect to run
+      .update(<TripPlannerPage mapData={mapData} />);
+    expect(Object.keys(cbHash)).toEqual([
```
meagonqz: Could this `expect` be after line 65? I'm a little confused by the code below not having an assertion.

apps/site/assets/ts/trip-planner/components/TripPlannerPage.tsx
katehedgpeth: Do you mean remove them from the main function, or have them in both places? my understanding is that when called with only a function, `useEffect` will run on every render. We don't have anything meaningful to pass as a memo argument, so i added the `removeEventListeners` inside it so it doesn't add a new listener every render.

apps/site/assets/js/trip-planner-location-controls.js
```diff
@@ -222,14 +222,18 @@ export class TripPlannerLocControls {
       label,
       title
     };
-    $(document).trigger("trip-plan:update-marker", { detail });
+    const event = new Event("trip-plan:update-marker");
```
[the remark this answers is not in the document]
katehedgpeth: I tested this on crossbrowsertesting and it seems to work in IE.

apps/site/assets/ts/trip-planner/__tests__/TripPlannerPageTest.tsx
katehedgpeth: using the `act` functions got the code coverage to recognize that these were being called (and therefore increase the code coverage), but I wasn't able to figure out how to verify that things changed as expected. I feel pretty confident because of the unit tests. Do you know if there's a way to check the component's state from outside?

apps/site/assets/ts/trip-planner/components/TripPlannerPage.tsx
meagonqz: I'm referring to returning a function at the end of the useEffect function, here's the documentation:
https://reactjs.org/docs/hooks-reference.html#cleaning-up-an-effect

It's a convention to do clean-up by returning a function, and the convention will do a cleanup of the listeners on the unmount.

apps/site/assets/ts/trip-planner/__tests__/TripPlannerPageTest.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+  it("registers event listeners", () => {
+    const cbHash: { [event: string]: Function } = {};
+    document.addEventListener = (name: string, cb: EventListener) => {
+      cbHash[name] = cb;
+    };
+    renderer
+      .create(<TripPlannerPage mapData={mapData} />)
+      // update triggers useEffect to run
+      .update(<TripPlannerPage mapData={mapData} />);
```
meagonqz: This `update()` can be removed, and the expect can be moved to line 66 without effecting coverage since the code below will cause a dispatch and state change.

apps/site/assets/ts/trip-planner/__tests__/TripPlannerPageTest.tsx
meagonqz: We discussed offline that we can't test the preferred way because the trip plan events aren't being triggered by react, so this works for test coverage and the reducer unit tests are good too. 

My comments are mostly cosmetic. Though if I came back to this code later, I'd probably look into the test if I didn't see an `expect` at the the end...since we're OK with this test, I think the cosmetic changes are worthwhile.

apps/site/assets/ts/trip-planner/__tests__/TripPlannerPageTest.tsx
katehedgpeth: I tried removing both the `.update` and the `act` lines, and that key test still passed, so I don't think it's the code below that's triggering it. Which makes sense, since those tests are fetching their callbacks from the hash.

I've updated to add an expect that the callbacks don't throw, so there's at least some testing going on. I confirmed that it does fail if the callback is given a bad argument.

katehedgpeth:
@meagonqz this is ready for another review

PR #9, opened 2019-05-31
OSM on Trip Planner Results

meagonqz:
#### Summary of changes
**Asana Ticket:** [🔍 OpenStreetMaps | Trip Planner - show multiple maps on page](https://app.asana.com/0/555089885850811/1120178496978620)

This PR is large so I'm delineating the changes here:

- Added `<TripPlannerResults />` component that gets its itinerary HTML serverside but handles open/close of the accordion in React so that the map will render normally. This is related to https://github.com/Leaflet/Leaflet/issues/3108 but the suggested solution doesn't work well for us without React.

- Made changes to the leaflet Icon struct so we could set sizes and anchors serverside.

- Made changes to our leaflet `<Map />` component so we could set the bounds by the markers and leave zoom unset (undefined, not null) to prevent unnecessary jumpy zoom behavior.

- Added an extra div with a className in our `<ExpandableBlock />` for styling.

- Added test coverage exception for `leaflet/components` directory in jest config because that was overlooked before and I was getting dinged for a 0.1% coverage miss.

#### Checklist
  - [x] `npm run dialyzer`

<br>
Assigned to: @NAME


apps/site/assets/css/_helpers.scss
```diff
@@ -35,12 +35,14 @@
 
 .callout-small {
   .icon-bus,
+  .icon-subway,
```
meagonqz: For @ingridpierre :)

ryan-mahoney:
I'm having a bit of trouble getting this to work locally.

I am seeing `http://localhost:8090/tripplanresults.js` despite restarting / rebuilding / clearing `priv`.

Also it appears to be breaking the v1 trip planner regardless of the _flag settings.

ryan-mahoney:
actually, I think I was seeing certain errors because webpack was already running on that port fot T-Alerts -- will take another look!

apps/site/assets/css/_trip-plan-results.scss
```diff
@@ -0,0 +1,75 @@
+.m-trip-plan-results__itinerary-container {
+  margin-bottom: $base-spacing * 1.5;
+
+  .c-expandable-block__header {
+    font-size: $base-spacing;
+    line-height: 1.5;
```
ryan-mahoney: does this need some sort of unit or relations to `$base-spacing`?

apps/site/assets/ts/jest.config.js
```diff
@@ -17,11 +17,17 @@ module.exports = {
       lines: 95,
       statements: -25 // No idea what good default is for this, this means "up to 25 uncovered statements allowed"
     },
+    // The Leaflet API is difficult to test, so we consider a lower
```
ryan-mahoney: 👍

apps/site/assets/ts/schedule/components/Map.tsx
```diff
@@ -33,23 +33,23 @@ export const getBounds = (markers: Marker[]): LatLngBounds => {
 export const iconOpts = (
   icon: string | null
 ): {
-  iconSize?: [number, number];
-  iconAnchor?: [number, number];
+  icon_size?: [number, number];
```
ryan-mahoney: what is the reasoning behind this change?

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -1,9 +1,10 @@
 defmodule SiteWeb.TripPlanView do
   use SiteWeb, :view
   require Routes.Route
+  alias Site.React
   alias Site.TripPlan.{Query, ItineraryRow}
   alias Routes.Route
-  alias Phoenix.HTML.Form
+  alias Phoenix.{HTML, HTML.Form}
```
ryan-mahoney: 👍 was not aware of this syntax

apps/site/assets/css/_trip-plan-results.scss
meagonqz: line-height is OK without units, [MDN on unitless line-height](
https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#Prefer_unitless_numbers_for_line-height_values). the grid buttons are relatively new and set a line-height this way. I think this is OK since line-height being unitless has similar reasoning as using $base-spacing.

apps/site/assets/ts/schedule/components/Map.tsx
meagonqz: So we could set sizes and anchors serverside.

The more generic `<Map />` (the one rendered by this map) now expects the icon_size/icon_anchor to possibly come in the serverside (before we didn't need to), so I made these underscores for consistency (and so that our shared leaflet `<Map />` didn't have to know about it when building icons).

meagonqz: I don't love the whole underscore -> camelcase issue in the first place but probably out of scope for this PR.

PR #12, opened 2019-06-03
Schedule Note

meagonqz:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Input box - Subway Schedule Notes](https://app.asana.com/0/555089885850811/1122431599573146)

Added `<ScheduleNote />` for subway schedules.
<br>
Assigned to: @katehedgpeth


apps/site/lib/site_web/templates/schedule/_cms_teasers.html.eex
```diff
@@ -1,3 +1,3 @@
-<%= if @featured_content do SiteWeb.PartialView.teaser(@featured_content, class: "m-schedule-line__teaser") end %>
+<%= if @featured_content do SiteWeb.PartialView.teaser(@featured_content, class: "m-schedule-page__teaser") end %>
```
katehedgpeth: I'm not sure i understand why we're separating these into two templates just to use different classes. is there harm in just updating the class in the existing template?

meagonqz: The schedule note is going at the top of the column in the new page instead of the teasers but the schedule note doesn't exist on the old page. The other template can be deleted once we remove the flag (the selector is specific to the schedule page).

meagonqz: I can move the classnames as an option to the render instead. It's a margin change for both desktop and mobile.

meagonqz:
@katehedgpeth I added variables for the classnames and updated with the icon provided by Ingrid this morning.

meagonqz:
@ingridpierre can you take a look? We might want to change the "Schedule Notes" `h2` to a `h3` because of MD devices.

ingridpierre:
- Little thing: I think all the hours need a space before "AM" and "PM" to match our guidelines. Also, peak should say "7 AM - 9 AM, 4 PM - 6:30", what's in there now doesn't match the SDP. This will also help somewhat with wrapping I hope.

- Also some small restructuring, related to the H3 comment. After talking to Liana, last week we were worried that once we remove the schedule, mobile users will really be confused by where to find schedule info. So we're moving the schedule note and schedule finder up the page. I also add a blue background to make it look more cohesive with the accordions in Desktop. New background color and placement in mobile shown here: https://zpl.io/Vx0LD0R

ingridpierre:
Note: Looks like we can't move the schedule note from this position until we refactor some of the other parts of the template. Will make a note in another ticket to move the schedule note up in mobile in the future, below the main H2 of the page.

PR #13, opened 2019-06-03
add schedule finder to schedule line page

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Input box - Create the box with a destination dropdown](https://app.asana.com/0/477545582537986/1122431599573147/f)

---

Screen Shot 2019-06-05 at 12 11 01 PM

---

Screen Shot 2019-06-05 at 12 11 25 PM

---

<br>
Assigned to: @meagonqz


apps/site/assets/css/_schedule-page.scss
```diff
@@ -34,3 +34,56 @@
 .m-schedule-page__sidebar-list {
   padding-left: $base-spacing * 1.5;
 }
+
+.schedule-finder {
+  background-color: $gray-bordered-background;
+  padding: $base-spacing;
+
+  .c-modal__trigger-wrapper {
+    display: inline; // otherwise clicking next to the button fires the event
```
meagonqz: 👍

apps/site/assets/css/_schedule-page.scss
```diff
@@ -34,3 +34,56 @@
...
+  padding: $base-spacing / 2;
+  position: relative;
+
+  &.error {
+    border: solid 2px $error-text;
+  }
+}
+
+.schedule-finder__select {
+  appearance: none;
```
meagonqz: Does this look OK in IE11? https://caniuse.com/#feat=css-appearance

apps/site/assets/css/_schedule-page.scss
```diff
@@ -34,3 +34,56 @@
...
+  background-color: $white;
+  border: 0;
+  width: 100%;
+
+  @include media-breakpoint-only(md) {
+    width: 90%;
+  }
+}
+
+.schedule-finder__arrow {
```
meagonqz: It'd be great for this to be an SVG. If you think it's out of scope, we can add it as a polish ticket.

@ingridpierre can we add the task of providing an SVG to your queue too? We can then replace it on the accordions too.

apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
```diff
@@ -0,0 +1,204 @@
...
+          className="schedule-finder__select"
+          onKeyUp={e =>
+            handleReactEnterKeyPress(e, () => {
+              submitButtonEl!.current!.click();
+            })
+          }
+        >
+          <option value="">Choose a direction</option>
+          <option value="0">
+            {directionNames[0]}  {directionDestinations[0]}
```
meagonqz: For some reason this causes some weird whitespace characters on SSR. Using a template string prevents it.

```suggestion
            {`${directionNames[0]}  ${directionDestinations[0]}`}
```

apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
```diff
@@ -0,0 +1,204 @@
...
+          </option>
+          <option value="1">
+            {directionNames[1]}  {directionDestinations[1]}
```
meagonqz:
```suggestion
            {`${directionNames[1]}  ${directionDestinations[1]}`}
```

apps/site/assets/css/_schedule-page.scss
ryan-mahoney: good question!  I'm going to throw it on dev-green and check.

apps/site/assets/css/_schedule-page.scss
ryan-mahoney: I will add the SVG -- at the time I did this, I didn't have the SVG

apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
ryan-mahoney: ahh interesting. For the space between the words I am using a special unicode character to achieve a wider space -- I think that is the cause... but I will try this

apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
```diff
@@ -0,0 +1,196 @@
...
+          }
+          onKeyUp={e =>
+            handleReactEnterKeyPress(e, () => {
+              submitButtonEl!.current!.click();
+            })
+          }
+        >
+          <option value="">Choose a direction</option>
+          <option value="0">
+            {directionNames[0]}&nbsp;&nbsp;&nbsp;{directionDestinations[0]}
```
meagonqz: This is harder to read 😬 , why not a template string?

ingridpierre:
image
The overlapping text here looks bad, can we add some separation between the end of the displayed text and the arrow icon? 

Can we have Inbound/Outbound in all caps so it looks like: "INBOUND Ruggles"

Here is an SVG for the small arrow:
https://drive.google.com/file/d/1hQTGZyOM5uP1RA3QUkLkFipr7hh_KHAv/view?usp=sharing

apps/site/assets/css/_schedule-page.scss
ingridpierre: SVG: https://drive.google.com/file/d/1hQTGZyOM5uP1RA3QUkLkFipr7hh_KHAv/view?usp=sharing

apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
ryan-mahoney: I tried it, but in FF I still saw crazy characters :/

ryan-mahoney:
@ingridpierre / @meagonqz feel free to take another look when you can

apps/site/assets/ts/schedule/__tests__/ScheduleFinderTest.tsx
```diff
@@ -0,0 +1,89 @@
...
+        directionDestinations={directionDestinations}
+        directionNames={directionNames}
+        stops={stops}
+      />
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+it("opens modal after displaying error", () => {
```
ryan-mahoney: @meagonqz I did all the test in once place here because there is some issue with the focus trap dependency of the modal that causes it to throw an error if you try to mount the same component (the depends on Modal) in the same test suite more than once :/ I looked into it briefly, but I think fixing that is not in the scope of this PR

ryan-mahoney:
@ingridpierre I fixed the issues you indicated, please take another look when you can. Thanks!

apps/site/assets/ts/schedule/__tests__/ScheduleFinderTest.tsx
meagonqz: Sounds good, can you add a comment on the top of this test indicating that?

ingridpierre:
1.) Small design change - **can we change the background color** to that light blue we're using for accordions, just so the sidebar looks nicer and more unified. Meag is also doing this in her schedule note PR.

-----

2.) In the **MD breakpoint** the right column is getting weird and narrow. I've been meaning for us to address this, but I figure now's my chance. Remove the 1 column offset and make the right sidebar 5 cols instead of four for MD layout. 

**CURRENT col-xs-12 col-md-4 col-md-offset-1**: 
image

**DESIRED col-xs-12 col-md-5**: 
image

-----

3.) **XS, SM schedule finder needs bottom margin** so it doesn't sit on top of the curated content.

-----

4.) How does this interact with the schedule note? We shouldn't show the finder tool on subway pages or anywhere the schedule note is being added (including schedules/Green and schedules/Green-branches).

ingridpierre:
5.) Instead of the error message "Choose a value for..." can we instead say "Please provide an origin", "Please provide a destination", or "Please provide an origin and destination"

ryan-mahoney:
@ingridpierre thanks for feedback -- applied design changes

ingridpierre:
@ryan-mahoney slight miscommunication, we still want the offset column on XL and LG, just not MD.

ingridpierre:
Another thing that's happening is that clicking directly on the arrow for the dropdown doesn't do anything currently. Users will probably be confused (even when they don't have to like on accordions and things, many folks tend to click directly on icons inside buttons/inputs).

ryan-mahoney:
@meagonqz tested in IE11:

Screen Shot 2019-06-07 at 11 04 17 AM

ryan-mahoney:
@ingridpierre made requested changes

apps/site/assets/ts/schedule/__tests__/ScheduleFinderTest.tsx
meagonqz: bump

apps/site/assets/ts/schedule/components/SchedulePage.tsx
```diff
@@ -22,11 +23,22 @@ const SchedulePage = ({
...
+    schedule_note: scheduleNote,
+    direction_names: directionNames,
+    direction_destinations: directionDestinations,
+    stops
   }
 }: Props): ReactElement<HTMLElement> => (
   <>
     {!!scheduleNote && <ScheduleNote scheduleNote={scheduleNote} />}
+    {routeType !== 0 &&
+    routeType !== 1 && ( // don't show for subway
```
meagonqz: scheduleNote is only truthy for subway at the moment, so this could be a scheduleNote OR scheduleFinder instead. That could change so I guess this defensive logic is OK.

PR #26, opened 2019-06-06
add predictions to CR timetable

meagonqz:
#### Summary of changes
**Asana Ticket:** [🐞 Provide better messages in Commuter Rail timetable tooltips](https://app.asana.com/0/477545582537986/1108435465356012)

This wasn't done before for a couple reasons but also because I don't know if there will be performance issues with pulling these predictions on vehicle updates. The existing ticket is a 3 because we might prefer to do prediction streaming from the API. We could see if this suffices for now?

<br>
Assigned to: @katehedgpeth

### DIFF
--- a/apps/site/lib/site_web/channels/vehicle_channel.ex
+++ b/apps/site/lib/site_web/channels/vehicle_channel.ex
@@ -41,7 +41,7 @@ defmodule SiteWeb.VehicleChannel do
     route = Routes.Repo.get(vehicle.route_id)
     stop_name = get_stop_name(vehicle.stop_id)
     trip = Schedules.Repo.trip(vehicle.trip_id)
-    prediction = List.first(Predictions.Repo.all(trip: vehicle.trip_id))
+    prediction = List.first(Predictions.Repo.all(trip: vehicle.trip_id, stop: vehicle.stop_id))
 
     %{
       data: %{vehicle: vehicle, stop_name: stop_name},
--- a/apps/site/lib/site_web/channels/vehicle_channel.ex
+++ b/apps/site/lib/site_web/channels/vehicle_channel.ex
@@ -41,6 +41,7 @@ defmodule SiteWeb.VehicleChannel do
     route = Routes.Repo.get(vehicle.route_id)
     stop_name = get_stop_name(vehicle.stop_id)
     trip = Schedules.Repo.trip(vehicle.trip_id)
+    prediction = List.first(Predictions.Repo.all(trip: vehicle.trip_id))
 
     %{
       data: %{vehicle: vehicle, stop_name: stop_name},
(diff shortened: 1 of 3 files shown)


apps/site/lib/site_web/channels/vehicle_channel.ex
```diff
@@ -41,6 +41,7 @@ defmodule SiteWeb.VehicleChannel do
     route = Routes.Repo.get(vehicle.route_id)
     stop_name = get_stop_name(vehicle.stop_id)
     trip = Schedules.Repo.trip(vehicle.trip_id)
+    prediction = List.first(Predictions.Repo.all(trip: vehicle.trip_id))
```
katehedgpeth: I don't think this is right -- I think you need to at least pass the stop id as well.

meagonqz: This worked for CR w/o the stop id. Since these are live updates, it feels like it should always work for a given trip, but maybe better to be defensive.

I did check this out on Red line pages without the stop and it looks like it works correctly as well, but I discovered a different issue with language that I'm going to bring up to product anyway.

meagonqz:
@ingridpierre This PR currently would add predictions to tooltips on maps with live vehicle updates (Commuter rail timetable and schedule pages). 

The subway language between stops for schedule pages could use some product review. For subway, when a vehicle is in between stops, the prediction says "Expected departure" which is from the next stop. Checking the API, subway predictions only seem to have `departure_time` but not `arrival_time`, and maybe this is OK.

For CR stops, it correctly says "Expected arrival" when in transit and "Expected departure" when it's arrived at the station.

Screen Shot 2019-06-07 at 3 01 22 PM

meagonqz:
I've noticed some intermittent `nil` predictions come in through the Vehicle stream, so we might not be ready to do this.

The user experience from that being that a user can have a tool tip with a prediction open, and a moment later it'll re-render (flashing) without that prediction information, and then another moment later re-render again with a prediction.

ingridpierre:
@meagonqz  I've also observed the predictions flashing in and out, but on all modes not just CR.

I think the original ticket was tto add this info to CR only because the format "xx:xx pm" makes sense there because since it takes longer for CR vehicles to move between stops. 

On the other modes like bus and subway I don't think we want this information at all because it's not in the 'XX min' format used for those predictions, and the stops are so close that the prediction is almost always just like <1 or 2 minutes in the future.

Can we just have these changes affect CR and have the other modes tooltips exclude predictions (just show "On the way to..." for bus  and "On the way.."/"Arrived at.." for subway)?

meagonqz:
@ingridpierre yes, we could have these predictions on tooltips just for commuter rail. 

I'm not sure if this is viable with the predictions flashing in and out though. I don't know how that would be caused by this addition either.

meagonqz:
Closing for now and creating a task for investigation.

PR #28, opened 2019-06-07
OSM Trip Planner - hide alerts behind a link

meagonqz:
#### Summary of changes
**Asana Ticket:** [🔍 OpenStreetMaps | Trip Planner - Hide Alerts behind a link](https://app.asana.com/0/555089885850811/1126250841147222)

<br>
Assigned to: @ingridpierre

### DIFF
--- a/apps/site/assets/js/alert-item.js
+++ b/apps/site/assets/js/alert-item.js
@@ -1,17 +1,29 @@
 const ITEM_SELECTOR = "js-toggle-alert";
 
-export default () =>
+export const onload = () =>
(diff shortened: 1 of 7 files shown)


meagonqz:
@ingridpierre Sadly I noticed another issue, where you can't expand alerts that have bodies. We could remove the arrow and have the body always shown in this case where they have already clicked "view alert". 

I'd use our existing Alert React component if it were possible but it's not, the alerts rendering is too deep/complicated in the templates for the trip plan results body.

ingridpierre:
lgtm!  Glad we got the updating message working, view/hide was way preferable to "toggle alert" 👍

PR #30, opened 2019-06-10
Map on Green Line parent page

meagonqz:
#### Summary of changes
**Asana Ticket:** [Schedules | Map disappears on Green Line parent page](https://app.asana.com/0/555089885850811/1125936468243532)

This fixes the issue with the map disappearing, but the data we are sending doesn't seem valid. We shouldn't be receiving or sending "update" events with an empty array as the data. It seems like it might be problem on the API side but it could use a closer look.

<br>
Assigned to: @NAME

### DIFF
--- a/apps/site/assets/ts/schedule/components/Map.tsx
+++ b/apps/site/assets/ts/schedule/components/Map.tsx
@@ -85,6 +85,9 @@ export const reducer = (state: Marker[], action: Action): Marker[] => {
       );
 
     case "update":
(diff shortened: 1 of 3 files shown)


meagonqz:
I couldn't replicate the issues on the 111.

I could add a guard here to exit if there is an empty update, but as far as I can see, it never should be. With GL there is not now. It might be worth looking into the 111 separately. 

Should we patch for empty updates and then create a ticket for the 111?

PR #31, opened 2019-06-11
standardize alert banner parsing

katehedgpeth:
#### Summary of changes
**Asana Ticket:** [🚒 Website Alert Banner - Use banner text field](https://app.asana.com/0/555089885850811/1126432560732675)

Previously, we had a separate struct for banners, which had different keys from the normal `%Alert{}` struct. When we redesigned the alerts page, we stopped using that banner struct directly -- we still kept it but we were only using it to store the ID of the alert it represented. The alert was parsed as a normal alert, and so it kept the text from the `header` attribute -- but banners are supposed to use the text from the `banner` attribute.

This refactor removes the `Banner` struct entirely, and updates how we select text for the `header` field for alerts. We store only the ID of the banner alert in the repo instead of storing the full alert separately.

<br>
Assigned to: @ryan-mahoney

### DIFF
--- a/apps/site/lib/site_web/templates/layout/_alert_announcement.html.eex
+++ b/apps/site/lib/site_web/templates/layout/_alert_announcement.html.eex
@@ -1,3 +1,3 @@
 <%= link to: (@alert_banner.url || alert_path(@conn, :index)), class: "alert-announcement__header" do %>
-  <%= SiteWeb.AlertView.render "_banner.html", alert: %{Alerts.Repo.by_id(@alert_banner.id) | priority: :system} %>
+  <%= SiteWeb.AlertView.render "_banner.html", alert: %{Alerts.Repo.by_id(@alert_banner.id) | header: @alert_banner.title, priority: :system} %>
 <% end %>


ryan-mahoney:
@katehedgpeth used this branch to investigate potential banner issues. Copied from [Asana ticket](https://app.asana.com/0/477545582537986/1126432560732676/f):

```
- created a banner, saw it appear on site within 1 minute
- changes banner text, saw it update within 1 minute
- removed the banner, saw it disappear from site within 1 minute
- created a new banner, saw it appear within 1 minute
- create a second banner, saw that the first banner was downgraded and the new banner appeared within 1 minute
- deleted all banner alerts, saw that no banners appeared within 1 minute
```

katehedgpeth:
@paulswartz This updated parser uses the text of the banner field for banner alerts if it exists. It's my understanding that the banner field is only populated on banner alerts -- that's how we determine that an alert is a banner alert. Is that incorrect?

paulswartz:
No, that's correct, but the banner text should only be appearing as the website banner. The `header` should be appearing everywhere else the alert appears: /alerts/, the schedule pages, &c.

katehedgpeth:
ok, I misunderstood. that makes this change a LOT simpler.

PR #36, opened 2019-06-12
conditionally trigger modal only for click event

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Input box - Choose an origin stop](https://app.asana.com/0/555089885850811/1122431599573149)

- open modal when select is clicked
- do not open modal when activated via keyboard

<br>
Assigned to: @NAME


apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
```diff
@@ -31,13 +31,22 @@ interface State {
...
 
 const SelectContainer = ({
   children,
-  error
+  error,
+  handleClick
 }: SelectContainerProps): ReactElement<HTMLElement> => (
-  <div className={`schedule-finder__select-container ${error && "error"}`}>
+  <div
+    className={`schedule-finder__select-container ${error && "error"}`}
```
meagonqz: This `error && "error"` returns `false` if there isn't an error, and so there's an extra class "false" which we don't want. Unlikely that we'd ever name a class `false` but worth preventing if a dependency or something did.

ingridpierre:
Stop list should change depending on direction selected.
---

1. Currently list of stops is ignoring my direction (both in the modal and the regular dropdown). In the below example I have selected outbound to harvard, but stop list is showing inbound to dudley.image

2. Relatedly, greyed out stations like Stoughton should be selectable in this example: image

Ferry schedule finder is incomplete
---
Not seeing what I should for Ferry schedules:
image



**Stop List Styling**
----

1. Wrapping stop name text needs some added indentation for consistency with left edge.image

2. Stop names on the right should have right padding that matches the right edge of the close button, see example above "...Shopping Mall"image

3. Would like CR zones to be regular font-weight rather than bold:
image

ingridpierre:
@ryan-mahoney lgtm! 👍

apps/site/assets/ts/components/Modal.tsx
```diff
@@ -61,7 +69,8 @@ const ModalContent = ({
   ariaLabel,
   focusElementId,
   bodyPadding,
-  scrollBarPadding
+  scrollBarPadding,
+  className
```
meagonqz: Probably should default this to an empty string if not provided, you can see the `undefined`s in the snapshot

PR #38, opened 2019-06-12
Tweaks to Modal

meagonqz:
#### Summary of changes
**Asana Ticket:** No ticket, but part of https://github.com/mbta/dotcom/pull/36

I noticed some changes that we could make to the Modal now that we're working with a more concrete example.

If you add the modal open class to the HTML element instead of the body-wrapper div, it prevents the background body from scrolling to the top if you're further down the page when you open the modal.

Using useLayoutEffect prevents the Modal from "jumping in" from the left since the background content styling was changing at the same time.

<br>
Assigned to: @ryan-mahoney

### DIFF
--- a/apps/site/assets/ts/components/Modal.tsx
+++ b/apps/site/assets/ts/components/Modal.tsx
@@ -1,8 +1,8 @@
 import React, {
   ReactElement,
   useState,
+  useLayoutEffect,
   MouseEvent,
-  useEffect,
   createRef
 } from "react";
 import { createPortal } from "react-dom";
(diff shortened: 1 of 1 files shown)


apps/site/assets/ts/components/__tests__/__snapshots__/ModalTest.tsx.snap
```diff
@@ -20,7 +20,9 @@ exports[`it renders 1`] = `
         containerInfo={
           <body>
             <div
+              aria-hidden="true"
```
meagonqz: In this test, this makes sense because it's rendering a snap after being clicked/opened via enzyme.

meagonqz:
I'm going to do testing to see how this works out on mobile and what to do when the modal needs to scroll.

meagonqz:
I remember now that in mobile, we do a full size modal so the scrolling is not an issue. This prevention of scrolling wouldn't work in iOS. There's this package for iOS https://github.com/willmcpo/body-scroll-lock that is the proposed solution for iOS modals but we don't need it yet.

The modal on your branch has overflow which is scrolling fine.

so this is ready @ryan-mahoney 👍

PR #50, opened 2019-06-19
Upcoming Departures in Schedule Modal

meagonqz:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Finder Tool - Upcoming departures](https://app.asana.com/0/555089885850811/1121929590554556)

<br>
Assigned to: @NAME


apps/site/lib/site_web/controllers/mode_controller.ex
```diff
@@ -36,6 +40,36 @@ defmodule SiteWeb.ModeController do
...
   end
 
+  @spec predictions_api(Conn.t(), map) :: Conn.t()
+  def predictions_api(conn, %{
+        "id" => route_id,
+        "origin_stop" => origin_stop,
+        "direction_id" => direction_id
+      }) do
+    schedules =
+      StopController.get_schedules(route_id, origin_stop, direction_id: direction_id)
```
meagonqz: Need to move this out into a common module

meagonqz:
This requires a large refactor because of https://github.com/mbta/dotcom/pull/48 so I removed the labels for review until it is up to date.

meagonqz:
@ryan-mahoney this is ready for a first pass. Also, if you have any opinions on where the "get_schedules" from the stop_controller should go...

apps/site/assets/ts/__v3api.d.ts
```diff
@@ -73,16 +73,19 @@ export interface Prediction {
   track: string | null;
 }
 
-export interface Route {
-  alert_count: number;
+export interface BareRoute {
```
katehedgpeth: I think this should be the other way around -- `Route` should be the base interface and it should get extended and renamed when necessary.

apps/site/lib/predicted_schedule/schedules.ex
```diff
@@ -0,0 +1,43 @@
+defmodule PredictedSchedule.Schedules do
+  @moduledoc """
+  Provides current predicted schedules
+  """
+
+  alias Site.TransitNearMe
+
+  def get_schedules(route_id, stop_id, opts \\ []) do
+    schedules_fn = &Schedules.Repo.by_route_ids/2
+    now = Util.now()
```
katehedgpeth: this should be passed to the function as an option

apps/site/lib/predicted_schedule/schedules.ex
```diff
@@ -0,0 +1,43 @@
...
+
+      [] ->
+        # if there are no schedules left for today, get schedules for tomorrow
+        PredictedSchedule.group(
+          [],
+          schedules_fn.(
+            [route_id],
+            stop_ids: stop_id,
+            direction_id: direction_id,
+            date: TransitNearMe.tomorrow_date(now)
```
katehedgpeth: can this get moved to a helper somewhere, maybe in `Util`?

apps/site/lib/predicted_schedule/schedules.ex
```diff
@@ -0,0 +1,43 @@
+defmodule PredictedSchedule.Schedules do
+  @moduledoc """
+  Provides current predicted schedules
+  """
+
+  alias Site.TransitNearMe
+
+  def get_schedules(route_id, stop_id, opts \\ []) do
```
katehedgpeth: `PredictedSchedules.Schedules.get_schedules` implies that this is somehow only dealing with schedules, and excluding predictions. I would expect a function like this to be `PredictedSchedules.get`.

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleModalContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { BareRoute as Route, RouteType } from "../../../__v3api";
+import { SimpleStop, StopPrediction } from "../__schedule";
+import isSilverLine from "../../../helpers/silver-line";
+
+const stopInfo = (
+  selectedOrigin: string,
+  stops: SimpleStop[]
+): SimpleStop | undefined => stops.find(({ id }) => id === selectedOrigin);
+
+const stopNameLink = (
```
ryan-mahoney: I had converted some of these functions to their own small components in previous PR -- sure you don't want to keep those?

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleModalContent.tsx
```diff
@@ -0,0 +1,140 @@
...
+        className={`h1 schedule-finder__modal-route-pill u-bg--${
+          isSilverLine(id) ? "silver-line" : "bus"
+        }`}
+      >
+        {name}
+      </div>
+    </div>
+  ) : null;
+
+type fetchAction =
```
ryan-mahoney: so glad you took the reducer approach!

ryan-mahoney:
@meagonqz looks pretty good! I support Kate's comments :)

apps/site/assets/ts/__v3api.d.ts
meagonqz: This would touch a lot of code and I wanted to avoid touching unnecessary files. I can open another PR after to fix this.

If I had opened all my fixups as individual PRs it would have been a lot.

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleModalContent.tsx
meagonqz: We don't that in other places and I don't think it's necessary.

apps/site/lib/predicted_schedule/schedules.ex
meagonqz: This was existing in this function I moved, I will fix-up these three comments though.

apps/site/lib/predicted_schedule/schedules.ex
meagonqz: My concern with `Util` is discoverability, does something like Predicted_Schedule.Util work? It does return a list of PredictedSchedules. 

I am OK with `Util` though if you think that's better.

meagonqz: Oh, I missed that this was pointing to line 36.

meagonqz:
@ingridpierre this is ready for product review now

in case you got caught in any of my rebases / force pushes, fwiw you can run this to get to the latest code:
(after `git checkout mh-modal-depart`), `git reset --hard origin/mh-modal-depart`

apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -194,10 +194,12 @@ defmodule SiteWeb.StopController do
 
   @spec schedules_for_route(Route.t(), Stop.id_t()) :: route_with_directions | nil
   defp schedules_for_route(%Route{} = route, stop_id) do
+    now = Util.now()
```
ryan-mahoney: should this also be an opt... or get passed in from a higher level? (for testing purposes)

ryan-mahoney: actually, since this is a controller, is now already set in the `conn` somewhere?

apps/site/lib/site_web/controllers/mode_controller.ex
```diff
@@ -36,6 +39,39 @@ defmodule SiteWeb.ModeController do
     |> render("index.html")
   end
 
+  @spec predictions_api(Conn.t(), map) :: Conn.t()
+  def predictions_api(conn, %{
+        "id" => route_id,
+        "origin_stop" => origin_stop,
+        "direction_id" => direction_id
+      }) do
+    now = Util.now()
```
ryan-mahoney: same as other comment... `now` might already be set in the `conn`

ingridpierre:
image
Destinations should be black, for this version it's not interactive yet and not look like a link.

ingridpierre:
Looks good after some of the changes we discussed in person! 👍 

I will add another polish ticket for the max modal width, to add more predictions, and hopefully figure out the interesting sort issues there.

PR #51, opened 2019-06-19
origin modal filter items

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Input box - Stops filter textbox](https://app.asana.com/0/555089885850811/1122431599573150)

Screen Shot 2019-06-20 at 11 47 00 AM

<br>
Assigned to: @meagonqz


apps/site/assets/ts/components/Modal.tsx
```diff
@@ -48,44 +61,21 @@ const ModalContent = ({
...
-    // Deactivate
-    return () => {
-      if (bodyWrapper) {
-        bodyWrapper.setAttribute("aria-hidden", "false");
-        htmlElement.classList.remove("modal-open");
-        bodyWrapper.style.paddingRight = bodyPadding;
-      }
-      trap.deactivate();
-    };
-  });
```
meagonqz: If you add a dependency array of [] here (as the last arg to the hook) then I think it solves your problem without getting rid of the focus trap and adding extra logic.

You will want to do something like ` focusElementId={state.modalId === "origin" ? "origin-search" : ...}`

ryan-mahoney: incredible. Thank you so much for pointing this out!

ingridpierre:
image

Changes requested:
- Bold for title
- Since the search isn't a search it's a filter it would be nice to change the placeholder to "Filter stops and stations"
- Since selecting required, "Select from the list below" is more appropriate than "Or select.."
- Can we make just the div with the stop list scroll rather than the entire modal? I want to draw more attention to the filter since people tended to miss it before.

ingridpierre:
image
- Intermittently seeing the modal ignore my direction selection. In other words, the stop list sometimes shows the stops for the wrong direction.

ryan-mahoney:
I figured out the sorting issue: the component needs to take into consideration the default direction sorting that is set on the controller level -- fixed.

PR #54, opened 2019-06-20
don't render empty description lists

katehedgpeth:
#### Summary of changes
**Asana Ticket:** [🐞  Hide entire Content List ¶ if list of content is empty](https://app.asana.com/0/555089885850811/1122465604344968)

Prevents header of description lists from being rendered if the list has no contents.

<br>
Assigned to: @ryan-mahoney

### DIFF
--- a/apps/site/lib/site_web/views/content_view.ex
+++ b/apps/site/lib/site_web/views/content_view.ex
@@ -8,7 +8,7 @@ defmodule SiteWeb.ContentView do
 
   alias Content.Field.{File, Image, Link}
   alias Content.Paragraph
-  alias Content.Paragraph.{Callout, ColumnMulti, FareCard}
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/views/content_view.ex
```diff
@@ -8,7 +8,7 @@ defmodule SiteWeb.ContentView do
 
   alias Content.Field.{File, Image, Link}
   alias Content.Paragraph
-  alias Content.Paragraph.{Callout, ColumnMulti, FareCard}
+  alias Content.Paragraph.{Callout, ColumnMulti, DescriptionList, FareCard}
```
amaisano: Wrong Paragraph. This is supposed to be for `ContentList.t()` -- empty content lists are still appearing on the website.

PR #57, opened 2019-06-25
move schedule finder on mobile

katehedgpeth:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Finder Tool - Move the finder on mobile](https://app.asana.com/0/555089885850811/1127978533448182)

Added a second instance of the Schedule Finder to the page for mobile, similar to the treatment for schedule notes on mobile.

<br>
Assigned to: @ryan-mahoney // @ingridpierre

### DIFF
--- a/apps/site/assets/css/_schedule-page.scss
+++ b/apps/site/assets/css/_schedule-page.scss
@@ -110,6 +110,12 @@
       display: none;
     }
   }
+
+  &--accordion {
+    background-color: $brand-primary-lightest-contrast;
+    padding-bottom: 0;
+    padding-top: 0;
+  }
 }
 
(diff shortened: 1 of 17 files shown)


ingridpierre:
image
Not related to this pr, but if it's fast can we change the error text to say "direction" instead of destination? Currently the mismatch is confusing.

I'm not sure we can help it or not, but in desktop the "flash" of the schedule finder disappearing above the main column is pretty jarring. More so than in mobile because I can see the duplicated element too. Is there any way to help with that, or is it just a side effect of the other javascript things we're seeing?

ryan-mahoney:
@ingridpierre created a ticket for the item you mentioned

PR #59, opened 2019-06-25
Improve Predictions.Repo.all performance

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [🔧 Latency | Improve performance of TransitNearMe module](https://app.asana.com/0/555089885850811/1128108664139983)

This is a difficult problem because what we really want to do is retrieve fewer predictions, but there is really no current way to get a maximum of 2 predictions for each headway that passes through a stop.

Still, there were some optimizations that should make things faster.

1. remove a call to `Task.async_stream`.  Rather than speeding things up, this call seems to have resulted in a lot of inter-process communication overhead.

2. parse JSON with the newer `jason` elixir library which is faster and uses less memory.

---

The following flame graphs were generated with this command: `:eflame.apply(:normal, "stacks.out", Predictions.Repo, :all, [[stop: "place-welln"]])`

### Before

Screen Shot 2019-06-25 at 2 37 45 PM

### Remove Async

Screen Shot 2019-06-25 at 2 37 57 PM

### Add jason

Screen Shot 2019-06-25 at 2 38 07 PM

<br>
Assigned to: @meagonqz

### DIFF
--- a/apps/json_api/lib/json_api.ex
+++ b/apps/json_api/lib/json_api.ex
@@ -42,7 +42,7 @@ defmodule JsonApi do
 
   @spec parse(String.t()) :: JsonApi.t() | {:error, any}
   def parse(body) do
-    with {:ok, parsed} <- Poison.Parser.parse(body),
+    with {:ok, parsed} <- Jason.decode(body),
          {:ok, data} <- parse_data(parsed) do
       %JsonApi{
         links: parse_links(parsed),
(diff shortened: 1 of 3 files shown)


paulswartz:
There's an `IO.inspect` call in the flame chart: is that in the code somewhere or leftover from local debugging?

ryan-mahoney:
@paulswartz yes, that was in there at the time I was running the eflame, but subsequently removed

ryan-mahoney:
in the end, the major change here made the graph look better... but the performance worse :/

PR #68, opened 2019-06-28
remove call to services.repo

ryan-mahoney:
#### Summary of changes
NO TICKET

This call was resulting in some crashes on dev/dev-green so I just want to remove this so it doesn't sneak into a deploy next week.  Also, we are not using this yet, so it is harmless to remove it for now.

@meagonqz

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line_controller.ex
@@ -61,10 +61,6 @@ defmodule SiteWeb.ScheduleController.LineController do
         fare_link: ScheduleView.route_fare_link(conn.assigns.route),
         holidays: conn.assigns.holidays,
         route: Route.to_json_safe(conn.assigns.route),
-        services:
-          conn.assigns.route.id
-          |> Services.Repo.by_route_id()
-          |> Enum.sort_by(&Date.to_string(&1.start_date)),
         schedule_note: ScheduleNote.new(conn.assigns.route),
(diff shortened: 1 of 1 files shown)


apps/fares/lib/fare_info.ex
```diff
@@ -5,7 +5,7 @@ defmodule Fares.FareInfo do
 
   alias Fares.Fare
 
-  @july_1_2019 1_561_953_600
+  @july_1_2019 1_561_939_199
```
meagonqz: 1561953600 converts to Monday July 01, 2019 00:00:00 (am) in time zone America/New York (EDT)
https://www.epochconverter.com/timezones?q=1561953600&tz=America%2FNew_York

🤔

paulswartz: It probably doesn't matter too much, but the fares will take effect ~3am.

meagonqz: ^👍 

I think it'd make more sense to the average user to have the fares on the site switch at 12am. If someone ends up paying less at 1am, I can't imagine they'd complain. I do think we could educate riders about the 3am service day but not within this work.

The changed time stamp 1561939199 converts to Sunday June 30, 2019 19:59:59 (pm) in time zone America/New York (EDT) so that's early.

ryan-mahoney: I'm going to leave this alone for now. Will change at midnight EDT per Dave Barker's request.

PR #77, opened 2019-07-02
feat: don't require WIREMOCK_PROXY_URL to be set

paulswartz:
The current code requires WIREMOCK_PROXY_URL to be set to a particular value
for the tests to pass. This pulls the configuration into the test so that the
environment isn't required to be in a particular state.

Also removes `async: true` since it modifies the global state.

<br>
Assigned to: @meagonqz


ryan-mahoney:
@meagonqz @paulswartz what is happening w/ this PR? Looks like it has been lingering :)

paulswartz:
It was failing Backstop, and I can't download the images to see what the issue was. I also couldn't get Backstop passing locally.

ryan-mahoney:
@meagonqz I'm going to try to resolve the issues here and ping you when it passes CI

meagonqz:
I never noticed this because I only look for my avatar in the PR view so I thought it was assigned to Ryan. Seeing if this will pass CI after the update.

PR #83, opened 2019-07-05
schedule, get route patterns

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Directionality - Get trip pattern data](https://app.asana.com/0/555089885850811/1127979261717805)

- get all route patterns for a route, grouped by direction
- get all shapes for route, organized in a map by shape.id
- allow shape repo call to not require a `direction_id`
- remove some file that appeared to have been checked-in accidentally

<br>
Assigned to: @phildarnowsky


apps/routes/lib/repo.ex
```diff
@@ -47,11 +47,13 @@ defmodule Routes.Repo do
...
-  @spec get_shapes(String.t(), 0 | 1, boolean) :: [Shape.t()]
-  def get_shapes(route_id, direction_id, filter_negative_priority? \\ true) do
+  @spec get_shapes(String.t(), Keyword.t(), boolean) :: [Shape.t()]
+  def get_shapes(route_id, opts, filter_negative_priority? \\ true) do
+    opts = Keyword.put(opts, :route, route_id)
+
     shapes =
-      cache({route_id, direction_id}, fn _ ->
-        case Shapes.all(route: route_id, direction_id: direction_id) do
+      cache(opts, fn _ ->
```
gshkolnik: are the same `opts` always going to be in the same order? If not, we might be hurting our cache hit rate.

phildarnowsky: Easy to fix by changing `opts` to `Enum.sort(opts)`

apps/route_patterns/lib/repo.ex
```diff
@@ -10,6 +10,7 @@ defmodule RoutePatterns.Repo do
       nil -> [route: route_id]
       direction_id -> [route: route_id, direction_id: direction_id]
     end
+    |> Keyword.put(:sort, "sort_order")
```
phildarnowsky: Is this what you were referring to earlier when you said the default sort order seemed to be by typicality?

apps/routes/lib/repo.ex
```diff
@@ -47,11 +47,13 @@ defmodule Routes.Repo do
     end
   end
 
-  @spec get_shapes(String.t(), 0 | 1, boolean) :: [Shape.t()]
-  def get_shapes(route_id, direction_id, filter_negative_priority? \\ true) do
+  @spec get_shapes(String.t(), Keyword.t(), boolean) :: [Shape.t()]
+  def get_shapes(route_id, opts, filter_negative_priority? \\ true) do
```
phildarnowsky: so the idea is that, now that `direction_id` is optional, we can call `get_shapes` without it to get all shapes in both directions for the route in question when we fetch all the data for this new feature?

apps/site/lib/site_web/controllers/schedule/line.ex
```diff
@@ -76,25 +89,63 @@ defmodule SiteWeb.ScheduleController.Line do
...
+
+    Map.put(route_pattern, :shape_id, shape_id)
+  end
+
+  @spec get_route_shape_map(Route.id_t()) :: map
+  def get_route_shape_map(route_id) do
+    route_id
+    |> get_route_shapes()
+    |> Enum.map(fn shape -> {shape.id, shape} end)
+    |> Map.new()
```
gshkolnik: no need to do `Enum.map`, `Map.new` does that already.

apps/route_patterns/lib/repo.ex
ryan-mahoney: yes

apps/routes/lib/repo.ex
ryan-mahoney: yes

PR #85, opened 2019-07-06
Use Repo.teasers for event index

amaisano:
#### Summary of changes
**Asana Ticket:** [Use teaser request for Events](https://app.asana.com/0/555089885850811/958950737688914)

- dependent on #84 
- make event times naive once parsed to avoid double TZ shift
- we only have the start time at the moment, not the whole range

### DIFF
--- a/apps/site/lib/site_web/controllers/project_controller.ex
+++ b/apps/site/lib/site_web/controllers/project_controller.ex
@@ -181,28 +181,15 @@ defmodule SiteWeb.ProjectController do
   end
 
   @spec get_events_async(integer, :past | :upcoming) :: (() -> [Teaser.t()])
-  def get_events_async(id, :past) do
+  def get_events_async(id, timeframe) do
     fn ->
       Repo.teasers(
         type: :event,
         related_to: id,
         items_per_page: 10,
-        date_op: "<",
+        date_op: (timeframe == :past && "<") || ">=",
         date: [value: "now"],
-        sort_order: "DESC"
-      )
-    end
-  end
-
-  def get_events_async(id, :upcoming) do
-    fn ->
-      Repo.teasers(
-        type: :event,
-        related_to: id,
-        items_per_page: 10,
-        date_op: ">=",
-        date: [value: "now"],
-        sort_order: "ASC"
+        sort_order: (timeframe == :past && "DESC") || "ASC"
       )
     end
   end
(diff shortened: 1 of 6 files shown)


apps/site/lib/site_web/controllers/project_controller.ex
```diff
@@ -182,8 +180,32 @@ defmodule SiteWeb.ProjectController do
...
+      Repo.teasers(
+        type: :event,
+        related_to: id,
+        items_per_page: 10,
+        date_op: ">=",
+        date: [value: "now"],
+        sort_order: "ASC"
+      )
+    end
+  end
```
ryan-mahoney: a bit repetitive, how do you feel about:
```
def get_events_async(id, timeframe) do
    fn ->
      Repo.teasers(
        type: :event,
        related_to: id,
        items_per_page: 10,
        date_op: if timeframe == :past, do: "<", else: ">=",
        date: [value: "now"],
        sort_order: if timeframe == :past, do: "DESC" else: "ASC"
      )
    end
  end
```

apps/site/lib/site_web/templates/event/_event_list.html.eex
```diff
@@ -3,14 +3,23 @@
...
+        range = case event do
+          %Content.Teaser{} = item -> %{start: item.date, stop: item.date_end}
+          %Content.Event{} = item -> %{start: item.start_time, stop: item.end_time}
+        end
+
+        path_fn = case event do
+          %Content.Teaser{path: path} -> fn -> cms_static_page_path(@conn, path) end
+          %Content.Event{} -> fn -> event_path(@conn, :show, event) end
+        end
+      %>
```
ryan-mahoney: I didn't read too deeply into this, but I though that the list would always be `Teaser.t()` and not `Event.t()`?

amaisano: We're keeping support for when this template i used by the very old (and not often used) Upcoming Board Meetings paragraph, which serves us `Event.t()` structs. See [this commit](https://github.com/mbta/dotcom/pull/85/commits/64c7215f9ff8225c760ed0bbb5ba80c9d4213884) for a little more detail.

amaisano: FYI: https://app.asana.com/0/385363666817452/1130867967246842

amaisano:
@ryan-mahoney now that the dependent branch is merged into master I changed the base of this. I think you already approved it via comment but now I need an official approval (pending tests). Thanks.

PR #90, opened 2019-07-09
Keep redirect [sub] params [if whitelisted]

amaisano:
#### Summary of changes
**Asana Ticket:** [AFC 2.0 Charlie | TNM vanity and UTM params](https://app.asana.com/0/555089885850811/1122614077520942)

When the CMS API calls `view/2` on a Drupal path that is actually a redirect, we lost some of the params (both top-level and nested) due to the redirect using a Map and Strings vs. a List and Atoms to define the params.

This PR adds clauses and guards for redirects where the params are in this alternate format, and continues to sanitize nested and non-nested params.

I've also confirmed the CMS merges normal params from the request with those of the redirect location (it just had trouble with nested param format).

**Test Link** (note: connect to the live CMS when testing):

http://localhost:4001/transit-near-me-charlie?location[lattitude]=12345&location[longitude]=12345&badkey[badsubkey]=badbadbad

**Should redirect to:**

http://localhost:4001/transit-near-me?location%5Blattitude%5D=12345&location%5Blongitude%5D=12345&utm_campaign=charlie&utm_medium=transit-near-me&utm_source=cubic

### DIFF
--- a/apps/content/lib/cms/http_client.ex
+++ b/apps/content/lib/cms/http_client.ex
@@ -26,6 +26,8 @@ defmodule Content.CMS.HTTPClient do
 
   @impl true
   def view(path, params) do
+    # IO.inspect({path, params}, label: "Incoming view request")
+
     params = [
       {"_format", "json"}
       | Enum.reduce(params, [], &stringify_params/2)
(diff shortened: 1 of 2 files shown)


apps/content/lib/cms/http_client.ex
```diff
@@ -73,13 +75,21 @@ defmodule Content.CMS.HTTPClient do
...
 
   # Convert nested key values to their own keys, if whitelisted
-  @spec list_to_params(String.t(), param_list, {safe_key(), String.t()}) :: param_list
+  @spec list_to_params(String.t(), param_list, {safe_key(), String.t()} | map) :: param_list
   defp list_to_params(key, acc, {sub_key, sub_val}) when sub_key in @safe_keys do
     stringify_params({key <> "[#{sub_key}]", sub_val}, acc)
   end
 
+  # Nested params from 302 redirects come in as a single-pair Map.t() element
+  defp list_to_params(key, acc, val) when is_map(val) do
```
ryan-mahoney: when the test runs, this function is never called

amaisano: See latest commit. Turns out we no longer need this clause :)

PR #92, opened 2019-07-10
make :route_patterns a dependency of :site app

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [Look into schedule 500 error on dev](https://app.asana.com/0/555089885850811/1130522906709257)

The application build is crashing because `:route_patterns` was never added as a dependency to the `:site` application after the modules were referenced.

A similar issue happened recently with the `:services` application.

<br>
Assigned to: @meagonqz


ryan-mahoney:
verified on dev-green

PR #93, opened 2019-07-10
Log messages to calculate predictions cache miss ratio

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [🔧 Performance: Log cache/hit ratio for predictions repo](https://app.asana.com/0/555089885850811/1129115952715485)

Added two log messages. One gets logged every time `Predictions.Repo.all` gets called, the other gets logged only when that function has a cache miss. From this, we can calculate our cache miss (or hit) ratio via Splunk.

<br>
Assigned to: @ryan-mahoney

### DIFF
--- a/apps/predictions/lib/repo.ex
+++ b/apps/predictions/lib/repo.ex
@@ -13,6 +13,8 @@ defmodule Predictions.Repo do
   ]
 
   def all(opts) when is_list(opts) and opts != [] do
+    Logger.info("predictions_repo_all")
+
     @default_params
     |> add_optional_param(opts, :route)
     |> add_optional_param(opts, :stop)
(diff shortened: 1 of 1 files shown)


apps/predictions/lib/repo.ex
```diff
@@ -13,6 +13,8 @@ defmodule Predictions.Repo do
   ]
 
   def all(opts) when is_list(opts) and opts != [] do
+    Logger.info("predictions_repo_all")
```
ryan-mahoney: I think for the purpose of simplifying the Splunk work, it still makes sense to use something like `prediction_repo_all_cache=call` `prediction_repo_all_cache=cache_miss`... so we are still querying for one `field` and then doing some calculation on it's value.  In my limited experience it can be tricky to compare the results of two distinctly different splunk queries

phildarnowsky:
@ryan-mahoney updated.

PR #94, opened 2019-07-10
Pass 1 on performance improvements to ServiceSchedules

meagonqz:
#### Summary of changes
**Asana Ticket:** [🔧 Performance of scheduled departures (MOB)](https://app.asana.com/0/555089885850811/1130531308814807)

<br>
Assigned to: @phildarnowsky


ryan-mahoney:
there is still some timing / logging in here... if we are going to merge, we should probably take that out

ryan-mahoney:
also, before merging we should get this on dev green and maybe send the web-crawler to run against the site and take a look at the memory utilization.  I would hate to merge this in only to find that the site is crashing because we have exhausted the available memory

PR #96, opened 2019-07-15
If ¶s are present, use dynamic sidebar content

amaisano:
#### Summary of changes
**Asana Ticket:** [Remove hard-coded project sidebar](https://app.asana.com/0/555089885850811/1117410708814904)

When paragraphs are present, don't use the hard-coded sidebar; use the dynamic sidebar that generic pages use (includes ordered paragraphs of content lists).
- added comments
- simplified if/then logic/conditions

### DIFF
--- a/apps/site/lib/site_web/templates/project/show.html.eex
+++ b/apps/site/lib/site_web/templates/project/show.html.eex
@@ -1,12 +1,5 @@
-<%
-  {sidebar_class, sidebar_right} = case {@updates, @diversions} do
-    {[], []} -> {"c-cms--no-sidebar", false}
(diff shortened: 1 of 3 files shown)


apps/site/lib/site_web/templates/project/show.html.eex
```diff
@@ -1,21 +1,27 @@
+<%
```
phildarnowsky: @amaisano as long as we're moving this block around anyway, could we move it into the controller rather than have the logic in the template?

phildarnowsky:
@amaisano looks good but this time I'll wait for Semaphore to finish

amaisano:
@phildarnowsky I fixed that rebase issue. Should be good for a final look.

PR #99, opened 2019-07-16
Schedules | Finder Tool - Trip Details

meagonqz:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Finder Tool - Trip Details](https://app.asana.com/0/555089885850811/1121929590554561)

Add trip details. Includes adding fares for each trip leg and adding a total time duration for each trip.

<br>
Assigned to: @ryan-mahoney


ingridpierre:
Incredible work! Feels great to see it come together like this. A couple comments below:

----

1. Getting a 404 when I click "view fares". Example link: http://localhost:4001/fares/bus_subway?destination=110&origin=place-dudly
----
2. Let's remove the horizontal line above the table/header that says 'Stops' and 'Arrival'
----

3. Can we add the down/up disclosure arrows to the end of the schedule rows? They look really similar to the live predictions rows above (which aren't clickable or expandable).
----
 
4. There's a tiny little jitter "dance" that the letters inside the pills do when you click the row open and closed. This is a very minor nitpick, sorry.
----

5. Question: My original intent was that people should only be able to click open one schedule details list at a time. But now that I'm playing around in it, would making this change cause scrolling issues if you click open a row and something collapses above it?
----

6.  Can we make the ferry table rows look more like Commuter Rail? Right now entire ferry line name appear on every row in addition to the headsign. Should just have the ferry icon instead. Compare:
image
image
----

7.  Noticed some odd wrapping, when the destination/headsign gets too long for one line.
image

apps/site/lib/site_web/controllers/schedule/line_controller.ex
```diff
@@ -43,28 +44,75 @@ defmodule SiteWeb.ScheduleController.LineController do
...
+  def duration_for_service(schedules) do
+    first = List.first(schedules).time
+    last = List.last(schedules).time
+    %{schedules: schedules, duration: Timex.diff(last, first, :minutes)}
+  end
+
+  def formatted_time(%{schedules: schedules, duration: duration}) do
+    time_formatted_schedules =
+      schedules
+      |> Enum.map(&Map.update!(&1, :time, fn time -> Timex.format!(time, "{0h12}:{m} {AM}") end))
```
ryan-mahoney: may make sense to deal w/ date formatting on client since we generally know `Timex` is slow

meagonqz: I'm not sure we should make it a hard fast rule to do date formatting on the client.

We already use this function for trip duration on schedule pages (see TripInfo module) that have trips listed.

Doing the trip duration clientside correctly would probably cause me to bring in a new library. So bundle size is a concern in this case. In other cases, we'd have to handle timezone correctly (usually EST but parsing dates clientside needs to be correct on that) which is also often easier with a library.

I think because this is just comparing the first -> last date from every trip, it's okay. I will add timing to get a better idea, and see if DateTime can do it instead.

ryan-mahoney: Phil and I looked at this last time we met to review performance... and it is surprisingly slow :/

meagonqz: This code to get the trip duration was not part of the React work previously, so I didn't think you would have checked the old EEX templates for this.

meagonqz: Ah I see I misplaced where you placed this comment. I did introduce a Timex.diff call somewhere else.

This code is exactly as it was, just moved around, not new, so I won't change it in this PR as this PR doesn't address the performance stuff.

meagonqz:
@ingridpierre I've addressed your comments beside the one about only one Trip being expanded at a time. We should move that into a separate task as I didn't realize it from the mockup that was expected. It shouldn't be a terribly hard ticket, but this ticket was large and I now have some extensive rebasing to do.

meagonqz:
@ingridpierre This is ready for review and rebased off master  / the spinner PR. It looks like the modal is pretty jumpy with the spinner when you switch from services with many trips -> less trips, but that existed before my change, we should add that as followup polish.

ingridpierre:
I think this looks beautiful, thanks for your attention to detail here. The spinner is great, super useful to have it around - I didn't catch the jumpiness myself.

I caught a few extant data-related bugs (unrelated to this ticket) that I'm adding to Asana.

I would say this is ready to go, but I managed to get snagged on an error that I can't evaluate on my own. Will send the console output over slack.

ingridpierre:
Forgot to mention steps to reproduce the bug:

1. Go to Newbury/Rockport schedule page
2. Open up any schedule from the schedule finder.
3. Choose a holiday schedule from the picker
4. Modal will crash and sidebar disappears

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const ScheduleTable = ({
+  schedule
+}: {
+  schedule: ServiceScheduleInfo;
+}): ReactElement<HTMLElement> => {
+  if (schedule.trip_order.length === 0) {
+    return (
+      <div className="callout schedule-table--empty">
+        There are no scheduled departures for this service
```
meagonqz: @ingridpierre LMK if you want to change this language for when there is no service (but a service category exists). I took it from the current Schedule tab.

ingridpierre: @meagonqz  How about: "There is no scheduled service for this time period." 

Alternatively, I'd love to say "There is no scheduled service on [date]" or "There is no scheduled service between [date range]". Just to help connect the dots for folks if they happen to visit the page on a holiday and it defaults to a "no service" message.

ingridpierre:
Looked great when I looked at it in the office, glad we caught that bug. Little language tweak then 👍.

PR #100, opened 2019-07-17
Pwd schedule finder fetches schedules async

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Finder Tool - fetch schedules for service as-needed](https://app.asana.com/0/555089885850811/1131031450608147)

Changed the new schedule finder from loading all data up front to loading dynamically, on request.

<br>
Assigned to: @ryan-mahoney


apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
```diff
@@ -129,8 +188,9 @@ const ServiceSelector = ({
             className="schedule-finder__select"
             defaultValue={defaultServiceId}
             onChange={(): void => {
+              /* istanbul ignore next */
               if (ref && ref.current) {
```
ryan-mahoney: you might be able to do `ref.current!.value` (or something similar) if `current` is guaranteed to be set

apps/site/lib/site_web/controllers/schedule/schedule_api.ex
```diff
@@ -0,0 +1,27 @@
+defmodule SiteWeb.ScheduleController.ScheduleApi do
+  use SiteWeb, :controller
+
+  def show(conn, %{"id" => id, "date" => date, "direction_id" => direction_id}) do
```
ryan-mahoney: minor comment: might be slightly more clear to resolve `"id" => route_id`

ryan-mahoney:
not sure if we have a spinner design yet, but there are some decent pure CSS ones here: https://projects.lukehaas.me/css-loaders/

apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
phildarnowsky: I'll give it a try, not familiar with the `!` operator in this context but I'm assuming from your comment that it's similar to Ruby's `try`?

ryan-mahoney: I think this is like a hint to the typescript interpreter in a case where, yes, technically it is possible for a value to be null in general, but in this specific logic branch it is guaranteed to have a value

phildarnowsky: [googles a little] Interesting...so it's just to force certain behavior out of the type checker.

apps/site/lib/site_web/controllers/schedule/schedule_api.ex
```diff
@@ -0,0 +1,27 @@
...
+    {:ok, date} = Date.from_iso8601(date)
+    schedule_data = get_schedules(id, date, direction_id)
+
+    json(conn, schedule_data)
+  end
+
+  @spec get_schedules(binary, any, any) :: %{by_trip: map, trip_order: [any]}
+  def get_schedules(route_id, date, direction_id) do
+    services =
```
ryan-mahoney: are these `services` or `schedules`?

apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
phildarnowsky: @ryan-mahoney I think it might be possible for `current` not to be set, so I'll leave it.

apps/site/lib/site_web/controllers/schedule/schedule_api.ex
phildarnowsky: Well we're getting them out of a `Schedules.Repo` so it would be weird if they weren't `schedules`. I cut-and-paste this without reading it closely, will correct.

meagonqz: They are individual schedules (to comprise a full trip of schedules) organized by service.  A schedule is a "route A stops at this stop at this time", a trip is a collection of schedules (from origin -> last stop), and service is a collection of trips.

just a FYI clarification since the structure of the data from the API makes these kind of interchangeable terms, even in the frontend code.

phildarnowsky:
@ryan-mahoney  Here's the spinner that I went with as a placeholder.
spinner_demo

PR #101, opened 2019-07-18
Intelligent / overridable Content List CTAs

amaisano:
#### Summary of changes
**Asana Ticket:** [Automatically add "View More" link to content lists](https://app.asana.com/0/555089885850811/1117410708814919)

**NOTE:** Depends on mbta/cms#283

Since we are no longer hard-coding sidebar lists like Project Updates and they now come direct from the CMS, we can parse new CTA-related fields from this paragraph type to generate a relevant "View all..." link below the list of teasers.

Authors are also able to override this behavior by either hiding the CTA all together, or overriding either or both the CTA text or CTA url.

This makes content lists much more powerful for teasing a subset of content that lives elsewhere on the website (or even on an external website).

### DIFF
--- a/apps/site/lib/site_web/templates/content/_content_list.html.eex
+++ b/apps/site/lib/site_web/templates/content/_content_list.html.eex
@@ -7,7 +7,7 @@
       type: Keyword.get(@content.recipe, :type, :generic),
       conn: @conn) %>
 
-<%= if list_cta?({@content.ingredients.type, @content.cta}, @content.teasers, @conn.path_info) do %>
+<%= if list_cta?(@content.ingredients.type, @content.cta, @content.teasers, @conn.path_info) do %>
   <% cta = setup_list_cta(@content, @conn.path_info) %>
 
   <p>
--- a/apps/site/lib/site_web/templates/content/_content_list.html.eex
+++ b/apps/site/lib/site_web/templates/content/_content_list.html.eex
@@ -3,7 +3,14 @@
 <% end %>
 
 <%= render("_teaser_list.html",
-      list: @content,
       teasers: @content.teasers,
       type: Keyword.get(@content.recipe, :type, :generic),
       conn: @conn) %>
+
+<%= if list_cta?({@content.ingredients.type, @content.cta}, @content.teasers, @conn.path_info) do %>
+  <% cta = setup_list_cta(@content, @conn.path_info) %>
+
+  <p>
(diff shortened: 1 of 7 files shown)


ingridpierre:
Currently, I'm not seeing a link for "View project updates". Is that intentional? I remember we had discussed that project update lists should have links to their relevant index pages.

image

ingridpierre:
Also, like you did for the other content types, can you add a sidebar for a project updates content list to the demo page as well?

ingridpierre:
Can you add another project updates list but promote to the sidebar?

I tried to add one in the sandbox CMS with Aayesha, but it kept throwing an error we weren't able to resolve.

amaisano:
@ingridpierre 

1. When a list of project updates is placed out of context (not placed on a project page), we cannot automatically know what URL to use for the CTA, hence the CTA is excluded. In this situation, the author can provide the CTA link/text manually for that one instance of the content list of updates.

2. I think I provided the wrong URL in the CMS ticket description. Please use https://pr-283-mbta.pantheonsite.io/admin/content/paragraphs/6/edit instead (updated ticket testing instructions). I'm available here if you want to walk through it together with me.

apps/site/lib/site_web/templates/content/_content_list.html.eex
```diff
@@ -6,3 +6,11 @@
       teasers: @content.teasers,
       type: Keyword.get(@content.recipe, :type, :generic),
       conn: @conn) %>
+
+<%= if list_cta?({@content.ingredients.type, @content.cta}, @content.teasers, @conn.path_info) do %>
```
phildarnowsky: @amaisano it looks like we're putting `@content.ingredients.type` and `@content.cta` into a tuple here, only to just pattern-match it back into two separate params in `list_cta`. Unless there's some reason to keep the tuple I'm not seeing, how about just passing them separately?

apps/site/lib/site_web/views/content_view.ex
```diff
@@ -134,6 +137,49 @@ defmodule SiteWeb.ContentView do
...
+  # Either has required link values or has a readily available default link path
+  def list_cta?(_, _, _) do
+    true
+  end
+
+  @spec setup_list_cta(ContentList.t(), [String.t()]) :: Link.t()
+  def setup_list_cta(list, conn_path) do
+    current_path =
+      [""]
+      |> Enum.concat(conn_path)
```
phildarnowsky: It might be more idiomatic to write

```
current_path =
  ["" | conn_path]
  |> Enum.join("/")
```

PR #106, opened 2019-07-19
OpenStreetMaps | Map zoom reset

meagonqz:
#### Summary of changes
**Asana Ticket:** [🔍 OpenStreetMaps | Map zoom reset](https://app.asana.com/0/555089885850811/1125936468243528)

We were calculating the bounds based on the markers and since the markers are live updated, it would sometimes change the bounds if the markers changed enough in position on the map.

This sets the bound to whatever the initial bounds are from the initial markers. It doesn't update the bounds when vehicles move or are added/removed from the map.

<br>
Assigned to: @NAME

### DIFF
--- a/apps/site/assets/ts/schedule/components/Map.tsx
+++ b/apps/site/assets/ts/schedule/components/Map.tsx
@@ -1,4 +1,10 @@
-import React, { ReactElement, useReducer, useEffect, Dispatch } from "react";
+import React, {
+  ReactElement,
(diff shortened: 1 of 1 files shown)


ingridpierre:
Experientially, this is like night and day. Glad to be rid of this bug. Checked a few common heavily loaded routes and had no trouble zooming, and panning around the map uninterrupted.

👍

PR #107, opened 2019-07-22
Schedule Finder: aria-controls fix-up

meagonqz:
#### Summary of changes
**Asana Ticket:** No ticket

Noticed a missing id for an `aria-controls` element when looking over my PR from last week.

<br>
Assigned to: @phildarnowsky

### DIFF
--- a/apps/site/assets/ts/schedule/__tests__/__snapshots__/ScheduleTableTest.tsx.snap
+++ b/apps/site/assets/ts/schedule/__tests__/__snapshots__/ScheduleTableTest.tsx.snap
@@ -90,6 +90,7 @@ exports[`ScheduleTable it renders 1`] = `
           </tr>
           <tr
             className="schedule-table__subtable-container"
+            id="trip-40606114"
           >
             <td
               className="schedule-table__subtable-td"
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/controllers/schedule/schedule_api.ex
```diff
@@ -56,6 +56,7 @@ defmodule SiteWeb.ScheduleController.ScheduleApi do
 
   def enhance_services([]), do: []
 
+
```
phildarnowsky: How about rebasing to remove this commit entirely, since now it's not doing anything but adding some whitespace?

phildarnowsky: ...and also the next commit.

meagonqz: I always squash and merge so those commits go away...

It's not a habit on our team to rebase out commits in a topic branch. I can do it now though.

meagonqz: If you see in the final diff, there's no extra space, this commit was pre-format (there's a commit called "format")

meagonqz:
@phildarnowsky I rebased out those commits

PR #111, opened 2019-07-23
Schedule Finder | indicate school trips

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Finder Tool - show which trips are school trips](https://app.asana.com/0/555089885850811/1125331776732576)

- added caching to the route patterns repo
- added `route_pattern_id` to the `Trip` struct
- provided `route_pattern_id` to custom schedule finder data
- rendered the `S` symbol for school trips based on `RoutePattern.time_desc`
- removed some unused props that were left in previous commits
- hid some async/state work from codecov until we have a better approach to testing

---

It is difficult to test because currently there are no school trips, but this is what it looks like (with fake data)

Screen Shot 2019-07-25 at 2 04 36 PM

--- 

<br>
Assigned to: @meagonqz / @ingridpierre


apps/site/lib/site_web/controllers/schedule/schedule_api.ex
```diff
@@ -61,9 +61,14 @@ defmodule SiteWeb.ScheduleController.ScheduleApi do
     |> Stream.map(fn {trip_id, service} -> {trip_id, fares_for_service(service)} end)
     |> Stream.map(fn {trip_id, service} -> {trip_id, duration_for_service(service)} end)
     |> Stream.map(fn {trip_id, service} -> {trip_id, formatted_time(service)} end)
+    |> Enum.map(fn {trip_id, service} -> {trip_id, route_pattern(service)} end)
```
meagonqz: Can we use `Stream.map` to match the above? for the lazy eval.

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleTable.tsx
```diff
@@ -1,12 +1,36 @@
...
+}
+
+const isSchoolTrip = (
+  routePatternsById: {
+    [key: string]: RoutePatternWithShape;
+  },
+  routePatternId: string
+): boolean =>
+  (
+    (routePatternsById[routePatternId] &&
```
meagonqz: you really hate ternaries huh? 😉

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleTable.tsx
```diff
@@ -1,12 +1,36 @@
...
+    ""
+  ).match(/school/gi) !== null;
+
+const ScheduleTable = ({
+  schedule,
+  routePatterns
+}: Props): ReactElement<HTMLElement> => {
+  const routePatternsById = routePatterns.reduce(
+    (accumulator, routePattern) =>
+      Object.assign({}, accumulator, { [routePattern.id]: routePattern }),
```
meagonqz: and the spread operator 😉 
(how much am I allowed to sass you on this PR and still have you know I'm kidding? I'll stop at this one)

apps/site/assets/ts/schedule/components/schedule-finder/TableRow.tsx
```diff
@@ -61,6 +65,9 @@ const BusTableRow = ({
         onClick={onClick}
         onKeyPress={e => handleReactEnterKeyPress(e, onClick)}
       >
+        <td className="schedule-table__td--tiny">
+          <strong>{isSchoolTrip && "S"}</strong>
```
meagonqz: If it isn't a school trip, this creates an extra table data element with an empty strong, we probably shouldn't render the `<td>` unless we need it.

Do we want to add some sort of hidden label here for SR that says "School trip"? I'm not sure what we do for this iteration for sighted users to know what "S" means.

apps/site/lib/site_web/controllers/schedule/schedule_api.ex
ryan-mahoney: sure -- wasn't fully clear on the advantage -- will update

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleTable.tsx
ryan-mahoney: hahaha

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleTable.tsx
ryan-mahoney: oh man -- will update this -- I keep forgetting about the spread operator!

apps/site/assets/ts/schedule/components/schedule-finder/TableRow.tsx
ryan-mahoney: re: accessibility, will update -- makes perfect sense.

re: the table, it's a little tricky, because if any one of the trips happens to be a school trip, then they all need a `<td>` cell and header. Are you thinking to check if any trip is a school trip, and if not, don't generate the column?

ingridpierre:
I think I heard chatter that we could add in an old spring rating to test this with school trip data?

Is it possible to get an old rating loaded in and showing up under "other schedules" or something?

apps/site/assets/ts/schedule/components/schedule-finder/TableRow.tsx
```diff
@@ -61,6 +67,11 @@ const BusTableRow = ({
         onClick={onClick}
         onKeyPress={e => handleReactEnterKeyPress(e, onClick)}
       >
+        {anySchoolTrips && (
+          <td className="schedule-table__td--tiny">
+            <strong aria-label="school trip">{isSchoolTrip && "S"}</strong>
```
meagonqz: I see your point about each row needing this element. I like the approach to check anySchoolTrips. 

I think we should also have two states for the inner content, so if this individual <td> is not a school trip, then there's no <strong></strong> element (or it has different content or label). I'm not sure what SR would say with an empty <strong></strong> element, though? so perhaps this is OK?

ryan-mahoney: I thought this was going to be more work, then I did it and it was easy :) are you cool w/ keeping this implementation?

paulswartz: Please don't use the label or text "school trip".

ryan-mahoney: yes, we are removing this

ryan-mahoney: (sorry, the "keep this" was about something else, not the wording)

meagonqz: I'm good with the way that you're only including that row if any of the trips are *supplemental service. 

For a service with some supplemental trips, we might need to make a tweak to the empty state for non-supplemental trips (maybe just moving isSchoolTrip && outside of the <strong>, or having the label say something different)

ryan-mahoney:
@ingridpierre updated image

apps/site/assets/ts/schedule/components/schedule-finder/TableRow.tsx
```diff
@@ -61,6 +67,11 @@ const BusTableRow = ({
         onClick={onClick}
         onKeyPress={e => handleReactEnterKeyPress(e, onClick)}
       >
+        {anySchoolTrips && (
+          <td className="schedule-table__td--tiny">
+            <strong>{isSchoolTrip && "S"}</strong>
```
meagonqz: Can we move the isSchoolTrip outside of the strong tag?

ryan-mahoney: sorry -- I didn't get what you were saying in the other comment -- updated!

PR #116, opened 2019-07-24
Repo.teasers should support multiple `type` values

amaisano:
#### Summary of changes
**Asana Ticket:** [Events are taking the place of projects](https://app.asana.com/0/555089885850811/1130887606820083)

Changes `Repo.teasers` `type` option from single atom to list of atoms.
- `type: [:project, :news_entry]` will be converted to `{"type[]", "project"}, {"type[]", "news_entry}`

NOTE: Must be deployed _first_, then mbta/cms#286 can be deployed.

### DIFF
--- a/apps/content/lib/paragraph/content_list.ex
+++ b/apps/content/lib/paragraph/content_list.ex
@@ -43,7 +43,7 @@ defmodule Content.Paragraph.ContentList do
       terms: terms,
       term_depth: field_value(data, "field_term_depth"),
       items_per_page: field_value(data, "field_number_of_items"),
-      type: data |> field_value("field_content_type") |> content_type(),
+      type: [data |> field_value("field_content_type") |> content_type()],
       type_op: field_value(data, "field_type_logic"),
       promoted: field_value(data, "field_promoted"),
       sticky: field_value(data, "field_sticky"),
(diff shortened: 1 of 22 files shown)


apps/content/lib/paragraph/content_list.ex
```diff
@@ -202,6 +202,13 @@ defmodule Content.Paragraph.ContentList do
     |> combine()
   end
 
+  # If no types are specified, discard the :type key.
+  defp combine(%{type: [nil]} = ingredients) do
```
ryan-mahoney: are you sure you don't want to match `[]` (empty list) vs `[nil]` a list that contains a nil value?

apps/site/lib/site_web/controllers/schedule/cms.ex
```diff
@@ -11,8 +11,7 @@ defmodule SiteWeb.ScheduleController.CMS do
   alias Content.{Repo, Teaser}
 
   @featured_opts [
-    type: :news_entry,
-    type_op: "not in",
+    type: [:project, :project_update],
```
ryan-mahoney: 👍

apps/content/lib/paragraph/content_list.ex
amaisano: Thanks for bumping this. It was not a good idea to wrap the result of `content_type/1` in a list regardless of if it was nil or not, so I adjusted things to avoid the need for this extra function clause.

PR #117, opened 2019-07-25
Schedule Finder | Fix stop list for opposite Direction

meagonqz:
#### Summary of changes
**Asana Ticket:** [⏱🐞 Schedule Finder | Stop Selection Bug](https://app.asana.com/0/555089885850811/1132542813294288)

We were reversing the list of stops from the selected direction (selected on the whole page via the Direction Selector) which was not correct since some stops are unidirectional. This retrieves the correct list of stops for the other direction.

<br>
Assigned to: @NAME


meagonqz:
@ingridpierre This is ready for product review

I still need to fix up some tests.

ingridpierre:
@meagonqz This looks great, looked at it on our longest routes and ones with complicated variants. For what the original ticket was trying to accomplish, I was able to see the stops I would expect and observed that the lists changed significantly depending on the route and direction. 👍 

However, I was able to break the tool again using a particular minor edge case (saturday-only stop). I'll file a new ticket for that as well.

PR #120, opened 2019-07-25
Redesigned "More Projects" list

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [🗂️ Project Index |  Project List - Show list Items](https://app.asana.com/0/555089885850811/1131928348122856)

Redid the "More Projects" (formerly "All Projects") list on the project index page in accordance with the new design.

Note that this still needs:
* a placeholder image for projects without one
* a generic subway-mode icon
* the projects endpoint on the CMS to include statuses

<br>
Assigned to: @meagonqz


apps/site/assets/ts/projects/components/MoreProjectsTable.tsx
```diff
@@ -0,0 +1,40 @@
...
+import MoreProjectsRow from "./MoreProjectsRow";
+import { Project } from "./Project";
+
+interface Props {
+  projects: Project[];
+}
+
+const MoreProjectsTable = ({ projects }: Props): ReactElement<HTMLElement> => (
+  <table className="c-more-projects-table" aria-label="More Projects">
+    <colgroup>
```
meagonqz: I think you can remove this colgroup element.

The thead with tr & th for the table headers is all you need.

A colgroup is for a header that spans more than one column. Kind of an odd table situation that you don't see often, but if you had three columns that applied to one thing and then three columns to the right that applied to another thing, you'd group 3 and 3 in their own  colgroups.

apps/site/assets/ts/projects/components/RouteIcon.tsx
```diff
@@ -0,0 +1,49 @@
...
+  silver: silverLineIcon,
+  subway: silverLineIcon
+};
+
+const RouteIcon = ({ tag }: Props): ReactElement<HTMLElement> => {
+  const iconFunction = tagsToIcons[tag];
+  return iconFunction("c-svg__icon c-more-projects-table__route-icon");
+};
+
+export { RouteIcon as default };
```
meagonqz: This can be `export default RouteIcon`

phildarnowsky:
note to self: can filter out unused properties on the backend to make the payload smaller

phildarnowsky:
@meagonqz this current version passes CI and should address all your feedback.

phildarnowsky:
@meagonqz if the rumors I'm hearing are true, we might have status fields in production by Monday. I could get behind hiding the column behind a flag.

ingridpierre:
@phildarnowsky thanks so much for putting this up on dev - biggest thing I think is we're missing some responsiveness in the table.

- Missing some responsive layouts. For MD-SM, see: https://zpl.io/VQRxN5R For XS: https://zpl.io/beKr3Ez
- Table header for more projects should have background color of $brand-primary-lightest
- Table header text needs about .75 rem of top and bottom padding.
- Mode icons are rendering a bit taller than they are wide.
- Needs placeholder image for projects without a designated feature image.

ingridpierre:
Small adjustment: we also want the titles to be h3s not just bold body text. It will be easier for screen readers to traverse and they'll look consistently styled with the other article titles across the site.

phildarnowsky:
@meagonqz @ingridpierre FYI I'm going to remove the "Status" `<th>` with the intention of looping back to that later once we're actually publishing statuses in production.

phildarnowsky:
@ingridpierre the latest version is deploying to dev-green and should be up there by about 3:30 or 3:35.

ingridpierre:
@phildarnowsky looks good except the placeholder image appears broken to me.
image

ingridpierre:
👍 lgtm!

PR #125, opened 2019-07-30
Schedule Finder | Polish

meagonqz:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Finder Tool - Polish](https://app.asana.com/0/555089885850811/1131943808986526)

I ran the Modal flickering that can be see on LG devices (see ticket for an example) through the Chrome Performance tool that can take snapshots. It showed that the modal was going from old data -> spinner -> old data -> new data.

Since the modal went from big to small, the transition was more noticeable. This would be more noticeable on a 3G network on an older mobile device. I was able to emulate that scenario by using the Chrome Performance tool to add network and CPU throttling.

Looking at the code, 
the state was set to loading
-> a render to show the spinner happened,
the state was set to not loading
-> a render to hide the spinner happened and show the (still old) data,
then the state was set with the new data
-> a render to show the new data happened

`useReducer` is common when making external API calls as it can be helpful with debugging and when combining or caching API data across different calls. This also could have used a single `useState({ isLoading: boolean, data: Payload | null })` to set isLoading to false and data to the new data at the same time.

<br>
Assigned to: @phildarnowsky

### DIFF
--- a/apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
@@ -1,11 +1,4 @@
-import React, {
-  Dispatch,
-  ReactElement,
-  SetStateAction,
-  useEffect,
-  useState,
-  useReducer
-} from "react";
+import React, { ReactElement, useEffect, useState, useReducer } from "react";
 import SelectContainer from "./SelectContainer";
 import { ServiceWithServiceDate } from "../../../__v3api";
 import {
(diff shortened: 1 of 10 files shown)


apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
```diff
@@ -111,23 +162,43 @@ export const ServiceSelector = ({
...
+
+      if (!selectedService) {
+        return;
+      }
+      fetchData(routeId, stopId, selectedService, directionId, dispatch);
+    },
+    [services, routeId, directionId, stopId, selectedServiceId]
   );
 
+  //   const [isLoading, setIsLoading] = useState(true);
```
phildarnowsky: @meagonqz this block of commented-out stuff can go

meagonqz: Yep I already removed this in a downstream commit from this one 👍

meagonqz:
@ingridpierre this is ready for product review

ingridpierre:
@meagonqz this lgtm 👍

PR #129, opened 2019-07-31
Fix Elixir tests to work with new CI env vars

phildarnowsky:
#### Summary of changes
No ticket. In the process of changing around some env vars on CI to try to address the Trip Planner/Backstop issue, I broke some Elixir tests. This fixes them.

<br>
Assigned to: @meagonqz

### DIFF
--- a/apps/v3_api/test/headers_test.exs
+++ b/apps/v3_api/test/headers_test.exs
@@ -4,20 +4,23 @@ defmodule V3Api.HeadersTest do
   use ExUnit.Case, async: true
 
   test "always adds api header" do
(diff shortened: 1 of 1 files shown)


apps/v3_api/test/headers_test.exs
```diff
@@ -4,20 +4,23 @@ defmodule V3Api.HeadersTest do
   use ExUnit.Case, async: true
 
   test "always adds api header" do
-    assert Headers.build("API_KEY", use_cache?: false) == [
```
meagonqz: It looks like the env var must have been set differently (possibly via config?) in CI before if this was previously passing. Doesn't seem important to me, though.

PR #135, opened 2019-08-02
Paginate project list

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [🗂️ Project Index |  Project List - Pagination functionality](https://app.asana.com/0/555089885850811/1131928348122855)

Instead of displaying just the first 50 project teasers in the "More Projects" table, we display 10 to start out, and add a button to fetch more from the server.
<br>
Assigned to: @ryan-mahoney

### DIFF
--- a/apps/site/assets/ts/projects/components/MoreProjectsTable.tsx
+++ b/apps/site/assets/ts/projects/components/MoreProjectsTable.tsx
@@ -12,32 +12,24 @@ interface State {
   fetchInProgress: boolean;
 }
 
-export const fetchMoreProjects = (
+export const fetchMoreProjects = async (
   state: State,
   setState: Dispatch<State>
-): void => {
-  if (window.fetch) {
-    setState({
-      projects: state.projects,
-      fetchInProgress: true
-    });
-
-    const offset = state.projects.length;
-    window
-      .fetch(`/project_api?offset=${offset}`)
-      .then(response => {
-        if (response.ok && response.body) {
-          return response.json();
-        }
-        return [];
-      })
-      .then(newProjects => {
-        setState({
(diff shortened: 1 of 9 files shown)


apps/site/assets/ts/projects/components/MoreProjectsTable.tsx
```diff
@@ -7,41 +7,94 @@ interface Props {
   placeholderImageUrl: string;
 }
 
+interface State {
+  projects: Project[];
+  fetchInProgress: boolean;
+}
+
+export const fetchMoreProjects = (
```
ryan-mahoney: a few ideas for this function:

- return early (instead of wrapping in a big if statement)
- use spread operator when changing one value in state object
- async/await instead of promise (maybe this is just personal preference, but I find it easier to read)

```export const fetchMoreProjects = async (
  state: State,
  setState: Dispatch<State>
): Promise<void> => {
  if (!window.fetch) {
    return;
  }
  setState({ ...state, fetchInProgress: true });

  const offset = state.projects.length;
  const response = await window.fetch(`/project_api?offset=${offset}`);
  const projects: Project[] =
    response.ok && response.body ? Array.from(await response.json()) : [];

  setState({
    projects: state.projects.concat(projects),
    fetchInProgress: false
  });
};```

ingridpierre:
One small thing I'm seeing:
In desktop (LG and up) The More Projects Table goes from partial width to full width when you press the button. Causes a little jumpiness, it should be full width at all times for those breakpoints.

phildarnowsky:
@ingridpierre the latest revision should fix the table width

ingridpierre:
@phildarnowsky lgtm! 👍

PR #136, opened 2019-08-05
instrumentation for nodejs process

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** NO TICKET

Some prod servers ran out of memory over the weekend, but it doesn't see like the cause was Erlang process.  Maybe it was the nodejs processes?  It is hard to say because we don't have any log data about them.

This change gives us:
- a log of the process when it starts
- a log of the memory utilization of each process each minute
- a log of the time to handle each request
- stderr logging
- exposure of this log to cloudwatch via elastic beanstalk custom configuration

<br>
Assigned to: @NAME

### DIFF
--- a/apps/site/lib/site/react/worker.ex
+++ b/apps/site/lib/site/react/worker.ex
@@ -29,15 +29,14 @@ defmodule Site.React.Worker do
       ""
       |> receive_response(&handle_json/1)
 
-    case response do
-      %{data: %{"error" => nil} = data} ->
-        {:reply, {:ok, data}, state}
-
-      %{data: %{"error" => _error} = data} ->
-        {:reply, {:error, data}, state}
-
-      _ ->
-        {:noreply, state}
+    if get_in(response, [:data]) do
+      if get_in(response, [:data, "error"]) do
+        {:reply, {:error, response.data}, state}
+      else
+        {:reply, {:ok, response.data}, state}
+      end
+    else
+      {:noreply, state}
     end
   end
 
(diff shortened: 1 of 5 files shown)


ryan-mahoney:
currently having difficulties w/ write permissions in production environments

apps/site/react_renderer/app.js
```diff
@@ -11,6 +11,20 @@ import ScheduleDirection from "../assets/ts/schedule/components/ScheduleDirectio
 import TripPlannerResults from "../assets/ts/trip-plan-results/components/TripPlannerResults";
 import MoreProjectsTable from "../assets/ts/projects/components/MoreProjectsTable";
 
+// log the process id when a process is started
+process.stdout.write(`node_logging node_process process=${process.pid}\n`);
```
phildarnowsky: @meagonqz we're logging just enough messages, and their format is regular enough, that I'd like to see a helper that we could invoke here like `log("node_process", {process: process.pid})`, which would do all the formatting for us.

apps/site/react_renderer/app.js
```diff
@@ -11,6 +11,20 @@ import ScheduleDirection from "../assets/ts/schedule/components/ScheduleDirectio
...
+
+// log the memory usage each minute
+const memoryUsage = () => {
+  const used = process.memoryUsage().heapUsed / 1024 / 1024;
+  process.stdout.write(
+    `node_logging node_memory process=${process.pid} memory_in_mb=${Math.round(
```
phildarnowsky: @meagonqz as a further example, the helper I'm thinking about here would here be invoked `log("node_memory", {process: process.pid, memory_in_mb: Math.round(used * 100) / 100})`

apps/site/lib/site/react/worker.ex
```diff
@@ -27,30 +27,64 @@ defmodule Site.React.Worker do
...
+        |> Enum.map(&handle_data(&1, handle_fn))
+        |> Enum.find(&(!is_nil(&1)))
+      end)
 
-        if String.ends_with?(str, "\n") do
-          new_resp
-        else
-          receive_response(new_resp)
-        end
+    Logger.warn("node_logging req_time milliseconds=#{time / 1_000}")
```
meagonqz: I moved the req time logging here because having it in node was too slow. 

Tests were timing out on {:render, Component, props} intermittently. I think it was pushing certain tests over the threshold for workers (5000ms), some of these renderings are already slow as we know.

We were also getting the huge HTML blob and the final node_logging line in the same stream response so elixir had to do a lot of string parsing.

amaisano:
Was this put into the `warn` level purposefully? What about putting it at the `info` level? I'm seeing a lot of output from this now in CI and locally.

PR #140, opened 2019-08-05
Projects Index | Text Partial

amaisano:
#### Summary of changes
**Asana Ticket:** [Projects Index | Text Partial](https://app.asana.com/0/555089885850811/1133902923804293)

Instead of hard coding the text intro to this page, use a CMS paragraph that authors have access to edit using the paragraph ("partial") system.

NOTE: Required an update to a few repo and view functions to work properly.

image

### DIFF
--- a/apps/cms/lib/api/static.ex
+++ b/apps/cms/lib/api/static.ex
@@ -417,7 +417,11 @@ defmodule CMS.API.Static do
     end
   end
 
-  def view("/admin/content/paragraphs/" <> _paragraph_id, _) do
+  def view("/admin/content/paragraphs/25", params) do
+    redirect("/paragraphs/custom-html/projects-index", params, 301)
+  end
+
+  def view("/paragraphs/custom-html/projects-index", _) do
     {:ok, paragraph_response()}
   end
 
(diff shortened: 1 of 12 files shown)


apps/cms/lib/api/static.ex
```diff
@@ -417,7 +417,11 @@ defmodule CMS.API.Static do
     end
   end
 
-  def view("/admin/content/paragraphs/" <> _paragraph_id, _) do
```
meagonqz: Do you no longer need this generic case for a paragraph by node_id?

amaisano: So, you can still call a paragraph by ID, but you have to include that full path ("admin/content/etc/**25**"). True, we lose the ability to simply call `paragraph(25)` but now all paragraphs have fully-adjustable aliases like pages do, so it will nearly always redirect.

@ryan-mahoney , do you think we should keep/continue to support the "permalink" style call of `paragraph(25)` along with the wildcard alias?

ryan-mahoney: if we were using that frequently I would say it might be nice to keep it around for simplicity... but since we are not, I think it's okay to do away w/ it.

PR #142, opened 2019-08-06
Project Index |  Project Banner

meagonqz:
#### Summary of changes
**Asana Ticket:** [🗂️ Project Index |  Project Banner](https://app.asana.com/0/555089885850811/1132297856215673)

<br>
Assigned to: @phildarnowsky

### DIFF
--- a/apps/site/assets/ts/projects/components/Banner.tsx
+++ b/apps/site/assets/ts/projects/components/Banner.tsx
@@ -16,6 +16,7 @@ export const cmsRouteToClass = (route: Route): string => {
   if (route.group === "custom") return routeToCSSClass(route.mode!);
   if (route.group === "mode" && route.id !== null)
     return routeToCSSClass(route.id);
+  if (route.group === "line") return `${routeToCSSClass(route.id)}-line`;
   return "unknown";
 };
 
--- a/apps/site/assets/ts/projects/components/Banner.tsx
+++ b/apps/site/assets/ts/projects/components/Banner.tsx
@@ -28,12 +28,12 @@ const bannerBgClass = (banner: Project): string => {
 };
 
 const bannerContentClass = (banner: Project): string =>
-  `m-banner__content m-banner__content--default m-banner__content--left u-bg--${bannerBgClass(
+  `m-banner__content m-banner__content--responsive-side-by-side m-banner__content--left u-bg--${bannerBgClass(
     banner
   )}`;
 
 const bannerUpdated = (banner: Project): ReactElement<HTMLElement> => (
-  <span>Updated {formattedDate(banner.date)}</span>
+  <span className="m-banner__date">Updated {formattedDate(banner.date)}</span>
 );
 
 const BannerContent = ({
@@ -62,18 +62,34 @@ const Banner = ({
     href={banner.path}
     className="m-banner m-banner--responsive m-banner--responsive-lg-margin m-banner--small-margin-top m-banner--default"
   >
-    <div
-      className="m-banner__image m-banner__image--responsive m-banner__image--default"
-      style={{
-        backgroundImage: ` url(${bannerImageURL(banner, placeholderImageUrl)})`
-      }}
-    >
-      {banner.image ? <div className="sr-only">{banner.image.alt}</div> : null}
-      <div className="container hidden-xs-down">
+    <div className="hidden-xs-down">
+      <div className="m-banner__image m-banner__image--responsive-side-by-side m-banner__image--default">
         <BannerContent banner={banner} />
+        {(banner.image && (
+          <img
+            className="m-banner__image--by-side"
+            src={banner.image.url}
+            alt={banner.image.alt}
+          />
+        )) || (
+          <img
+            className="m-banner__image--by-side"
+            src={placeholderImageUrl}
+            alt="MBTA logo"
+          />
+        )}
       </div>
     </div>
     <div className="hidden-sm-up">
+      <div
+        className="m-banner__image m-banner__image--responsive m-banner__image--default"
+        style={{
+          backgroundImage: ` url(${bannerImageURL(
+            banner,
+            placeholderImageUrl
+          )})`
+        }}
+      />
       <BannerContent banner={banner} />
     </div>
   </a>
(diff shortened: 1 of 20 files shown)


apps/site/assets/ts/projects/components/__projects.d.ts
```diff
@@ -15,7 +15,44 @@ export interface Image {
...
+  banner_type: "default" | "important";
+  text_position: "left" | "right";
+  category: string;
+  routes: Route[];
+  title: string;
+  updated_on: string;
+  utm_url: string;
+}
+
+export interface ProjectTeaser {
```
phildarnowsky: This `interface` seems redundant with `Project` above.

phildarnowsky: This `interface` seems redundant with `Project` above.

apps/site/lib/site_web/controllers/project_controller.ex
```diff
@@ -218,6 +219,9 @@ defmodule SiteWeb.ProjectController do
 
   @spec simplify_teaser(map()) :: map()
   defp simplify_teaser(teaser) do
-    Map.take(teaser, ~w(id image path title routes date status)a)
+    teaser
+    |> Map.put(:path, project_path(SiteWeb.Endpoint, :show, teaser))
+    |> Map.put(:formatted_date, ViewHelpers.format_full_date(teaser.date))
```
phildarnowsky: In `projects/components/MoreProjectsRow.tsx` there's an existing `formattedDate` function that you could pull out into a helper and re-use, rather than formatting the date on the backend and passing it in.

apps/site/lib/site_web/views/helpers/cms_router_helpers.ex
```diff
@@ -85,7 +85,7 @@ defmodule SiteWeb.CmsRouterHelpers do
   end
 
   @spec project_path(
-          Conn.t(),
+          Conn.t() | any,
```
phildarnowsky: Why `any`? Is it not possible to be more specific about the type here? Seems like it would be desirable to do so if possible.

apps/site/assets/ts/projects/components/Banner.tsx
```diff
@@ -0,0 +1,72 @@
...
+);
+
+const Banner = ({ banner }: Props): ReactElement<HTMLElement> => (
+  <a
+    href={banner.path}
+    className="m-banner m-banner--responsive m-banner--small-margin-top m-banner--default"
+  >
+    <div
+      className="m-banner__image m-banner__image--responsive m-banner__image--default"
+      style={{ backgroundImage: ` url(${banner.image.url})` }}
```
phildarnowsky: It's not guaranteed that `image` will be set, we have a placeholder image that we can display if the image is missing. See how `placeholderImageUrl` is used in `MoreProjectsRow.tsx` for example.

apps/site/assets/ts/projects/components/Banner.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const bannerBgClass = (banner: ProjectTeaserWithDate): string => {
+  if (banner.routes.length === 0) return "u-bg--unknown";
+  return bannerRouteClass(banner);
+};
+
+const toCSSClass = (name: string): string =>
+  name
+    .toLowerCase()
+    .replace(" ", "-")
```
phildarnowsky: We should use a regex here, because according to [the api docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace), "If `pattern` is a string, only the first occurrence will be replaced." And since we have to use a regex anyway, might as well combine both calls into one:

```
name
  .toLowerCase()
  .replace(/[ _]/g, "-")
```

apps/site/lib/site_web/views/helpers/cms_router_helpers.ex
meagonqz: I couldn't find a type for SiteWeb.Endpoint. I did some internet searching and didn't find anything helpful either.

paulswartz: `module`

apps/site/assets/ts/projects/components/Banner.tsx
meagonqz: I took this from https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/stop/components/RoutePillList.tsx#L15

Since there was extra CMS specific logic for groups I didn't take the whole function. There shouldn't really be a route that has more than one hyphen but I can change it in both places.

apps/site/assets/ts/projects/components/__projects.d.ts
meagonqz: I can fix this up.

ProjectTeaser would be the "pure" CMS API interface for a teaser of type project whereas Project above shows the version we're sending down which is "simplified". I like having the Elixir structs match the Typescript interfaces but this already is polluted by the simplified version via optional parameters. We do have like Simple<Blah> interfaces so maybe I'll rename Project to SimpleProject.

apps/site/assets/ts/projects/components/Banner.tsx
meagonqz: 👍 

I'm not sure if we want to use the same placeholder for the banner but I'll use this for now. @ingridpierre

ingridpierre:
Some notes: 
- Should capitalize "Updated" on "Updated on..."
- Banner should be full width, with no margins, only for the MD-SM breakpoints. For larger and smaller breakpoints they take up the normal 12/12 columns with margins on the outside outside.
- Waiting on the changes we discussed over slack reflective of: https://app.zeplin.io/project/5c422a32bfb873bf2bf338c1/dashboard?seid=5d278a0bdadd1f77b1777da2
- Concerned about what the banner will look like when the project updates sidebar comes in, we may need to make height/width adjustments to accommodate the loss of columns on LG and larger devices.

meagonqz:
@ingridpierre could you clarify the comment below? thanks!

"Banner should be full width, with no margins, only for the MD-SM breakpoints. For larger and smaller breakpoints they take up the normal 12/12 columns with margins on the outside outside." Do you want XS to be different than this mockup: https://app.zeplin.io/project/5c422a32bfb873bf2bf338c1/screen/5d279ca2d363eb9dcbf71f64? The XS without margin is what's the homepage ATM.

ingridpierre:
@meagonqz sorry for the confusing comment, there are two elements in XS. The solid color title block has a margin, and the image doesn't. What's in the mockup is accurate (except for the typo).

meagonqz:
@ingridpierre and @phildarnowsky this is ready for another look

ingridpierre:
@meagonqz lgtm! 👍

PR #146, opened 2019-08-09
Feature projects on new project index

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [🗂️ Project Index |  Project Feature](https://app.asana.com/0/555089885850811/1132297856215674)

Added "Featured Projects" to the new project index page, i.e. those that have the `sticky` flag set except for the one being used for the banner.

<br>
Assigned to: @meagonqz

### DIFF
--- a/apps/site/assets/ts/projects/components/FeaturedProjectsList.tsx
+++ b/apps/site/assets/ts/projects/components/FeaturedProjectsList.tsx
@@ -1,0 +1,53 @@
+import React, { ReactElement } from "react";
+import { SimpleProject as Project } from "./__projects";
+import FeaturedProjectsRow from "./FeaturedProjectsRow";
+
+interface Props {
+  projects: Project[];
+  placeholderImageUrl: string;
+}
+
+const FeaturedProjectsList = ({
+  projects,
+  placeholderImageUrl
+}: Props): ReactElement<HTMLElement> => {
+  const doGroupPairwise = (
+    projectsToGroup: Project[],
+    acc: Project[][]
+  ): Project[][] => {
+    if (projectsToGroup.length === 0) {
+      return acc;
+    }
+
+    const nextPair = projectsToGroup.slice(0, 2);
+    const rest = projectsToGroup.slice(2);
+    return doGroupPairwise(rest, acc.concat([nextPair]));
+  };
+
+  const groupPairwise = (projectsToGroup: Project[]): Project[][] =>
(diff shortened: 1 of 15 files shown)


ryan-mahoney:
Screen Shot 2019-08-09 at 1 13 35 PM

looks like there is a left-margin (or padding) on the two images below the main image that is not in the design

apps/site/assets/ts/projects/components/FeaturedProjectsList.tsx
```diff
@@ -0,0 +1,53 @@
...
+interface Props {
+  projects: Project[];
+  placeholderImageUrl: string;
+}
+
+const FeaturedProjectsList = ({
+  projects,
+  placeholderImageUrl
+}: Props): ReactElement<HTMLElement> => {
+  const doGroupPairwise = (
```
ryan-mahoney: do `doGroupPairwise` and `groupPairWise` need to be defined inside `FeatureProjectList`?  If not, might be better to put them above rather than inside

apps/site/assets/ts/projects/components/FeaturedProjectsRow.tsx
```diff
@@ -0,0 +1,34 @@
...
+interface Props {
+  placeholderImageUrl: string;
+  projectPair: Project[];
+}
+
+const FeaturedProjectsRow = ({
+  projectPair,
+  placeholderImageUrl
+}: Props): ReactElement<HTMLElement> => {
+  const projectToComponent = (project: Project): JSX.Element => (
```
ryan-mahoney: also might be better to define `projectToComponent` outside `FeaturedProjectsRow`

apps/site/assets/ts/projects/components/RoutePill.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { ReactElement } from "react";
```
ryan-mahoney: I think we might already have a component that does this somewhere... although I maybe it's similar but not the same. See `apps/site/assets/ts/stop/components/RoutePillList.tsx`

phildarnowsky: @ryan-mahoney yeah, this `RoutePillList` is similar in intent to the other but implemented somewhat differently; its appearance is also somewhat different.

phildarnowsky:
@ryan-mahoney ready for another look

liana:
This looks good to me. 👍

PR #149, opened 2019-08-12
Add cdn.knightlab.com to EmbeddedMedia

meagonqz:
#### Summary of changes
**Asana Ticket:** [❗ 🚧 🔶 OL Cars | Embedded iframe has funky sizing](https://app.asana.com/0/555089885850811/1134554424225557)

This allows these iframes to be full width (break out of our columns). 

The scrollbars are due to the Knightlab/Juxtapose javascript in the iframe setting an explicit width/height on an element within it. I don't see any easy way around this. After this change, content authors could use option #2 on https://mbta.com/projects/orange-line-improvement-program/update/new-orange-line-cars?preview=&vid=28918&nid=4453. It seems to resize itself best in our flexbox layout.

The image may not break out of the column at certain viewpoints. If you inspect the element, you will see it has added padding (top or left) because Juxtapose is trying to retain the original aspect ratio of the images. The padding keeps the image from getting stretched.

Deployed to dev-green at https://green.dev.mbtace.com/projects/orange-line-improvement-program/update/new-orange-line-cars?preview=&vid=28918&nid=4453

<br>
Assigned to: @NAME

### DIFF
--- a/apps/site/lib/site/content_rewriters/embedded_media.ex
+++ b/apps/site/lib/site/content_rewriters/embedded_media.ex
@@ -12,7 +12,8 @@ defmodule Site.ContentRewriters.EmbeddedMedia do
     "livestream.com",
     "mbtace.com/shuttle-tracker",
     "youtube.com",
-    "youtu.be"
+    "youtu.be",
+    "cdn.knightlab.com"
   ]
 
   defstruct alignment: :none,


meagonqz:
> This helps a bit but the end result is still a little wonky within the grid system. I think if the authors choose an exact 4:3 or 16:9 (needs experimentation) image, we won't get all that extra padding or scrollbars, since the `c-media` system defaults to a 16:9 ratio on iframes (padding-bottom `56.25%`). The image they have is a standard DSLR 3:2 ratio, but could perhaps be cropped. Alternatively, we can add a CSS ratio for 3:2 and map this by iframe `src` domain...

I think it would be best to crop the images to get a different aspect ratio. I could add some code to match on knightlabs and set `padding-bottom: 66.66%` but that would expect all future knightlabs embedded media to be 3:2 as well.

aayesha:
Hmm, I think option #2 on https://www.mbta.com/projects/orange-line-improvement-program/update/new-orange-line-cars?preview=&vid=28918&nid=4453 works for now. we can experiment with resizing later. thanks!

PR #150, opened 2019-08-13
Change 424 bus from inner express to local

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [❗🚍 Better Bus | Remove 424 from hard-coded list of express routes for Sept 1](https://app.asana.com/0/555089885850811/1133711653869789)

This changes the 424 from an inner express bus to a local. I considered making the code changes detailed in the ticket, but on investigation decided that it was more complex than I wanted to do for something with a hard deadline like this.

<br>
Assigned to: @meagonqz


phildarnowsky:
@meagonqz Note that @aayesha pointed out that the 448 and 449 are being cut by the Better Bus Project, so on her suggestion I'm removing them from the list too.

PR #152, opened 2019-08-13
Fix alert icons in Transit Near Me

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [🐞 Routes with high priority alerts aren't showing alert icon in TNM](https://app.asana.com/0/555089885850811/1135457050018732)

A bug in Transit Near Me was preventing routes with alerts from showing alert icons. Fixed.

<br>
Assigned to: @ryan-mahoney


phildarnowsky:
@ryan-mahoney After discussing how this wasn't worth fixing, I happened upon the buggy section of code, and saw that it was a small enough fix that it probably was worth doing after all, even if it's only going to be in production for a few days.

PR #153, opened 2019-08-13
automatically transpile async/await in JS

ryan-mahoney:
#### Summary of changes
NO TICKET

@meagonqz observed that project pagination was not working on browsers that do not support async / await. This should allow us to keep using it but have it transpiled to promises at build time.

<br>
Assigned to: @meagonqz

### DIFF
--- a/apps/site/assets/.babelrc.js
+++ b/apps/site/assets/.babelrc.js
@@ -7,5 +7,7 @@ module.exports = {
         corejs: 2
       }
     ]
-  ]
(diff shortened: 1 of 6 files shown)


meagonqz:
covered by babel-polyfill and babel-preset-env.
issue was related to whatwg-fetch missing for this page and using response.body which doesnt exist in that polyfill, addressed in https://github.com/mbta/dotcom/pull/161

PR #159, opened 2019-08-14
Transit Near Me | Use realtime data API

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [TNM Performance: pull new realtime schedule data asynchronously from front-end](https://app.asana.com/0/555089885850811/1134294684192689)

- Transit Near Me controller now only fetches stops and their distances
- remove existing API for fetching an entirely new set of Transit Near Me data
- call the Realtime Schedule API in TNM on load and then every 15 seconds
- convert Realtime data into both TNM formats on client side in JS
- bring in `dot-prop-immutable` library (it's small!) for updating deeply nested TNM data
- fix broken elixir tests
- add new jest tests

KNOWN ISSUES:
- in some cases predictions may be incorrect because they were added in an not correctly added to the route_pattern -- this is a known issue and will be fixed on API side. Doesn't happen ofter, but I've seen it on some Red Line predictions

<br>
Assigned to: @meagonqz / @phildarnowsky


apps/site/test/site/lib/react_test.exs
```diff
@@ -64,38 +56,11 @@ defmodule Site.ReactTest do
...
-                     routeSidebarData: "crash",
-                     stopSidebarData: []
-                   })
+                   React.render("TransitNearMe", %{})
         end)
 
-      assert log =~ ~r/react_renderer component=TransitNearMe.* is not a function/
-    end
-
-    test "it translates zero-width space HTML entities into UTF-8 characters", %{
```
meagonqz: Did you mean to remove this test?

ryan-mahoney: It needs to be moved somewhere else because there is no data w/ a zero-width space that would be rendered on the page initially.

I'm thinking to add a new task to the backlog to put this test into another component. I don't think it is specific to TNM, but that is where we first encountered this issue.

apps/site/assets/ts/tnm/components/TransitNearMe.tsx
```diff
@@ -101,11 +95,26 @@ const TransitNearMe = ({
...
+  );
+  const selectedStop = getSelectedStop(
+    state.stopsWithRoutes,
+    state.selectedStopId
+  );
+
+  // call once on initial load
+  useEffect(
+    () => {
+      fetchRealtimeSchedules(stopsWithDistances.stops, dispatch);
```
meagonqz: So I suppose a trade-off of doing this all clientside is that SSR rendering is moot for the cards. The user experience of the sometimes long page load is worse though.

apps/site/assets/ts/tnm/components/TransitNearMe.tsx
```diff
@@ -42,33 +45,25 @@ const validateModeFilter = (acc: Mode[], mode: string): Mode[] =>
...
-        if (response.ok) return response.json();
-        throw new Error(response.statusText);
-      })
-      .then((routes: RouteWithStopsWithDirections[]) =>
-        dispatch(routeSidebarDataAction(routes))
-      )
-      .catch(() => {});
-  }
+): Promise<undefined> => {
+  if (!window.fetch) return;
```
meagonqz: More appropriate to make the SSR comment here: we trade off SSR rendering of the route cards for a faster initial page load without the cards.

ryan-mahoney: at an earlier stage we determined that it currently isn't possible for us to build all the data to render all the cards fast enough in a single request.  The main feature we are getting with this change is the ability to do the data fetching in smaller chunks.  However, in a future optimization we might send down the data for the first few cards so that they are rendered ~instantly, but that is out of scope for first pass

ryan-mahoney: I think once this new foundation is in place there are a few potential next steps we could take to improve the user experience... but I'm hesitant to put more into this PR as it's already take so long :/

apps/site/assets/ts/tnm/helpers/process-realtime-data.ts
```diff
@@ -0,0 +1,316 @@
...
+
+      return {
+        prediction: shortPrediction,
+        scheduled_time: scheduledTime,
+        delay
+      } as PredictedOrScheduledTime;
+    }
+  ),
+
+  train_number: ""
```
meagonqz: this train number would be schedule.trip.name, but we only would display it for CR
https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/schedule/components/schedule-finder/TableRow.tsx#L164

meagonqz: https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/components/Headsign.tsx#L110 I think setting that value will just work

apps/site/assets/ts/tnm/components/TransitNearMe.tsx
```diff
@@ -42,33 +45,25 @@ const validateModeFilter = (acc: Mode[], mode: string): Mode[] =>
...
 
-  return new Promise(resolve => resolve());
+  for (const stop of stops) {
```
meagonqz: Did we do any testing on how long this takes to do network requests serially instead of either getting all the data in one network call, or doing an async mapping with a Promise.all resolution?

I suspect the amount of data might mean it's an imperceivable difference but it would be good to know since it's the problem we're solving and offloading some work to the FE.

apps/site/lib/site/transit_near_me.ex
```diff
@@ -42,31 +42,20 @@ defmodule Site.TransitNearMe do
     schedules_fn: &Schedules.Repo.schedule_for_stop/2
   ]
 
-  @spec build(Address.t(), Keyword.t()) :: t() | error
+  @spec build(Address.t(), Keyword.t()) :: map
```
meagonqz: Could we update the defstruct above to not include schedules and keep this as `t()`?

meagonqz:
Another issue I noticed is that the map doesn't always center correctly now. maybe some missing data?:

https://mbta.com/transit-near-me?_utf8=%E2%9C%93&location%5Baddress%5D=Boston+Logan+Airport&location%5Blatitude%5D=42.365396&location%5Blongitude%5D=-71.017547
vs
https://green.dev.mbtace.com/transit-near-me?_utf8=%E2%9C%93&location%5Baddress%5D=Boston+Logan+Airport&location%5Blatitude%5D=42.365396&location%5Blongitude%5D=-71.017547

ryan-mahoney:
> FWIW I think it's still possible that the route cards could be displayed out of order because the data is being dispatched and concat'd as it comes in, not once the Promise.all resolves. (i.e. South Stations has a lot of data and it might take longer for that data to be parsed).
> 
> For data to be definitely in order (diff style):
> 
> ```
>     const json = Promise.all(responses.map(async response => ...
>  +      return data;
>  -      dispatch(realtimeScheduleDataAction(data));
>       })
>     );
>  +  json.map(data => dispatch(realtimeScheduleDataAction(data)));
> ```
> 
> I don't think this is a big deal.

I man... I had done something like this... I must have lost it during a merge conflict. Will add

PR #169, opened 2019-08-20
Make SVG icons accessible

amaisano:
#### Summary of changes
**Asana Ticket:** [Mode icons are not accessible on projects index and station pages](https://app.asana.com/0/555089885850811/1134068764318785)

- Changed `aria-hidden` to `false` whenever certain mode/accessibility/etc icons are rendered
- Default is still `aria-hidden=false`
- ARIA will announce embedded SVG `<title>` value
- Update snapshots and tests

Note: Opted _not_ to use the black-colored hover-tooltip as per @meagonqz 's comment below.

### DIFF
--- a/apps/site/assets/svgo.yml
+++ b/apps/site/assets/svgo.yml
@@ -1,5 +1,5 @@
 plugins:
-  - removeTitle: true
+  - removeTitle: false
   - removeAttrs:
       attrs:
         - 'id'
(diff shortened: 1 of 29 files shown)


aayesha:
Had Rob Dias from the Massachusetts Office on Disability check https://www.mbta.com/stops with these screen readers:
- Jaws 18 (2 year old version)
- IE 11 on Windows 7
- Voice-Over on iPhone

He said, "...the icons are labeled. They are not showing up as traditional graphics which is good…they are labeled, no issues on my end."

PR #171, opened 2019-08-21
Rail replacement shuttles should use :free_fare

amaisano:
#### Summary of changes
**Asana Ticket:** [Fares | Users know bus shuttles are free](https://app.asana.com/0/555089885850811/1133954832480177)

- any route with `description: :rail_replacement_bus` will return a `:free_fare` base fare instead of a typical bus fare
- affects fare output on both old and new trip info components

image

image

### DIFF
--- a/apps/site/lib/site/base_fare.ex
+++ b/apps/site/lib/site/base_fare.ex
@@ -8,7 +8,7 @@ defmodule Site.BaseFare do
   the express fare is always returned.
   """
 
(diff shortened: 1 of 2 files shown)


amaisano:
@jfabi I think that would be a great idea. Since that attribute isn't currently parsed/mapped to the `Route.t()` object however I decided to leverage the way we do it for certain transfers.

PR #172, opened 2019-08-21
Fix readability of weekend-selected dates on calendar

amaisano:
#### Summary of changes
**Asana Ticket:** [Schedules | selected date on the calendar hard to read](https://app.asana.com/0/555089885850811/1132719231879592)

Increases specificity of `.schedule-selected` selector so that selected weekend dates use dark blue styling, making the white text readable.

image

### DIFF
--- a/apps/site/assets/css/_schedule.scss
+++ b/apps/site/assets/css/_schedule.scss
@@ -396,6 +396,7 @@ a {
 }
 
 .schedule-selected,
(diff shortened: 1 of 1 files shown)


amaisano:
Decided to avoid using `@extend` as it creates some odd compilation side affects that were unexpected. Using `@mixin` instead gives us the same end result without ordering glitches.

See https://webinista.com/updates/dont-use-extend-sass/.

PR #173, opened 2019-08-21
Projects | Render all three components as one render

meagonqz:
#### Summary of changes
**Asana Ticket:** No ticket

Cleanup in preparation for handling filters, etc.

<br>
Assigned to: @phildarnowsky

### DIFF
--- a/apps/site/assets/ts/projects/components/MoreProjectsTable.tsx
+++ b/apps/site/assets/ts/projects/components/MoreProjectsTable.tsx
@@ -41,48 +41,53 @@ const MoreProjectsTable = ({
   });
 
   return (
-    <div>
-      <table className="c-more-projects-table" aria-label="More Projects">
-        <thead className="c-more-projects-table__thead hidden-md-down">
-          <tr>
-            <th
-              scope="col"
-              className="c-more-projects-table__th c-more-projects-table__th-project"
-            >
-              Project
-            </th>
-            <th
-              scope="col"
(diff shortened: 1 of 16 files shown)


apps/site/assets/ts/projects/components/MoreProjectsTable.tsx
```diff
@@ -41,48 +41,53 @@ const MoreProjectsTable = ({
...
-            <MoreProjectsRow
-              key={project.id}
-              placeholderImageUrl={placeholderImageUrl}
-              {...project}
-            />
-          ))}
-        </tbody>
-      </table>
+    <div className="container">
+      <h2 className="hidden-sm-down featured-project-subheader">
```
meagonqz: This in the elixir template was set as "className" instead of "class" so it didn't take effect. Do we want this hidden on sm down @phildarnowsky?

apps/site/lib/site_web/templates/project/_all_projects.html.eex
```diff
@@ -1,8 +1,8 @@
-<h2 className="hidden-sm-down featured-project-subheader">More Projects</h2>
```
meagonqz: Here's the "className" reference if that helps.

apps/site/assets/ts/projects/components/MoreProjectsTable.tsx
phildarnowsky: @meagonqz not sure why I put that there, this header should be visible at all breakpoints.

PR #177, opened 2019-08-22
Don't reset map center to downtown on first zoom

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [🐞 Trip planner map issue w/ double clicking](https://app.asana.com/0/555089885850811/1135810086712490)

Here we use a trick similar to https://github.com/mbta/dotcom/pull/37, in which we save off the (calculated) center of the map and save it to the state before zooming causes a re-render. Also we don't set an explicit center if bounds are set on the map, to avoid introducing another bug in the process of fixing the above.

<br>
Assigned to: @meagonqz


phildarnowsky:
closing in favor of simpler solution

PR #185, opened 2019-08-28
Project page filters

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [🗂️ Project Index |  Filter - Mode filters](https://app.asana.com/0/555089885850811/1136817875129090)

Added mode-filter buttons that determine which projects and project updates are shown on this page, by mode.

<br>
Assigned to: @meagonqz


phildarnowsky:
@meagonqz this should address all your feedback, I expect all tests to pass except for one of the Elixir tests I fixed this morning.

ingridpierre:
An adjustment we'd like to the logic here would be to include more projects in any remaining feature/banner slots, ordered by most recently updated. Currently we only include projects labeled as sticky, but when a filter is active we'd like to see sticky + most recent in the Featured Projects.

ingridpierre:
In the XS breakpoint it seems that the banner element disappears.

ingridpierre:
In XS, the filter buttons should use 100% of the width (2x2 stretched to fit) , I think the fixed size here is creating some undesirable gaps for mobile devices:

image

ingridpierre:
Looks like the banner color isn't coming through here, should be purple I imagine, but also we should have a default color (primary blue) when there's no mode or line tagged (to see this I filtered by CR projects). 
image

ingridpierre:
Also not seeing banner elements show up at all in XS breakpoint.

ryan-mahoney:
@ingridpierre thanks for reviewing, looking into those issues now

ryan-mahoney:
@phildarnowsky I spoke to Ingrid about this yesterday in Slack.

The color issue was resolved. The only remaining issue is the first item should appear in a banner style even at XS breakpoint.

phildarnowsky:
@ingridpierre The version currently on dev-green should remedy the XS-banner problem.

ingridpierre:
@phildarnowsky 👍 looks fabulous!

PR #190, opened 2019-08-29
Project Index |  Search - Add Algolia Search

meagonqz:
#### Summary of changes
**Asana Ticket:** [🗂️ Project Index |  Search - Add Algolia Search](https://app.asana.com/0/555089885850811/1134953213042611)

<br>
Assigned to: @ryan-mahoney


meagonqz:
@aayesha This is up on green.dev

apps/site/assets/js/algolia-result.js
```diff
@@ -56,30 +77,47 @@ export const TEMPLATES = {
...
+    hitFeatureIcons: getFeatureIcons(hit, index),
     id: hit.place_id || null
   });
 }
 
-export function getIcon(hit, type) {
+export function getIcon(hit, type, searchType) {
+  if (searchType) {
+    if (searchType === "projects") {
+      console.log(hit.related_transit_gtfs_id, "related transit id");
```
meagonqz: I'll remove these console statements before merging.

ingridpierre:
@meagonqz this looks great! i think I was hoping for the icons to be on the right like they are on /stops search.

ingridpierre:
I'm pinging Aayesha to re-index our Algolia results so that the search terms match on recently added items, though I'm worried now about the need to constantly do this to keep search working on this page.

Is there a way we can automate that?

meagonqz:
@ingridpierre what does that mean for the lefthand icon? Should there be none or should it always be the little "i" for info? In cases where a result doesn't have any related data, what should the icons be? (currently the little "i" on the left is for no data).

meagonqz:
@ingridpierre re: indexing, I don't know. I think that drupal automatically reindexes its algolia index every 3 hours, but @amaisano would know more.

aayesha:
I'm reindexing the live CMS now. Looks like there's a bug: [Algolia-Drupal re-indexing operation is broken](https://app.asana.com/0/search/1137779314850650/1137779314850649)

aayesha:
But if @ingridpierre  is testing on green.dev, that's not fed by the live CMS. I'm not sure if it comes from the test CMS. @amaisano, if I reindex test, will that address Ingrid's issues?

amaisano:
Do not index anything but the live site, or else the live Algolia index will be completely replaced by the stale content on TEST or DEV Pantheon sites. The dev dotcom sites all use the live site's Drupal algolia data.

ingridpierre:
@meagonqz  sorry I missed you question above, since we want the icons for this one to go to the right, I don't think the (i) icon is necessary. We'll just leave it blank and both changes will make it easier to scan this particular list since the # of icons is changing from line to line.

ingridpierre:
@amaisano @aayesha  Sorry for the confusion here, I forgot that we were set up to automatically index results every 3 hours. Given that's the case, what do you make of https://green.dev.mbtace.com/projects/systemwide-commuter-rail-improvements not showing up in search results by the title? Example below:

image
image

aayesha:
The Systemwide CR Improvements project page got published but then was **unpublished**, which is why it's not showing up in search results maybe? and green.dev is referencing an older version for the index page maybe?

PR #197, opened 2019-09-05
cache realtime data, reduce TNM polling time

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [TNM: cache result by stop for 30 seconds](https://app.asana.com/0/555089885850811/1138499454939089)

- use `RepoCache` to cache complete JSON response for stops for 30 seconds
- currently, we are only sending a single stop from TNM, so caching the list of stops should be fine for achieving re-use across separate requests.
- the JSON responses are currently anywhere from 1kb to 20kb, as such, I'm not expecting this cache to grow too large as it seems not span more than 1000 stops in a 30 second window
- also, we have recently doubled the dotcom memory and are still typically running at less than 50% of the available memory
- with this cache in place, I reduced the polling time of the frontend

---

Some local time measurements:
http://localhost:4001/api/realtime/stops/?stops=place-welln

Not Cached:
2.402423
1.998292
2.127799

Cached:
0.090732
0.090380
0.097794

<br>
Assigned to: @paulswartz

### DIFF
--- a/apps/site/assets/ts/tnm/components/TransitNearMe.tsx
+++ b/apps/site/assets/ts/tnm/components/TransitNearMe.tsx
@@ -148,7 +148,7 @@ const TransitNearMe = ({
         stopsWithDistances.stops.map(stop => stop.id),
         dispatch
       ),
-    45000
+    30000
   );
 
   return (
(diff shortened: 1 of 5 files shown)


ryan-mahoney:
test is failing -- will look into it tomorrow AM

PR #198, opened 2019-09-06
Teaser endpoint argument changes

amaisano:
#### Summary of changes
**Asana Ticket:** [Teasers endpoint serving incorrect Content List results](https://app.asana.com/0/555089885850811/1137138645355009)

Prepare for changes in https://github.com/mbta/cms/pull/301 (**deploy this FIRST**)
- Always fill both slots for term arguments in request: `/cms/teasers/<tag>/<route>`
- If either argument is unnecessary, fill with default wildcard `"any"`
- Simplify logic for building this part of the content list recipe

### DIFF
--- a/apps/cms/test/paragraph/content_list_test.exs
+++ b/apps/cms/test/paragraph/content_list_test.exs
@@ -81,30 +81,21 @@ defmodule CMS.Partial.Paragraph.ContentListTest do
       opts =
         cms_map(
           terms: nil,
+          routes: nil,
           term_depth: 4
         )
 
       assert opts == []
     end
 
-    test "If the default term depth is found, discard depth and compose arguments" do
-      opts =
-        cms_map(
-          terms: [123, 321],
-          term_depth: 4
-        )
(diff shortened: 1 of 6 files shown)


apps/cms/test/paragraph/content_list_test.exs
```diff
@@ -206,8 +198,8 @@ defmodule CMS.Partial.Paragraph.ContentListTest do
     {"#{k}", [%{"value" => v}]}
   end
 
-  defp cms_field({:terms = k, terms}) do
-    {"field_#{k}", Enum.map(terms, &%{"target_id" => &1})}
```
ryan-mahoney: this looks like it used to build a tuple that contained a list of possibly many items, but now it will just contain a list of one item.  Is this the desired change?

amaisano: Correct. The CMS API is being updated to limit each field to only a single value array. To date, nothing the CMS used more than one term -- the JSON response will never have more than one result.

PR #201, opened 2019-09-09
fix project pill className for all subway lines

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [🔥 Project Index | Another invisible pill!](https://app.asana.com/0/555089885850811/1138833128646984)

Updated:

Screen Shot 2019-09-09 at 1 39 26 PM

<br>
Assigned to: @amaisano

### DIFF
--- a/apps/site/assets/ts/projects/__tests__/RoutePillListTest.tsx
+++ b/apps/site/assets/ts/projects/__tests__/RoutePillListTest.tsx
@@ -38,6 +38,16 @@ describe("routeToModeName", () => {
 
     expect(routeToModeName(greenLineBranchRoute)).toEqual("green-line");
   });
+
+  it("returns the correct name for all subway line indicator", () => {
+    expect(
+      routeToModeName({
+        mode: "subway",
+        id: "subway",
+        group: "line"
+      })
+    ).toEqual("subway");
+  });
 });
 
 describe("component", () => {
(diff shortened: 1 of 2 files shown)


apps/site/assets/ts/projects/__tests__/RoutePillListTest.tsx
```diff
@@ -38,6 +38,16 @@ describe("routeToModeName", () => {
 
     expect(routeToModeName(greenLineBranchRoute)).toEqual("green-line");
   });
+
+  it("returns the correct name for all subway line indicator", () => {
+    expect(
+      routeToModeName({
+        mode: "subway",
+        id: "subway",
+        group: "line"
```
amaisano: Interesting. I would have expected subway to be classified in the `mode` group, not the `line` group. Is this what we're getting from the CMS or is this just something you wanted to test?

amaisano: image

ryan-mahoney: the function doesn't actually use that field, I just got it from copy/paste from previous test

amaisano: I'm still a little confused on how a subway tag from the CMS could come out as "subway--line" when there is no metadata on the tag with "line" in the first place, but this work does resolve the issue to LGTM.

ryan-mahoney: there is a function that converts this data into class names and this cases was falling into a branch where `-line` was being added as a suffix

PR #208, opened 2019-09-10
fix issues causing 500 errors in realtime API

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** NO TICKET

This is an odd issue... basically, deploys have been failing because during the heath check stage of new servers instances, enough 500 errors are generated that the server instance fails the check and the deploy rolls back.

The odd thing about this is that it seems that the errors are not appearing in Splunk. The number of 500 errors recorded there vs the number of 500 responses I can see in the AWS Elastic Beanstalk Heath console do not correlate.

I found this issue by downloading a log file from Beanstalk directly. That should be investigated in . a separate ticket.

What is happening here is somehow a value that is not a valid stop_id is being passed and the application is crashing because that is an unhandled case.

<br>
Assigned to: @phildarnowsky

### DIFF
--- a/apps/site/lib/site/realtime_schedule.ex
+++ b/apps/site/lib/site/realtime_schedule.ex
@@ -27,7 +27,7 @@ defmodule Site.RealtimeSchedule do
   @predicted_schedules_per_stop 2
 
   @default_opts [
-    stops_fn: &StopsRepo.get!/1,
+    stops_fn: &StopsRepo.get/1,
     routes_fn: &RoutesRepo.by_stop_with_route_pattern/1,
     predictions_fn: &PredictionsRepo.all_no_cache/1,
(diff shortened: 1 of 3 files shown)


apps/site/lib/site/realtime_schedule.ex
```diff
@@ -227,6 +236,10 @@ defmodule Site.RealtimeSchedule do
   end
 
   @spec build_output(map, [route_with_patterns_t], map, map, map, DateTime.t()) :: [map]
+  defp build_output(%{message: _message}, _, _, _, _, _) do
```
phildarnowsky: @ryan-mahoney it's not clear what this clause is for

ryan-mahoney: removed

PR #209, opened 2019-09-11
Fix wrapping time in schedule finder table

amaisano:
#### Summary of changes
**Asana Ticket:** [Schedules | Finder Tool - Wrapping times in mobile](https://app.asana.com/0/555089885850811/1137523575743818)

By adding the proper `colspan` of `3` to the "Trip Length" / "Fare" row, the table flexes correctly to fit the times on a single line. **Bonus**: Remove `whitespace: nowrap` on headsigns due to longer headsigns (eg, "Haverhill (Shuttle))" causing page overflow on XS.

**Before / After:**

image

**Before / After:**

image

### DIFF
--- a/apps/site/assets/css/_schedule-page.scss
+++ b/apps/site/assets/css/_schedule-page.scss
@@ -417,6 +417,7 @@
   padding: $base-spacing / 4 0;
 
   &--right-adjusted {
+    padding-left: $base-spacing / 2;
     text-align: right;
     vertical-align: top;
     white-space: pre;
--- a/apps/site/assets/css/_schedule-page.scss
+++ b/apps/site/assets/css/_schedule-page.scss
@@ -418,6 +418,8 @@
 
   &--right-adjusted {
     text-align: right;
+    vertical-align: top;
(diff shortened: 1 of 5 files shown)


ingridpierre:
Looks like Middleborough could use some love here. I believe there's some function or behavior somewhere to split up long CR names like this if they need to wrap?
image

ingridpierre:
Also seeing this on various buses as well (434 and 435 are my usual suspects to test):
image

amaisano:
@ingridpierre I made some improvements to two places:

- FORCE time to ALWAYS be non-wrapping (even when other columns are maxed out) -- nice catch
- Add the `breakTextAtSlash` helper function to stop name and headsign output instances

I added a new screenshot above to illustrate these improvements (taken at 320px wide). They don't 100% fix super long words so there are still some instances of text overflow. Also, due to this slight overflow, the down caret is off the page and not visible (the one for expanding a trip).

ingridpierre:
@amaisano Thanks, that's looking much better! 👍

PR #211, opened 2019-09-11
fix ios 10 safari issue

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [🐞 Safari | Website doesn't render correctly](https://app.asana.com/0/555089885850811/1139044049600985)

For whatever reason, this CSS rule was causing the page to overflow in iOS 10 and be aligned to the bottom of the page with no scroll bar.

<br>
Assigned to: @amaisano

### DIFF
--- a/apps/site/assets/css/_fare-card.scss
+++ b/apps/site/assets/css/_fare-card.scss
@@ -44,7 +44,7 @@ $fare-card-icon-offset: 1.375 * $base-spacing;
   &--linked {
     @extend %button-focus-color;
 
-    p a {  
+    a {
       position: relative;
       z-index: 2;
     }
--- a/apps/site/assets/css/_fare-card.scss
+++ b/apps/site/assets/css/_fare-card.scss
@@ -44,6 +44,11 @@ $fare-card-icon-offset: 1.375 * $base-spacing;
   &--linked {
     @extend %button-focus-color;
 
+    p a {  
+      position: relative;
+      z-index: 2;
+    }
+
     .u-linked-card__primary-link {
       color: $black;
       text-decoration: none;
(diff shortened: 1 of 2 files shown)


apps/site/assets/css/_utilities.scss
```diff
@@ -163,9 +163,9 @@
       z-index: 1;
     }
   }
+}
 
-  a:not(&__primary-link) {
-    position: relative;
-    z-index: 2;
-  }
+.u-linked-card a:not(&__primary-link) {
```
meagonqz: Probably need to add `// scss-lint:disable` here

apps/site/assets/css/_fare-card.scss
```diff
@@ -44,6 +44,11 @@ $fare-card-icon-offset: 1.375 * $base-spacing;
   &--linked {
     @extend %button-focus-color;
 
+    p a {
```
amaisano: this will not work. the homepage banners use this utility and the secondary links are _not_ within `<p>` tags. Why can this not just be any link within the parent selector?

amaisano: This utility is used for things besides just fare cards. Hence it's placement in `utilities`

apps/site/assets/css/_utilities.scss
ryan-mahoney: updated

apps/site/assets/css/_fare-card.scss
```diff
@@ -44,7 +44,7 @@ $fare-card-icon-offset: 1.375 * $base-spacing;
   &--linked {
     @extend %button-focus-color;
 
-    p a {  
+    a {
       position: relative;
```
amaisano: These cannot go here. They must go within the global utility selectors as more than just fare cards use this setup. Please return these to the `.u-linked-card` utility helper.

PR #215, opened 2019-09-12
Schedule Page | Direction Selector

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Directionality - Direction header box style](https://app.asana.com/0/555089885850811/1118913618186296)

- added styles
- changed it to display headsign instead of route pattern name
- only show route pattern selector for bus
- only show origin for bus if the headsigns for the direction have duplicate names
- somewhat accessible, although would be great to see what IHCD thinks when this entire page is done
- depends on latest version of API

---

Screen Shot 2019-09-20 at 2 55 27 PM

Screen Shot 2019-09-20 at 2 55 56 PM

---

<br>
Assigned to: @meagonqz / @ingridpierre

### DIFF
--- a/apps/site/assets/ts/schedule/components/direction/ScheduleDirectionMenu.tsx
+++ b/apps/site/assets/ts/schedule/components/direction/ScheduleDirectionMenu.tsx
@@ -227,6 +227,7 @@ const ExpandedMenu = ({
           id="route-pattern_uncommon"
           role="menuitem"
           tabIndex={0}
+          aria-label="click for additional routes"
           className="m-schedule-direction__menu-item"
           onClick={handleClick}
           onKeyUp={(e: ReactKeyboardEvent) =>


ingridpierre:
**Let's please match the design and style this component differently.** 

- Arrow should come just after the name of the route
- No border
- Keep the focus styling where you see a rounded blue background on hover.
- Style it with header styling (so it gets helvetica)
- inbound/outbound direction label: .u-small-caps

I wonder if ultimately it shouldn't be semantically a select since it's more similar to a floating menu like:
image

What do you think, especially in terms of what's appropriate for screenreaders to interpret?

------

Looks like all the commuter rail lines and the subway lines (except green) changed from being bidirectional to having variants when we start using route patterns across the board. I *think* that **for the subway lines  and CR we have to use the direction_destinations field for this label instead**. I should have anticipated this in the notes for the ticket, happy to discuss if this changes or complicates how we'll pull the trip shapes for the future schedule work.

image

Instead of this:
image

ingridpierre:
Thanks for your close attention on accessibility, really pleased with how nicely it's working so far.
_Paraphrasing our earlier convo, we'd like to use route info on CR and Subway (direction_destinations) rather than route patterns. Let's try your idea of adding "from [origin]" when typicality 1 headsigns are present and identical._

1. **Direction and headsign** 
.m-schedule-direction__direction (ex. INBOUND)  and .m-schedule-direction__route-pattern (Central Square) should be Helvetica. See the Schedule Finder's modal header as an existing style to compare to.

2. **Headsign in menu** 
.m-schedule-direction__menu-item-headsign should have font-weight: 500, font-size: 1rem

3. **Route pattern description**
.m-schedule-direction__menu-item-description should have color: $gray-light by default, and $gray on hover/focus. Differs from design, my original color values here weren't meeting contrast guidelines, sorry.

4. **Button**
Looks like some containers are colliding with the button. FYI rte 434 has our longest headsign.
image

5. *Add margin above the button in SM - XS.* 1rem

6. *Add margin below schedule finder in SM - XS breakpoints* 1rem

7.  Globally, not just here I'd like all our '.btn's to have font-weight: 500. That white/blue combo fails AAA contrast so this will help a tiny bit.

ingridpierre:
Pretty confusing interaction here. On SL1 I clicked "Uncommon destinations" and the hidden routes were not appended to the bottom of the list where I expected them, but rather the middle of the list.


image

ingridpierre:
Looks like... something weird happened to alerts maybe on this branch? Seeing duplicates stack up both with and without the flag disabled.

image

ryan-mahoney:
> Pretty confusing interaction here. On SL1 I clicked "Uncommon destinations" and the hidden routes were not appended to the bottom of the list where I expected them, but rather the middle of the list.
> 
> image

ahh I see -- sorting issue -- will fix

ryan-mahoney:
> Looks like... something weird happened to alerts maybe on this branch? Seeing duplicates stack up both with and without the flag disabled.
> 
> image

interesting... wonder if that is related to this branch using the updated API version.  Will look into it.

ingridpierre:
@ryan-mahoney Could we add some logic to only show the "from" attribute when headsigns are duplicative? I don't think its useful for every variant the way it's showing up now.

ryan-mahoney:
> @ryan-mahoney Could we add some logic to only show the "from" text when headsigns are duplicative? I don't think its useful for every variant the way it's showing up now.

yes... we could do that :)

ryan-mahoney:
@ingridpierre  I think the alert issue might just be because we are looking on dev-green

apps/site/assets/css/_buttons.scss
```diff
@@ -8,6 +8,7 @@
   // Bootstrap override (uses absolute base font size otherwise).
   // Allows button to inherit size of surrounding/body text.
   font-size: inherit;
+  font-weight: $font-weight-medium;
```
ryan-mahoney: Ingrid requested to change this globally

apps/site/assets/css/_select-custom.scss
```diff
@@ -0,0 +1,38 @@
+.c-select-custom {
```
ryan-mahoney: in an earlier version of this code I had made the custom select CSS more module so it can be re-used... but then I didn't end up re-using it.  I could remove this... or make it it's own small PR... but I consider it to be fairly harmless, so I am leaving it in for now :)

apps/site/assets/static/images/icon-change-direction.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48">
```
meagonqz: Can we add a title to this? We did that for all of our other SVGs for accessibility (lol as you might remember)

apps/site/assets/ts/schedule/__tests__/ScheduleTableTest.tsx
```diff
@@ -46,7 +48,7 @@ const crRoutePatterns = [
     id: "CR-Fairmount-0-0",
     direction_id: 0
   }
-] as RoutePatternWithShape[];
+] as EnhancedRoutePattern[];
```
meagonqz: Shouldn't this crRoutePattern have a headsign value too?

apps/site/assets/ts/schedule/components/direction/ScheduleDirectionMenu.tsx
```diff
@@ -1,44 +1,359 @@
...
+  )!.headsign;
+
+/* istanbul ignore next */
+const externalCloseEvent = (dispatch: Dispatch<Action>): void => {
+  document.addEventListener(
+    "click",
+    (event: Event): void => {
+      if (!event.target) return;
+      const element = event.target as HTMLElement;
+      if (element.closest(".js-m-schedule-click-boundary")) return;
```
meagonqz: Closest isnt supported in IE11. Do we have or can we add a polyfill?

https://caniuse.com/#search=closest

meagonqz:
@ryan-mahoney which w3c spec/reference did you use for this? so we are on the same page as to what kind of menu this is etc

ryan-mahoney:
> @ryan-mahoney which w3c spec/reference did you use for this? so we are on the same page as to what kind of menu this is etc

I was looking at https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html

meagonqz:
Okay, great!

https://www.w3.org/WAI/tutorials/menus/ has more detail on menus...the success criteria mentions having another way to access these pages since menus can be hard for some folks, as well as something about having a clear indicator to the SR of what the current value is.

for 1) another way to access these pages we could create a ticket for as it'd probably require design but for 2) it looks like we can add `aria-current="page"` to whichever menu item is the current page

apps/site/assets/ts/schedule/components/direction/ScheduleDirectionMenu.tsx
```diff
@@ -1,44 +1,359 @@
...
+            key={routePattern.id}
+            routePatternIds={routePatternIds}
+            routePattern={routePattern}
+            selected={selectedRoutePatternId === routePattern.id}
+            focused={index === focusIndex}
+            duplicated={isDuplicateHeadsign(routePattern, routePatterns)}
+            dispatch={dispatch}
+          />
+        ))}
+      {hasMoreRoutePatterns(routePatterns) && showAllRoutePatterns === false && (
```
meagonqz: I don't know if this item as a button that gets replaced with extra items could be more accessible -- we could wait on feedback on it from IHCD.
I think for now we could add a hidden div / label that says something like "click for additional routes" or something.

ryan-mahoney: good call. Added an `aria-label` and the experience was much better w/ screen reader

PR #223, opened 2019-09-18
capture and render overnight parking information

amaisano:
#### Summary of changes
**Asana Ticket:** [🚙 Add overnight parking data](https://app.asana.com/0/555089885850811/1134739080933662)

- adds new elixir key to parking capacity struct
- mirrored changes in react components
- adjusted output according to asana for each overnight value in elixir prior to react
- prints "Unknown" if overnight data is not [yet] available in the api, or other invalid key is found.

image

### DIFF
--- a/apps/site/assets/ts/stop/__tests__/sidebar/__snapshots__/ParkingTest.tsx.snap
+++ b/apps/site/assets/ts/stop/__tests__/sidebar/__snapshots__/ParkingTest.tsx.snap
@@ -111,6 +111,7 @@ Array [
               <strong>
                 Overnight Parking: 
               </strong>
+              Available
             </li>
             <li>
               <strong>
@@ -312,6 +313,7 @@ exports[`Parking via enzyme-to-json it displays parking info when opened and inn
                 <strong>
                   Overnight Parking: 
                 </strong>
+                Available
               </li>
               <li>
                 <strong>
--- a/apps/site/assets/ts/stop/__tests__/sidebar/__snapshots__/ParkingTest.tsx.snap
+++ b/apps/site/assets/ts/stop/__tests__/sidebar/__snapshots__/ParkingTest.tsx.snap
@@ -109,6 +109,11 @@ Array [
             </li>
             <li>
               <strong>
+                Overnight Parking: 
+              </strong>
+            </li>
+            <li>
+              <strong>
                 Type: 
               </strong>
               Garage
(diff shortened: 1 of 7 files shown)


apps/site/assets/ts/stop/components/sidebar/Parking.tsx
```diff
@@ -44,6 +44,10 @@ const renderCapacity = (lot: ParkingLot): ReactElement<HTMLElement> => {
             <strong>Accessible spaces: </strong>
             {accessible}
           </li>
+          <li>
```
meagonqz: Since overnight is optional and might not be in the data, we should show this HTML segment if overnight is available.
you can do something like
```{ overnight && <li><strong>Overnight Parking: ... </li> }```, where the false case {undefined} wouldn't render anything via jsx in React

It looks like the other lis in this file could use that treatment too since they are listed as optional, so you could fix those up too 🙏 .

Unless overnight data is always present, then we should change the interface to say `overnight` instead `overnight?`

File: apps/site/assets/ts/stop/__tests__/sidebar/__snapshots__/ParkingTest.tsx.snap [hunk 2]
```diff
@@ -303,6 +308,11 @@ exports[`Parking via enzyme-to-json it displays parking info when opened and inn
                 </strong>
                 4
               </li>
+              <li>
```
meagonqz: You could add to the parking data structure in `ParkingTest.tsx` so that it outputs data in the snapshot.

apps/stops/lib/stop.ex
```diff
@@ -190,6 +194,17 @@ defmodule Stops.Stop.ParkingLot.Capacity do
...
+  @spec pretty_overnight_msg(String.t() | nil) :: String.t()
+  defp pretty_overnight_msg("no"), do: "Not available"
+  defp pretty_overnight_msg("yes"), do: "Available"
+  defp pretty_overnight_msg("yes-except-snow"), do: "Available, except during snow emergencies"
+  defp pretty_overnight_msg("no-except-snow"), do: "Not available, except during snow emergencies"
+
+  defp pretty_overnight_msg("yes-snow-unknown"),
+    do: "Available. During snow emergencies, check posted signs."
+
+  defp pretty_overnight_msg(_), do: "Unknown"
```
meagonqz: Oh, hm, just seeing that we do this now, so we'd always have a value for overnight parking. In that case you can change the interface to remove the ? instead of my comment above.

apps/site/assets/ts/stop/components/sidebar/Parking.tsx
meagonqz: Sorry, I missed the part where we were adding a default value for no data.

PR #225, opened 2019-09-19
Adjust some tests to match recent GTFS changes

phildarnowsky:
No ticket. Some GTFS changes were pushed out today that broke a number of our tests. Corrected.

<br>
Assigned to: @meagonqz


apps/predictions/test/repo_test.exs
```diff
@@ -37,7 +37,7 @@ defmodule Predictions.RepoTest do
 
     test "filters by min_time" do
       min_time = Util.now() |> Timex.shift(minutes: 15)
-      predictions = Repo.all(route: "Orange", min_time: min_time)
+      predictions = Repo.all(route: "39", min_time: min_time)
```
meagonqz: Was this changed because there were no predictions for orange line or?

phildarnowsky: @meagonqz yes, although I just did try changing it back to `Orange` and now it works fine...going to leave it on `39`.

PR #231, opened 2019-09-26
Green line route switcher

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Green line special case for route selection](https://app.asana.com/0/555089885850811/1141160481766011)

Added Green line route switching capability

<br>
Assigned to: @NAME

### DIFF
--- a/apps/site/assets/css/_schedule-direction.scss
+++ b/apps/site/assets/css/_schedule-direction.scss
@@ -106,8 +106,8 @@
 
 .m-schedule-direction__menu-item-icon {
   margin-right: $base-spacing * .5;
-  width: $base-spacing * 1.25;
   transform: translateY($base-spacing * .25);
+  width: $base-spacing * 1.25;
 }
 
 .m-schedule-direction__menu-item-description {
--- a/apps/site/assets/css/_schedule-direction.scss
+++ b/apps/site/assets/css/_schedule-direction.scss
@@ -27,6 +27,7 @@
 .m-schedule-direction__route-pattern {
   align-items: center;
   display: flex;
+  font-family: $headings-font-family;
   font-size: $font-size-h3;
   font-weight: bold;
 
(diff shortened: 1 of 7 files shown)


ingridpierre:
1.  .m-schedule-direction__route-pattern should be Helvetica.
2. Hooray! We get to use my new docs maybe:

**Green Line Direction picker documentation**
https://zeroheight.com/2fedee66c/p/33616f/b/59a3ae/t/869217

Referring to the examples contained in the link above, looks like we're using route-long-name instead of the direction-destinations field. For example, we should say "[Green line B] Boston College"  and "[Green Line B] Park St" rather than "Green Line B" for both the eastbound and westbound label. 

Since it's semantically important, the branch icons need aria descriptions so they say "Green Line B".

3. From an accessibility standpoint, we've heard some complaints when focus doesn't return to the direction picker after it's been interacted with and it transforms the page. But since we're navigating to a new page entirely, I'm wondering if changing the focus is still a good idea?

phildarnowsky:
@ryan-mahoney took another look, don't have anything further to add at the moment.

ingridpierre:
@phildarnowsky is this still in review? can we merge this work?

phildarnowsky:
@ingridpierre I don't believe @ryan-mahoney has made the changes I called for, correct me if I'm wrong Ryan.

ryan-mahoney:
yes, there is still some work remaining

ingridpierre:
1. The Green Line icons should not be aria-hidden. We should have them read out "Green Line", "Green Line B", etc since they are pretty important in this context.

2. Seeing an issue with branch direction names. It's coming through as Eastbound: North Station. Should say Eastbound: Park Street.

See `direction_destination` here, subway lines should use that field instead of headsign, shape names, or route pattern name:
https://api-v3.mbta.com/routes/Green-B

image

ryan-mahoney:
@ingridpierre please take another look. I fixed the Green B issue and added screen-reader only text to address the icons being invisible to screen-readers

phildarnowsky:
@ryan-mahoney LGTM

PR #233, opened 2019-09-26
Schedules | Line Diagram - Stop Components

meagonqz:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Line Diagram - Stop Components](https://app.asana.com/0/555089885850811/1141229989083697)

The "stop_data" object is derived from the "stop_bubbles" {stop_bubbles, route_stops} data structure that exists on the current page -- but oddly enough it appears like the "stop_data.branch" value is incorrect. I'm not sure if there is some intention there because I didn't change this function. It can be looked at when we build out the line itself -- the data for the correct branch IS there, within "route_stop.branch" so it shouldn't be a big issue. @ryan-mahoney for context

<br>
Assigned to: @phildarnowsky


apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -102,6 +104,49 @@ const ScheduleDirection = ({
...
+          ({ route_stop: routeStop, stop_data: stopData }: LineDiagramStop) => (
+            <div className="m-schedule-line-diagram__stop">
+              <div className="m-schedule-line-diagram__card-left">
+                <div className="m-schedule-line-diagram__stop-name">
+                  {routeStop.name}
+                </div>
+                <div>{stopData.type === "terminus" && stopData.branch}</div>
+                <div className="m-schedule-line-diagram__connections">
+                  {routeStop.connections.map((route: Route) =>
+                    route.type === 3 && !route.name.startsWith("SL") ? (
```
phildarnowsky: @meagonqz let's extract these lines into a helper method

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -0,0 +1,41 @@
+defmodule SiteWeb.ScheduleController.LineApi do
+  @moduledoc """
+    API for retrieving Map data by route and variant
```
phildarnowsky: @meagonqz did you mean "retrieving LineDiagram data"?

apps/site/lib/site_web/controllers/schedule/line/diagram_helpers.ex
```diff
@@ -0,0 +1,326 @@
+defmodule SiteWeb.ScheduleController.Line.DiagramHelpers do
```
phildarnowsky: @meagonqz is this file basically a cut-and-paste of the big chunk of code that formerly was in `ScheduleController.Line`, or did you also make more significant changes in the process?

meagonqz: Yes, this code was just moved over into it's own module so it could be shared (and minimize the size and complexity of that Line `plug`)

phildarnowsky: that's what I was hoping to hear

apps/site/lib/site_web/controllers/schedule/line_controller.ex
```diff
@@ -82,11 +83,19 @@ defmodule SiteWeb.ScheduleController.LineController do
...
-        shape_map: conn.assigns.shape_map
+        shape_map: conn.assigns.shape_map,
+        line_diagram: Enum.map(conn.assigns.all_stops, &update_route_stop_data/1)
       }
     )
   end
 
+  def update_route_stop_data({data, %RouteStop{} = map}) do
+    %{
+      stop_data: %{branch: List.first(Keyword.keys(data)), type: List.first(Keyword.values(data))},
```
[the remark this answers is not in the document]
meagonqz: I punted on making this exact because I know it's for the line diagram but that's not this task.

It's a list of tuples like [{:stop, nil}] or [{:line, Braintree}, {:stop, Ashmont}] etc and Poison won't handle encoding tuples, so I made it into a list of maps for JS with presumed keys to identify each tuples intention.

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
meagonqz: I'd like to hold off on that since this is scaffolding and this part needs additional UI work anyway (like handling multiple commuter rail icons). It will change so I don't see the need to move it into a separate file yet.

apps/site/lib/site_web/controllers/schedule/line_api.ex
meagonqz: Copy pasta, thanks

PR #238, opened 2019-10-03
Stop list diagram changes with new directionality interface

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [React wrapper for existing line diagram and remove flag](https://app.asana.com/0/555089885850811/1141392483330380)

Here we hook up the old static-HTML stop line diagram to the newfangled directionality interface, allowing us to get rid of the flag gating the new interface off and cut over to it permanently.

<br>
Assigned to: @meagonqz

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/green.ex
+++ b/apps/site/lib/site_web/controllers/schedule/green.ex
@@ -70,6 +70,7 @@ defmodule SiteWeb.ScheduleController.Green do
     |> await_assign_all_default(__MODULE__)
     |> put_view(ScheduleView)
     |> LineController.assign_schedule_page_data()
+    |> LineController.assign_stop_list_html()
     |> render("show.html", [])
   end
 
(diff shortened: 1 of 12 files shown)


apps/site/lib/site_web/controllers/schedule/green.ex
```diff
@@ -41,6 +41,18 @@ defmodule SiteWeb.ScheduleController.Green do
 
   def show(conn, _parmas), do: line(conn, [])
 
+  def line_diagram_api(conn, _) do
+    conn
+    |> call_plug(SiteWeb.ScheduleController.HoursOfOperation)
```
meagonqz: Do we need to call all these plugs for the green line version of the API? I noticed the one in line_controller doesn't

phildarnowsky: I want to say that I did verify that we do need all of them, but I'll double-check.

phildarnowsky:
Closing this, it would possibly take longer to resolve the conflicts than it would to write and deploy the stuff that's supposed to replace this.

PR #239, opened 2019-10-03
Adjust fares for Foxboro pilot

amaisano:
#### Summary of changes
**Asana Ticket:** [Foxboro | Fix pricing ($20 special event vs. Zone 4 vs. Interzone 4)](https://app.asana.com/0/555089885850811/1140953760634051)

Examines trip name and route id to determine if fares should be adjusted for Foxboro related trips.
- Trip argument added to `base_fare` function. Contains train # for commuter rail.
- If trip train name is on exception list and traveling to/from zone 1A, force it to zone 1
- If any fare request involves the route id of "CR-Foxboro," return the $20 fare
- Fare Finder for CR now prints all Zone 4 fares if a Foxboro station is selected
- Fare Finder continues to output the $20 fare if a Foxboro station is selected
- Updated Schedule Finder to work with this as well

**Schedule Finder**
image

**Schedule Tab**
image

### DIFF
--- a/apps/fares/lib/fares.ex
+++ b/apps/fares/lib/fares.ex
@@ -13,10 +13,8 @@ defmodule Fares do
   @outer_express_routes ~w(352 354 505)
   @outer_express_route_set MapSet.new(@outer_express_routes)
 
-  @foxboro_interzone ~w(741 743 745 750 752 754 756)
-  @foxboro_interzone_set MapSet.new(@foxboro_interzone)
-
-  @default_trip %Trip{name: "", id: ""}
+  @foxboro_reverse_commute ~w(741 743 745 750 752 754 756)
+  @foxboro_reverse_commute_set MapSet.new(@foxboro_reverse_commute)
 
   @type ferry_name ::
           :ferry_cross_harbor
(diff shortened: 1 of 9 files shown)


ingridpierre:
Not seeing fares come out of the schedule finder at all the way I would expect.

See this outbound train from Talbot Ave, going from Zone 1A - Zone 4. We're saying if it travels for 1 stop to the next (1A- 1A) it will cost $6.25 , and if it travels all the way to Foxboro in Zone 4 (1A-Zone 4) it will cost $4.50. This should instead say 2.40 and 8.75 respectively.

image

ingridpierre:
FYI - the same can be said of the schedule tab, seeing 1A zone travel appear as $6.50 (SS to Foxboro wrongly returns no results). However, since we're retiring this tab with the Foxboro pilot, I only mention this if it's helpful for troubleshooting.

image

ingridpierre:
@amaisano This is set from my perspective. Pending approval from other stakeholders on any other business rules for Foxboro fares I am not aware of (please update me if anything changes so I can clearly document that elsewhere), but the above issue was addressed. 👍

PR #240, opened 2019-10-07
Foxboro | Franklin timetable "via Fairmount Line"

meagonqz:
#### Summary of changes
**Asana Ticket:** [🦊 🕸️❗Foxboro | Franklin timetable should include "via Fairmount Line"  on some trains](https://app.asana.com/0/555089885850811/1142379675259929)

There are 4 stops outbound and 3 stops (no Hyde Park) inbound in the Franklin "schedules". See details in ticket. Summary is that some Franklin non-via schedules include Hyde Park outbound but not inbound. We produce the timetable through the schedules from the API.

<br>
Assigned to: @NAME


ingridpierre:
@meagonqz can we add some sr-only tags here and hide the literal text in the "via ..." cells?

ingridpierre:
@meagonqz lgtm! 👍 We'll also be putting these through the wringer with IHCD when we restart work with them so this just helps us get set up for that later.

PR #241, opened 2019-10-08
Hide CR schedule tab (redirect to timetable)

amaisano:
#### Summary of changes
**Asana Ticket:** [Foxboro | Franklin AND Fairmount schedule tab doesn't display inbound trips when origin = Foxboro](https://app.asana.com/0/555089885850811/1143031265509297)

- redirects CR route schedules to timetable tab
- hides CR Schedule tab
- hides "Schedules from here" for CR stop list

image

### DIFF
--- a/apps/site/assets/css/_schedule.scss
+++ b/apps/site/assets/css/_schedule.scss
@@ -1033,7 +1033,8 @@ del .trip-list-realtime {
       }
     }
 
-    .info-tab {
+    .info-tab,
+    .schedule-tab {
       @include media-breakpoint-up(sm) {
         display: none;
       }
--- a/apps/site/assets/css/_schedule.scss
+++ b/apps/site/assets/css/_schedule.scss
@@ -1033,8 +1033,7 @@ del .trip-list-realtime {
       }
     }
 
-    .info-tab,
-    .schedule-tab {
+    .info-tab {
       @include media-breakpoint-up(sm) {
         display: none;
       }
--- a/apps/site/lib/site_web/controllers/schedule/trip_view_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/trip_view_controller.ex
@@ -4,6 +4,7 @@ defmodule SiteWeb.ScheduleController.TripViewController do
   """
   use SiteWeb, :controller
   alias Routes.Route
+  alias Zones.Repo
 
   plug(SiteWeb.Plugs.Route)
   plug(SiteWeb.Plugs.DateInRating)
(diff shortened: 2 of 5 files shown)


aayesha:
Deferring to @ingridpierre  on this one for review.

apps/site/lib/site_web/controllers/schedule/trip_view_controller.ex
```diff
@@ -16,6 +16,10 @@ defmodule SiteWeb.ScheduleController.TripViewController do
   plug(SiteWeb.ScheduleController.ScheduleError)
   plug(:zone_map)
 
+  def show(conn, %{"route" => "CR-" <> _ = route}) do
```
meagonqz: This works, but I think to make it more future-proof, we should match on the route type instead of string matching on the name and use the `timetable_path` helper.

```
  def show(
        %{assigns: %{route: %Route{type: 2, id: route_id}}, query_params: query_params} = conn,
        _
      ) do
    redirect(conn, to: timetable_path(conn, :show, route_id, query_params))
  end
```

apps/site/assets/css/_schedule.scss
```diff
@@ -1033,7 +1033,8 @@ del .trip-list-realtime {
       }
     }
 
-    .info-tab {
+    .info-tab,
+    .schedule-tab {
```
meagonqz: There is still a schedule tab showing on sm down so I think there is something left to adjust. I think it would be better to do it in the view functions instead of hiding via css. See my other comment on the test.

apps/site/test/site_web/views/schedule_view_test.exs
```diff
@@ -1025,7 +1025,7 @@ defmodule SiteWeb.ScheduleViewTest do
   end
 
   describe "route_header_tabs/1" do
-    test "returns 5 tabs for commuter rail (1 hidden by css)", %{conn: conn} do
+    test "returns 5 tabs for commuter rail (2 hidden by css)", %{conn: conn} do
```
meagonqz: I think we should change this function under test here to not return the schedule tab at all for commuter rail, it doesn't look like it would be difficult: https://github.com/mbta/dotcom/blob/master/apps/site/lib/site_web/views/schedule_view.ex#L386 

One was hidden by css before in this test because one was only shown on mobile to allow for better responsiveness. We had to adjust the size of the tabs on mobile so that 4 tabs would fit across.

ingridpierre:
@amaisano 👍  looks great -  very pleased to see this tab and those links go. The links will return when we have that line diagram updated with the new stop card style.

PR #245, opened 2019-10-09
Fix new Fairmount line diagram

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [🦊 🕸️❗Foxboro | Fairmount line diagram displays Dedham Corp Center twice/disconnected](https://app.asana.com/0/555089885850811/1142034065520052)

The revised `Shape`s for the Fairmount Line don't cooperate well with our branch-deduction code that ultimately dictates the shape and content of the line diagram. In MVP fashion, fixed by adding a special case for the Fairmount line that tweaks the pertinent data by hand in such a way that the diagram comes out correctly.

<br>
Assigned to: @meagonqz

### DIFF
--- a/apps/stops/lib/route_stop.ex
+++ b/apps/stops/lib/route_stop.ex
@@ -99,6 +99,32 @@ defmodule Stops.RouteStop do
     do_list_from_shapes(shape.name, Enum.map(stops, & &1.id), stops, route)
   end
 
+  def list_from_shapes(shapes, stops, %Route{id: "CR-Fairmount"} = route, 0) do
+    mainline = Enum.find(shapes, &(&1.name == "South Station - Readville via Fairmount"))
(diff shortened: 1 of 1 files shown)


apps/stops/lib/route_stop.ex
```diff
@@ -99,6 +99,30 @@ defmodule Stops.RouteStop do
...
+
+    distinct_stop_ids =
+      mainline.stop_ids |> Enum.concat(foxboro_extension.stop_ids) |> Enum.uniq()
+
+    [do_list_from_shapes(mainline.name, distinct_stop_ids, stops, route)]
+    |> merge_branch_list(0)
+  end
+
+  def list_from_shapes(shapes, stops, %Route{id: "CR-Fairmount"} = route, 1) do
+    mainline = Enum.at(shapes, 0)
```
phildarnowsky: note to self: don't rely on the order, no reason to assume it won't change, instead look at the data

PR #246, opened 2019-10-09
Foxboro | Frank in CR-Fairmount timetable

meagonqz:
#### Summary of changes
**Asana Ticket:** [Foxboro | Fairmount timetable needs FRANK](https://app.asana.com/0/555089885850811/1143031265509300)
Added FRANK notation to timetable for Foxboro pilot

<br>
Assigned to: @amaisano

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
@@ -140,7 +140,7 @@ defmodule SiteWeb.ScheduleController.TimetableController do
     case ScheduleView.is_atleast_oct_21_2019?(date) do
       true ->
         %{
-          {"726"} => ScheduleView.timetable_note(%{route: route, direction_id: 1}),
+          {"726"} => ScheduleView.timetable_note(%{route: route, direction_id: 1, date: date}),
           {"726", "place-FS-0049"} => "FRANK"
         }
 
--- a/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
@@ -109,7 +109,7 @@ defmodule SiteWeb.ScheduleController.TimetableController do
   end
 
   def trip_messages(%Routes.Route{id: "CR-Franklin"}, 1, date) do
-    case is_atleast_oct_21_2019(date) do
+    case ScheduleView.is_atleast_oct_21_2019?(date) do
       true ->
         ["740", "746", "748", "750", "754", "726", "758"]
         |> Enum.flat_map(&franklin_via_fairmount(&1, 1))
@@ -125,7 +125,7 @@ defmodule SiteWeb.ScheduleController.TimetableController do
   end
 
   def trip_messages(%Routes.Route{id: "CR-Franklin"}, 0, date) do
-    case is_atleast_oct_21_2019(date) do
+    case ScheduleView.is_atleast_oct_21_2019?(date) do
       true ->
         ["741", "743", "747", "749", "755", "757", "759"]
         |> Enum.flat_map(&franklin_via_fairmount(&1, 0))
@@ -136,10 +136,13 @@ defmodule SiteWeb.ScheduleController.TimetableController do
     end
   end
 
-  def trip_messages(%Routes.Route{id: "CR-Fairmount"}, 1, date) do
-    case is_atleast_oct_21_2019(date) do
+  def trip_messages(%Routes.Route{id: "CR-Fairmount"} = route, 1, date) do
+    case ScheduleView.is_atleast_oct_21_2019?(date) do
       true ->
-        %{{"726", "place-FS-0049"} => "FRANK"}
+        %{
+          {"726"} => ScheduleView.timetable_note(%{route: route, direction_id: 1}),
+          {"726", "place-FS-0049"} => "FRANK"
+        }
 
       false ->
         %{}
@@ -148,14 +151,6 @@ defmodule SiteWeb.ScheduleController.TimetableController do
 
   def trip_messages(_, _, _) do
     %{}
-  end
-
-  def is_atleast_oct_21_2019(date) do
-    case Date.compare(date, ~D[2019-10-21]) do
-      :lt -> false
-      :eq -> true
-      :gt -> true
-    end
   end
 
   defp franklin_via_fairmount(train, 1) do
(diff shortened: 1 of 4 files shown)


apps/site/lib/site_web/templates/schedule/_timetable_note.html.eex
```diff
@@ -0,0 +1,2 @@
+<div>
```
meagonqz: Didn't mean to commit this, will remove.

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
```diff
@@ -67,49 +70,62 @@ defmodule SiteWeb.ScheduleController.TimetableController do
...
 
-  defp trip_messages(%Routes.Route{id: "CR-Franklin"}, 1) do
-    %{
-      {"790", "place-rugg"} => "Via",
-      {"790", "place-bbsta"} => "Fairmount"
-    }
+  def trip_messages(%Routes.Route{id: "CR-Fairmount"}, 1, date) do
+    case is_atleast_oct_21_2019(date) do
+      true ->
+        %{{"726", "place-FS-0049"} => "FRANK"}
```
meagonqz: @ingridpierre I am starting with the other PR for the accessible labels but will get to this soon. I plan to put the whole schedule note from below in this cell as sr-only.

apps/site/lib/site_web/views/schedule_view.ex
```diff
@@ -469,4 +469,22 @@ defmodule SiteWeb.ScheduleView do
       []
     end
   end
+
+  @spec timetable_note(Conn.t()) :: Safe.t()
+  def timetable_note(%{route: %Route{id: "CR-Fairmount"}}) do
```
amaisano: Should we avoid putting this note under the timetable for outbound? There is no FRANK cell for outbound.

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
```diff
@@ -67,49 +70,62 @@ defmodule SiteWeb.ScheduleController.TimetableController do
...
+    end
   end
 
-  defp trip_messages(_, _) do
+  def trip_messages(_, _, _) do
     %{}
   end
 
+  def is_atleast_oct_21_2019(date) do
+    case Date.compare(date, ~D[2019-10-21]) do
```
amaisano: Could be rewritten as `do: Date.compare(date, ~D[2019-10-21]) != :lt`

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
amaisano: This is purple. In the paper schedule, that matches, but in @ingridpierre 's design, it's in the default blue color. Not sure if that's an issue or if things have changed since the design was mocked.

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
```diff
@@ -67,49 +70,62 @@ defmodule SiteWeb.ScheduleController.TimetableController do
     end
   end
 
-  @spec trip_messages(Routes.Route.t(), 0 | 1) :: %{{String.t(), String.t()} => String.t()}
-  defp trip_messages(%Routes.Route{id: "CR-Haverhill"}, 0) do
+  @spec trip_messages(Routes.Route.t(), 0 | 1, Date.t()) :: %{
```
amaisano: Is `Date.t()` the same as `Calendar.date()`?

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
```diff
@@ -54,7 +54,10 @@ defmodule SiteWeb.ScheduleController.TimetableController do
     |> assign(:trip_schedules, trip_schedules)
     |> assign(:vehicle_schedules, vehicle_schedules)
     |> assign(:prior_stops, prior_stops)
-    |> assign(:trip_messages, trip_messages(conn.assigns.route, conn.assigns.direction_id))
+    |> assign(
+      :trip_messages,
+      trip_messages(conn.assigns.route, conn.assigns.direction_id, conn.assigns.date)
```
amaisano: looks like this update may have broken a test at `test/site_web/controllers/schedule_controller_test.exs:100`

apps/site/lib/site_web/views/schedule_view.ex
meagonqz: I wasn't sure. The ticket/design didn't indicate. I could do that, @ingridpierre / @katrinaml ?

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
meagonqz: The VIA messages are all purple as it stands and that matches paper schedule, so I assumed it should be the same. 
Up to @ingridpierre, would be a bit more work though to change this.

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
meagonqz: https://hexdocs.pm/elixir/Calendar.html seems to be an interoperability API.

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
meagonqz: True, I prefer the case syntax in elixir code for conditions instead of a !=

I could do the case with a default _ arg for eq/gt but decided not to.

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
meagonqz: I fixed this by excluding the franklin/fairmount from these tests since they are date dependent (and would start failing on Oct 21 or would have to be hardcoded to a date, both seem meh) and I already covered the date changing case in another test.

apps/site/lib/site_web/views/schedule_view.ex
ingridpierre: Makes sense to me - I'll take note of this since it's likely a bit brittle with rating changes.

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
ingridpierre: We can keep the purple, they're not links and they don't activate tooltips so this happens to be something of an an improvement.

ingridpierre:
Similar comment to the other foxboro PR "Via Franklin Line" should be in the sr-only text.

PR #247, opened 2019-10-09
Tighten up CR Fare Finder and Foxboro Special Event fare

amaisano:
#### Summary of changes
**Asana Ticket:** [Exclude Zone 4 fares from CR Fare Finder prior to Foxboro Pilot launch date](https://app.asana.com/0/555089885850811/1144031495137785)

Follow on to https://github.com/mbta/dotcom/pull/239
- place special events fare first
- change label of fare to include "Special Event" verbiage
- link Gillette stadium to destination page
- correct media information for $20 fare (new `:special_event` media ticket)
- date-gate Zone 4 fares prior to Foxboro Pilot launch

image

### DIFF
--- a/apps/site/lib/site_web/controllers/fare/commuter.ex
+++ b/apps/site/lib/site_web/controllers/fare/commuter.ex
@@ -50,6 +50,6 @@ defmodule SiteWeb.FareController.Commuter do
   defp foxboro?(a, b) when @foxboro in [a, b], do: true
   defp foxboro?(_, _), do: false
 
-  @spec foxboro_pilot?(Calendar.date()) :: boolean
+  @spec foxboro_pilot?(Date.t()) :: boolean
   defp foxboro_pilot?(current_date), do: Date.compare(current_date, @pilot_launch_date) != :lt
 end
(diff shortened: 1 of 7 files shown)


apps/site/lib/site_web/controllers/fare/commuter.ex
```diff
@@ -45,4 +49,7 @@ defmodule SiteWeb.FareController.Commuter do
   @spec foxboro?(String.t(), String.t()) :: boolean()
   defp foxboro?(a, b) when @foxboro in [a, b], do: true
   defp foxboro?(_, _), do: false
+
+  @spec foxboro_pilot?(Calendar.date()) :: boolean
```
meagonqz: I see why you asked the question on the other PR. I think technically either is OK. I don't see this type in use elsewhere in our codebase but Date.compare would work with it. We expect that this date object is actually a date like ~D[2019-10-21] not a datetime or any other type since we assign it in https://github.com/mbta/dotcom/blob/master/apps/site/lib/site_web/router.ex#L23 as Timex.to_date or Util.service_date

PR #248, opened 2019-10-09
Filter out Franklin shape that was breaking line diagram

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [🦊 🕸️❗Foxboro | Franklin line diagram is very wrong](https://app.asana.com/0/555089885850811/1142034065520053)

The problem was an extraneous shape; filtering it out restores the line diagram to its proper form.

<br>
Assigned to: @amaisano

### DIFF
--- a/apps/stops/lib/route_stop.ex
+++ b/apps/stops/lib/route_stop.ex
@@ -125,6 +125,18 @@ defmodule Stops.RouteStop do
     |> merge_branch_list(1)
   end
 
+  def list_from_shapes(
+        shapes,
+        stops,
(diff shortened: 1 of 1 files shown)


phildarnowsky:
@amaisano going to rebase this and force-push momentarily, FYI

amaisano:
@phildarnowsky is there a test suite for the line diagram that could benefit from an addition of making sure this route is filtered out / there are the correct # of branches?

phildarnowsky:
@amaisano there's not a test currently for that, no--I plan to do so when we come back and "do this right", if we do

PR #250, opened 2019-10-15
Line diagram alerts

amaisano:
#### Summary of changes
**Asana Ticket:** [Schedules | Line Diagram - Stop Component - Alert Icon](https://app.asana.com/0/555089885850811/1141229989083701)

If a stop has any high priority alerts, print the alert icon similar to Trip Planner itinerary. This is just for placement. Final styling to be done on a later pass, along with other stop card elements.

image

### DIFF
--- a/apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
+++ b/apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
@@ -249,6 +249,252 @@ export const lineDiagram = [
       branch: null
     },
     alerts: []
+  },
+  {
+    stop_data: [{ type: "terminus", branch: null }],
+    route_stop: {
+      zone: null,
+      stop_features: ["bus", "access", "parking_lot"],
+      station_info: {
+        type: "station",
+        "station?": true,
+        platform_name: null,
+        platform_code: null,
+        parking_lots: [
+          {
+            utilization: null,
+            payment: {
+              monthly_rate: "Monthly passes are not available at MBTA garages.",
+              mobile_app: null,
+              methods: ["Tap Card", "Credit/Debit Card", "Cash"],
+              daily_rate:
+                "Mon-Fri: $9 | Sat-Sun: $3 (Prices are valid for up to 14 hours of parking. After the first 14 hours, the daily rate is $15.)"
+            },
+            note: null,
+            name: "Alewife Garage",
+            manager: {
+              url: "http://www.parkmbta.com/",
+              phone: "617-222-3200",
+              name: "Republic Parking System",
+              contact: "MBTA Customer Support"
+            },
+            longitude: -71.142483,
+            latitude: 42.395428,
+            capacity: {
+              type: "Garage",
+              total: 2471,
+              overnight: "Available",
+              accessible: 32
+            },
+            address: null
+          }
+        ],
+        parent_id: null,
+        note: null,
+        name: "Alewife",
+        longitude: -71.142483,
+        latitude: 42.395428,
+        "is_child?": false,
+        id: "place-alfcl",
+        "has_fare_machine?": true,
+        "has_charlie_card_vendor?": true,
+        fare_facilities: ["fare_media_assistant", "fare_vending_machine"],
+        description: null,
+        closed_stop_info: null,
+        child_ids: [
+          "141",
+          "70061",
+          "Alewife-01",
+          "Alewife-02",
+          "door-alfcl-alewife",
+          "door-alfcl-busway",
+          "door-alfcl-cambridgepark",
+          "door-alfcl-russell",
+          "door-alfcl-steel"
+        ],
+        bike_storage: ["bike_storage_cage"],
+        address:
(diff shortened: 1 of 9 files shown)


apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -1,12 +1,26 @@
...
 
 interface Props {
   lineDiagram: LineDiagramStop[];
 }
 
+const maybeAlert = (alerts: Alert[]): ReactElement<HTMLElement> | null => {
+  const highPriorityAlerts = alerts.filter(alert => alert.priority === "high");
+  if (highPriorityAlerts.length === 0) return null;
+  return (
+    <i
```
meagonqz: I know this is styling which you said is to come in another pass, but can we start out with the alert svg icon instead of using font icons? There's already a helper for it. https://github.com/mbta/dotcom/blob/3ceb65e08aa2beef103db8df8e8b67354a91816f/apps/site/assets/ts/helpers/icon.tsx#L23
And leave out the tooltip for now, add a sibling "sr-only" label of "service alert or delay".

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -21,7 +24,17 @@ defmodule SiteWeb.ScheduleController.LineApi do
         stops_by_route_fn: &StopsRepo.by_route/3
       })
 
-    json(conn, Enum.map(line_data, &update_route_stop_data/1))
+    conn =
+      conn
+      |> assign(:route, LineHelpers.get_route(route_id))
+      |> assign_alerts([])
```
meagonqz: I forgot this helper existed!

It looks like we might want to filter by direction here. I don't think direction_id is set in the conn here like it is in the query params. If it's not already set in the conn, then we should do something like you did for `:route` and then the helper should filter by direction.

https://github.com/mbta/dotcom/blob/4014bdd53b6a93d3ab8e3c6c7b8826af3eb568c9/apps/site/lib/site_web/controllers/helpers.ex#L86

apps/site/lib/site_web/controllers/schedule/line_controller.ex
```diff
@@ -84,15 +87,19 @@ defmodule SiteWeb.ScheduleController.LineController do
...
+        line_diagram:
+          Enum.map(conn.assigns.all_stops, fn stop ->
+            update_route_stop_data(stop, conn.assigns.alerts, conn.assigns.date)
+          end)
       }
     )
   end
 
-  def update_route_stop_data({data, %RouteStop{} = map}) do
+  def update_route_stop_data({data, %RouteStop{id: stop_id} = map}, alerts, date) do
```
meagonqz: We should make this a shared function so that someone else doesn't miss the line api version. 

I'd add update_route_stop_data to LineApi and import it here since line api is more specific and more likely to be long lived. You could instead make it a shared module RouteStopData to hold this function. I would personally hold off making a module since I suspect the structure of the data will change as we flesh out this new line diagram. up to you!

apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
```diff
@@ -63,7 +63,8 @@ export const lineDiagram = [
       ],
       closed_stop_info: null,
       branch: null
-    }
+    },
+    alerts: []
```
meagonqz: We should add an example that has an alert. To get some alert data, you can grab the data from React DevTools (which I can show you in person), or you could hit the line api in your browser.

apps/site/assets/ts/schedule/components/LineDiagram.tsx
amaisano: What about using `renderSVG("c-svg__icon-alerts-triangle", alertIcon)`?

meagonqz: That would produce the same thing as `alertIcon("c-svg__icon-alerts-triangle")`. I think renderSVG came before the extra icon helpers, it would be easier to change all of the alert icons at once with alertIcon(className)...but we already have exceptions so it doesn't matter.

https://github.com/mbta/dotcom/blob/3ceb65e08aa2beef103db8df8e8b67354a91816f/apps/site/assets/ts/helpers/icon.tsx#L23-L24

apps/site/lib/site_web/controllers/schedule/line_controller.ex
```diff
@@ -181,7 +178,7 @@ defmodule SiteWeb.ScheduleController.LineController do
 
   defp tab_name(conn, _), do: assign(conn, :tab, "line")
 
-  defp alerts(conn, _), do: assign_alerts(conn, [])
+  defp alerts(conn, _), do: assign_alerts(conn, filter_by_direction?: true)
```
meagonqz: This filters by direction by default if direction_id is available but I don't mind this explicitness.

amaisano: Oh, haha.  ¯\_(ツ)_/¯

PR #254, opened 2019-10-18
Style basics of stop components, fix presentational bugs

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [⏱Schedules | Line Diagram - Stop Component Style](https://app.asana.com/0/555089885850811/1141229989083699)

Added spacing between feature icons and connection icons. Also fixed the following presentational bugs:

* Multiple CR/ferry icons were appearing at stops, whereas only one is desired no matter how many lines connect there.
* Silver Line connections now uses pills rather than icons.
* If a stop had both bus and non-bus connections, the bus connections would get stretched vertically. Fixed.
<br>
Assigned to: @amaisano

### DIFF
--- a/apps/site/assets/ts/schedule/components/LineDiagram.tsx
+++ b/apps/site/assets/ts/schedule/components/LineDiagram.tsx
@@ -60,21 +60,25 @@ const LineDiagram = ({
             <div className="m-schedule-line-diagram__connections">
               {filteredConnections(routeStop.connections).map((route: Route) =>
                 route.type === 3 ? (
-                  <span
-                    key={route.id}
-                    className={`c-icon__bus-pill m-schedule-line-diagram__connection ${busBackgroundClass(
-                      route
-                    )}`}
-                  >
-                    {route.name}
-                  </span>
+                  <a href={`/schedules/${route.id}/line`}>
+                    <span
+                      key={route.id}
+                      className={`c-icon__bus-pill m-schedule-line-diagram__connection ${busBackgroundClass(
+                        route
+                      )}`}
+                    >
+                      {route.name}
+                    </span>
+                  </a>
                 ) : (
-                  <span
-                    key={route.id}
-                    className="m-schedule-line-diagram__connection"
-                  >
-                    {modeIcon(route.id)}
-                  </span>
+                  <a href={`/schedules/${route.id}/line`}>
+                    <span
+                      key={route.id}
+                      className="m-schedule-line-diagram__connection"
+                    >
+                      {modeIcon(route.id)}
+                    </span>
+                  </a>
                 )
               )}
             </div>
--- a/apps/site/assets/ts/schedule/components/LineDiagram.tsx
+++ b/apps/site/assets/ts/schedule/components/LineDiagram.tsx
@@ -50,13 +50,12 @@ const LineDiagram = ({
 }: Props): ReactElement<HTMLElement> | null => {
   const routeType = lineDiagram[0].route_stop.route!.type;
 
-  return(
+  return (
     <>
       <h3>
-        {routeType === 0 || routeType === 1 || routeType === 2 ?
-          "Stations" :
-          "Stops"
-        }
+        {routeType === 0 || routeType === 1 || routeType === 2
+          ? "Stations"
+          : "Stops"}
       </h3>
       {lineDiagram.map(
         ({ route_stop: routeStop, alerts: stopAlerts }: LineDiagramStop) => (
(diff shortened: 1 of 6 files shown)


phildarnowsky:
note to self: needs linting

ingridpierre:
Seeing the following server error on /schedules/CR-Fairmount/line on this branch:
```
: UndefinedFunctionError at GET /schedules/CR-Fairmount/line
function nil.stop_ids/0 is undefined
```

phildarnowsky:
@ingridpierre are you still getting that error? it looks fine on my machine, perhaps we could do a screenshare to diagnose it on yours if necessary.

ingridpierre:
1. Can we remove the dreaded double border here - would a negative 1px vertical margin work?
image
----
2. Looks like there's still a good bit of work to do for Green Line/Branches as well, not seeing connection icons that I would expect. We should separately estimate that work so it doesn't hold up this ticket.
----
3. Let's **only show those purple zone tags on CR line pages**, zones aren't applicable to other modes so we can save a lot of space and noise if we lose em.
----
4. The tap targets are pretty small, but I still think a**ll the connections icons should be links** to the corresponding /schedules/X/line tab.
----
5. Semantically I think this page is potentially easier to traverse for screen readers if we **turn the stop names into headers (h4)**. It will help with your ability to scan stop names with all those connection icons. Since we shouldn't skip headers in the hierarchy **let's also add an h3 above the stop list list labeled "Stops" (for bus and ferry) or "Stations" (for CR and Subway).** Example:
image
----
6. Will chat with you offline about what will happen once realtime information is added to this. Mobile is quite crowded and I have an idea for what to do there.

ingridpierre:
7. Too much padding below the first row of icons, should be same for each row that wraps:
IMG_7081

ingridpierre:
8. The size of the connection and station feature icons can probably be significantly reduced.
See: https://www.mbta.com/stops/subway to see a similar treatment at 1rem size. Can we have the bus pulls, and mode connection icons shrink a bit?

ingridpierre:
9. Looks like ferry stops aren't really showing at all, OK if we have to investigate that separately, but want to make sure it doesn't get missed.

phildarnowsky:
@ingridpierre re #2: is any of your Green Line concern perhaps covered by the ticket Ryan is doing or is this separate issues?

phildarnowsky:
@ingridpierre re #6: maybe let's talk Thursday just after/in place of design review?

phildarnowsky:
@ingridpierre re #9: could you expand a little on what you mean?

phildarnowsky:
@ingridpierre latest version in dev-green responds to your feedback, except for those points I had questions on above.

ingridpierre:
Changes look good. Small adjustment to bus pills, they should get a tiny bit more padding to match the height of the other icons. Currently the pills are a little short:

**CURRENT:**
image


**DESIRED:**
image

ingridpierre:
> @ingridpierre re #9: could you expand a little on what you mean?

added it as a separate ticket here: https://app.asana.com/0/477545582537986/1146873289024569/f

amaisano:
Better @ingridpierre? 
image

ingridpierre:
Based on your screenshot I think this is set from a design perspective. Thanks @amaisano

apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -12,6 +12,24 @@ interface Props {
...
+  const firstFerryIndex = connections.findIndex(
+    connection => connection.type === 4
+  );
+
+  return connections.filter(
+    (connection, index) =>
+      (connection.type !== 2 && connection.type !== 4) ||
+      (connection.type === 2 && index === firstCRIndex) ||
+      (connection.type === 4 && index === firstFerryIndex)
+  );
```
amaisano: This is for de-duping similar mode icons, right?

apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -24,72 +42,82 @@ const maybeAlert = (alerts: Alert[]): ReactElement<HTMLElement> | null => {
...
+                {maybeAlert(stopAlerts)}
+                <a href={`/stops/${routeStop.id}`}>
+                  <h4>{routeStop.name}</h4>
+                </a>
+              </div>
+              <div className="m-schedule-line-diagram__connections">
+                {filteredConnections(routeStop.connections).map(
+                  (route: Route) =>
+                    route.type === 3 ? (
+                      <a href={`/schedules/${route.id}/line`}>
```
amaisano: The parent `<a>` tag could be lifted up as a shared parent. Seems the props on it are the same in either condition.

apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -24,72 +42,82 @@ const maybeAlert = (alerts: Alert[]): ReactElement<HTMLElement> | null => {
...
+                      "c-svg__icon-parking-default m-schedule-line-diagram__feature-icon"
+                    )
+                  : null}
+                {routeStop.stop_features.includes("access")
+                  ? accessibleIcon(
+                      "c-svg__icon-acessible-default m-schedule-line-diagram__feature-icon"
+                    )
+                  : null}
+                {routeStop.route!.type === 2 && routeStop.zone && (
+                  <span className="c-icon__cr-zone m-schedule-line-diagram__feature-icon">{`Zone ${
```
amaisano: There seems to be a greedy CSS issue happening here which is crushing Zone pills into squares. See image below:

image

amaisano: I think the font size for the zone needs to come down, too.

apps/site/assets/ts/schedule/components/LineDiagram.tsx
phildarnowsky: Correct. The rule is that, if there are ferry or CR connections, we want to display exactly one mode icon, regardless of how many connections.

phildarnowsky:
@amaisano the two latest commits should address your feedback

PR #258, opened 2019-10-30
Update docs based on onboarding experience

digitalcora:
README changes:

* Specify that an API key should be obtained from `dev.api.mbtace.com` rather than the production site

* Move Java install to the testing docs as it seems this is only needed for Wiremock, which itself is only needed for Backstop

* Add workaround needed for compiling Erlang on OS X 10.15

* Remove manual Caskroom setup — Homebrew does this automatically

* Specify that Chrome must be explicitly installed to use ChromeDriver since it is not auto-installed as a dependency

ENVIRONMENT changes:

* Remove mention of Google billing information since it appears this is not needed (when using an MBTA Google account anyway)

TESTING changes:

* Remove instruction to manually start `phx.server` for testing since all tests pass without doing this (was necessary before?)

* Remove mention of spurious Dialyzer "can never match" errors since these don't appear to be present anymore

* Add explicit Homebrew commands for installing Docker and Wiremock, including `--no-quarantine` flag required for Java on OS X 10.15+

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,25 @@
 [![Build Status](https://semaphoreci.com/api/v1/mbta/dotcom/branches/master/badge.svg)](https://semaphoreci.com/mbta/dotcom)
+
 # Dotcom
 
 The new face of https://www.mbta.com/
-  - [Getting Started](#getting-started)
-  - [Running the Server](#running-the-server)
-  - [Environment Variables](#environment-variables)
-    - [`V3_API_KEY`](#v3apikey)
-    - [`V3_URL`](#v3url)
-    - [`GOOGLE_API_KEY`](#googleapikey)
-    - [`DRUPAL_ROOT`](#drupalroot)
-    - [`ALGOLIA_APP_ID`, `ALGOLIA_SEARCH_KEY`, and `ALGOLIA_ADMIN_KEY`](#algoliaappid-algoliasearchkey-and-algoliaadminkey)
+
+- [Getting Started](#getting-started)
+- [Running the Server](#running-the-server)
+- [Environment Variables](#environment-variables)
+    - [`V3_API_KEY`](#v3_api_key)
+    - [`V3_URL`](#v3_url)
+    - [`GOOGLE_API_KEY`](#google_api_key)
+    - [`DRUPAL_ROOT`](#drupal_root)
+    - [`ALGOLIA_APP_ID`, `ALGOLIA_SEARCH_KEY`, and `ALGOLIA_ADMIN_KEY`](#algolia_app_id-algolia_search_key-and-algolia_admin_key)
     - [Making the variables available to the app](#making-the-variables-available-to-the-app)
-  - [Additional documentation](#additional-documentation)
+- [Additional documentation](#additional-documentation)
+
 ## Getting Started
 
-1. Request a V3 API key at https://api-v3.mbta.com/. Note that, at any given time, the site may not be compatible with the
-very latest API version. As of this writing, the site is compatible with API version 2019-04-05.
+1. Request a V3 API key at https://dev.api.mbtace.com/, and request an increased
+rate limit for it (someone with access will need to approve this). Note that, at
+any given time, the site may not be compatible with the very latest API version.
 
 1. Install [Homebrew](https://docs.brew.sh/Installation.html):
     ```
@@ -28,7 +32,6 @@ very latest API version. As of this writing, the site is compatible with API ver
 
      ```
      brew install coreutils automake autoconf openssl libyaml readline libxslt libtool unixodbc
-     brew cask install java
      ```
 
    * Add asdf plugins
@@ -48,6 +51,10 @@ very latest API version. As of this writing, the site is compatible with API ver
      ```
 
      If you run into problems, you might have to update the `import-release-team-keyring` script.
+
+   * If running OSX 10.15 Catalina, run `export MACOSX_DEPLOYMENT_TARGET=10.14`.
+     This works around a [known issue](https://github.com/asdf-vm/asdf-erlang/issues/116)
+     with compiling versions of Erlang prior to 22.1.4.
 
    * Run the install:
 
@@ -74,9 +81,10 @@ very latest API version. As of this writing, the site is compatible with API ver
 
 1. Install chromedriver (for Elixir acceptance tests using Wallaby)
     ```
-    brew tap caskroom/cask
     brew cask install chromedriver
     ```
+   Note: `chromedriver` requires Chrome to be installed. If you don't already
+   have it, `brew cask install google-chrome` is an easy way to install it.
 
 1. Install our Elixir dependencies. From the root of this repo:
     ```
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -1,16 +1,10 @@
 ## Tests
 
-Run `mix phx.server` to start a server in one window, then open a
-separate window and run `npm test` from the main folder. This will execute
-the following scripts in succession:
-
-* `mix test` — Phoenix tests
+* `mix test` — Elixir tests
 * `npm run test:js` — JS tests
 * `npm run backstop` — Backstop tests (see section below for details)
 
-Note that all tests will run even if one of them fails, and there is no final
-summary at the end, so pay attention to the output while they're running to
-make sure everything passes.
+`npm test` runs all of these in succession.
 
 ### Dialyzer
 
@@ -18,11 +12,6 @@ Dialyzer is a static analysis tool which looks at type information. We use it
 verify our type specifcations and make sure we're calling functions properly.
 
 * `mix dialyzer` — Runs the actual type checks.
-
-Currently, there are some spurious "The variable _ can never match ..."
-errors that can be ignored.
-`npm run dialyzer` will filter out these errors, though there are still a few
-which show up.
 
 ### Pronto
 
@@ -33,11 +22,8 @@ Installing Pronto on Max OS can be challenging because some of its dependencies 
 Follow these instructions to install:
 
 ```
-brew install cmake
-brew install pkg-config
-gem install pronto
-gem install pronto-scss
-gem install pronto-credo
+brew install cmake pkg-config
+gem install pronto pronto-scss pronto-credo
 ```
 
 Run it by calling `pronto run` in the `mbta/dotcom` directory. If there is no output, that means it passed.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1,21 +1,26 @@
...
+- [Additional documentation](#additional-documentation)
+
 ## Getting Started
 
-1. Request a V3 API key at https://api-v3.mbta.com/. Note that, at any given time, the site may not be compatible with the
-very latest API version. As of this writing, the site is compatible with API version 2019-04-05.
+1. Request a V3 API key at https://dev.api.mbtace.com/, and request an increased
+rate limit for it (someone with access will need to approve this). Note that, at
+any given time, the site may not be compatible with the very latest API version.
+As of this writing, the site is compatible with API version `2019-07-01`.
```
amaisano: Too specific - I'd leave this last line out.

README.md
```diff
@@ -70,12 +78,12 @@ very latest API version. As of this writing, the site is compatible with API ver
...
       ruby           <version> (set by ~/dotcom/.tool-versions)
      ```
-     
+
      If you are missing any versions, you should re-run `asdf install`. Related [Github issue about asdf-erlang](https://github.com/asdf-vm/asdf-erlang/issues/57)
 
 1. Install chromedriver (for Elixir acceptance tests using Wallaby)
     ```
-    brew tap caskroom/cask
+    brew cask install google-chrome # if you don't have Chrome already
```
amaisano: Can Chrome be acquired via browser download/installer or does it need to be explicitly added via brew?

README.md
```diff
@@ -84,11 +92,6 @@ very latest API version. As of this writing, the site is compatible with API ver
     mix deps.get
     ```
 
-1. Install npm globally
-   ```
-   npm install -g npm@6.7.0
```
amaisano: I think we need to explicitly all use this version. I recall lots of unintended package-lock changes due to people having the wrong or different versions in different commits. See https://github.com/mbta/dotcom-legacy/pull/2825

docs/TESTING.md
```diff
@@ -63,12 +49,24 @@ The tests are run against a live application, built in production mode. To make
...
+* Docker
+  * `brew cask install docker`
+  * Start Docker Desktop from the dock or Applications; this only has to be done
+    once, after which it will auto-start on login by default
+* Wiremock
+  * `brew cask install java --no-quarantine`
+    * This option is currently required on OS X 10.15+ due to Gatekeeper
+      changes. Ref: https://github.com/Homebrew/homebrew-cask/issues/70798
+  * `brew cask install wiremock-standalone`
+* Ensure `WIREMOCK_PATH` points to the Wiremock JAR file; with brew cask this
```
amaisano: Should this be listed as an envar or is this used inline with each run?

docs/TESTING.md
```diff
@@ -63,12 +49,24 @@ The tests are run against a live application, built in production mode. To make
...
+  will be something like `/usr/local/Cellar/wiremock-standalone/<VERSION>/libexec/wiremock-standalone-<VERSION>.jar`
 
-* Make sure docker is [installed](https://docs.docker.com/docker-for-mac/install/) and running: we run the tests in docker to ensure we get consistent images across platforms.
-* Run `npm run backstop` which starts up wiremock, a mocked phoenix server, and when those are up, kicks off backstop testing in docker
+Once all the above are in place: `npm run backstop`
```
amaisano: I personally haven't gotten this macro to work. I've always had to do the three-bash approach listed here: https://github.com/mbta/wiki/blob/56ab636371b94604b62446e2f032909d16205dd2/website/testing/backstop.md#starting-the-phoenix-server-and-wiremock-separately

README.md
amaisano: Node _may_ have fixed this since then, but not 100% sure.

README.md
digitalcora: The diff doesn't show it well, but this sentence is directly from the current README which also specifies a version, albeit an older one. I updated it with the reasoning that all of my tests passed when using the new version. If we want to delete the version entirely I'm in favor 🙂

README.md
digitalcora: It just needs to be installed, via any means. Perhaps we can safely leave out this instruction, I thought it might be good to specify somewhere that `chromedriver` needs Chrome to be installed manually since it doesn't list it as a dependency in Homebrew.

docs/TESTING.md
digitalcora: Ah, do you mean the wording of the sentence doesn't make it clear that `WIREMOCK_PATH` is in fact an environment variable and not some other kind of thing?

README.md
digitalcora: It just needs to be installed, via any means. Perhaps we can safely leave out this instruction, I thought it might be good to specify somewhere that `chromedriver` needs Chrome to be installed manually since it doesn't list it as a dependency in Homebrew.

README.md
digitalcora: I've just checked, and the version of `npm` (6.4.1) that ships with our current version of Node does not generate the lockfile any differently than `npm` 6.7.0, at least on my machine.

However they do _both_ generate the lockfile slightly differently than the current committed version; a few packages listed as `optional` become non-optional. The difference appears to be the same whether `npm` 6.4.1 or 6.7.0 is used. I'll PR this change in a separate branch if we think it's harmless (I think it likely is, from experience).

docs/TESTING.md
digitalcora: Hm! I wasn't aware this separate doc existed. Should the docs in this repo be cut down to the basics, and link to that doc for full details? Or vice-versa: should we integrate those docs into this repo? Of course since Backstop is currently broken and not active on CI anyway, perhaps it's a moot point. I made these changes before I knew that.

README.md
amaisano: I'm in favor of removing info that gets dated quickly.

README.md
amaisano: Ok. I would move it as a note prior to the command.

docs/TESTING.md
amaisano: Yes, and it doesn't seem to be listed in the envars section(s).

docs/TESTING.md
amaisano: I defer to @ryan-mahoney and @phildarnowsky as to where this actually lives. I'm not sure our current policy on the generic WIKI vs. repo-specific documentation.

README.md
amaisano: Update: The separate PR was closed for now. Let's keep this as is (explicitly requiring this version of npm), and revisit a little later on if new devs continue to experience discrepancies.

README.md
ryan-mahoney: I thought this was required to get Chrome to work headlessly... but may be mistaken

docs/TESTING.md
ryan-mahoney: @digitalcora since you are doing this now (and doing a great job!) I trust your judgment in terms of how to handle it. We don't have a policy

README.md
digitalcora: There shouldn't be any difference between installation methods; `brew cask` essentially automates the downloading and installing of the same file you'd get from the Chrome web site.

Will change this to a note!

docs/TESTING.md
digitalcora: It is listed in `ENVIRONMENT.md` if that's what you mean? This doc itself (`TESTING`) doesn't have a section about env vars.

docs/TESTING.md
digitalcora: I think it'd definitely be nice to get everything in one place; the existence of the `docs` directory in this repo (and absence of a link to the wiki) led me to think these were _the_ docs for the website project, when in reality many more docs existed in the wiki. Since the latter seems much more detailed and complete, I might advocate for merging this content into that, and having only the README with quick-start instructions here.

If that sounds good, I can either do that as a separate PR or revise this one.

digitalcora: New thought: a factor in this decision might be which docs we'd like to be public-facing and which we don't, given the wiki is a private repo and this one is open-source.

digitalcora: Although, there is the complication that the wiki is a private repo and this one is open-source, so we might want to keep some docs here if they would facilitate people outside the MBTA getting a fork of the website running.

docs/TESTING.md
phildarnowsky: FWIW my first guess on seeing an all-caps identifier in this context is that it would be an environment variable, but we could be clear while maintaining concision by saying "Ensure the environment variable `WIREMOCK_PATH`..."

docs/TESTING.md
phildarnowsky: I am of two minds about this. On the one hand, I strongly dislike Github wikis (not least because I can never find the damn things), and I also dislike having multiple sources of documentation split up by no particular logic. On the other hand, moving everything from the wiki to the `docs` directory sounds like make-work, and it might be good to have a few things we hold back from the general public.

So let's keep these docs here for now, and I will add a ticket to investigate ways to automate moving wiki docs into the `docs` directory.

docs/TESTING.md
```diff
@@ -1,16 +1,10 @@
 ## Tests
 
-Run `mix phx.server` to start a server in one window, then open a
-separate window and run `npm test` from the main folder. This will execute
-the following scripts in succession:
-
-* `mix test` — Phoenix tests
+* `mix test` — Elixir tests
```
phildarnowsky: @digitalcora We should also mention `test:jest`, which runs more UI-focused frontent tests, as well as test coverage.

docs/TESTING.md
digitalcora: @phildarnowsky Sorry, when I said "the wiki" I meant the `mbta/wiki` repository which Adamo linked in his first comment (which is not itself a GitHub Wiki™). I think we are all in agreement that docs should be a committed part of _some_ repo; the split right now is between the docs that live in `mbta/wiki` and the docs that live here in the `docs` directory.

docs/TESTING.md
digitalcora: I did change the wording here 👍

docs/TESTING.md
phildarnowsky: @digitalcora aha, my mistake. In that case let's continue not to mess with this for now and we'll discuss the question further at our next refinement.

docs/TESTING.md
digitalcora: The listed command `npm run test:js` runs both the Mocha and Jest tests. (but, this is only a reformatting of the info that was already here; should it be changed?)

phildarnowsky: on second thought NVM

digitalcora: Ok, then if your approval stands, I will merge this 😃

PR #260, opened 2019-10-31
Fix misaligned bus icon on Projects page

digitalcora:
**Asana Ticket:** [🐞 Bug | System icons misaligned on Project index](https://app.asana.com/0/477545582537986/1145601078296161/f)

In 0f80d0a a number of classnames that started with `c-` were renamed to start with `m-`, but we missed this one spot.

Fixed appearance:

Screen Shot 2019-10-31 at 3 46 09 PM

Note that since the bus icon now has the correct margins, it can sometimes wrap onto a second line when it didn't before:

Screen Shot 2019-10-31 at 3 47 07 PM

This seems okay to me but, backing up a bit, it seems odd that _only_ buses are represented by an icon here, where other transit modes are spelled out in text. The rest of the page (including the filter buttons at the top and the project list at the bottom) consistently uses icons for all transit modes. Maybe @ingridpierre can shed light on this decision.

### DIFF
--- a/apps/site/assets/ts/projects/__tests__/__snapshots__/FeaturedProjectsListTest.tsx.snap
+++ b/apps/site/assets/ts/projects/__tests__/__snapshots__/FeaturedProjectsListTest.tsx.snap
@@ -41,7 +41,7 @@ exports[`renders 1`] = `
           >
             <span
               aria-hidden="false"
-              className="notranslate c-svg__icon c-featured-project__route-icon"
+              className="notranslate c-svg__icon m-featured-project__route-icon"
               dangerouslySetInnerHTML={
                 Object {
                   "__html": "SVG",
--- a/apps/site/assets/ts/projects/__tests__/__snapshots__/ProjectUpdateListTest.tsx.snap
+++ b/apps/site/assets/ts/projects/__tests__/__snapshots__/ProjectUpdateListTest.tsx.snap
@@ -98,7 +98,7 @@ Array [
       >
         <span
           aria-hidden="false"
-          className="notranslate c-svg__icon c-featured-project__route-icon"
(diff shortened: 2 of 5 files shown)


digitalcora:
I'm not sure why a coverage change is being reported here — the line that supposedly lost coverage is in a Phoenix controller somewhere and doesn't relate to this change. Maybe it's a side effect of the tests being run against a live environment?

amaisano:
Coverage seems to have passed. We typically only concern ourselves with the DIFF tab: https://codecov.io/gh/mbta/dotcom/compare/f4f4da36123925cf1855a15bfdc2a67388048bae...4fe68eb8ffa8d9720326b33995ce4c6e4d62c814/diff

ingridpierre:
@digitalcora I totally understand the concern here, but it's intentional. There's actually a further perceived discrepancy I hear from other folks, which is that subway lines are described by their line and CR lines only get the mode "pill" shape.

This has to do with how our wayfinding standards IRL work, and also with space concerns. The bus icon is what is intended as often the related or affected buses on projects are too numerous to list. Wayfinding standards [are really clearly documented by that team](https://cdn.mbta.com/sites/default/files/engineering/001-design-standards-and-guidelines/2015-03-01-signage-system-overview.pdf) so the Designers tend to follow those when talking about the styling of route info, since they're established outside the department and closely reflect what customers see when they're not looking at their devices. 

**TL;DR:** why we use an icon rather than a pill that says "Bus" or something:

Here's an image of the wayfinding standards to show what this looks like out in the world. Subway line pills, followed by commuter rail mode pill, and bus icon to indicate bus connections.
image (8)

digitalcora:
Thank you @ingridpierre 🙌

PR #261, opened 2019-10-31
Commit current package-lock from `npm i`

digitalcora:
[(relevant discussion)](https://github.com/mbta/dotcom/pull/258#discussion_r341324581)

This is the package-lock generated by running our base `npm install` using NPM 6.4.1, which is the version that ships with Node 8.15.0, which is the version we have specified via asdf... at least, this is how it's generated on my machine, so I'd appreciate it if someone else could try an `npm i` on this branch and see if any changes show up!

The file is also generated the same way when I use NPM 6.7.0, the version recommended to manually install in the README.

This does not appear to involve the issue described in mbta/dotcom-legacy#2825.

### DIFF
--- a/apps/site/assets/package-lock.json
+++ b/apps/site/assets/package-lock.json
@@ -7049,8 +7049,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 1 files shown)


digitalcora:
Closing, as it turns out not everyone actually generates the same lockfile even when using identical versions of Node/NPM. This may be down to differences in exactly what already exists in your `node_modules` at the time you run the command. The diff appears harmless, at any rate, so we don't have to dig deep into this right now.

PR #263, opened 2019-11-05
Add basic antispam measure to contact form

digitalcora:
**Asana Ticket:** [Feedback Form | Address spam issues](https://app.asana.com/0/477545582537986/1145937042305257/f)

This adds an invisible field to the contact form that must _not_ be filled in for the submission to succeed. The image below has the field un-hidden for illustrative purposes (this is also how a screen reader might "see" it, which is why the label is important).

Screen Shot 2019-11-05 at 11 57 43 AM

---

I was unable to get Wallaby testing working for this — I could be missing something, but there seems to be an issue with the library where telling it to fill in a non-visible field just doesn't work, but acts as though it worked. I used this code:

```elixir
@tag :wallaby
test "Catches spam bots that fill every available field", %{session: session} do
  session
  |> visit("/customer-support")
  |> fill_in(css("#comments"), with: "Support Form Integration Test")
  |> click(css("label[for=\"service-Suggestion\"]"))
  |> fill_in(css("#leave-this-alone", visible: false), with: "spam!")
  |> click(@submit_button)
  |> assert_text("We had an issue submitting your feedback")
end
```

This "succeeds" at filling in the field, but fails on the `assert_text`, and inspecting the `conn` reveals that the field has an empty value as though it wasn't filled in.

### DIFF
--- a/apps/site/lib/site_web/controllers/customer_support_controller.ex
+++ b/apps/site/lib/site_web/controllers/customer_support_controller.ex
@@ -92,12 +92,13 @@ defmodule SiteWeb.CustomerSupportController do
         [
           &validate_comments/1,
           &validate_service/1,
+          &validate_antispam/1,
           &validate_name/1,
           &validate_email/1,
           &validate_privacy/1
         ]
       else
-        [&validate_comments/1, &validate_service/1]
+        [&validate_comments/1, &validate_service/1, &validate_antispam/1]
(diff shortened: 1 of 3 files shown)


phildarnowsky:
@digitalcora Looks good to me, though for an example I would have gone with "You're not my real dad."

PR #264, opened 2019-11-07
Fix missing retail location vendors

amaisano:
#### Summary of changes
**Asana Ticket:** [retail sales listings should include vendor name](https://app.asana.com/0/555089885850811/1145005926601595)

V3 API change: use updated key for location name (`long_name`) -- `name` no longer exists.

**BEFORE:**
image

**AFTER:**
image

### DIFF
--- a/apps/fares/test/retail_locations_data_test.exs
+++ b/apps/fares/test/retail_locations_data_test.exs
@@ -20,6 +20,12 @@ defmodule Fares.RetailLocationsDataTest do
       end
     end
 
+    test "all locations have a non-nil name" do
+      for %Location{name: name} <- Data.get() do
+        assert is_binary(name)
+      end
+    end
+
     test "build_r_tree returns a tree with all location data" do
       tree = Data.build_r_tree()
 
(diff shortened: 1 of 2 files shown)


apps/fares/test/retail_locations_data_test.exs
```diff
@@ -20,6 +20,12 @@ defmodule Fares.RetailLocationsDataTest do
       end
     end
 
+    test "all locations have a non-nil name" do
+      for %Location{name: name} <- Data.get() do
+        assert is_binary(name)
+      end
+    end
```
digitalcora: I know it's usually good advice to write a failing test before fixing a bug, and this test would have failed in this specific scenario... but I feel like we'll probably never see _this_ specific scenario again, and that's all the test covers. It's also testing the output of a live external service, which has its own issues of making the test suite a bit slower and more unreliable. I'd advocate for not adding this, but would like to hear what y'all think.

(in terms of _process_ that would have prevented this bug, did we check our code against the [changelog](https://www.mbta.com/developers/v3-api/changelog) before moving to the new API version? it does call out this `name` change)

amaisano: Our pacing with the GTFS Creator repo is not always in sync. We could indeed improve our protocol between teams here to avoid this issue in the future. @phildarnowsky and I have discussed having the API team run our dotcom tests, but beyond that a changelog review would be great.

paulswartz: You don't have to update to new API versions: it's why we maintain the old ones rather than making arbitrary changes.

amaisano: @phildarnowsky where/when do API version updates happen on prod?

PR #265, opened 2019-11-07
Add server check for support form uploads

digitalcora:
The client side already has validation for trying to attach non-images, so this is mostly an anti-spambot measure to prevent people from throwing viruses into the ticketing system (even though this would require a support person to intentionally download and open them).

The `filename` and `mime_type` of a Plug.Upload are both provided by the user and thus untrusted, but since we ultimately attach the upload to an email using its `filename`, that is what will control how the recipient's computer tries to open it.

**Asana Ticket:** [Feedback Form | Address file type security issue](https://app.asana.com/0/477545582537986/1148399311722825/f)

### DIFF
--- a/apps/site/lib/site_web/controllers/customer_support_controller.ex
+++ b/apps/site/lib/site_web/controllers/customer_support_controller.ex
@@ -2,6 +2,8 @@ defmodule SiteWeb.CustomerSupportController do
   use SiteWeb, :controller
   require Logger
 
+  @allowed_attachment_types ~w(image/bmp image/gif image/jpeg image/png image/tiff image/webp)
+
   plug(Turbolinks.Plug.NoCache)
   plug(:set_service_options)
   plug(:assign_ip)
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -137,6 +138,19 @@ defmodule SiteWeb.CustomerSupportController do
...
   defp validate_privacy(_), do: "privacy"
 
+  @spec validate_photos(map) :: :ok | String.t()
+  defp validate_photos(%{"photos" => photos}) when is_list(photos) do
+    if Enum.all?(photos, &valid_upload?/1), do: :ok, else: "photos"
+  end
+
+  defp validate_photos(_), do: :ok
+
+  @allowed_attachment_types ~w(image/bmp image/gif image/jpeg image/png image/tiff image/webp)
```
phildarnowsky: we customarily put module attributes at the head of the file, right after the `use/require/import/alias` stuff--in this case that would go on line 4, after the line reading `require Logger` but before the first call to `plug`

digitalcora: Makes sense! Wasn't sure what the preference was — I'd have expected Credo to complain, but maybe it doesn't handle this yet.

PR #266, opened 2019-11-12
Fix Jest watch mode not working

digitalcora:
I was on my way to write up a tech-debt Asana ticket for this issue, but wondered if it could be as simple as something being outdated. Turns out, it was! (the specific crash was [this one](https://github.com/facebook/jest/issues/8084)) This only took a few minutes to address.

For ease of access, this also adds a `npm run test:jest:watch` script. The default mode only runs tests related to files that have changed.

### DIFF
--- a/apps/site/assets/package-lock.json
+++ b/apps/site/assets/package-lock.json
@@ -1480,83 +1480,1572 @@
       }
     },
     "@jest/core": {
-      "version": "24.3.1",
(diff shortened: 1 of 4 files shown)


digitalcora:
This is the Jest change I mentioned in standup that removed test setup files from the coverage stats: https://github.com/facebook/jest/pull/7790

PR #267, opened 2019-11-13
Add (empty) shuttles page, tab for same.

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Shuttles | Add tab to RL, OL, and GL schedules](https://app.asana.com/0/555089885850811/1148592031135764)

Added a "Shuttles" page to the rapid-transit schedules. Currently it only contains a blank layout, ready to populate. The tab for this page only appears if (a) you have a flag set and (b) there is an alert for the line in question.

Note to reviewer: I am punting for the moment on testing the tab logic. This is because the business logic is not yet clear around whether and/or when we should show this page when there's an active but _upcoming_ shuttle alert, and also because I would like it merged ASAP so that we have a common template to add our various features to. I will add tests in a separate ticket that I'll circle back to before this goes to production.

<br>
Assigned to: @digitalcora


apps/site/lib/site_web/controllers/schedule/shuttles_controller.ex
```diff
@@ -0,0 +1,23 @@
...
+  @moduledoc false
+
+  use SiteWeb, :controller
+
+  alias SiteWeb.ScheduleView
+
+  plug(SiteWeb.Plugs.Route)
+  plug(SiteWeb.ScheduleController.Defaults)
+  plug(:alerts)
+  plug(SiteWeb.Plugs.AlertsByTimeframe)
```
digitalcora: The mix of patterns here is a bit hard to follow — these two plugs are related and one depends on the other, but one is implemented as a `Plug` and one as a call to a shared helper function. I wonder if it would be much work to make `assign_alerts` into a `Plug`? It only seems to be used in a few places.

apps/site/lib/site_web/controllers/schedule/green.ex
```diff
@@ -55,6 +55,13 @@ defmodule SiteWeb.ScheduleController.Green do
     |> render("show.html", [])
   end
 
+  def shuttles(conn, _params) do
+    conn
+    |> assign(:tab, "shuttles")
+    |> put_view(ScheduleView)
+    |> render("show.html", [])
+  end
```
digitalcora: Any particular reason we're implementing this page two different ways? For the green line it's a function added to the existing GL schedule controller (here), but for other lines it's a standalone controller.

apps/site/assets/ts/shuttles-entry.ts
```diff
@@ -0,0 +1,3 @@
+import shuttles from "./shuttles/shuttles-loader";
+
+shuttles();
```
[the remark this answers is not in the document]
phildarnowsky: I actually am not sure either on why we split it up this way, @ryan-mahoney do you know what the purpose of that is?

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { ReactElement } from "react";
```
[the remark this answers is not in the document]
phildarnowsky: @digitalcora when I remove `React` from the import list, I get the following error on line 5, which is the single line where we have TSX markup: `'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.`

apps/site/assets/ts/shuttles/shuttles-loader.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { render } from "react-dom";
+import ShuttlesPage from "./components/ShuttlesPage";
+
+const shuttles = (): void => {
+  const shuttlesRootEl = document.getElementById("react-projects-root");
```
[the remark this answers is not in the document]
phildarnowsky: I certainly wasn't looking at the projects index to remember how to set up a React page *cough* *cough*

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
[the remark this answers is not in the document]
digitalcora: ohhh, I see. I imagine the TSX compiles to a `React` function call of some sort.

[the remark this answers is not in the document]
digitalcora: ohhh, I see. I imagine the TSX compiles to a `React` function call of some sort.

apps/site/assets/ts/shuttles/shuttles-loader.tsx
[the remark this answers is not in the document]
phildarnowsky: OK. I was discombobulated for a moment about why this still seemed to work even with the wrong root element, until I remembered the fact that we _also_ render React components server-side (for purposes of e.g. SEO).

apps/site/lib/site_web/controllers/schedule/shuttles_controller.ex
phildarnowsky: I'll look into it.

apps/site/lib/site_web/controllers/schedule/green.ex
phildarnowsky: Not especially, except that in doing so I'm replicating the existing pattern. It might be desirable to split up this big controller but that's a job I'm willing to leave for another time.

apps/site/lib/site_web/controllers/schedule/shuttles_controller.ex
phildarnowsky: There actually does seem to be a good reason to do this, to wit, we call `assign_alerts` with different parameters from one controller to the other, and I'm not aware of a simple way to implement that with `Plug`s--you'd have to assign the pertinent arguments to a known assign in the `conn` and then pull that back out in the `Plug`. But if I'm missing a better way to do it, I'd like to hear about it.

digitalcora: `plug` accepts a second argument, which if present is handed as a second argument to the plug's `init` and `call` functions. For example:

```elixir
plug SiteWeb.Plugs.AssignAlerts, filter_by_direction?: true
```

We are already doing this in `SiteWeb.Plugs.Banner`, but it appears no controller that uses the banner actually uses the options.

apps/site/lib/site_web/controllers/schedule/green.ex
digitalcora: Following the existing pattern seems as good a justification as any, under the circumstances.

apps/site/lib/site_web/controllers/schedule/shuttles_controller.ex
phildarnowsky: TIL. OK, seems like a simple refactoring.

phildarnowsky: @digitalcora I see that there is at least one place that uses `assign_alerts` outside of a plug context (i.e. just as a regular function call), and it's not clear to me if changing that to a plug will break the logic, so on reconsideration I'm inclined to leave this as-is.

digitalcora: Sounds fine. I'd ordinarily advocate taking a closer look but we should prioritize getting this skeleton in place to be able to build on it.

PR #272, opened 2019-11-14
Update shuttle-tab logic, add tests

phildarnowsky:
Here we tweak the logic controlling whether or not the shuttle tab
appears, so that it only appears in the case of an ongoing (or
ongoing_upcoming) alert. We also add test coverage for this logic.

#### Summary of changes
**Asana Ticket:** [Shuttles | Properly test appearance of shuttles tab](https://app.asana.com/0/555089885850811/1149377816922177)

<br>
Assigned to: @digitalcora

### DIFF
--- a/apps/site/lib/site_web/controllers/helpers.ex
+++ b/apps/site/lib/site_web/controllers/helpers.ex
@@ -76,10 +76,11 @@ defmodule SiteWeb.ControllerHelpers do
         opts
       ) do
     filter_by_direction? = Keyword.get(opts, :filter_by_direction?, true)
+    alerts_by_route_id_and_type_fn = Keyword.get(opts, :repo_fn, &Repo.by_route_id_and_type/3)
 
     alerts =
       route_id
-      |> Repo.by_route_id_and_type(route_type, date_time)
+      |> alerts_by_route_id_and_type_fn.(route_type, date_time)
       |> Match.match([%InformedEntity{route_type: route_type}, %InformedEntity{route: route_id}])
 
     filtered_alerts =
--- a/apps/site/test/site_web/views/schedule_view_test.exs
+++ b/apps/site/test/site_web/views/schedule_view_test.exs
@@ -11,6 +11,7 @@ defmodule SiteWeb.ScheduleViewTest do
   import SiteWeb.ScheduleView
   import Phoenix.HTML.Tag, only: [content_tag: 3]
   import Phoenix.HTML, only: [safe_to_string: 1]
+  import SiteWeb.ControllerHelpers, only: [assign_alerts: 2]
 
   @trip %Schedules.Trip{name: "101", headsign: "Headsign", direction_id: 0, id: "1"}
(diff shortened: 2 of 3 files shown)


digitalcora:
Will review in a bit, though with my new API knowledge I'm noticing this might not be the "correct" logic for showing the tab after all... basically, a shuttle alert being active is not a guarantee that we have the required map shapes to render this page. This is sort of a design call, but I have a feeling it would be a bad experience if someone clicked the **Shuttles** tab and just saw "whoops, we don't have anything to show you here after all".

phildarnowsky:
@digitalcora yes, good catch, that's a case we should account for. I think though we can address that particular case when we add the logic that actually fetches and renders the shapes.

apps/site/lib/site_web/controllers/helpers.ex
```diff
@@ -76,10 +76,11 @@ defmodule SiteWeb.ControllerHelpers do
         opts
       ) do
     filter_by_direction? = Keyword.get(opts, :filter_by_direction?, true)
+    repo_fn = Keyword.get(opts, :repo_fn, &Repo.by_route_id_and_type/3)
 
     alerts =
       route_id
-      |> Repo.by_route_id_and_type(route_type, date_time)
+      |> repo_fn.(route_type, date_time)
```
digitalcora: The name `repo_fn` seems like it could be more descriptive here, but other than going full `alerts_by_route_id_and_type_fn` I've got nothing. It would be nice if we had a more structured means of doing this kind of "dependency injection" in the first place, but in the short term I do appreciate the testing enabled by doing it here.

apps/site/test/site_web/views/schedule_view_test.exs
```diff
@@ -1055,6 +1070,76 @@ defmodule SiteWeb.ScheduleViewTest do
...
+          conn
+          |> put_req_header("cookie", "shuttles=true")
+          |> fetch_cookies
+          |> assign(:route, %Route{id: "83", type: 3})
+          |> assign(:tab, "alerts")
+          |> assign(:tab_params, [])
+          |> assign(:date_time, DateTime.utc_now())
+          |> assign_alerts(filter_by_direction?: true, repo_fn: fake_repo_fn)
+          |> route_header_tabs()
+          |> safe_to_string()
```
digitalcora: Can we extract this lengthy stack of prerequisites to a helper function(s) within the test file? In particular I think it would be good to have an explicit `enable_shuttles_laboratory` function so we know what that logic is doing and can cleanly remove it when we get rid of the flag.

apps/site/lib/site_web/controllers/helpers.ex
phildarnowsky: @digitalcora let's go with `alerts_by_route_id_and_type_fn` then, that's a perfectly cromulent name.

digitalcora: Ok, have fun with the line wrapping 😄

PR #274, opened 2019-11-14
add initial shuttle page overview with direction buttons

thecristen:
#### Summary of changes
* Adds route data to the shuttles page React component
* Adds helper function to access JSON-safe route data
* Creates buttons dynamically from route direction_destinations and adds an event handler and initial effect
* Adds initial tests for ShuttlesPage and ShuttlesOverview

**Asana Ticket:** [Shuttles | Direction selector](https://app.asana.com/0/477545582537986/1148592031135784/f)

### DIFF
--- a/apps/site/assets/ts/shuttles/__tests__/ShuttlesOverviewTest.tsx
+++ b/apps/site/assets/ts/shuttles/__tests__/ShuttlesOverviewTest.tsx
@@ -1,0 +1,44 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import { mount, ReactWrapper } from "enzyme";
+import { createReactRoot } from "../../app/helpers/testUtils";
+import ShuttlesOverview from "../components/ShuttlesOverview";
+import { Route } from "../../__v3api";
+
+const route: Route = {
+  description: "",
+  direction_destinations: { 0: "Oak Grove", 1: "Forest Hills" },
+  direction_names: { 0: "Inbound", 1: "Outbound" },
+  id: "Orange",
+  name: "Orange",
+  long_name: "Orange Line",
+  type: 1
+};
+
+it("it renders", () => {
+  createReactRoot();
+  const tree = renderer
+    .create(<ShuttlesOverview places={route.direction_destinations} />)
+    .toJSON();
+  expect(tree).toBeTruthy();
+  expect(tree).toMatchSnapshot();
+});
+
+describe("ShuttlesOverview direction control", () => {
+  let wrapper: ReactWrapper;
+  beforeEach(() => {
+    wrapper = mount(<ShuttlesOverview places={route.direction_destinations} />);
+  });
+
+  it("shows 3 filtering choices", () => {
+    expect(wrapper.find("button")).toHaveLength(3);
+  });
+
+  it("displays a button for each direction", () => {
+    Object.values(route.direction_destinations).forEach(headsign => {
+      expect(
+        wrapper.containsMatchingElement(<button>{headsign}</button>)
+      ).toEqual(true);
+    });
+  });
+});
--- a/apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
+++ b/apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
@@ -1,0 +1,22 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import { createReactRoot } from "../../app/helpers/testUtils";
+import ShuttlesPage from "../components/ShuttlesPage";
+import { Route } from "../../__v3api";
+
+const route: Route = {
+  description: "",
+  direction_destinations: { 0: "Oak Grove", 1: "Forest Hills" },
+  direction_names: { 0: "Inbound", 1: "Outbound" },
+  id: "Orange",
+  name: "Orange",
+  long_name: "Orange Line",
+  type: 1
+};
+
+it("it renders", () => {
+  createReactRoot();
+  const tree = renderer.create(<ShuttlesPage route={route} />).toJSON();
+  expect(tree).toBeTruthy();
+  expect(tree).toMatchSnapshot();
+});
--- a/apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
+++ b/apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import renderer from "react-test-renderer";
+import { mount, ReactWrapper } from "enzyme";
 import { createReactRoot } from "../../app/helpers/testUtils";
 import ShuttlesPage from "../components/ShuttlesPage";
 import { Route } from "../../__v3api";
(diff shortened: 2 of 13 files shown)


phildarnowsky:
@thecristen I'm going to tag this `needs design review` as well, which we do when we have a feature with user-facing UI changes. However, in this case, given that we've broken up this page into many small parts, it may be more efficient to do a single, or a handful, of design reviews, followed by a polish PR, rather than one design review for each individual PR. I will defer to however @ingridpierre wants to play this one.

apps/site/assets/ts/shuttles/components/ShuttlesOverview.tsx
```diff
@@ -0,0 +1,43 @@
...
+interface Props {
+  places: {
+    0: string;
+    1: string;
+  };
+}
+
+const ShuttlesOverview = ({ places }: Props): ReactElement<HTMLElement> => {
+  const [selectedDirectionId, setSelectedDirectionId] = useState("");
+  useEffect(() => {}, [selectedDirectionId]);
```
phildarnowsky: @thecristen this `useEffect` doesn't seem to do anything

apps/site/assets/ts/shuttles/components/ShuttlesOverview.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const directionButton = (
```
phildarnowsky: Small point here: we capitalize any TS function that implements a UI element, whether it's a `ReactElement` or not. So in this case it's be `DirectionButton`.

apps/site/assets/ts/shuttles/components/ShuttlesOverview.tsx
```diff
@@ -0,0 +1,43 @@
...
+    >
+      {headsign}
+    </button>
+  );
+
+  return (
+    <>
+      <div role="group" aria-label="Filter shuttle map by direction">
+        {directionButton()}
+        {Object.entries(places).map(([directionId, headsign]) =>
```
phildarnowsky: Let's throw a `sort` in here, since IIRC the order of the elements of the return value of `entries` is not guaranteed. It'll _probably_ always be 0 then 1...but with one method call we can be certain.

apps/site/assets/ts/shuttles/__tests__/ShuttlesOverviewTest.tsx
```diff
@@ -0,0 +1,43 @@
...
+  long_name: "Orange Line",
+  type: 1
+};
+
+it("it renders", () => {
+  createReactRoot();
+  const tree = renderer
+    .create(<ShuttlesOverview places={route.direction_destinations} />)
+    .toJSON();
+  expect(tree).toBeTruthy();
```
phildarnowsky: @thecristen let's use a "snapshot" test here. (See anyplace in the frontend tests we've called `expect(tree).toMatchSnapshot()`. The idea with a snapshot test is, we have a file checked into the repo that contains the markup that this component should render to, and the test verifies the actual markup generated against the content of this "snapshot" file.

Of course, the snapshot file doesn't exist until we create it, and thus running `npm run test:jest` will fail on this test. But you can have the snapshot automatically created by running `npm run test:jest:update`, which will run the test suite and generate/regenerate snapshots as needed.

The resulting snapshot file should be added to the PR as well.

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
```diff
@@ -0,0 +1,21 @@
...
+  id: "Orange",
+  name: "Orange",
+  long_name: "Orange Line",
+  type: 1
+};
+
+it("it renders", () => {
+  createReactRoot();
+  const tree = renderer.create(<ShuttlesPage route={route} />).toJSON();
+  expect(tree).toBeTruthy();
```
phildarnowsky: @thecristen and let's do a snapshot here as well

apps/site/assets/ts/shuttles/components/ShuttlesOverview.tsx
thecristen: true - it's intended as a placeholder that will handle whatever is needed to update the future overview map.

ingridpierre:
For Green Line main page (/schedules/Green), let's not use the direction_destinations field, and instead use the general direction name of "Eastbound" and "Westbound".

image

ingridpierre:
@thecristen Let's add a 'noindex' tag here so these pages don't get indexed by google. Currently the tab can be accessed if you have the URL and its then "available" even for things like buses. I would hate for someone to google "red line shuttle" and be taken here when theres nothing active.

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
```diff
@@ -6,13 +6,18 @@ interface Props {
...
-  <div className="shuttles__main col-sm-12 col-lg-7">
-    <h1>Shuttle Maps</h1>
-    <ShuttlesOverview places={route.direction_destinations} />
-    <div>Station Detail and dropdown</div>
-    <div>Map2</div>
-  </div>
-);
+const ShuttlesPage = ({ route }: Props): ReactElement<HTMLElement> => {
+  const places =
+    route.id === "Green" ? route.direction_names : route.direction_destinations;
```
digitalcora: Would it be reasonable to just always use `direction_names`? (this might be sort of a design question)

apps/site/assets/ts/shuttles/components/ShuttlesOverview.tsx
phildarnowsky: OK, in that case let's add a comment to that effect.

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
phildarnowsky: @ingridpierre ?

phildarnowsky:
@thecristen looking good once the comment is in the place I called out. I don't remember if we talked about your level of familiarity with complex rebases, so let's hold off on merging this PR until I'm back and we can talk about a strategy to rebase this.

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
thecristen: This change was made in collaboration with @ingridpierre so I'm confident `direction_names` is just needed here in this single case (showing Eastbound/Westbound instead of a concatenation of up to four different GL terminals), while the other pages can show destinations (Park St, Boston College, and so on).

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
```diff
@@ -1,5 +1,6 @@
 import React from "react";
 import renderer from "react-test-renderer";
+import { mount, ReactWrapper } from "enzyme";
```
phildarnowsky: @thecristen I'd like commit titles to contain enough information that, when you're skimming the logs and looking at just the titles, you have enough context to tell what in the app is changing.

In this case, if I were to read just the title of the commit, I would only know that it was a special case for the Green Line for something-or-another, and there are lots of ways in which the Green Line is an edge case, so the context is not as clear as is desirable.

What I suggest is a commit title like:

`Handle Green Line in shuttle direction selector`

and then in the body of the commit message you can explain the specifics of the fix, i.e. in this case using the generic labels "Eastbound/Westbound".

digitalcora:
@thecristen If you're rewording commit messages anyway, I'd recommend adjusting some of the longer message "titles" to be [50 characters or less](https://chris.beams.io/posts/git-commit/#limit-50) to avoid this kind of truncation:

Screen Shot 2019-11-20 at 12 03 34 PM

In my editor I have an extra margin at 50 characters (in addition to 72) for the "commit message" syntax; you can probably set yours to do the same!

PR #276, opened 2019-11-15
Attempt Codecov Fix

ianwestcott:
#### Summary of changes
**Asana Ticket:** [Fix Codecov for Dotcom](https://app.asana.com/0/1113179098808463/1149791411919817)

There are multiple ways to configure GitHub integrations with Codecov: One option is to use a [“bot user”](https://docs.codecov.io/docs/team-bot) that authenticates, and another is to use the [GitHub application integration](https://github.com/marketplace/codecov). We were using both in our org, and it was messy. 

So I’m trying to move away from the bot user. Apparently that means removing this line that explicitly tells Codecov to use the bot user instead of the integration.

<br>
Assigned to: @amaisano

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,4 @@
 codecov:
-  bot: mbtace
   notify:
     require_ci_to_pass: true
 comment:
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,3 @@
-codecov:
-  notify:
-    require_ci_to_pass: true
 comment:
   behavior: default
   layout: header, diff


codecov.yml
```diff
@@ -1,7 +1,3 @@
-codecov:
-  bot: mbtace
-  notify:
-    require_ci_to_pass: true
```
digitalcora: This option seems useful in its own right, or do we have to remove it if we want to use the non-bot approach? Would deleting only line 2 fix the issue?

ianwestcott: It's the default setting: https://docs.codecov.io/docs/codecov-yaml#section-default-yaml

I'll admit that I mostly only removed it to trigger a new build. But generally I prefer not to declare settings in configuration if they are the default value. If a config only reflects the non-default values, it's easier to read and understand.

ianwestcott: Plus in our case it was nested under the `notify` section (perhaps incorrectly?), so I'm not sure it was being used anyway.

digitalcora:
@ianwestcott If you bring this up-to-date with master it should pass CI (flaky Elixir tests aside).

PR #277, opened 2019-11-15
Disable Pronto GitHub PR status reporting

digitalcora:
This is a quick fix for Pronto breaking all our builds because its access token was removed. The relevant step is now green on this branch in Semaphore.


ianwestcott:
Kicked off a rebuild

PR #284, opened 2019-11-22
Copy existing alerts section to Shuttles tab

digitalcora:
**Asana Ticket:** [Shuttles | Show existing full alert component on this tab](https://app.asana.com/0/477545582537986/1148592031135794/f)

This is copied straight from the current alerts section in `schedule/_line.html.eex`, so it should show the same set of alerts.

### DIFF
--- a/apps/site/lib/site_web/templates/schedule/_shuttles.html.eex
+++ b/apps/site/lib/site_web/templates/schedule/_shuttles.html.eex
@@ -10,6 +10,10 @@
 </script>
 
 <div class="row">
+  <div class="shuttles__alerts col-md-12">
+    <%= SiteWeb.AlertView.group(alerts: @alerts, route: @route, date_time: @date_time, priority_filter: :high) %>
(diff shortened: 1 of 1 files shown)


apps/site/lib/site_web/templates/schedule/_shuttles.html.eex
```diff
@@ -10,6 +10,10 @@
 </script>
 
 <div class="row">
+  <div class="shuttles__alerts col-md-12">
+    <%= SiteWeb.AlertView.group(alerts: @alerts, route: @route, date_time: @date_time, priority_filter: :high) %>
```
amaisano: Does the `priority_filter` allow shuttle notifications to appear? Or is that going to be adjusted in a future commit anyway?

digitalcora: It does, and it will be 🙂

PR #287, opened 2019-11-22
Shuttles station selector

thecristen:
#### Summary of changes
**Asana Ticket:** [Shuttles | Station selector dropdown](https://app.asana.com/0/555089885850811/1148592031135787?du=1149164108974353&invite=575f3f3f994707d96407b76ff2caf727)

* 😅  Updates one development dependency so that I could use the [invoke function](https://airbnb.io/enzyme/docs/api/ReactWrapper/invoke.html) in my Jest tests
* 🔧  Refactors the existing shuttles work (along with accompanying tests and snapshots)
* ➕  Adds and uses helper function to get shuttle data to the templates
* ⚡️ Includes a notional map component

Assigned to: @digitalcora


apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -114,6 +115,13 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
     RouteStops.by_direction(stops[route.id], shapes, route, direction_id)
   end
 
+  @spec get_shuttle_data(Plug.Conn.t()) :: [Phoenix.HTML.safe()]
+  def get_shuttle_data(conn) do
+    now = conn.assigns[:date_time] || Util.now()
+    {:ok, data} = ShuttleDiversion.active([conn.assigns.route.id], now)
+    data
+  end
+
```
thecristen: I'm working on writing a test for this and I'm not quite sure what I want this function to do if there's no route or if it doesn't return `:ok`.

apps/site/assets/ts/shuttles/components/__shuttles.d.ts
```diff
@@ -0,0 +1,20 @@
+export interface ShuttleDestinations {
+  0: string;
+  1: string;
+}
+
+export interface ShuttleStop {
```
digitalcora: `ShuttleStop` makes it sound like these are the stops for the shuttles themselves, but they encompass all stops (rail and bus) related to a shuttle diversion. Maybe this could just be `Stop`? The shapes are just `Shape`, after all.

apps/site/assets/ts/shuttles/components/__shuttles.d.ts
```diff
@@ -0,0 +1,20 @@
+export interface ShuttleDestinations {
```
digitalcora: This seems unused?

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -114,6 +115,13 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
     RouteStops.by_direction(stops[route.id], shapes, route, direction_id)
   end
 
+  @spec get_shuttle_data(Plug.Conn.t()) :: [Phoenix.HTML.safe()]
+  def get_shuttle_data(conn) do
+    now = conn.assigns[:date_time] || Util.now()
```
digitalcora: I think we can make this `conn.assigns.date_time` (without the or-default) since we expect it to always be there when the `DateTime` plug is used; it has its own defaulting logic.

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
```diff
@@ -1,21 +1,110 @@
...
+  }>({ latitude: null, longitude: null });
+
+  useEffect(
+    () => {
+      if (selectedDirectionId === "") {
+        setDisplayedShapes([...diversions.shapes]);
+        setDisplayedStops([...diversions.stops]);
+      } else {
+        setDisplayedShapes(
+          diversions.shapes.filter(s => s.direction_id === +selectedDirectionId)
```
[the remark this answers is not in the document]
thecristen: Yes, that's the reason.

apps/site/assets/ts/shuttles/__tests__/StationDropdownTest.tsx
```diff
@@ -0,0 +1,38 @@
...
+describe("StationDetail station control", () => {
+  let wrapper: ReactWrapper;
+  beforeEach(() => {
+    wrapper = mount(
+      <StationDropdown stations={stations} onSelect={s => true} />
+    );
+  });
+
+  it("shows appropriate number of stations", () => {
+    expect(wrapper.find("option")).toHaveLength(stations.length + 1);
```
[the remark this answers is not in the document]
thecristen: Confirmed that it _shouldn't_ be in here, and will be removed!

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { ReactElement } from "react";
+import { ShuttleStop, Shape } from "./__shuttles";
+
+interface Props {
+  shapes: Shape[];
+  stops: ShuttleStop[];
+  center?: {
+    latitude: number | null;
+    longitude: number | null;
```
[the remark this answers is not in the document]
thecristen: Yeah this bit is a total mess. Will resolve in the upcoming map PR.

apps/site/assets/ts/shuttles/components/StationDropdown.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface Props {
+  stations: ShuttleStop[];
+  onSelect: (stationId: string) => void;
+}
+
+const StationDropdown = ({
+  stations,
+  onSelect
+}: Props): ReactElement<HTMLElement> => {
+  const [selectedStationId, setSelectedStationId] = useState("");
```
phildarnowsky: @thecristen another `""` that should probably be `null`

apps/site/assets/ts/shuttles/components/StationDropdown.tsx
```diff
@@ -0,0 +1,42 @@
...
+  useEffect(
+    () => {
+      onSelect(selectedStationId);
+    },
+    [onSelect, selectedStationId]
```
phildarnowsky: @thecristen I think you may want to leave `onSelect` out of this list of triggers

apps/site/assets/ts/shuttles/components/StationDropdown.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  return (
+    <div className="c-select-custom__container">
+      <select
+        className="c-select-custom"
+        defaultValue={selectedStationId}
+        onChange={e => setSelectedStationId(e.target.value)}
+      >
+        <option value="">Choose a station</option>
+        {stations.sort().map(s => (
```
phildarnowsky: @thecristen this is why I'm into explicit variable names: sure, you save a few characters, but here `s` is a station, not a shape--which would not be the first thing I thought of having read some of the code above where it does stand for shape. In general, if my collection is named `foos`, I call the iterator `foo`.

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
```diff
@@ -0,0 +1,49 @@
...
+    className="m-shuttles-map"
+    style={{
+      border: "1px dotted gray",
+      backgroundColor: "#ee000011",
+      height: "444px",
+      fontSize: "12px"
+    }}
+  >
+    Center?: {JSON.stringify(center)}
+    <details>
```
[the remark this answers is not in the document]
phildarnowsky: @thecristen that is, this is just a temporary "map" here so we can, internally, verify that the right data is getting through to the frontend?

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
phildarnowsky: @thecristen @digitalcora I'm looking at the definition of `ShuttleDiversion.active` and it looks like, despite what the typespec implies, it will always either return `{:ok, %ShuttleDiversion{...}}`, or it will crash due to a pattern-match fail.

If I follow, if we get to this code and there is no active diversion, the most likely explanation is that the end-user went directly to the shuttles page, without following a link there. From a UX point of view, then, the best thing to do would probably be to redirect them to the main page for the line.

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
```diff
@@ -1,21 +1,110 @@
...
+      </div>
+
+      <ShuttlesMap shapes={displayedShapes} stops={displayedStops} />
+
+      <div className="c-heading-with-control">
+        <h3>Station Detail</h3>
+
+        <StationDropdown
+          stations={diversions.stops.filter(s => s.type === "rail_affected")}
+          onSelect={stationId => {
```
phildarnowsky: @thecristen as a rule I like to avoid any significant amount of code in an inline event callback, could you please extract this into a method called `centerMapOnStation` or whatnot?

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
digitalcora: I believe the typespec is correct. If any of the `with <-` pattern-matches fail, the term that failed to match will be returned, which will be `{:error, any}` as per the typespecs of the called functions. This would occur if there was an HTTP error fetching data from the API. In this case, I think we should have an error state for the page that explains something went wrong, and it might go right if you refresh. But I'm not sure if we have a convention for such things on other pages.

phildarnowsky: you're right @digitalcora I was reading this wrong entirely, never mind

thecristen: An error state would be good; I've added a failing test and also am unsure what convention I ought to follow here :)

apps/site/assets/css/_shuttles-page.scss
```diff
@@ -38,4 +38,27 @@
...
+.c-heading-with-control {
+  align-items: baseline;
+  display: flex;
+  justify-content: space-between;
+
+  @include media-breakpoint-only(xs) {
+    display: initial;
+  }
+
+  > * {
```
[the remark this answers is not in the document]
thecristen: Fair, I ended up removing this selector as its changes were minimally noticable anyhow...

apps/site/assets/ts/shuttles/__tests__/ShuttlesMapTest.tsx
```diff
@@ -0,0 +1,29 @@
...
+  createReactRoot();
+  const tree = renderer
+    .create(
+      <ShuttlesMap
+        shapes={diversionsData.shapes}
+        stops={diversionsData.stops}
+      />
+    )
+    .toJSON();
+  expect(tree).toBeTruthy();
```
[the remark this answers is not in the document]
thecristen: Removed redundant truthy checks

apps/site/assets/ts/shuttles/__tests__/ShuttlesMapTest.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { createReactRoot } from "../../app/helpers/testUtils";
+import ShuttlesMap from "../components/ShuttlesMap";
+
+import _diversionsData from "./diversionData.json";
+import { Shape, ShuttleStop } from "../components/__shuttles";
+
+interface Diversion {
+  shapes: Shape[];
+  stops: ShuttleStop[];
+}
```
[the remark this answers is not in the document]
thecristen: Good call, extracted this out and am reusing accordingly

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
```diff
@@ -33,18 +45,24 @@ const destinations = (r: Route) => Object.values(r.direction_destinations);
...
+    .create(<ShuttlesPage route={route} diversions={diversionsData} />)
+    .toJSON();
   expect(tree).toBeTruthy();
   expect(tree).toMatchSnapshot();
 });
 
-describe("it passes to ShuttlesOverview", () => {
-  let wrapper: ReactWrapper;
-  let greenWrapper: ReactWrapper;
+describe("it shows", () => {
```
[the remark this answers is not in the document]
thecristen: Fair note, for some reason my eyes just blocked out the `describe`s and `it`s. Hopefully have improved the test names now, and also aimed for a more honest description of the Green line test.

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
```diff
@@ -73,3 +91,53 @@ describe("it passes to ShuttlesOverview", () => {
...
+  it("direction filters stops on both maps", () => {
+    const stopsLength = diversionsData.stops.length;
+    wrapper.find(ShuttlesMap).forEach(map => {
+      expect(map.prop("stops").length).toBe(stopsLength);
+    });
+
+    act(() => {
+      wrapper
+        .findWhere(n => n.key() === "Alewife")
+        .first()
```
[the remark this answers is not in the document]
thecristen: Maybe. Ended up taking a different approach anyhow

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
```diff
@@ -73,3 +91,53 @@ describe("it passes to ShuttlesOverview", () => {
...
+        .invoke("onClick")("0");
```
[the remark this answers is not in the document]
thecristen: It turns out `simulate("click")` works fine in this case, and with my latest commit technically we don't need to update Enzyme here anymore... 🤦‍♀ 

Though [this issue](https://github.com/airbnb/enzyme/issues/2173) suggests `simulate()` will be deprecated someday, and consensus in the Github issues seems to be to avoid using simulate and just call whatever function it's triggering directly... \shrug

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
```diff
@@ -73,3 +91,53 @@ describe("it passes to ShuttlesOverview", () => {
...
+    });
+
+    wrapper.update();
+    wrapper.find(ShuttlesMap).forEach(map => {
+      expect(map.prop("stops").length).toBeLessThan(stopsLength);
```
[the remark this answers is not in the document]
thecristen: Added more specific checks here, yay

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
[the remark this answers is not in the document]
thecristen: Well, extracted another interface for this (Coordinates), should be less of a mess now

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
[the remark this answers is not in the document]
thecristen: Done and done

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
```diff
@@ -1,21 +1,110 @@
...
+    directionId: string,
+    headsign: string
+  ): JSX.Element => (
+    <button
+      key={`${headsign}`}
+      type="button"
+      className={`btn btn-secondary btn-sm ${
+        selectedDirectionId === directionId
+          ? "c-btn-group-stackable__btn c-btn-group-stackable__btn--active"
+          : "c-btn-group-stackable__btn"
```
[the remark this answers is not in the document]
thecristen: Yep

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
thecristen: That's great, done

apps/site/assets/package.json
```diff
@@ -35,8 +35,8 @@
     "@babel/preset-env": "^7.4.2",
     "@babel/register": "^7.4.0",
     "@types/core-js": "^2.5.0",
-    "@types/enzyme": "^3.1.18",
-    "@types/enzyme-adapter-react-16": "^1.0.4",
+    "@types/enzyme": "^3.10.3",
```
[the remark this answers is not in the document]
thecristen: I managed to refactor the test and no longer need to use `invoke`, so I am considering reverting this altogether...would that be preferred here?

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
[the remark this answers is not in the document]
thecristen: @phildarnowsky Exactly that. If preferred I can remove it from this PR

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
```diff
@@ -1,21 +1,110 @@
...
+}: Props): ReactElement<HTMLElement> => {
   const places =
     route.id === "Green" ? route.direction_names : route.direction_destinations;
 
+  const [selectedDirectionId, setSelectedDirectionId] = useState("");
+  const [displayedShapes, setDisplayedShapes] = useState([
+    ...diversions.shapes
+  ]);
+  const [displayedStops, setDisplayedStops] = useState([...diversions.stops]);
+  const [mapCenter, setMapCenter] = useState<{
```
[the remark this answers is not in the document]
thecristen: Done

apps/site/assets/ts/shuttles/components/StationDropdown.tsx
thecristen: @phildarnowsky sadly the linter gives me a warning in that case

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
thecristen: Sounds good @phildarnowsky , I've gone ahead and extracted this into a method.

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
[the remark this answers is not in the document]
digitalcora: Good to know! When you say "call the function directly" you mean using `invoke`, or is there another way?

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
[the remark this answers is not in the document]
digitalcora: Incidentally, it looks like `MapData` handles this by just not allowing coordinates to be null (which I think makes sense; a map always has a center point, a marker always has a location, etc.)

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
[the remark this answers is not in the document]
thecristen: Yeah. Probably more relevant for class-based React components, but I think one can define a function on the class component and invoke it by `component.foo()`. Or if the relevant function can be imported, it can be used in the test that way too. Anyway...

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
```diff
@@ -73,3 +85,116 @@ describe("it passes to ShuttlesOverview", () => {
     });
   });
 });
+
+describe("the selected", () => {
```
digitalcora: Just to put what we were talking about yesterday in writing: we should document (in a comment here perhaps) exactly why we're doing this odd arrangement of `describe` vs `it` so no one gets confused when they try to refactor it in the future.

apps/site/assets/ts/shuttles/__tests__/diversionData.json
```diff
@@ -0,0 +1 @@
...9999999985,-71.12802999999998],[42.39759999999998,-71.12909999999998],[42.39772999999998,-71.13087999999998],[42.39792999999998,-71.13449999999997],[42.39796999999998,-71.13519999999997],[42.397869999999976,-71.13607999999996],[42.39782999999998,-71.13635999999997],[42.39774999999998,-71.13669999999996],[42.39765999999998,-71.13707999999997],[42.39760999999998,-71.13717999999997],[42.39755999999998,-71.13724999999997],[42.397489999999976,-71.13728999999996],[42.397219999999976,-71.13736999999996],[42.396949999999975,-71.13748999999996],[42.396679999999975,-71.13770999999996],[42.396479999999976,-71.13801999999995],[42.39635999999997,-71.13828999999996],[42.39634999999997,-71.13831999999995],[42.39624999999997,-71.13895999999995],[42.39623999999996,-71.13913999999995],[42.39618999999996,-71.13954999999996],[42.39614999999996,-71.14207999999995]],"is_shuttle_route":false,"id":"931_0010-70082-70072-2","direction_id":1},{"polyline":[[42.39615,-71.14208],[42.39619,-71.13955000000001],[42.39624,-71.13914000000001],[42.39625,-71.13896000000001],[42.396350000000005,-71.13832000000001],[42.39636000000001,-71.13829000000001],[42.39648000000001,-71.13802000000001],[42.39668000000001,-71.13771
```
digitalcora: We should refresh this snapshot, if we haven't already, so it correctly excludes the extra shapes that were in there before.

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
```diff
@@ -1,21 +1,115 @@
...
   const places =
     route.id === "Green" ? route.direction_names : route.direction_destinations;
 
+  const [selectedDirectionId, setSelectedDirectionId] = useState<
+    MaybeDirectionId
+  >(null);
+  const [displayedShapes, setDisplayedShapes] = useState([
+    ...diversions.shapes
+  ]);
+  const [displayedStops, setDisplayedStops] = useState([...diversions.stops]);
```
digitalcora: Is `[...diversions.stops]` not equivalent to `diversions.stops`? (and likewise for `shapes`)

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
```diff
@@ -1,21 +1,115 @@
...
+      type="button"
+      className={`btn btn-secondary btn-sm c-btn-group-stackable__btn ${selectedDirectionId ===
+        directionId && "c-btn-group-stackable__btn--active"}`}
+      onClick={() => setSelectedDirectionId(directionId)}
+    >
+      {headsign}
+    </button>
+  );
+
+  const centerMapOnStation = (stationId: string): void => {
```
digitalcora: We're inconsistent in some of this code between calling them `stations` or `stops`. I'd say we should stick with `stop` since that is the name of the domain model.

apps/site/test/site_web/controllers/schedule/line_test.exs
```diff
@@ -446,4 +448,19 @@ defmodule SiteWeb.ScheduleController.LineTest do
...
+      conn: conn
+    } do
+      conn =
+        conn
+        |> Conn.assign(:route, %Route{id: "Red", type: 1})
+        |> Conn.assign(:date_time, ~N[2019-11-25T12:00:00])
+
+      expected = %Site.ShuttleDiversion{shapes: [], stops: []}
+      data = get_shuttle_data(conn)
+      assert data == expected
```
digitalcora: I'm not sure this test is valuable to us as-is — we'd expect the data to be empty for any date in the past, since the relevant Alert would no longer exist, and we can't pick a date in the future since the alerts won't exist yet.

I think what we want is for `ShuttleDiversion` itself to have comprehensive tests, which is something I'll be working on soon. If that's the case, we can be confident that the very straightforward `get_shuttle_data` will work as we expect. But if we still wanted to unit-test it, we could mock `ShuttleDiversion` (look up `with_mock` for how we're currently doing this, which is not great, but it's something).

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
thecristen: Yes! Because in JavaScript arrays are mutable, and in this case I wanted to copy the array (in this case I happen to use the spread operator `...` to achieve that) and not maintain a reference to the original.

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
thecristen: I'm going to table this for my next PR because yes.

apps/site/test/site_web/controllers/schedule/line_test.exs
thecristen: Unit tests on `ShuttleDiversion` sounds like a good idea. As for mocking `ShuttleDiversion`, I'm not quite grasping how that would help with this particular test. Let's discuss soon. :)

apps/site/assets/ts/shuttles/__tests__/diversionData.json
thecristen: Good call, will update this fixture along with the associated snapshots!

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
thecristen: Will address in my next PR.

apps/site/test/site_web/controllers/schedule/line_test.exs
digitalcora: `mock` works like this:

```elixir
import Mock

diversion_fn = fn _, _ -> {:ok, %ShuttleDiversion{...}} end

with_mock ShuttleDiversion, active: diversion_fn do
  # in here, calls to `ShuttleDiversion.active` are instead sent to `diversion_fn`
end
```

This is slightly dangerous because it doesn't verify that `diversion_fn` actually matches the signature of `ShuttleDiversion.active` — if the latter were modified (maybe a third argument is added), the test would continue to pass, but would no longer be indicative of what would happen in a real un-mocked environment.

apps/site/assets/ts/shuttles/components/ShuttlesPage.tsx
digitalcora: Ah, right, I missed that the intent here was to copy the array — the way I was familiar with was `array.slice()`.

PR #290, opened 2019-11-26
Add maps to shuttles tab

thecristen:
#### Summary of changes
**Asana Ticket:** [Shuttles | Pull shapes (Subway/Shuttles) onto map](https://app.asana.com/0/1149164109017504/1148592031135761?du=1149164108974353) and [Shuttles | Station detail map](https://app.asana.com/0/1149164109017504/1148592031135778?du=1149164108974353)

All shapes are pulled into the two maps. Second map is zoomed into the selection from the station selector, and both maps filter all shapes based on the selected direction.

This needs one more bit of code to take into account the correct line colors from the latest merge to master. With corrected line colors the end result looks like this:

Shuttles-Maps-RL


ingridpierre:
I love this! I have a few changes I'd like to see, but already I love how this is shaping up.

1) **red X-icon**: I don't think we should use that red color + icon to show the affected subway stations. In the shuttle diagram in the sidebar, we're using the same icon to mean something different. I prefer to continue to show a T logo icon for stations, even if they are closed to for that line's service they are still T stations and it makes them clearer against the shuttle icons. 
2) The **terminus stops** for the subway line shapes should not show an X icon, but use the T station icon. Technically the are open, to the active part of the line.
3) **Stroke changes:** in addition to color, I think the width is a little chunky, especially compared to the existing line map. This makes it tough to read some of the street names under the shuttle shapes at different zoom levels. Can we try:
- Reduce subway lines stroke-width to 6px. 
- Update shuttle lines stroke-width: 4px, and stroke-dasharray: 4px, 8px.

ingridpierre:
Also thought I'd note it, though I don't know what it means: locally I get quite a lot of error output in my server complaining about undefined shapes, though the console in Chrome is clean.

ingridpierre:
Would it be quite difficult to get a tooltip for each icon?

For example, here's a really rich tooltip in use on the transit-near-me maps.
image

And the much simpler one on /schedules/../line:
image

I think we could cause customer confusion if we don't help label these stops somehow.

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
```diff
@@ -1,46 +1,103 @@
...
+      latitude: 42.360718
+    };
+
+    /* eslint-disable @typescript-eslint/camelcase */
+    const mapData: MapData = {
+      height: 400,
+      width: 400,
+      default_center: mapCenter,
+      markers: makeMarkers(stops),
+      stop_markers: makeMarkers(stops),
```
digitalcora: Do we need to use both `markers` and `stop_markers`? How do these differ?

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
```diff
@@ -1,46 +1,103 @@
...
+      polylines: makePolylines(shapes),
+      tile_server_url: "https://mbta-map-tiles-dev.s3.amazonaws.com",
```
digitalcora: This strikes me as something that should not be hard-coded, although annoyingly, it looks like this is only properly defined on the server (since with the other maps the server is responsible for everything).

As a quick solution, we could pass the URL into the component as a separate prop, and get the value from `Application.fetch_env!(:site, :tile_server_url)` in the server-rendered template.

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
```diff
@@ -1,46 +1,103 @@
...
+}: Props): ReactElement<HTMLElement> | null => {
+  if (typeof window !== "undefined") {
+    /* eslint-disable-next-line global-require */
+    const getBounds = require("../../leaflet/bounds").default;
+    const bounds = getBounds(
+      makeMarkers(stops.filter(stop => stop.type === "rail_affected"))
+    );
+    const mapCenter = center || {
+      longitude: -71.05891,
+      latitude: 42.360718
```
digitalcora: What location is this? Could we put it in a descriptively-named constant?

apps/site/assets/ts/shuttles/components/StationDropdown.tsx
```diff
@@ -15,7 +15,9 @@ const StationDropdown = ({
   const [selectedStationId, setSelectedStationId] = useState(stations[0].id);
   useEffect(
     () => {
-      onSelect(selectedStationId);
+      if (selectedStationId) {
```
digitalcora: Shouldn't there always be a selected ID? Is this left over from when we had  the "Select a station" entry?

thecristen:
@ingridpierre Haven't updated this branch yet but I tried the suggested stroke widths and dasharrays, as seen here:

image

thecristen:
Oh check this out -- on the Green route, the polylines are filtered but the stops are not! @digitalcora might you know what's behind this? 

In this case only the D line has shuttle service -- @ingridpierre, I'm guessing we should still show the B, C, and E lines here?

image

digitalcora:
@thecristen I think I do, unfortunately... this synthesized "combo" green line is quite the headache. We need some extra logic to detect if, when multiple routes are requested, only a subset of the routes have diversions — and if so, add in the normal non-diverted shapes for the other routes to the returned collection.

apps/site/lib/site/shuttle_diversion.ex
```diff
@@ -10,12 +10,13 @@ defmodule Site.ShuttleDiversion do
...
     @moduledoc "Derivative of an API `shape` tailored for drawing a shuttle diversion map."
 
-    @enforce_keys [:id, :direction_id, :polyline, :is_shuttle_route]
+    @enforce_keys [:id, :direction_id, :color, :polyline, :is_shuttle_route]
     defstruct @enforce_keys
 
     @type t :: %__MODULE__{
             id: String.t(),
             direction_id: 0 | 1,
+            color: String.t(),
```
digitalcora: Sorry for the crossed wires, but I merged a PR last week that was supposed to serve as the source for the line color, just in a different place: https://github.com/mbta/dotcom/pull/289

Thoughts on this approach vs. that one?

thecristen: I'm not clear how #289 offers an alternate way of getting the color onto the `Shape`? Or was the intended approach to not put the color onto the shape, but use some other way?

apps/site/lib/site/shuttle_diversion.ex
```diff
@@ -123,6 +124,7 @@ defmodule Site.ShuttleDiversion do
           id: shape.id,
           direction_id: shape.attributes["direction_id"],
           polyline: decode_polyline(shape.attributes["polyline"]),
+          color: hd(shape.relationships["route"]).attributes["color"],
```
thecristen: @digitalcora This builds on the work done in #289 that puts the color onto the route!

digitalcora: This would have worked regardless of #289 being merged — that PR added a `color` attribute onto our Elixir struct representation of a `Route`, but this code is operating directly on JSON responses from the API and so doesn't use that.

apps/site/lib/site/shuttle_diversion.ex
digitalcora: It adds a `color` attribute onto Routes, and we have the current Route available on the shuttles page. So: for each shape, if this is not a shuttle shape, use the color defined on the current route, else use a frontend-defined "shuttle color".

digitalcora: I can see how this way might be easier since it places the color closer to the shape itself, which is where we eventually want it for the `Map`, and the `color` defined in the API for the shuttle shapes happens to be the color we want to use on the map. In that case maybe we should revert #289 since it now serves no purpose.

thecristen: 😅 I also ended up doing that because I wasn't finding any color attribute on the route. But I try it now, and it's there... whoops. I'd still need to attach the color to the shape, but I think it might be easier for me to do that in the same manner I attach the headsign onto the stop (for the tooltip), so I think I'll do that and revert this extra stuff.

thecristen: This has been corrected, thank you!

apps/site/lib/site/shuttle_diversion.ex
thecristen: Corrected!

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
thecristen: I think I had plucked it from a test. I've adjusted this to use the center of the map bounds if no explicit center is defined. 👍

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
```diff
@@ -1,46 +1,103 @@
...
 
 interface Props {
   shapes: Shape[];
   stops: Stop[];
   center?: Coordinates;
 }
 
+/* eslint-disable @typescript-eslint/camelcase */
+const makeMarkers = (stops: Stop[]): MapMarker[] =>
+  stops.map((stop: Stop, index) => {
```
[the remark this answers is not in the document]
thecristen: I think I might have added the `Stop` type before adding in `Stop[]` here. Anyway I don't see the harm in leaving both here.

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
thecristen: It turns out `stop_markers` is redundant/unneeded so I will remove that.

apps/site/assets/ts/shuttles/components/ShuttlesMap.tsx
thecristen: This has been tweaked to use `Application.fetch_env!(:site, :tile_server_url)`, thanks!

apps/site/assets/ts/shuttles/__tests__/ShuttlesPageTest.tsx
```diff
@@ -38,163 +41,111 @@ const greenRoute: Route = {
...
-    let map = wrapper.find(ShuttlesMap).at(1);
-    expect(map.prop("center")).not.toEqual({
-      latitude: first_station.latitude,
-      longitude: first_station.longitude
-    });
-    expect(map.prop("center")).toEqual({
-      latitude: other_station.latitude,
-      longitude: other_station.longitude
-    });
+it("sthe shuttles page detail map by default centers first affected stop", () => {
```
digitalcora: `sthe`

PR #296, opened 2019-12-03
Fix combined GL shuttle map not including all shapes

digitalcora:
**Asana Ticket:** [🐞 Shuttles | Combined green line tab doesn't show all green lines](https://app.asana.com/0/477545582537986/1152219630248609/f)

In the case where shuttle data was requested for multiple routes but only some of those routes had a shuttle diversion, the returned stops would be from all requested routes, but the shapes would only be those corresponding to the diverted routes.

This occurred because we were building the shapes based on trips that had an atypical route pattern, but for routes unaffected by a diversion, none of the trips would have an atypical route pattern.

We resolve this by adding an extra set of trips to the shape building process, consisting of those which 1) are typical, and 2) do not belong to the routes affected by active shuttle alerts.

(this also includes, as a separate commit, a tiny fix to our `TripsHack` workaround to account for a new shuttle date)


apps/site/lib/site/shuttle_diversion.ex
```diff
@@ -214,4 +210,37 @@ defmodule Site.ShuttleDiversion do
...
+  end
+
+  defp trips_on_unaffected_routes(route_ids, trips, time) do
+    # Since trips can only be retrieved by service date and not by a finer-grained time window,
+    # the overall collection of trips may include ones that are both typical and atypical for the
+    # routes experiencing a diversion, if the diversion occurs only during part of the day. So we
+    # need to reject trips that are typical, but belong to a route that has an active diversion.
+
+    affected_route_ids =
+      route_ids
```
thecristen: Do we need to compute `affected_route_ids` this way as opposed to using the trips from `trips_on_affected_routes/1` and getting the route ids from there?

digitalcora: I think it'd end up being the same set of IDs; there's no cost to doing it one way or the other, but I do like using the alert data to make decisions where possible, since they're the most reliable indicator we have that something is actually a shuttle service (when we have things not "reachable" from an alert, like the shapes, we have to make an educated guess).

PR #299, opened 2019-12-05
Rework how we handle holidays in the schedule finder

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Schedule finder | Determine holiday services using typicality](https://app.asana.com/0/555089885850811/1152227770350433)

* We take advantage of the fact that holiday data will soon be more consistent (try this against dev-green) to determine which services represent holidays, and which holidays they represent, in a simpler and more robust manner.
* We drop some of the text out of the `option`s for holidays, bringing them in accordance with the new information design in the finder.

### DIFF
--- a/apps/site/assets/ts/helpers/__tests__/service-test.ts
+++ b/apps/site/assets/ts/helpers/__tests__/service-test.ts
@@ -1,5 +1,4 @@
 import {
-  serviceDays,
   serviceDate,
   groupServiceByDate,
   groupByType,
@@ -39,45 +38,6 @@ const serviceWithDate: ServiceWithServiceDate = {
   service_date: "2019-06-25",
   name: "weekday"
 };
-
-describe("serviceDays", () => {
-  it("returns an empty string for weekends", () => {
-    const saturday: Service = {
-      ...service,
-      description: "Saturday schedule",
-      type: "saturday",
-      valid_days: [6]
-    };
-    expect(serviceDays(saturday)).toBe("");
-
-    const sunday: Service = {
-      ...service,
-      description: "Sunday schedule",
-      type: "sunday",
-      valid_days: [7]
-    };
-    expect(serviceDays(sunday)).toBe("");
-  });
-
-  it("returns a single day in parentheses", () => {
-    const friday: Service = { ...service, valid_days: [5] };
-    expect(serviceDays(friday)).toBe("Friday");
-  });
-
-  it("returns consecutive days as Monday - Friday", () => {
-    expect(serviceDays(service)).toBe("Monday - Friday");
-    const monToWed: Service = { ...service, valid_days: [1, 2, 3] };
-    expect(serviceDays(monToWed)).toBe("Monday - Wednesday");
-
-    const wedToFri: Service = { ...service, valid_days: [3, 4, 5] };
-    expect(serviceDays(wedToFri)).toBe("Wednesday - Friday");
-  });
-
-  it("lists all non-consecutive days", () => {
-    const monWedFri: Service = { ...service, valid_days: [1, 3, 5] };
-    expect(serviceDays(monWedFri)).toBe("Monday, Wednesday, Friday");
-  });
-});
 
 const currentService = {
   ...serviceWithDate,
--- a/apps/site/assets/ts/helpers/__tests__/service-test.ts
+++ b/apps/site/assets/ts/helpers/__tests__/service-test.ts
@@ -75,61 +75,74 @@ describe("serviceDate", () => {
 describe("groupServiceByDate", () => {
   it("groups current schedules with the proper service period", () => {
     const groupedService = groupServiceByDate(currentService);
-    expect(groupedService).toEqual({
-      type: "current",
-      servicePeriod: "ends July 28",
-      service: currentService
-    });
+    expect(groupedService).toEqual([
+      {
+        type: "current",
+        servicePeriod: "ends July 28",
+        service: currentService
+      }
+    ]);
   });
 
   it("groups upcoming schedules as future with proper service period", () => {
     const groupedService = groupServiceByDate(upcomingService);
-    expect(groupedService).toEqual({
-      type: "future",
-      servicePeriod: "starts July 1",
-      service: upcomingService
-    });
+    expect(groupedService).toEqual([
+      {
+        type: "future",
+        servicePeriod: "starts July 1",
+        service: upcomingService
+      }
+    ]);
   });
 
   it("groups holiday schedules with proper service period", () => {
-    const groupedService = groupServiceByDate({
-      ...service,
-      service_date: "06-25-19"
-    });
-    expect(groupedService).toEqual({
-      type: "holiday",
-      servicePeriod: "on June 25",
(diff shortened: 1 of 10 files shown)


thecristen:
Could use a different way of displaying when there are _no_ holidays, e.g. hide the `optgroup` label.
image

apps/site/assets/ts/helpers/__tests__/service-test.ts
```diff
@@ -115,61 +80,74 @@ describe("serviceDate", () => {
...
-    expect(groupedService).toEqual({
-      type: "holiday",
-      servicePeriod: "on June 25",
-      service: { ...service, service_date: "06-25-19" }
-    });
+    const groupedService = groupServiceByDate(holidayService);
+    expect(groupedService).toEqual([
+      {
+        type: "holiday",
+        servicePeriod: "St. Transit's Day, June 25",
```
thecristen: Nice touch.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
```diff
@@ -182,11 +165,20 @@ export const ServiceSelector = ({
               return (
                 <optgroup key={group} label={optGroupTitles[group]}>
                   {servicesByOptGroup[group].map((service: ServiceByOptGroup) =>
-                    serviceDescription(
-                      service.service,
-                      group,
-                      service.servicePeriod,
-                      multipleWeekdays
+                    group === "holiday" ? (
+                      <HolidayServiceOption
```
thecristen: Is there a particular reason you chose to have `<HolidayServiceOption>` and `<ServiceOption>` be two separate components, instead of handling this `group === "holiday"` condition inside a consolidated `<ServiceOption>`?

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOption.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const isMultipleWeekday =
+    multipleWeekdays &&
+    service.type === "weekday" &&
+    service.typicality !== "holiday_service";
+  return (
+    <option value={service.id}>
+      {isMultipleWeekday
+        ? `${serviceDays(service)} schedule`
+        : service.description}
+      {group !== "holiday" ? ", " : " "}
```
thecristen: image

What happened here? I'm wondering if there's an inconsistency between the `group` value of "holiday" and the `service.typicality` value of "holiday_service" -- the output suggests that in this particular example for the day after thanksgiving, `isMultipleWeekday` is true, `serviceDays(service)` outputs nothing, and line 65 outputs `" "`. But it shows under the Holiday Schedules optgroup, but not rendering with `HolidayServiceOption`.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
phildarnowsky: It's basically because the logic for `HolidayServiceOption` is simple and that for `ServiceOption` is complicated, and also because when we call out storm service I'm contemplating having a third `StormServiceOption`. In the near future we'll be simplifying `ServiceOption` as well and perhaps at that point it'll make sense to combine the three together.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOption.tsx
ingridpierre: @thecristen For this branch I'm seeing an output that looks very different from what you screenshotted here. For example I see: "Thanksgiving Day, Nov 28" rather than the "Holiday (Sunday schedule) Thanksgiving Day" in your screenshot.

Am I missing something?

I do see "schedule Day after Thanksgiving Day" come through as well though for me it's in future schedules like so:
image

ingridpierre:
Took a look just now, but I had some questions in slack that may affect this work. Let's check in next week about the following:

https://mbtace.slack.com/archives/C8THJ1NJY/p1575681138010800

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOption.tsx
thecristen: @ingridpierre The screenshot is what I saw/still see on dev green: https://green.dev.mbtace.com/schedules/751/line 🤷‍♀

phildarnowsky: @ingridpierre @thecristen ah, so I see I left out one important thing, when I said to try this against dev-green, I meant the dev-green _API_. That is, I ran the server locally, but with the env var `V3_URL=https://green.dev.api.mbtace.com`.

When I do that, this is the selector I see for that route:

Screen Shot 2019-12-09 at 1 54 32 PM

phildarnowsky:
@ingridpierre are you cool with @thecristen 's suggestion above, to leave out the holiday `<optgroup>` when there are no holidays?

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOption.tsx
thecristen: Ohhhh, thanks that makes a huge difference. :sweat_smile: Now I can see what you see for the SL4 as well. 👍 

What I get for the 39 bus is pretty different, (for example Monday - Friday instead of "Weekday") -- is this discrepancy expected?

image

ingridpierre:
@phildarnowsky yes, I think we can leave it off when it's empty. Same for Future Schedules, while we're at it.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOption.tsx
ingridpierre: We want it to say "Weekday". Ideally like this ([see my slack thread](https://mbtace.slack.com/archives/C8THJ1NJY/p1575681138010800))
Pasted Graphic 13

so that it ultimately looks like this:
Pasted Graphic 14

phildarnowsky: @ingridpierre I just realized that one thing we don't have a card for, is to make the current schedules consistently read "Weekday", "Saturday", or "Sunday" as appropriate, i.e. none of this "Monday-Friday schedule" business anymore. Is that something we need for Thursday or can we live with the current rendering a little longer (assuming that we get rid of redundant schedules)?

phildarnowsky:
@thecristen refactored a bit, and now we also leave out optgroups with no service.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOption.tsx
thecristen: I found it concerning that the output isn't consistent here between Monday-Friday and Weekday, but let's address that in a followup PR.

phildarnowsky:
@ingridpierre do you have anything further on this PR?

ingridpierre:
@phildarnowsky sorry for the delay. this lgtm now 👍

PR #303, opened 2019-12-06
Omit past services from schedule finder

phildarnowsky:
Services that are over are just noise in the schedule finder: leave them
out.

#### Summary of changes
**Asana Ticket:** [Schedule finder | Don't display past schedules](https://app.asana.com/0/555089885850811/1152227770350445)

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line_controller.ex
@@ -44,13 +44,6 @@ defmodule SiteWeb.ScheduleController.LineController do
   end
 
   def assign_schedule_page_data(conn) do
-    service_date = Util.service_date()
-
-    services =
-      conn.assigns.route.id
-      |> ServicesRepo.by_route_id()
-      |> dedup_services()
-
     assign(
       conn,
       :schedule_page_data,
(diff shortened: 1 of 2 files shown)


paulswartz:
How many services does this filter out?

phildarnowsky:
@paulswartz Right now on dev-green I'm seeing 12 services out of a possible 166 with an end date in the past. In prod, 12 out of 71.

ingridpierre:
TL;DR: This change looks great, but has exposed some (hopefully?) small work still needed to properly classify past/current/future services. 

See: [Potential follow-up ticket](https://app.asana.com/0/385363666817452/1152870274795137/f)
----
This pr massively improved a ton of the routes I was able to look at. I  noticed that we are now correctly listing Saturday schedules before Sunday schedules (previously they were out of order), which is also great. 

However: there are some misclassified current/future services, observable on routes like the SLs, 39.

Basically, it seems we're calling some schedules "future" when they're actually part of the current rating. [I've created a followup ticket here.](https://app.asana.com/0/385363666817452/1152870274795137/f) I think it's related to this PR since it's about how we classify services as current/future/past, but I'm cool if you think it should be addressed separately, since there are some open questions there too.

image
image

phildarnowsky:
@ingridpierre I'm thinking to address the current/future confusion in a separate PR; it's not as closely related to what we're doing here as it might appear.

ingridpierre:
@phildarnowsky if that's the case then I think this is good to go from my perspective, definitely cleans up a good amount of noise as-is!

PR #305, opened 2019-12-10
Omit services that are subsets of other services

phildarnowsky:
We have logic in the service picker that gets rid of services with the
same start date, end date, and valid days as another service. However,
there are other cases in which we have two services, one of which is a
subset of another, rather than an exact match. These subsets are
redundant too: omit them.

#### Summary of changes
**Asana Ticket:** [Services Picker | De-dupe services that differ only by what garage they run out of](https://app.asana.com/0/555089885850811/1152227770350437)

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line_controller.ex
@@ -83,8 +83,38 @@ defmodule SiteWeb.ScheduleController.LineController do
     )
   end
 
-  @spec dedup_services([Service.t()]) :: [Service.t()]
-  def dedup_services(services) do
+  # If we have two services A and B with the same type and typicality,
+  # with the date range from A's start to A's end a subset of the date
+  # range from B's start to B's end, either A is in the list of services
+  # erroneously (for example, in the case of the 39 in the fall 2019
+  # rating), or A represents a special service that's not a holiday (for
+  # instance, the Thanksgiving-week extra service to Logan on the SL1 in
+  # the fall 2019 rating).
+  #
+  # However, in neither of these cases do we want to show service A. In the
+  # first case, we don't want to show A because it's erroneous, and in the
+  # second case, we don't want to show A for parity with the paper/PDF
+  # schedules, in which these special services are not generally called
+  # out.
+
+  @spec dedup_similar_services([Service.t()]) :: [Service.t()]
+  def dedup_similar_services(services) do
+    services
+    |> Enum.group_by(&{&1.type, &1.typicality})
+    |> Enum.flat_map(fn {_, service_group} ->
+      service_group
+      |> Enum.reject(fn service ->
+        Enum.any?(service_group, fn other_service ->
+          other_service != service &&
+            Date.compare(other_service.start_date, service.start_date) != :gt &&
+            Date.compare(other_service.end_date, service.end_date) != :lt
+        end)
+      end)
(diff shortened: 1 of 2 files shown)


digitalcora:
As a general comment: the task here is _"de-dupe services that differ only by what garage they run out of"_, but what the code implements is _"filter out services that are a date-wise subset of other services"_. Do these always coincide? They seem mostly unrelated, and (admittedly with a limited knowledge of how services work) I can envision wanting to _not_ filter out a service that is a date-wise subset of another.

phildarnowsky:
@digitalcora so the chain of events that led to the redundant schedules on the 39 is detailed in https://app.asana.com/0/385363666817452/1152870274795137. It's complicated but tl;dr buses for the 39 run out of two different garages, and due to an edge case in scheduling, service out of the two garages doesn't get combined in the usual way on the data side of the house. This resulted in two nearly-identical services, that differed only in ID and start date.

I don't believe there's a valid use case for multiple services with one a date-wise subset of the other and the same type and typicality--maybe if the `valid_days` were different, which is something I could check here. I'll ask the data team and see if that's correct.

paulswartz:
What happens with the Silver Line routes here? They run additional service, and are the reason that the 39 has the additional service IDs.

phildarnowsky:
@paulswartz How is the extra Thanksgiving-season service modeled?

paulswartz:
It was a separate service ID, active Monday, Tuesday, Wednesday, from 2019-11-25 to 2019-11-27. It's marked in HASTUS as typical Weekday service, which is also how we marked it in GTFS.

phildarnowsky:
@paulswartz OK. How was this carved out of normal weekday service, with `removed_dates` or what?

paulswartz:
Yep, the normal schedule had a start date of 2019-10-30 (when the particular GTFS started) and an end date of 2019-12-20 (last Friday of the rating). It then removed the week of Thanksgiving with `removed_dates`.

phildarnowsky:
@paulswartz @digitalcora  OK. So suppose we alter this PR to take `removed_dates` into account when deciding if one service is a datewise subset of another. That way we'd have filtered out the duplicate services on the 39, but not the Thanksgiving-week special service on the SL1.

@ingridpierre We need to figure out how to represent service like the Thanksgiving-week special service on the SL1 we're talking about here. What I'm thinking is adding it to "Current Service"/"Future Service" labelled "Special Weekday Service: November Xth to November Yth". This is a case where there really is more than one valid weekday service for a route and rating.

ingridpierre:
@phildarnowsky The goal of this work is to bring us more in line with what we publish in the pdf, which sticks to weekday/saturday/sunday and a common list of affected holiday dates.

 For example, the way the holiday is notated on the pdf for the SL1/2/3 is shown below. It does not include those other nuanced dates like 11/29, or any other days around thanksgiving.

image
image

As you can see from the above, though we have more data for non-holiday-holiday dates they were not considered significant enough to be communicated to the public by Schedules within schedules. I think it's fine if this work does not accommodate them in the picker, until we are told otherwise.

digitalcora:
@phildarnowsky Are you planning to handle added/removed dates in this PR, or separately?

phildarnowsky:
@digitalcora Based on @ingridpierre 's comment above, if we want to omit things like the Thanksgiving-season special service from the picker, I'm thinking _not_ to handle the removed-dates case until further notice.

PR #306, opened 2019-12-11
Show stop municipality if address is not available

digitalcora:
**Asana Ticket:** [🚉  Add municipality to stop pages](https://app.asana.com/0/477545582537986/1122431599573158/f)

This incidentally renames `AddressBlock` to `LocationBlock` since this component will not always contain an address, but does always contain items related to the stop's physical location.

**Note:** If testing this branch locally, the server must be restarted for the new info to show up. This is because we're proactively caching the final parsed version of all stop data on startup, so changes to the parsing don't take effect until a restart.

Screen Shot 2019-12-11 at 11 11 36 AM


ingridpierre:
Lgtm!! 👍 

Sorry for the wait, this looked great even before I took a detailed peek, but I figured I'd share the following so you can see some cases where this helps a ton with orientation (especially via search). 

I think this example could also be good to demo when this is merged this week. See below where we have 6 different "Washington St @ School St"  stops in different cities, that each lacked addresses. Only very few stops have physical addresses so this improved thousands of pages!

image

Stop IDs:
11187 | Washington St @ School St
1287 | Washington St @ School St
13618 | Washington St @ School St
35236 | Washington St @ School St
3940 | Washington St @ School St
90620 | Washington St @ School St

digitalcora:
@ingridpierre Sounds like a good idea, I will totally steal that screenshot 😁

PR #309, opened 2019-12-13
Tell bots not to index line page with date param

phildarnowsky:
No ticket.

A number of web crawler bots (the main offender is Bingbot but there are
others) are crawling our line pages, which is fine. However, they're
apparently trying to crawl schedules with varying `date`s in the query
params, with dates from years in the past to years in the future. This
is causing enough 500 errors on deploy to cause deploys to fail.

Here we update `robots.txt` to disallow crawling of line pages with
dates in the query params.

In dev-green this essentially put a stop to 500 errors:

Screen Shot 2019-12-13 at 5 26 16 AM

### DIFF
--- a/apps/site/assets/static/robots.txt
+++ b/apps/site/assets/static/robots.txt
@@ -6,3 +6,4 @@ Disallow: /trip_planner?*
 Disallow: /trip-planner?*
 Disallow: /search?*
 Disallow: /*?*preview*&vid=*
+Disallow: /schedules/*/line?*date*


phildarnowsky:
@digitalcora @thecristen @ryan-mahoney @paulswartz can whoever is up and online first please look at this one urgently? It's holding up dotcom prod deploys and hence the new rating.

paulswartz:
What about fixing the crash? I feel like this used to work, with a message saying "there's no service on that date"

phildarnowsky:
@ryan-mahoney I am certain, there are indeed simulators and I tried it. It only blocks requests if the `date` param is present.

phildarnowsky:
@paulswartz I have a separate ticket to look for and fix the crash. In the meantime I don't see much point in indexing pages for dates years in the past; if there is one, we can always change `robots.txt` again after fixing the crash.

phildarnowsky:
@ryan-mahoney I'm going to fix the crash separately, and I want to get this merged ASAP in the meantime so we can deploy dotcom so we can deploy the rating.

ryan-mahoney:
please create a ticket for the issue if it doesn't already exist :)

PR #310, opened 2019-12-13
Don't 500 on line page with date outside rating

phildarnowsky:
As of 65e97eb765d0b0a97e85a09cda3b325905c352d9, the line page accepts a
`date` query param, which enables Backstop testing on that page to work
properly again. However, it also introduced a regression, where if the
date given has no service available in the API, rather than just
returning live predictions, PredictedSchedule.group/3 was choking on the
struct representing the "no service" error. Fixed.

#### Summary of changes
**Asana Ticket:** [Date in params outside of rating shouldn't blow up line page](https://app.asana.com/0/385363666817452/1153779766848073)

### DIFF
--- a/apps/site/lib/predicted_schedule.ex
+++ b/apps/site/lib/predicted_schedule.ex
@@ -74,6 +74,10 @@ defmodule PredictedSchedule do
     |> Enum.sort_by(sort_fn)
   end
 
+  defp create_map({:error, [%JsonApi.Error{code: "no_service"}]}) do
+    %{}
+  end
+
   defp create_map(predictions_or_schedules) do
     Map.new(predictions_or_schedules, &group_transform/1)
   end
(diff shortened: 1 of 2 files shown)


phildarnowsky:
@thecristen @paulswartz As we're fetching them currently, the predictions ignore the `date` parameter and instead use the real current time. We could easily change the code to not return predictions if these are out of sync, I'm testing something like that right now. TBH I don't know how much these pages are still even user-facing but that's something I don't have time to research right now.

paulswartz:
To me, this doesn't seem like something that `PredictedSchedule.group/2` should be dealing with. What's fetching the schedules that get passed into that? Maybe it shouldn't bother fetching the predictions or building a group if the date is outside the rating?

phildarnowsky:
@paulswartz yes that's a better approach. new branch shortly.

PR #311, opened 2019-12-13
Don't show schedules or predictions outside rating

phildarnowsky:
As of 65e97eb, the line page accepts a date query param, which enables
Backstop testing on that page to work properly again. However, it also
introduced a regression, where if the date given has no service
available in the API, rather than just returning live predictions,
TransitNearMe.time_data_for_route_by_stop/3 was choking on the error
struct indicating "no service". Fixed.

#### Summary of changes
**Asana Ticket:** [Date in params outside of rating shouldn't blow up line page](https://app.asana.com/0/555089885850811/1153779766848073)

### DIFF
--- a/apps/site/lib/site/transit_near_me.ex
+++ b/apps/site/lib/site/transit_near_me.ex
@@ -408,16 +408,25 @@ defmodule Site.TransitNearMe do
     date = Keyword.get(opts, :date, Util.service_date())
     schedules_fn = Keyword.get(opts, :schedules_fn, &Schedules.Repo.by_route_ids/2)
 
-    route_id
-    |> expand_route_id()
-    |> schedules_fn.(direction_id: direction_id, date: date)
-    |> get_predicted_schedules([route: route_id, direction_id: direction_id], opts)
-    |> Enum.group_by(&PredictedSchedule.route(&1).id)
-    |> Enum.flat_map(&schedules_for_route(&1, %{}, [], opts).stops_with_directions)
-    |> convert_route_time_data_to_map()
-    |> Map.new(fn {stop_id, time_data} ->
-      {stop_id, limit_route_time_data(time_data)}
-    end)
+    schedule_data =
+      route_id
+      |> expand_route_id()
+      |> schedules_fn.(direction_id: direction_id, date: date)
+
+    case schedule_data do
+      {:error, [%JsonApi.Error{code: "no_service"}]} ->
+        %{}
+
+      _ ->
+        schedule_data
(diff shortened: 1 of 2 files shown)


apps/site/lib/site/transit_near_me.ex
```diff
@@ -408,16 +408,25 @@ defmodule Site.TransitNearMe do
...
+    schedule_data =
+      route_id
+      |> expand_route_id()
+      |> schedules_fn.(direction_id: direction_id, date: date)
+
+    case schedule_data do
+      {:error, [%JsonApi.Error{code: "no_service"}]} ->
+        %{}
+
+      _ ->
```
paulswartz: if there's another kind of error (like a timeout), do you want to do something else?

apps/site/test/site/transit_near_me_test.exs
```diff
@@ -461,5 +461,34 @@ defmodule Site.TransitNearMeTest do
...
+             source: %{"parameter" => "date"}
+           }
+         ]}
+      end
+
+      actual =
+        TransitNearMe.time_data_for_route_by_stop(@route.id, 1,
+          schedules_fn: schedules_fn,
+          predictions_fn: predictions_fn,
+          now: @now
```
thecristen: should this be `date: @now` instead of `now: @now` ?

apps/site/lib/site/transit_near_me.ex
phildarnowsky: For now, at least, I think the best thing to do is let it crash.

apps/site/test/site/transit_near_me_test.exs
phildarnowsky: no, this is a different parameter with a misleadingly similar name

phildarnowsky:
@paulswartz yes, it's common for the Elixir tests in dotcom CI to have timeouts, and this appears to be the case here. It's on the list of tech debt to fix up.

digitalcora:
I'm curious why we're not considering this a "let it crash" scenario. The `date` param is (I thought) exclusively a debugging tool — nobody using it should be surprised if they get an error page when operating beyond the time range of any known service.

phildarnowsky:
@digitalcora there's a reason, it's complicated but ultimately we're not actually doing it for any user's benefit but rather we're blocked from deploying dotcom if we let this page crash with out-of-service dates. I don't have time at the moment but if you're interested I can explain in more detail after demos.

digitalcora:
> we're blocked from deploying dotcom if we let this page crash

Well, this wouldn't be the weirdest thing I've learned about the site 😆

PR #312, opened 2019-12-16
Call out storm service

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Services Picker | Call out storm schedules](https://app.asana.com/0/477545582537986/1152227770350441)

Show storm schedules explicitly in the schedule picker. They can be distinguished by their typicality of 5, i.e. `unplanned_disruption`.


apps/site/lib/site_web/controllers/schedule/line_controller.ex
```diff
@@ -44,6 +44,8 @@ defmodule SiteWeb.ScheduleController.LineController do
   end
 
   def assign_schedule_page_data(conn) do
+    service_date = conn.assigns.date_time |> Util.service_date()
```
digitalcora: We have `conn.assigns.date` for just this purpose; no need to convert from a datetime 🙂

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOptGroup.tsx
```diff
@@ -13,6 +14,42 @@ interface Props {
...
+    return (
+      <HolidayServiceOption
+        key={service.service.id + service.servicePeriod}
+        service={service.service}
+        servicePeriod={service.servicePeriod}
+      />
+    );
+  }
+
+  if (service.service.typicality === "unplanned_disruption") {
```
digitalcora: Since the `ServiceOption` itself could make this distinction (as it has the full `service`), would it be better for this conditional to live there and not have a separate `StormServiceOption`?

ingridpierre:
@phildarnowsky Can we have the label read **"Storm** schedule, Date**" instead of "starts on Date". The reason being that we don't usually decide to run on a storm schedules for more than one day or more than 1 day in advance. Therefore 'starts on' is too vague and a little strange for this use-case.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOptGroup.tsx
phildarnowsky: This is basically an artifact of a refactoring that didn't work out, I'll try re-refactoring.

phildarnowsky:
@ingridpierre @digitalcora the three new commits should address your feedback

phildarnowsky:
@ingridpierre good on this one?

PR #313, opened 2019-12-17
Fix missing stops in Schedule Finder origin picker

digitalcora:
The root of this issue is that the stop data for the line page is being filtered by which stops are associated with a shape, which we're using to determine which stops are on which "branch" of a line. However, this is not a supported way to use shapes (the `stops` relationship on shapes may be deprecated in the future), and we think this is why some of the stops were missing.

Luckily the Schedule Finder origin picker only requires a flat list of stops on all branches/variants of the line, so in this case we can just send it the list of stops on the route, prior to the shape filtering.

**Asana Ticket:** [🐞 Schedule Finder origin picker is missing stops](https://app.asana.com/0/477545582537986/1146873289023569/f)


ingridpierre:
MASSIVE improvement here, so pleased with these changes! I was trying to find a good example that would make a great demo, I would love to be able to quantify just how many stops were missing. But everything here lgtm 👍 , so glad to see this fixed!

PR #315, opened 2019-12-17
Adjust style and behavior of Photo Gallery component

amaisano:
#### Summary of changes
**Asana Ticket:** [ Bug | Photo gallery: 6+ photos break padding, 7th breaks previous/next](https://app.asana.com/0/399597520748778/1122494857555453)

- Fix: spread thumbnails equally across available space using %
- Fix: use TOTAL thumbnail count instead of page thumbnail count to hide nav

image

### DIFF
--- a/apps/site/assets/css/_photo-gallery.scss
+++ b/apps/site/assets/css/_photo-gallery.scss
@@ -35,8 +35,13 @@
     display: none;
   }
 
+  // Remove inline, vertical space between rows of thumbnails
+  .js & {
+    line-height: 0;
+  }
+
   &.c-thumbnail-count--0,
-  &.c-thumbnail-count--1 {
+  &.c-thumbnail-total--1 {
     display: none;
 
     + .c-photo-gallery__nav-btns {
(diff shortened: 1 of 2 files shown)


aayesha:
Reviewed on Adamo's machine. 👍 Looks good! Thank you!!

apps/site/assets/css/_photo-gallery.scss
```diff
@@ -35,8 +35,12 @@
     display: none;
   }
 
+  .js & {
+    font-size: 0;
```
thecristen: What is this supposed to do?

amaisano: Instead of making each thumbnail link `display: block; float:left`, we're using `display: inline-block` which inherits the current font size, causing extra space to occur between wrapped lines. Setting font-size to zero eliminates this discrepancy, but only on browsers with JS enabled (non-JS browsers need font size to show the alt text for each thumbnail image).

thecristen: Gotcha - I'd add a tiny comment on this line to mention that. 

After tinkering with this a bit, I have some observations:

* Setting this to match the `<a>` margin you specified (.25%) instead of 0 also produces the desired outcome, and makes it a bit more clear what this is trying to achieve (match the spacing)
* This fix also works with line-height, which IMO makes more sense to change here than font-size (since there's no visible text here, and what we're looking at is essentially the line height of a line of images).

PR #317, opened 2019-12-19
Add tooltip to Line Diagram connections

thecristen:
#### Summary of changes
**Asana Ticket:** [Line Diagram | Connection Tooltips](https://app.asana.com/0/385363666817452/1141229989083703/f)

* Uses Bootstrap tooltip on connection link elements
* Adds method to output descriptive connection names
* Adds small spacing change to connection links to improve focus outline appearance

### DIFF
--- a/apps/site/assets/ts/helpers/icon.tsx
+++ b/apps/site/assets/ts/helpers/icon.tsx
@@ -107,3 +107,44 @@ export const caret = (
     />
   );
 };
+
+/* 
+A minimal component that gives its child component 
+a basic tooltip containing the specified text. Has no
+visual appearance itself.
+
+Note: data-selector="true" and data-original-title are
+specified here in order to preserve `title` text post-trigger
+*/
+export const TooltipWrapper: React.FC<{
+  children: JSX.Element;
+  tooltipText: string;
+  tooltipOptions?: object; // corresponds to Bootstrap tooltip options
+  href?: string;
+}> = ({ children, tooltipText, tooltipOptions, href }): JSX.Element => {
+  const { trigger, placement, animation } = Object.assign(
+    {
+      trigger: "hover focus",
+      placement: "top",
+      animation: "true"
+    },
+    tooltipOptions
+  );
+
+  const Tag = href ? "a" : "span";
+
+  return (
+    <Tag
+      href={href}
+      data-toggle="tooltip"
+      data-trigger={trigger}
+      data-placement={placement}
+      data-animation={animation}
+      data-selector="true"
+      data-original-title={tooltipText}
+      title={tooltipText}
+    >
+      {children}
+    </Tag>
+  );
+};
--- a/apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
+++ b/apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import renderer from "react-test-renderer";
+import { mount } from "enzyme";
 import LineDiagram from "../components/LineDiagram";
 import { StopType } from "../../__v3api";
 import { LineDiagramStop } from "../components/__schedule";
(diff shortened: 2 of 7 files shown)


apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -70,7 +80,17 @@ const LineDiagram = ({
...
-                    <a href={`/schedules/${route.id}/line`} key={route.id}>
+                    <a
+                      href={`/schedules/${route.id}/line`}
+                      key={route.id}
+                      data-toggle="tooltip"
+                      data-trigger="focus"
+                      data-placement="bottom"
+                      data-animation="false"
+                      data-selector="true"
+                      data-original-title={connectionName(route)}
```
digitalcora: I'm curious, why do we need both `title` and `data-original-title`?

apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
```diff
@@ -506,4 +507,16 @@ describe("LineDiagram", () => {
...
   });
+
+  it("has a tooltip for a transit connection", () => {
+    const wrapper = mount(<LineDiagram lineDiagram={lineDiagram} />);
+    const stopConnections = wrapper.find(
+      ".m-schedule-line-diagram__connections a"
+    );
+    stopConnections.forEach(connectionLink => {
+      const props = connectionLink.props();
+      expect(props.title).toBeTruthy();
```
digitalcora: Can we have something that asserts the title content is what we want it to be, given the data? I think we should have the `connectionName` logic tested somewhere since that's a key part of this feature.

apps/site/assets/ts/schedule/components/LineDiagram.tsx
thecristen: This reflects a total Bootstrap v.4.0.0-alpha.2 quirk: adding the `data-selector="true"` and `data-original-title` intercepts the tooltip's normal behavior of resetting the `title` to "" and inserting the title into `data-original-title` upon activation. I'd like to preserve the title rather than ever setting it to an empty string, for accessibility's sake.

ingridpierre:
Looking good so far! A couple things:
-  (paraphrasing IRL discussion) Commuter Rail icon should just say 'Commuter Rail'.
-  I think the facilities icons like parking and accessibility still need a tooltip. "Parking", and "Accessible" is what we use elsewhere.
image

I really like your touch here of making the tooltip pop downward so it doesn't cover the stop name, super thoughtful.

apps/site/assets/ts/schedule/components/LineDiagram.tsx
digitalcora: Ah, that does seem important. Do we know which version of Bootstrap fixes this? Can we call it out in a comment somewhere so we remember to remove it when the time comes?

thecristen:
@ingridpierre - added the new tooltips and adjusted the commuter rail tooltips!
@digitalcora - added those additional tests and comments!

apps/site/assets/ts/helpers/icon.tsx
```diff
@@ -107,3 +107,44 @@ export const caret = (
...
   );
 };
+
+/* 
+A minimal component that gives its child component 
+a basic tooltip containing the specified text. Has no
+visual appearance itself.
+
+Note: data-selector="true" and data-original-title are
+specified here in order to preserve `title` text post-trigger
```
digitalcora: If we can figure out which Bootstrap version we could remove this in, once we upgrade, it would be good to put that information here. Good comment as it stands though 👍

thecristen: This doesn't appear to the fixed in the latest version I tried, sadly. Will have to consider revisiting in the future...

digitalcora: That's too bad 😢 Bootstrap, why do you make it hard to be accessible

ingridpierre:
Lovely, lgtm! 👍

PR #319, opened 2019-12-20
Line Diagram - View schedules

thecristen:
#### Summary of changes
**Asana Ticket:** [Line Diagram | Pre-populate origin when clicking schedules-from-here link (CR, Bus, Ferry)](https://app.asana.com/0/385363666817452/1128109170076009/f)

Adds a "View schedule" link to every stop in the Line Diagram. The link opens a modal with the relevant schedules for that origin and the currently selected direction.

### Bus:
39-modal

### Commuter Rail:
CR-modal

### Ferry
~**Ferry case to be added soon**~ Made another ticket to handle the ferry case, which will need additional work to enable.


apps/site/assets/ts/schedule/__tests__/ScheduleDirectionTest.tsx
```diff
@@ -14,11 +14,12 @@ import { EnhancedRoute } from "../../__v3api";
 import { MapData } from "../../leaflet/components/__mapdata";
 import { RoutePatternsByDirection, ShapesById } from "../components/__schedule";
 import { lineDiagram } from "./LineDiagramTest";
+import { stops } from "./ScheduleModalTest";
 
 const body =
   '<div id="body-wrapper"><div id="react-root"></div><div id="map-root"></div></div>';
 
-const route = {
```
digitalcora: Extracting this shared data into a descriptively-named file might make the tests' "dependencies" easier to follow than exporting it from one test and using it in another.

apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
```diff
@@ -537,14 +541,60 @@ export const lineDiagram = [
   }
 ] as LineDiagramStop[];
 
+const stops = lineDiagram.map(({ route_stop }) => ({
```
digitalcora: It might be more intention-revealing to call this `simpleStops`, since it matches the shape of a `SimpleStop`. (or, could we typecast it as `SimpleStop[]`?)

apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
```diff
@@ -537,14 +541,60 @@ export const lineDiagram = [
...
+      />
+    );
+  });
+
+  afterEach(() => {
+    wrapper.unmount();
+  });
+
+  it("renders and matches snapshot", () => {
+    expect(wrapper.debug()).toMatchSnapshot();
```
digitalcora: In our current snapshot tests we seem to compare against `renderer.create(...).toJSON()` rather than `wrapper.debug()`. Is there an advantage to this way? (should we start doing it for all our future tests?)

apps/site/assets/ts/schedule/__tests__/ScheduleModalTest.tsx
```diff
@@ -18,7 +18,7 @@ const route: EnhancedRoute = {
   type: 1
 };
 
-const stops: SimpleStop[] = [
```
digitalcora: Similar comment here about exporting vs. extracting shared data.

apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -61,9 +79,21 @@ const connectionName = (connection: Route): string => {
...
+  lineDiagram,
+  route,
+  directionId,
+  routePatternsByDirection,
+  services,
+  stops
 }: Props): ReactElement<HTMLElement> | null => {
-  const routeType = lineDiagram[0].route_stop.route!.type;
+  const routeType = route.type;
+  const [modalState, toggleModal] = useState<{
```
digitalcora: I think `toggleModal` might be better named `setModalState`, since calling it does not necessarily toggle the modal – you have to provide it with a complete new state, and it wouldn't necessarily change the current open/closed state of the modal.

apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
thecristen: Both end up comparing JSON structures so I wouldn't think there'd be an advantage one way or the other; they're just tools that come with different testing libraries. This would be a broader discussion of which testing libraries we use and how we use them and I'd consider this comment outside the scope of a PR review

PR #321, opened 2019-12-23
Fix CMS paragraph order being scrambled

digitalcora:
The issue was that `async_with_timeout`, which accepts a list of functions, was implemented on top of `yield_or_default_many`, which accepts a map where the values are functions. The map values were being collected back into a list, but since maps are not ordered, this result list would not have the same ordering as the input list.

Because the previous approach resulted in this data flow:

```
list -> map -> list -> Task.yield_many -> list -> map -> list
```

We can simplify to this:

```
list -> Task.yield_many -> list
```

**Asana Ticket:** [🐞💧 Paragraphs display in wrong order (follow-up)](https://app.asana.com/0/385363666817452/1153851471078822/f)

### DIFF
--- a/apps/util/lib/util.ex
+++ b/apps/util/lib/util.ex
@@ -141,10 +141,11 @@ defmodule Util do
       when is_list(functions) and is_atom(module) do
     functions
     |> Enum.map(&Task.async/1)
+    |> Task.yield_many(timeout)
     |> Enum.with_index()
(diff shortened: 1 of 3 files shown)


thecristen:
Looks good -- can you `mix format` your changes so CI can run its tests?

digitalcora:
Whoops, could have sworn I did a format right before pushing. Fixed! 😅

PR #328, opened 2019-12-27
Tweak layout of Schedule Finder on small screens

digitalcora:
**Asana Ticket:** [Schedule Finder | Mobile Layout](https://app.asana.com/0/385363666817452/1154667123836713/f)

* The Schedule Finder box is no longer in a collapsible section
* Page header (e.g. "Info & Maps") is now displayed after the box

Screen Shot 2019-12-27 at 3 12 33 PM


ingridpierre:
@digitalcora Looks great! Glad to also be rid of the flashing above the map on desktop with slow connections. Looking forward to seeing how more folks interact with this live. 👍

digitalcora:
> Glad to also be rid of the flashing above the map on desktop

I might have fixed this accidentally — it was an issue I saw after I made some changes for this branch, and I assumed I had introduced it. Happy little accidents 🌳

digitalcora:
> It looks like what _this_ code is aiming to do, across a few files, is implement logic for "if there's no schedule note, use the mobile version of `ScheduleFinder`

That's not the case, so hopefully I can clear up some confusion here 😄

`ScheduleNote` is displayed in place of `ScheduleFinder` on subway line pages (and only there). We do this because subway lines don't (officially) have individually scheduled trips, so Schedule Finder would make no sense in that context. Instead there is a note that explains the headways of the line. This is the case regardless of mobile/desktop breakpoints. This PR does not touch that behavior.

The `classModifier` code you describe is there just to make the `ScheduleFinder` appear in one of two different places depending on whether the sidebar exists in the current breakpoint. You're correct that I was following the existing code; I saw we were already passing in a `classModifier` prop, and that we already had CSS for `schedule-finder--` classes, but no such prop was defined in the component itself. I assumed it was a remnant of something that used to work, and restored it to functionality. But, you're probably right that we could do this more simply using a descendant selector in the CSS, and not need the separate `schedule-finder--` classes at all.

digitalcora:
> But, you're probably right that we could do this more simply using a descendant selector in the CSS

I pushed a new commit that does this, let me know what you think @thecristen. There were no unique IDs or classes already that could be used to make the distinction, so I had to add some new ones.

PR #329, opened 2019-12-28
Depict branches on schedules line diagram

thecristen:
#### Summary of changes
**Asana Ticket:** [Line Diagram | Draw expanded branches](https://app.asana.com/0/385363666817452/1153833236795997/f)

Continuing the schedules line diagram redesign work,

| Red Line Northbound  | Red Line Southbound |
| --- | --- |
| image | image |
| **Providence/Stoughton CR**  | **Newburyport/Rockport CR** |
| image | image |

# Caveat: 

<details>
  <summary>The combined Green line page doesn't look consistent -- the underlying data is a bit different on this page currently (mainly having no stops marked as merge stops). We'd also want to consider a more detailed design (or decide if we don't want to support this page altogether).</summary>
  <figure></figure>
</details>

### DIFF
--- a/apps/site/assets/css/_schedule-page-line-diagram.scss
+++ b/apps/site/assets/css/_schedule-page-line-diagram.scss
@@ -6,16 +6,16 @@ $line-width: 10px;
     border-bottom: 0;
     display: flex;
     padding: $base-spacing;
-  
+
     &:last-child {
       border-bottom: $border;
     }
   }
-  
+
   .m-schedule-diagram__content {
     flex-grow: 1;
   }
-  
+
   .m-schedule-diagram__card {
     align-items: flex-start;
     display: flex;
@@ -27,7 +27,7 @@ $line-width: 10px;
 
   .m-schedule-diagram__stop-name {
     font-weight: bold;
-  
+
     h4 {
       margin-top: $base-spacing / -4;
     }
@@ -36,14 +36,14 @@ $line-width: 10px;
   .m-schedule-diagram__connections {
     .m-schedule-diagram__connection {
       margin: 0 ($base-spacing / 8);
-  
-      .c-svg__icon {
-        height: $base-spacing;
-        vertical-align: middle;
-        width: $base-spacing;
-      }
     }
-  
+
+    .c-svg__icon {
+      height: $base-spacing;
+      vertical-align: middle;
+      width: $base-spacing;
+    }
+
     a {
       display: inline-block;
       line-height: 0;
@@ -51,7 +51,7 @@ $line-width: 10px;
       vertical-align: top;
     }
   }
-  
+
   .m-schedule-diagram__features {
     .m-schedule-diagram__feature-icon {
       height: $base-spacing;
@@ -59,7 +59,7 @@ $line-width: 10px;
       margin-left: $base-spacing / 4;
       width: $base-spacing;
     }
-  
+
     .c-icon__cr-zone {
       font-size: $base-spacing * .7;
       vertical-align: text-top;
@@ -116,6 +116,8 @@ $line-width: 10px;
     circle {
       fill: $white;
       stroke: currentColor;
+      stroke-width: 0;
+      transform-origin: 50%;
     }
 
     path {
@@ -131,9 +133,18 @@ $line-width: 10px;
     transform: translateY(28px);
   }
 
+  .m-schedule-diagram__line--terminus circle {
+    stroke-width: 1.15px !important;
+  }
+
+  // first stop
+  .m-schedule-diagram__stop:first-child circle {
+    transform: translateY(3px) scale(1.15);
+  }
+
   // last stop
   .m-schedule-diagram__stop:last-child circle {
-    transform: translateY(34px);
+    transform: translateY(35px) scale(1.15);
   }
 }
 
@@ -144,7 +155,7 @@ $line-width: 10px;
     height: $base-spacing * 2;
 
     circle {
-      transform: translateY(34px);
+      transform: translateY(35px) scale(1.15);
     }
   }
 }
@@ -153,14 +164,17 @@ $line-width: 10px;
   // first stop on branch
   .m-schedule-diagram__line + .m-schedule-diagram__line--terminus {
     margin-top: $base-spacing * 2;
+
+    circle {
+      transform: translateY(3px) scale(1.15);
+    }
   }
 
-  // last stop on branch
-  .m-schedule-diagram__line + .m-schedule-diagram__line--merge {
-    height: $base-spacing * 2;
+  .m-schedule-diagram__lines--merging .m-schedule-diagram__line-bend {
+    transform: scaleY(-1) translateY(calc(-#{$base-spacing * 2} - 10px));
   }
 
-  .m-schedule-diagram__line--merge .m-schedule-diagram__line-bend {
-    transform: scaleY(-1) translateY(-50px);
+  .m-schedule-diagram__lines--merging .m-schedule-diagram__line:not(:first-child) {
+    height: calc(#{$base-spacing * 2} + 10px);
   }
 }
--- a/apps/site/assets/css/_schedule-page-line-diagram.scss
+++ b/apps/site/assets/css/_schedule-page-line-diagram.scss
@@ -1,89 +1,89 @@
-.m-schedule-diagram__stop {
-  border: $border;
-  border-bottom: 0;
-  display: flex;
-  padding: $base-spacing;
-
-  &:last-child {
-    border-bottom: $border;
-  }
-}
-
-.m-schedule-line-diagram__content {
-  flex-grow: 1;
-}
-
-.m-schedule-line-diagram__card {
-  align-items: flex-start;
-  display: flex;
-}
-
-.m-schedule-line-diagram__stop-name {
-  font-weight: bold;
-
-  h4 {
-    margin-top: $base-spacing / -4;
-  }
-}
-
-.m-schedule-line-diagram__connections {
-  .m-schedule-line-diagram__connection {
-    margin: 0 ($base-spacing / 8);
-
-    .c-svg__icon {
-      height: $base-spacing;
-      vertical-align: middle;
-      width: $base-spacing;
-    }
-  }
-
-  a {
-    display: inline-block;
-    line-height: 0;
-    max-height: $base-spacing * 1.5;
-    vertical-align: top;
-  }
-}
-
-.m-schedule-line-diagram__features {
-  .m-schedule-line-diagram__feature-icon {
-    height: $base-spacing;
-    line-height: $base-spacing;
-    margin-left: $base-spacing / 4;
-    width: $base-spacing;
-  }
-
-  .c-icon__cr-zone {
-    font-size: $base-spacing * .7;
-    vertical-align: text-top;
-    width: auto;
-  }
-}
-
-.m-schedule-line-diagram__card-left {
-  flex-grow: 1;
-}
-
-.m-schedule-line-diagram__footer {
-  text-align: center;
-
-  .btn {
-    padding: unset;
-  }
-}
-
-//
-//  Schedule line diagram
-//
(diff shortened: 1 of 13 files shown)


thecristen:
There's one set of failing backstop tests that I can't reproduce, where the 1 bus shape is missing from the schedule line page map for some reason 🤔

digitalcora:
> The combined Green line page doesn't look consistent -- the underlying data is a bit different on this page currently (mainly having no stops marked as merge stops).

How does the current combined-green-line page draw these branches, given it's working with the same data? (or is it actually _not_ working with the same data?)

thecristen:
> > The combined Green line page doesn't look consistent -- the underlying data is a bit different on this page currently (mainly having no stops marked as merge stops).
> 
> How does the current combined-green-line page draw these branches, given it's working with the same data? (or is it actually _not_ working with the same data?)

It's just not branched on the current one -- the stop icons are relied upon here to indicate shared stations.

image

digitalcora:
Ah, it's an independent straight line for each Green Line. I should have taken a glance at it before commenting, rather than relying on my sieve-like memory 😅

Is there a specific design clarification we need here, or is it just a question of whether we can work the required data into the existing backend within a reasonable time? From my recent adventures in this code I got the impression the combined GL was being handled _somewhere_... I can do some spelunking if you think that'd be helpful.

ingridpierre:
1. **Let's put the Green Line in it's own batch of follow-up work** to be addressed separately, I think we have the design intent and I'm happy to answer questions on that, but for the additional work I've added a ticket here: https://app.asana.com/0/385363666817452/1155323462935218/f 

2. Small thing, the **subway branches should be labeled as "Branch"** (e.g. Ashmont Branch, Green Line D Branch) rather than lines. Commuter Rail Lines are more commonly referred to as separate lines so they can stay labeled as lines, like they are now (e.g. Providence Line, Stoughton Line). Desired:
image
image

3. Line "ends" not appearing rounded. The origin/starts of lines look good though. Current:
image

apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -97,120 +128,196 @@ const LineDiagram = ({
     modalOpen: false
   });
 
+  const isBranchEnd = (index: number): boolean =>
```
digitalcora: Would it be reasonable to define these functions outside the component function, so they are not defined as part of rendering the component?

apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -97,120 +128,196 @@ const LineDiagram = ({
...
+    index ===
+      lineDiagram.lastIndexOf(
+        lineDiagram.filter(isOnSecondaryBranch)[
+          lineDiagram.filter(isOnSecondaryBranch).length - 1
+        ]
+      ) ||
+    (index > 0 &&
+      lineDiagram
+        .filter(l => isTerminusStop(l) && !isBeginningStop(l))
+        .includes(lineDiagram[index]));
```
digitalcora: These conditionals are a bit tricky to follow — I'm not even sure what that middle part with the `filter`s does. Could we assign each part of the condition to a descriptively-named variable? Or extract functions?

apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -97,120 +128,196 @@ const LineDiagram = ({
...
+
+                {isOnSecondaryBranch(lineDiagramStop) && (
+                  <div className="m-schedule-diagram__line m-schedule-diagram__line--branch">
+                    <svg>
+                      {isMergeStop(lineDiagramStop) && (
+                        <path
+                          d={`
+                            M-10,${isBranchStart(index) ? "-30" : "100"}
+                            h 15 
+                            v ${isBranchStart(index) ? "36" : "-76"}
```
digitalcora: Will the merge-line-drawing work if there are multiple branches? (e.g. Green Line)

apps/site/assets/css/_schedule-page-line-diagram.scss
```diff
@@ -1,15 +1,18 @@
...
 
   &:last-child {
     border-bottom: $border;
   }
 }
 
+.m-schedule-line-diagram__content {
+  flex-grow: 1;
+}
+
```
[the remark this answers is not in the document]
thecristen: All of the old `m-schedule-line-diagram` stuff is still in here, and I'm using `m-schedule-diagram` to distinguish the actual diagrammy bits (as opposed to the stop card, connections, features, etc)

apps/site/assets/ts/schedule/components/LineDiagram.tsx
thecristen: Even better - got rid of it!

apps/site/assets/ts/schedule/components/LineDiagram.tsx
thecristen: Ended up making this logic obsolete in my latest commit.

apps/site/assets/ts/schedule/components/LineDiagram.tsx
thecristen: The logic in the latest commit should work for any number of extra branches! 💃

apps/site/assets/css/_schedule-page-line-diagram.scss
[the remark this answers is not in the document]
thecristen: ..which is silly, so I'll go ahead and rename the classes to be more consistent.

ingridpierre:
Other updates look great!

Found a few more things, but feel free to push back or break this out into other work if it's not suited to this ticket:

1) Lowell Inbound, this is probably a weird data quirk because many trips skip a stop (Mishawum?), but we might need to do a little something to get rid of that little extra line:
image

2) Mattapan line does not appear at all, nor do any of the Ferries.

Still looking, will comment as I find things.

ingridpierre:
Franklin outbound, looking kind weird:
image
In ideal world, list would match what's on /schedules/CR-Franklin/timetable?direction_id=0. The branch should really just consist of Plimptonville + Foxboro stops. 

The segment that looks like this below should be removed if possible.
image

ingridpierre:
@thecristen For the Franklin Line, potentially another way to address it would be to have it use the other approach that we use for complex bus variants below. Not sure the pros/cons of having this one line switch to showing route patterns:

image

digitalcora:
@ingridpierre The Foxboro issue also exists on the old line diagram; we documented it [in this ticket](https://app.asana.com/0/385363666817452/1154464073332369/f).

apps/site/assets/ts/schedule/components/LineDiagram.tsx
```diff
@@ -158,76 +140,51 @@ const LineDiagram = ({
...
                   </div>
-                )}
+                ))}
               </div>
               <div className="m-schedule-line-diagram__content">
                 <div className="m-schedule-line-diagram__card">
                   <div className="m-schedule-line-diagram__card-left">
                     <div className="m-schedule-line-diagram__stop-name">
-                      {branchName(lineDiagramStop)}
+                      {StopBranchLabel(routeStop)}
```
digitalcora: Since this is defined like a component, should we invoke it as `<StopBranchLabel stop=routeStop>`?

apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
```diff
@@ -129,7 +130,64 @@ describe("LineDiagram", () => {
...
+      .find(".m-schedule-diagram__line--line")
+      .first()
+      .html();
+
+    expect(terminus).not.toEqual(stop);
+    expect(terminus).not.toEqual(line);
+    expect(terminus).not.toEqual(merge);
+    expect(stop).not.toEqual(merge);
+    expect(stop).not.toEqual(line);
+    expect(line).not.toEqual(merge);
```
digitalcora: I'm not quite sure what behavior this test aims to exercise — could we maybe use some specifics from the fixture data? For example: station X should appear as a terminus, station Y a merge, station Z a normal stop...

digitalcora:
@thecristen The branch points don't seem to be correctly located with the current version of this code. For example, on the red line, it looks like JFK is on a branch, but it's actually a shared stop (it is the last shared stop on the line). The branch point should be below JFK, not above it.

Screen Shot 2019-12-31 at 12 47 00 PM

thecristen:
> @thecristen For the Franklin Line, potentially another way to address it would be to have it use the other approach that we use for complex bus variants below...

The more I learn about the quirks on various Commuter Rail lines, the more I wonder whether using such an approach would be needed. Seems like it'd be helpful for **Franklin Line via Fairmount** and similar alternate routes.

digitalcora:
In Firefox, at least for me, the ends of the lines have visible "corners" poking out from the circle. It looks like the line is missing its end cap.

Screen Shot 2019-12-31 at 2 41 53 PM Screen Shot 2019-12-31 at 2 43 30 PM

thecristen:
Latest change should fix the line end issue, the Lowell Inbound quirk, and (most importantly) the merge point. Thanks @cora for your contribution on [Line Diagram | Draw expanded Green Line branches](https://app.asana.com/0/555089885850811/1155323462935218?du=1149164108974353&invite=575f3f3f994707d96407b76ff2caf727)!

Things we've noticed that are still outstanding:

- [ ] CR-Franklin is just _weird_, because (1) it's underlying data doesn't seem to be configured to branch in the right place, but also because (2) it's choosing the stops from the via Fairmount trips to display, but these might not always be the relevant ones to show...
- [ ] Mattapan woefully blank
- [ ] Ferries pages not showing diagrams either

Development-specific stuff that needs to be resolved:
- [ ] Test coverage is now too low -- I need to write some new/better test fixtures to facilitate this
- [ ] `apps/site/lib/site_web/controllers/schedule/line/diagram_helpers.ex:342 I: Function is too complex (CC is 11, max is 9)`

I need to add some better fixtures to write better tests and get coverage back to satisfactory!

apps/site/lib/site_web/controllers/schedule/line/diagram_helpers.ex
```diff
@@ -202,43 +223,59 @@ defmodule SiteWeb.ScheduleController.Line.DiagramHelpers do
...
 
-  def build_branched_stop(stop, branch_stops, {_, ["Green" <> _ | _] = green_branches}) do
-    # Green Line always evaluates all branches on all stops. If the stop should have a bubble for a branch,
-    # &stop_bubble_type/3 returns a valid tuple, otherwise it returns false. The bubble list then gets filtered to
-    # remove anything that's not a tuple.
+  def build_branched_stop(
+        stop,
+        branch_stops,
+        {_, ["Green" <> _ | _] = green_branches},
+        combine_green_branches
```
thecristen: @digitalcora This breaks a few tests since it changes this function's arity -- what're your thoughts on making `combine_green_branches` an optional argument and/or giving it a default value of `false`? Alternately I can add the fourth argument to the relevant places in the tests and that works fine too.

digitalcora: I could go either way. Follow your heart 💖

thecristen:
Note: Latest commit fixes the Mattapan page, covering tickets [Line Diagram | Mattapan](https://app.asana.com/0/385363666817452/1155608872540907/f) and [Line Diagram | (Mattapan) Pre-populate origin when clicking schedules-from-here link](https://app.asana.com/0/555089885850811/1154986721358490?du=1149164108974353&invite=575f3f3f994707d96407b76ff2caf727)

apps/site/assets/ts/schedule/components/LineDiagram.tsx
thecristen: Just realized I forgot to respond to this -- my thought on this is that this function generates boring JSX that doesn't need to be React-ified so to speak -- there doesn't seem to be anything to gain by defining a React component with props and whatnot for this :)

PR #331, opened 2019-12-31
Fix intermittent react renderer test failure

digitalcora:
Previously, in over half of test runs, a random test in `ReactTest` would fail with a `Task.await` timeout. The cause seems to be that we were starting a copy of the renderer process within one of the tests, which somehow interfered with the "real" process. At least, when this test is deleted, the intermittent failure disappears. The reason may have something to do with Poolboy, but it's hard to pursue further than that as Poolboy is written in Erlang and has zero documentation.

Since this test doesn't appear to provide much value, and neither does the one that stops the process, we can probably safely remove them. `React.stop` was only used in this test, so it can be deleted as well.

Finally, this change also removes a 10000ms timeout on the `Task.await` call that could never be reached. This is because the `GenServer.call` within the task was not given a timeout parameter, and so would time out after the default of 5000ms. This is also what `Task.await` defaults to, so removing the parameter aligns the timeout values.

### DIFF
--- a/apps/site/lib/site/react/react.ex
+++ b/apps/site/lib/site/react/react.ex
@@ -6,7 +6,6 @@ defmodule Site.React do
   use Supervisor
   alias Phoenix.HTML
 
-  @timeout 10_000
   @pool_name :react_render
   @default_pool_size 4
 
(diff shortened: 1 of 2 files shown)


digitalcora:
Ironically Elixir tests failed on this branch due to _other_ intermittent timeout errors, but not the one this change is intended to fix 😄

digitalcora:
> I notice that `Site.ReactTest` is marked as `async: true`. It sounds like that shouldn't be the case, as it's modifying global state.

This particular failure wasn't affected by whether `async: true` was present or not (that was one of the first things I tested). With the start/stop tests deleted, however, it should be safe to use `async` since there is no modification of global state.

>  Perhaps there are also other tests which are modifying global state and are marked as `async: true`?

Perhaps! Our remaining intermittent failures seem to be caused by HTTP request timeouts, but `async` could still be a suspect 🙂

PR #334, opened 2020-01-06
Increase HTTP timeout for API requests in test

digitalcora:
This is aimed at resolving the remaining intermittent test failures we are getting on CI, which are all HTTP timeouts on API requests.


digitalcora:
@phildarnowsky I'm gonna re-run this a bunch of times on CI before merging it, since those HTTP timeouts are only showing up in one out of every ~5 runs.

PR #336, opened 2020-01-06
Fix linter false positives on CI

digitalcora:
Since we were diffing the branch against the current tip of master, rather than the point where it diverged from master, linting would flag issues that had actually been resolved in master.

### DIFF
--- a/semaphore/scss_lint.sh
+++ b/semaphore/scss_lint.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
 fork_point=$(git merge-base --fork-point origin/master)
-changed=$(git diff --name-only $fork_point "*.ex" "*.exs")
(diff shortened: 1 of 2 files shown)


semaphore/scss_lint.sh
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 
-changed=$(git diff --name-only origin/master apps/site/assets/css)
+fork_point=$(git merge-base --fork-point origin/master)
+changed=$(git diff --name-only $fork_point "*.ex" "*.exs")
```
thecristen: I think you might have forgot to edit this one to be relevant to the CSS files rather than the Elixir ones?

digitalcora: Copy-paste mishap 😱

PR #337, opened 2020-01-07
Fix Franklin line diagram issues around Foxboro

digitalcora:
**Asana Ticket:** [Line Diagram | Outbound Franklin Line (Foxboro stops)](https://app.asana.com/0/385363666817452/1154464073332369/f)

* Fixes glitchy/disconnected line diagram in the outbound direction
* Fixes Ruggles and Back Bay not showing up on the inbound direction

It turns out we didn't have a special case for Franklin outbound at all (only inbound), so either the outbound line diagram glitch wasn't a regression, or the data previously didn't require a special case but now does.

While in here I discovered an adjacent issue where the _inbound_ map and line diagram was displaying the trunk stops from the Fairmount line instead of the actual Franklin line. I attempted to fix this so it matches the outbound line diagram.

### DIFF
--- a/apps/stops/lib/route_stop.ex
+++ b/apps/stops/lib/route_stop.ex
@@ -129,12 +129,26 @@ defmodule Stops.RouteStop do
         shapes,
         stops,
         %Route{id: "CR-Franklin"} = route,
-        1
+        0
       ) do
     shapes
-    |> Enum.reject(&(&1.name == "Forge Park/495 - South Station via Back Bay"))
+    |> Enum.reject(&(&1.name == "South Station - Foxboro via Fairmount"))
     |> Enum.map(&do_list_from_shapes(&1.name, &1.stop_ids, stops, route))
-    |> merge_branch_list(1)
+    |> merge_branch_list(0)
+  end
+
+  def list_from_shapes(
+        shapes,
+        stops,
+        %Route{id: "CR-Franklin"} = route,
+        1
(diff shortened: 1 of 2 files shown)


phildarnowsky:
LOL @ `fixboro`

ingridpierre:
Seems like the stops on the timetable got wacky, compare Readville and Endicott here. 
image

Also the "Via Fairmount Line" labels should be lining up in the timetable, but "Line" has drifted a bit. Looks like an extant bug in production. Can we take another look at this with this PR if it makes sense to you?
image

digitalcora:
@ingridpierre I'm not sure what you mean by "got wacky"... what should I be looking at with Readville and Endicott? I'm looking at the timetable in production and it seems to match the screenshot. (it is def broken, but seems unrelated to the line diagram bugs)

ingridpierre:
@digitalcora Apologies for being imprecise. You're right that what's on the timetable is a bug/regression unrelated to the work you've done on the diagram. Here you can see the desired appearance on the PDF provided by Keolis.

image

As you said above "line diagram was displaying the trunk stops from the Fairmount line instead of the actual Franklin line", looks like we're also doing this on the timetable too.

digitalcora:
@ingridpierre It looks like this is a distinct bug and doesn't involve any of the code that builds the line diagram, so I'd prefer to split it out.

PR #338, opened 2020-01-07
Update URL when line diagram direction changes

digitalcora:
This replicates the behavior of the old line diagram where the "reverse direction" button reloads the page with a new `direction_id` param.

**Asana Ticket:** [Line Diagram | Update direction_id in page URL when changing directions](https://app.asana.com/0/385363666817452/1155400692401305/f)

### DIFF
--- a/apps/site/assets/package.json
+++ b/apps/site/assets/package.json
@@ -28,6 +28,7 @@
     "sifter": "^0.5.1",
     "tether": "^1.3.4",
     "turbolinks": "git+https://github.com/paulswartz/turbolinks.git#82b188e",
+    "use-query-params": "^0.5.0",
     "whatwg-fetch": "^3.0.0"
   },
   "devDependencies": {
--- a/apps/site/assets/package-lock.json
+++ b/apps/site/assets/package-lock.json
@@ -1649,8 +1649,7 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "aproba": {
               "version": "1.2.0",
@@ -1671,14 +1670,12 @@
             "balanced-match": {
               "version": "1.0.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "bundled": true,
               "dev": true,
-              "optional": true,
               "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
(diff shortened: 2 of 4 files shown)


apps/site/assets/package.json
```diff
@@ -28,6 +28,7 @@
     "sifter": "^0.5.1",
     "tether": "^1.3.4",
     "turbolinks": "git+https://github.com/paulswartz/turbolinks.git#82b188e",
+    "use-query-params": "^0.5.0",
```
digitalcora: This package provides an actual `useQueryParams` hook for functional React components, but I couldn't figure out how to integrate it into our existing reducer-based logic for changing the line diagram direction. Fortunately it also exposes the URL-updating primitives it uses, which include `updateInLocation`.

thecristen: Hmm neat find. I agree there doesn't seem to be a clear way to integrate that hook into our current logic.

If you're only using `updateInLocation()` it seems that you could use the `serialize-query-params` dependency instead. Though I wonder just how much additional benefit that library offers, given that our current jQuery dependency can also [serialize query params](https://api.jquery.com/jQuery.param/). 

Though serializing is only half the task -- there's also getting and parsing the current location. This change appears to depend on [`query-string`](https://github.com/sindresorhus/query-string) for this, which doesn't indicate IE support, so that's concerning (though I haven't been able to check this in IE to verify). On the other hand, `query-string` was also made before [URLSearchParams](https://developers.google.com/web/updates/2016/01/urlsearchparams?hl=en#feature_detection_and_browser_support) was a big thing, which could also be used with a [polyfill to enable IE support](https://github.com/ungap/url-search-params) to parse the location. Anyway, I went down a rabbit hole here..... if it already Just Works in IE then I'd be quite reassured! 😅

digitalcora: > if it already Just Works in IE then I'd be quite reassured!

`URLSearchParams` works down to IE10, which is currently the oldest supported version (though it will actually drop out of support a few weeks from now). So regardless of whether `query-string` is using that or not, we should be covered.

thecristen: (I'm not sure where you're getting that `URLSearchParams` works down to IE10, as the page I'd linked to as well as the [Can I use entry indicate otherwise](https://caniuse.com/#search=URLSearchParams).)

Anyhow, unless you've already checked it in IE and verified it doesn't implode, I don't see how you can presume we should be covered on this. From `query-string`:

> This module targets Node.js 6 or later and the latest version of Chrome, Firefox, and Safari

digitalcora: oops, I was looking at the wrong compatibility table — `history.replaceState` is supported down to IE10, not `URLSearchParams`. However, `URLSearchParams` appears nowhere in the code of `query-string`, so there must be a different reason they don't claim support for IE. I think it might be the use of `Object.assign`.

That part of the README also says to use version 5 to get support for older browsers, and in fact the latest version of `serialize-query-params` does pull in version 5 and not 6. I'm not sure if it's actually the maintainer's policy to provide support for older browsers, or if that just happened to be the newest major version at the time.

Options as I see them:

1. Ship this as-is. It will work in IE, but if we're not careful with future upgrades of packages, it could break.

2. Explicitly depend on `query-string ^5.0.0`, and write our own `updateInLocation`.

3. Depend on nothing additional, use our existing `ts/helpers/query` (which has similar functionality to `query-string`), and write our own `updateInLocation`.

I'm inclined towards the second one. What do you think?

thecristen: Ah good find with the fact this uses the older version anyway! Looks like that should work fine then. I am OK with option 1, as I'd hope we'd be careful with any future upgrading we do. :)

PR #340, opened 2020-01-08
Clean up past time-dependent logic

digitalcora:
I've noticed a few places in the code where we were making decisions based on whether a certain date had passed (I even wrote one of them), and where the date has now passed so the conditional is no longer needed.

Method: I did a global search for the regex `201[56789]`, excluding tests and fixture data. This website didn't exist prior to 2015, so is unlikely to have any "feature gates" going further back.

Best reviewed as separate commits.

### DIFF
--- a/apps/site/lib/site/shuttle_diversion.ex
+++ b/apps/site/lib/site/shuttle_diversion.ex
@@ -38,46 +38,6 @@ defmodule Site.ShuttleDiversion do
           }
   end
 
-  defmodule TripsHack do
-    @moduledoc """
-    Proxy for `V3Api.Trips.by_route` that works around missing trip data for specific weekday
-    shuttle diversions on the D line, by borrowing the trips from an identical weekend diversion.
-    TODO: Delete after 2019-12-20.
-    """
-
-    @date_with_correct_trips "2019-12-21"
-    @dates_to_correct [
-      "2019-11-20",
-      "2019-11-21",
-      "2019-11-25",
-      "2019-11-26",
-      "2019-12-06",
-      "2019-12-13",
-      "2019-12-16",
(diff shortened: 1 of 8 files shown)


apps/site/lib/site/shuttle_diversion.ex
```diff
@@ -68,8 +68,9 @@ defmodule Site.ShuttleDiversion do
       ]
 
       with %JsonApi{data: trips} <- Trips.by_route(trips_route, trips_params),
-           route_stops when is_list(route_stops) <- Stops.by_routes(true_route_ids, 0),
-           unique_trips = unique_trips_by_shape(trips) do
+           route_stops when is_list(route_stops) <- Stops.by_routes(true_route_ids, 0) do
```
phildarnowsky: @digitalcora could you please rebase this and split it up so each change goes with the proper commit?

digitalcora: I considered this, but thought an approach that resulted in cleaner changesets within individual commits would be better. If someone were looking at one of these commits in the future, it might be unclear which changes were related to the purpose of the commit, and which were Credo failures caused by changing the overall file, but having nothing to do with the code that was changed.

I can switch it if you think the other way is better; I'd add a body to each of the commits like "also addresses Credo failures on changed files".

IMO there's no _great_ way to handle this other than a mega-commit that just fixes all our Credo failures everywhere. (this was one nice thing Pronto actually did for us: it was somehow able to filter Credo failures to just the ones involving _lines_ that were changed in the PR)

phildarnowsky: @digitalcora good point, let me think about it a minute

PR #342, opened 2020-01-09
Show predicted times in trip details

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Trip Details | Add realtime details to schedule finder so people know when trains will reach their destinations](https://app.asana.com/0/555089885850811/1144999616263105)

Changed the Trip Details in the Upcoming Departures section so that, rather than scheduled times, they display predicted times. This also includes special handling for CR when trains are more than 5 minutes late.


phildarnowsky:
@thecristen note that only the last 3 commits of this need to be reviewed, the preceding commits are all from Adamo's original branch and already approved.

phildarnowsky:
there is one legit Elixir test failure that I am looking into

apps/site/assets/css/_schedule-page.scss
```diff
@@ -431,3 +431,16 @@
 .schedule-table__scheduled-bus-pill {
   vertical-align: middle;
 }
+
+.schedule-table__upcoming-departures-header {
+  width: 100%;
```
thecristen: This shouldn't be needed, divs are 100% width by default

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
```diff
@@ -139,7 +139,30 @@ export const UpcomingDepartures = ({
   if (journeys !== null && hasPredictions(journeys)) {
     return (
       <>
-        <h3>Upcoming Departures</h3>
+        <div className="schedule-table__upcoming-departures-header">
+          <h3>Upcoming Departures</h3>
+          &nbsp;
```
thecristen: Let's avoid `&nbsp;`, especially since the design doesn't call for a small space here anyway

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
```diff
@@ -139,7 +139,30 @@ export const UpcomingDepartures = ({
...
+          <span className="schedule-table__live-clock" aria-hidden>
+            <span className="icon-realtime animate">
```
thecristen: This extra `span.icon-realtime.animate` wrapper is redundant as far as I can tell.

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
```diff
@@ -139,7 +139,30 @@ export const UpcomingDepartures = ({
...
+                  height="9"
+                  role="img"
+                  viewBox="0 0 9 9"
+                >
+                  <title>clock</title>
+                  <path
+                    fillRule="evenodd"
+                    d="M5 4V1.5a.5.5 0 0 0-1 0v3a.498.498 0 0 0 .5.5h2a.5.5 0 0 0 0-1H5zm-.5 5a4.5 4.5 0 1 1 0-9 4.5 4.5 0 0 1 0 9z"
+                  />
+                </svg>
```
thecristen: Might it make sense to import/export the clock icon from our static assets, similarly to what's done [in this helper file](https://github.com/mbta/dotcom/blob/95846cf829daba6f4a97e54a153b7a622c6ab851/apps/site/assets/ts/helpers/icon.tsx), rather than duplicating the SVG code here?

apps/site/assets/ts/schedule/components/__trips.d.ts
```diff
@@ -29,12 +30,14 @@ export interface TripInfo {
   status: string;
   duration: number;
   fare: Fare;
+  route_type: number;
```
thecristen: We probably want to use the [`RouteType` type](https://github.com/mbta/dotcom/blob/dbffbc3496dc08d25a7c5733e806aab3fd96b140/apps/site/assets/ts/__v3api.d.ts#L100) instead of number here.

ingridpierre:
@thecristen yes, thanks for the keen eye here! Should use the smaller middle dot · ( NOT this guy: • )

phildarnowsky:
@thecristen /cc @ingridpierre so the thing about the "Live" icon is that I spent a fair bit of time trying to get it to align right without any success--the browser _really_ wanted to align just the clock icon right, rather than the "LIVE" text, so you get something like the attached (depending on what approach exactly you try). You can probably CSS much better than I so if you want to take a crack at that, that would be awesome.
Screen Shot 2020-01-10 at 11 54 53 AM

thecristen:
@phildarnowsky I'll take a crack at the finicky icon later this afternoon when I'm back on my laptop! 😈

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
phildarnowsky: I will check

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
phildarnowsky: yes...I tried that and it turned into a whole thing so I moved on, but now that I have a working branch to compare it with it would make sense to try to refactor that again.

phildarnowsky:
@thecristen @ingridpierre this should cover everything

phildarnowsky:
note to self: no fares for any mode to CR--do I have that right @ingridpierre ?

ingridpierre:
Yep only CR gets the fares column. Still actively looking over this pr on dev green btw.

ingridpierre:
We should only show the predicted arrival time for Commuter Rail if it's running more than 5 minutes late. **Unless it's more than five minutes late, show the _scheduled_ time.**

For example, the first screenshot below should show the last two times as 6:17 and 6:22 (rather than show the predicted times here of 6:18 and 6:23). Trains are a little different when it comes to predictions since they can sit on platforms or otherwise make adjustments to stick to the schedule as closely as possible (and they shouldn't be marked as delayed if they're within that 5 min window).

**Upcoming departures (Late train 622)**
image

**Daily Schedules (Scheduled train 622)**
image

ingridpierre:
For XS devices (~320w) the destinations are wrapping and looking very cramped, making the table hard to read.  To help everything fit better here, can we add line breaks beneath the crossed out times like so:
image

Note: this work is different from the work we've already outlined to make the XS layout better  overall.

ingridpierre:
Regarding my comment above about arrival time display, @phildarnowsky can you help me understand next week where we're using departure times vs arrival times in these tables? The distinction is important and I want to make sure I'm not telling you to update the wrong thing in the wrong places.

phildarnowsky:
@ingridpierre I am deploying a version to dev-green that should address your feedback, it'll be up within about 20 minutes. I am also researching the question of when we use arrival times and when we use departure.

ingridpierre:
@phildarnowsky Thanks! I verified with Katrina and touched on predictions data with Eddie briefly, i think what I described above (show strikethrough times ONLY when difference is >5 mins late, and to show scheduled time otherwise), still holds here.

ingridpierre:
Thanks for that adjustment, this is good to go to get us unblocked on the remains of the schedule work! 👍

phildarnowsky:
@thecristen anything further on this puppy?

PR #344, opened 2020-01-13
Align CI setup with development setup

digitalcora:
This was a spare-time improvement following on from #332, with the key ideas: **1)** CI should automatically use the same versions of everything we use in development; **2)** we shouldn't spend time compiling the app for CI tasks that don't need it.

The following should be taken with a large chunk of salt since there is high variance in the reported times (sometimes the clock counts up for a minute or two before the first command starts), but the time difference for e.g. linting tasks is significant.

`master` | This Branch
:--:|:--:
Screen Shot 2020-01-09 at 4 33 44 PM | Screen Shot 2020-01-09 at 3 47 06 PM

The Dialyzer step is currently taking _longer_ than before due to an issue with the Semaphore cache not persisting the new PLTs. ~Semaphore support has acknowledged that it seems like a problem on their end but have not yet come up with a solution.~

After several weeks of investigation, Semaphore support finally figured out what was making the caching go wrong (disk space exhaustion from our massive repo, basically) and gave me some commands to run, which are included in this PR. Unfortunately since the cache is a shared resource across all branches, to see the benefit we likely have to merge this PR, wait for all extant branches to be up-to-date with master, then reset the cache and see if it works. If it doesn't, I definitely have the attention of Semaphore engineers already and can ask them what's going on.

### DIFF
--- a/semaphore/backstop.sh
+++ b/semaphore/backstop.sh
@@ -79,7 +79,7 @@ else
     BRANCH=$(git rev-parse --abbrev-ref HEAD)
     FILENAME="$BRANCH.tar.gz"
     tar -czvf "$FILENAME" apps/site/test/backstop_data
-    LINK=$(curl -F "file=@$FILENAME" https://file.io/\?expires\=1d | jq -r .link)
+    LINK=$(curl -F "file=@$FILENAME" https://file.io/\?expires=1d | jq -r .link)
     echo "Backstop report located at $LINK, available for 1 day for a single download."
   fi
   exit 1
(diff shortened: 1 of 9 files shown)


semaphore/backstop.sh
```diff
@@ -79,7 +79,7 @@ else
     BRANCH=$(git rev-parse --abbrev-ref HEAD)
     FILENAME="$BRANCH.tar.gz"
     tar -czvf "$FILENAME" apps/site/test/backstop_data
-    LINK=$(curl -F "file=@$FILENAME" https://file.io/\?expires\=1d | jq -r .link)
+    LINK=$(curl -F "file=@$FILENAME" https://file.io/\?expires=1d | jq -r .link)
```
digitalcora: ShellCheck pointed out we don't need to escape this as `=` is not a special character in this context.

PR #347, opened 2020-01-14
Nested paragraph un/publishing and preview support

amaisano:
#### Summary of changes
**Asana Ticket:** [Content authors can unpublish nested paragraphs without 500 error](https://app.asana.com/0/553506286097126/1117409475848341)

- Passes query parameters down from page response to paragraph parser
- Pass all nested paragraphs though paragraph parser, rather than direct rendering
- Prevent empty parent paragraphs from rendering when no published children are present
- Update logic for reusable paragraphs: top level type can be unpublished

Related CMS ticket: mbta/cms#318

### DIFF
--- a/apps/cms/lib/helpers.ex
+++ b/apps/cms/lib/helpers.ex
@@ -117,30 +117,30 @@ defmodule CMS.Helpers do
     end
   end
 
-  @spec parse_paragraphs(map, String.t()) :: [Paragraph.t()]
-  def parse_paragraphs(data, target_field \\ "field_paragraphs") do
+  @spec parse_paragraphs(map, map, String.t()) :: [Paragraph.t()]
+  def parse_paragraphs(data, query_params \\ %{}, target_field \\ "field_paragraphs") do
     data
     |> Map.get(target_field, [])
-    |> Enum.filter(&para_is_published/1)
-    |> Enum.map(&Paragraph.from_api/1)
+    |> Enum.filter(&para_is_published(&1, query_params))
+    |> Enum.map(&Paragraph.from_api(&1, query_params))
   end
 
-  @spec para_is_published(map) :: boolean
+  @spec para_is_published(map, map) :: boolean
   # Reusable paragraphs can be deleted, but their parent references may remain
-  defp para_is_published(%{"field_reusable_paragraph" => [nil]}) do
+  defp para_is_published(%{"field_reusable_paragraph" => [nil]}, _query_params) do
     false
   end
 
-  defp para_is_published(%{"field_reusable_paragraph" => reusable}) do
+  defp para_is_published(%{"field_reusable_paragraph" => reusable}, query_params) do
     [%{"status" => status, "paragraphs" => data}] = reusable
 
     case status do
       [%{"value" => false}] -> false
-      _ -> data |> List.first() |> para_is_published()
+      _ -> data |> List.first() |> para_is_published(query_params)
     end
   end
 
-  defp para_is_published(%{"status" => [%{"value" => value}]}) do
+  defp para_is_published(%{"status" => [%{"value" => value}]}, _query_params) do
     value
   end
 
--- a/apps/cms/lib/helpers.ex
+++ b/apps/cms/lib/helpers.ex
@@ -130,21 +130,33 @@ defmodule CMS.Helpers do
   end
 
   @spec para_is_published(map, map) :: boolean
-  # Reusable paragraphs can be deleted, but their parent references may remain
-  defp para_is_published(%{"field_reusable_paragraph" => [nil]}, _query_params) do
+  defp para_is_published(field_data, query_params)
+
+  # Reusable paragraphs instance aren't automatically removed when their child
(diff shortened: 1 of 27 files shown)


amaisano:
Codcov broken because master was merged w/o a successful CI pass.

phildarnowsky:
@amaisano I'm about to go out of town for a couple of days so you may want to get someone else to review this.

apps/site/lib/site_web/templates/cms/paragraph/_column_multi.html.eex
```diff
@@ -9,16 +9,19 @@
...
 
-<%= extend_width_if has_cards?, :cards do %>
-  <div class="c-multi-column__row row">
-    <%= for column <- @content.columns do %>
-      <div class="c-multi-column__column col-sm-6 col-md-<%= grid(@content) %>">
 
-        <%= for paragraph <- column.paragraphs do %>
-          <%= render_paragraph_content(paragraph, @conn) %>
-        <% end %>
+<%= if length(@content.columns) > 0 do %>
```
digitalcora: It's minor, but `Enum.any?(things)` is more efficient than `length(things) > 0`, which has to iterate the whole list to find the length.

apps/site/lib/site_web/templates/cms/paragraph/_description_list.html.eex
```diff
@@ -2,17 +2,19 @@
...
-        <dt class="dt">
-          <%= ContentRewriter.rewrite(description.term, @conn) %>
-        </dt>
-        <dd class="dd">
-          <%= ContentRewriter.rewrite(description.details, @conn) %>
-        </dd>
-      </div>
-    <% end %>
-  </dl>
+<%= if length(@content.descriptions) > 0 do %>
```
digitalcora: Same comment here re: `Enum.any?`

digitalcora:
As a general comment on the approach, I don't feel great about sending the full query params of the original web page request all the way down through the CMS functions, when those functions only care about one specific query param (and they don't care about a _query param_, per se, they care about whether the caller – which need not be a web controller! – is requesting a different mode of operation).

I think we should apply better separation-of-concerns here. Like:

```elixir
# in the web controllers:
cms_preview = (conn.query_params["preview"] == "true")
CMS.some_function(other_params, cms_preview)

# in the CMS modules:
def some_function(other_params, preview_mode) do
  # preview_mode is a bool here that indicates whether to preview
end
```

Does that sound reasonable? Is there an advantage to the current approach I'm not seeing?

amaisano:
@digitalcora sounds reasonable. My usage of the params was in flux for a while so I left them generally available. I can change to simpler/smaller arguments. Thanks.

apps/cms/lib/helpers.ex
```diff
@@ -141,18 +155,18 @@ defmodule CMS.Helpers do
...
     case parent_status do
       %{"value" => false} -> false
-      _ -> show_paragraph?(paragraph, query_params)
+      _ -> show_paragraph?(paragraph, preview_mode)
     end
   end
 
   # In "preview" mode, allow unpublished paragraphs to be rendered if requested
-  defp show_paragraph?(_, %{"preview" => _, "paragraphs" => _}) do
+  defp show_paragraph?(_, preview: true, paragraphs: true) do
```
digitalcora: It doesn't seem like these "preview modes" are actually independent of each other: if `preview` is true *and* `paragraphs` is true, this one bit of behavior changes, else it doesn't. Could we simplify to a single boolean?

amaisano: Authors can be in preview mode but request _not_ to see paragraph previews. I kept both elements are part of the match as a safeguard too, to ensure sending just `paragraphs` doesn't expose them (must be in `preview` mode).

apps/cms/lib/helpers.ex
```diff
@@ -117,30 +117,60 @@ defmodule CMS.Helpers do
     end
   end
 
-  @spec parse_paragraphs(map, String.t()) :: [Paragraph.t()]
-  def parse_paragraphs(data, target_field \\ "field_paragraphs") do
+  @doc """
+  Gathers information about desired view mode based on incoming
+  query parameters. Key presence is enough to count as true"
+  - page: whether the page is in preview mode (drafts are rendered)
+  - pagaragraphs: whether unpublished paragraphs should render
```
digitalcora: `pagaragraphs`, while it sounds awesome, I'm assuming is a typo 😆

apps/cms/lib/helpers.ex
```diff
@@ -117,30 +117,60 @@ defmodule CMS.Helpers do
     end
   end
 
-  @spec parse_paragraphs(map, String.t()) :: [Paragraph.t()]
-  def parse_paragraphs(data, target_field \\ "field_paragraphs") do
+  @doc """
+  Gathers information about desired view mode based on incoming
+  query parameters. Key presence is enough to count as true"
```
digitalcora: Stray quote mark `"`

PR #351, opened 2020-01-16
Handle added trips in finder_api

phildarnowsky:
FinderApi.departures implicitly assumed that all trips have associated
schedules, which is not true of added trips. This assumption led to a
nil creeping into an unexpected place, leading ultimately to an
UndefinedFunctionError. This case is now accounted for.

#### Summary of changes
**Asana Ticket:** [Upcoming Departures 500s if it encounters an added trip](https://app.asana.com/0/555089885850811/1157538555249409)


thecristen:
How should we handle this in the UI? It's great that these added trips show on the Upcoming Departures display, but those trips (correctly) have no schedule to display (see first one in screenshot). cc @ingridpierre 

image

PR #352, opened 2020-01-17
Schedule Finder | "View schedules" doesn't work for combo GL

thecristen:
#### Summary of changes
**Asana Ticket:** [Schedule Finder | "View schedules" doesn't work for combo GL](https://app.asana.com/0/385363666817452/1156654983519751/f)

We added some logic to enable us to use the appropriate route ID (e.g. "Green-D" instead of "Green") so that the appropriate API calls will return results!

Also includes some additional safeguards for added trips, where the modal opened from "View schedules" might show errors or abruptly close when a(n added) trip triggers backend errors or  provides a prediction for Upcoming Depatures but no schedule to populate the Accordion.

### DIFF
--- a/apps/site/assets/ts/schedule/components/schedule-finder/TripDetails.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/TripDetails.tsx
@@ -42,6 +42,9 @@ const TripSummary = ({
   </tr>
 );
 
+const allTimesHaveSchedule = (tripInfo: TripInfo): boolean =>
+  tripInfo.times.every(time => !!time.schedule);
+
 export const TripDetails = ({
   state,
   showFare
@@ -56,12 +59,14 @@ export const TripDetails = ({
     );
   }
 
+  const errorLoadingTrip = (
+    <p>
+      <em>Error loading trip details. Please try again later.</em>
+    </p>
+  );
+
   if (error) {
-    return (
-      <p>
-        <em>Error loading trip details. Please try again later.</em>
-      </p>
-    );
+    return errorLoadingTrip;
   }
 
   if (!tripInfo) return null;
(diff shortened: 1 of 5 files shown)


apps/site/lib/site_web/controllers/schedule/finder_api.ex
```diff
@@ -136,6 +136,13 @@ defmodule SiteWeb.ScheduleController.FinderApi do
         is_current: is_current
       )
 
+    route_ids =
+      if route_id == "Green" do
+        ["Green-B", "Green-C", "Green-D", "Green-E"]
```
digitalcora: This is accessible as `GreenLine.branch_ids()` so we don't have to hard-code it.

Also, could probably inline this:

```elixir
route_ids = if(route_id == "Green", do: GreenLine.branch_ids(), else: [route_id])
```

apps/site/assets/ts/schedule/components/schedule-finder/TripDetails.tsx
```diff
@@ -91,15 +93,19 @@ export const TripDetails = ({
...
+          departure.schedule ? (
+            <TripStop
+              departure={departure}
+              index={index}
+              showFare={showFare}
+              routeType={tripInfo.route_type}
+              key={departure.schedule.stop.id}
+            />
+          ) : (
+            errorLoadingTrip
```
digitalcora: I assume this is the scenario we've stumbled across elsewhere, where a trip can have a prediction but not a schedule? Is it possible that we could still show trip data in this case, or is it just not there?

thecristen: Yeah, added trips can have a prediction but no schedule, as noted in #351 . It might be possible to show trip data in this case but it would need a refactor of at least the `<TripStop />` component.

digitalcora: Ok – I can see how that work would be out-of-scope here, so maybe we can add a new Asana task for "schedule finder should show details of added-in trips"? Unless there already is one somewhere and I haven't seen it.

digitalcora:
0632d99 could use a more descriptive commit message, and also seems to have some unrelated changes included in it.

phildarnowsky:
@digitalcora ready for another look

phildarnowsky:
@digitalcora scratch that, I think I still need to bring test coverage up, and I just saw your comment from some hours ago

digitalcora:
@phildarnowsky Latest changes look good, other than that 🙂

PR #354, opened 2020-01-21
Schedule Finder | Show schedule note on subways

thecristen:
#### Summary of changes
**Asana Ticket:** [Schedule Finder | Show schedule note instead of trips for subway](https://app.asana.com/0/385363666817452/1156654983519748/f)

Note: The official hours of operation published by plans and schedules on the system map is: "Bus and Rapid Transit services run from 5:15 a.m. to 12:30 a.m., Monday– Saturday; Sunday from 6:00 a.m. to 12:30 a.m." -- I'm using this as a stopgap until we refactor and have a reasonable way to fetch the first and last trips down to the minute specific to each origin/direction/destination.

### DIFF
--- a/apps/site/assets/css/_schedule-page.scss
+++ b/apps/site/assets/css/_schedule-page.scss
@@ -58,6 +58,19 @@
   padding: $base-spacing * 2 $base-spacing;
 }
 
+.m-schedule-page__schedule-notes--modal {
+  margin-top: $base-spacing;
+  padding: $base-spacing;
+
+  .m-schedule-page__schedule-note-title {
+    display: none;
+
+    + .m-schedule-page__schedule-note .m-schedule-page__service {
+      margin-top: 0;
+    }
+  }
+}
+
 .m-schedule-page__schedule-notes--desktop {
   margin-top: $base-spacing * 2;
 
@@ -363,13 +376,29 @@
   font-variant-numeric: tabular-nums;
 }
 
-.schedule-finder__first-last-trip {
+.schedule-finder__first-last-trip--bordered {
   border-top: 3px solid $gray-lightest;
+  padding: ($base-spacing / 2) 0;
+}
+
+.schedule-finder__first-last-trip-label {
+  @extend %u-small-caps;
+  @extend %u-bold;
+  margin-right: $base-spacing;
+  white-space: nowrap;
+
+  @include media-breakpoint-down(xs) {
+    display: block;
+  }
+}
+
+.first-last-trips {
+  align-items: baseline;
   display: flex;
-  justify-content: space-around;
-  // scss-lint:disable DuplicateProperty IE11 doesn't support space-evenly
-  justify-content: space-evenly;
-  padding: $base-spacing / 2;
+}
+
+.first-last-trips__trip:not(:first-child) {
+  margin-left: $base-spacing;
 }
 
 .schedule-table__subtable {
--- a/apps/site/assets/ts/schedule/components/schedule-finder/FirstLastTimes.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/FirstLastTimes.tsx
@@ -24,6 +24,10 @@ const FirstLastTimes = ({
     return null;
   }
 
+  // For subway lines, the frequency of trips is lower on Saturdays, but
+  // the first and last trips are the same on Weekdays and Saturdays.
+  // When showing first and last trips use the weekday service,
+  // but label it as "Monday - Saturday".
   const timesList = [
     [
       "Monday - Saturday",
--- a/apps/site/assets/ts/schedule/components/schedule-finder/FirstLastTimes.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/FirstLastTimes.tsx
@@ -1,62 +1,109 @@
-import React, { ReactElement } from "react";
-import { SelectedDirection } from "../ScheduleFinder";
-import { HoursOfOperation } from "../__schedule";
+import React, { ReactElement, useReducer, useEffect } from "react";
+import { SelectedDirection, SelectedOrigin } from "../ScheduleFinder";
+import { reducer } from "../../../helpers/fetch";
+import { shortTime } from "../../../helpers/date";
 
-const getTimes = (
-  hoursOfOperation: HoursOfOperation,
-  serviceType: "week" | "saturday" | "sunday",
-  selectedDirection: SelectedDirection
-): [string, string] => [
-  hoursOfOperation[serviceType][selectedDirection!].first_departure,
-  hoursOfOperation[serviceType][selectedDirection!].last_departure
-];
+type fetchAction =
+  | { type: "FETCH_COMPLETE"; payload: [] }
+  | { type: "FETCH_ERROR" }
+  | { type: "FETCH_STARTED" };
+
+// First and last departures for two sets of services
+const fetchData = (
+  routeId: string,
+  selectedOrigin: SelectedOrigin,
+  selectedDirection: SelectedDirection,
+  dispatch: (action: fetchAction) => void
+): Promise<void> => {
+  dispatch({ type: "FETCH_STARTED" });
+  return (
+    window.fetch &&
+    window
+      .fetch(
+        `/schedules/schedule_api/hours?stop=${selectedOrigin}&route=${routeId}&direction=${selectedDirection}`
+      )
+      .then(response => {
+        if (response.ok) return response.json();
+        throw new Error(response.statusText);
+      })
+      .then(json => dispatch({ type: "FETCH_COMPLETE", payload: json }))
+      // @ts-ignore
+      .catch(() => dispatch({ type: "FETCH_ERROR" }))
+  );
+};
 
 interface Props {
-  hoursOfOperation: HoursOfOperation;
+  routeId: string;
+  selectedOrigin: SelectedOrigin;
   selectedDirection: SelectedDirection;
 }
 
+// TODO Add tests
 const FirstLastTimes = ({
-  hoursOfOperation,
+  routeId,
+  selectedOrigin,
   selectedDirection
 }: Props): ReactElement<HTMLElement> | null => {
-  if (selectedDirection === null) {
-    return null;
-  }
+  const [state, dispatch] = useReducer(reducer, {
+    data: null,
+    isLoading: true,
+    error: false
+  });
+
+  useEffect(
+    () => {
+      fetchData(routeId, selectedOrigin, selectedDirection, dispatch);
+    },
+    [routeId, selectedOrigin, selectedDirection]
+  );
 
   // For subway lines, the frequency of trips is lower on Saturdays, but
   // the first and last trips are the same on Weekdays and Saturdays.
   // When showing first and last trips use the weekday service,
   // but label it as "Monday - Saturday".
-  const timesList = [
-    [
-      "Monday - Saturday",
-      ...getTimes(hoursOfOperation, "week", selectedDirection)
-    ],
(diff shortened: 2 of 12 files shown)


apps/site/assets/ts/schedule/components/schedule-finder/ScheduleModalContent.tsx
```diff
@@ -130,6 +134,61 @@ const ScheduleModalContent = ({
...
       : directionDestinations[selectedDirection];
 
+  let ScheduledTripsOrScheduleNote: ReactElement<HTMLElement>;
+
+  if (scheduleNote) {
+    // we will show the schedule note along with the first and
+    // last trip times in lieu of the service picker and trip table.
+    const defaultFirstLastTimes: { [key: string]: [string, string] } = {
+      "Monday - Saturday": ["05:15 AM", "12:30 AM"],
+      Sunday: ["06:00 AM", "12:30 AM"]
```
digitalcora: What do you think of locating this data within `schedule_note.ex`, alongside the other static info that gets fed in as the `scheduleNote` property?

apps/site/assets/ts/schedule/schedule-loader.tsx
```diff
@@ -100,6 +101,7 @@ const renderDirectionAndMap = (
       ratingEndDate={ratingEndDate}
       stops={stops}
       today={today}
+      scheduleNote={scheduleNote!}
```
digitalcora: Why is the not-null override needed here? Are we guaranteed to always have a scheduleNote at this point?

ingridpierre:
This is a revision to the styling of the first/last trips in general, but seeing how the justified content spacing is turning out on midsize (MD/SM) devices is revealing potential accessibility/legibility issues. I worry that users using magnification or screen zoom may miss the last trip entirely if we don't push them together. **Let's make the label and its values be left aligned with 1 rem of spacing between each**

**CURRENT:**

image
image


**DESIRED:**

image

image

ingridpierre:
A few more things:
- Hoping that the comment above this one will solve the spacing issues seen in XS here, but we may have to force a line break after the label if not?
image

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleModalContent.tsx
thecristen: Yes, that sounds way better. I also just discovered `apps/schedules/lib/hours_of_operation.ex`, which I'm going to see if I can leverage here instead.

digitalcora: Indeed, it seems we already have an `HoursOfOperation` React component as well. This is all new stuff to me, glad we're uncovering it 😊

thecristen:
@ingridpierre I ended up needing that line break after all. Anyhow feel free to take another look! :)

@digitalcora I pretty much substantially revised my approach -- this change adds an hours_of_operation field to the ScheduleNote. It's leveraging the existing functionality from `Schedules.HoursOfOperation` though I chose to edit the resulting datetimes to just give me the time strings I needed... anyhow, feedback welcome. :)

apps/site/assets/ts/schedule/components/schedule-finder/FirstLastTimes.tsx
```diff
@@ -0,0 +1,60 @@
...
+  selectedDirection
+}: Props): ReactElement<HTMLElement> | null => {
+  if (selectedDirection === null) {
+    return null;
+  }
+
+  const timesList = [
+    [
+      "Monday - Saturday",
+      ...getTimes(hoursOfOperation, "week", selectedDirection)
```
digitalcora: The types define a `"saturday"`, but we seem to ignore that here and assume the `"week"` times are also valid for Saturday — curious why that is.

apps/site/lib/site_web/views/helpers/time_helpers.ex
```diff
@@ -1,9 +1,18 @@
 defmodule SiteWeb.TimeHelpers do
+  @moduledoc """
+  Shortcuts to manipulate DateTimes and Dates
+  """
```
digitalcora: In the case that the moduledoc fits on one line, we don't have to use the multi-line docstring format:

```elixir
@moduledoc "Shortcuts to manipulate DateTimes and Dates"
```

apps/site/lib/site/schedule_note.ex
```diff
@@ -7,24 +7,37 @@ defmodule Site.ScheduleNote do
...
+            alternate_text: nil,
+            hours_of_operation: []
 
   @type t :: %__MODULE__{
           peak_service: String.t(),
           offpeak_service: String.t(),
           exceptions: [exception],
-          alternate_text: String.t() | nil
+          alternate_text: String.t() | nil,
+          hours_of_operation: [hours_of_operation]
```
digitalcora: This type appears to be incorrect; `hours_of_operation` is set to the return value of the `hours` function, which is an `HoursOfOperation` struct. But, this struct doesn't align with its own typespec (`adjust_departures` changes the type of the fields). It looks like the purpose of "adjusting" the departures is to format the time values. Can we do this on the frontend? It seems like more of a view-layer concern.

apps/site/lib/site_web/views/cms_view.ex
```diff
@@ -4,7 +4,7 @@ defmodule SiteWeb.CMSView do
   """
   use SiteWeb, :view
 
-  import SiteWeb.TimeHelpers
+  import SiteWeb.TimeHelpers, except: [format_time: 1]
```
digitalcora: This seems potentially confusing as it's unclear which helpers we _do_ use, just from looking at this line. I think using `only` might be better here.

apps/site/test/site/lib/schedule_note_test.exs
```diff
@@ -36,6 +36,16 @@ defmodule Site.ScheduleNoteTest do
       )
     end
 
+    test "returns a schedule note with hours of operation for all lines" do
+      @all_lines
+      |> Enum.map(&ScheduleNote.new/1)
+      |> Enum.map(
+        assert(fn note ->
+          note.hours_of_operation
```
digitalcora: Is there a more detailed assertion we can make here? (that would nevertheless be resistant to changes in live data...)

apps/site/assets/css/_schedule-page.scss
```diff
@@ -363,13 +376,29 @@
...
 
-.schedule-finder__first-last-trip {
+.schedule-finder__first-last-trip--bordered {
   border-top: 3px solid $gray-lightest;
+  padding: ($base-spacing / 2) 0;
+}
+
+.schedule-finder__first-last-trip-label {
+  @extend .u-small-caps; // scss-lint:disable PlaceholderInExtend
+  @extend .u-bold; // scss-lint:disable PlaceholderInExtend
```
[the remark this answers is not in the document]
thecristen: On the chance that we decide in the future to change our bold style (say by tweaking the font-weight from bold to bolder, or adjusting font-size, or any other change) by redefining `.u-bold`, extending the class here will help us stay consistent.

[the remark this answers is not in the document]
digitalcora: OK — is there a more-recommended way for us to do this, such that we wouldn't have to suppress a warning?

apps/site/assets/ts/schedule/components/schedule-finder/FirstLastTimes.tsx
thecristen: Oh I should have added a comment on that! We shall be enlightened in an imminent commit :)

ingridpierre:
On closer inspection, do the first and last trip values not change here depending on the stop i've selected? They should differ from stop to stop. I think using the hours of operation got us further from the design intent here.

For example, on the Red Line, all the Northbound stops say their first trip happens at 5:16, and all the southbound stops say their first trip is at 5:13,  but that's not true. For example the actual first southbound trip doesn't leave Downtown Crossing until 5:36 AM.

apps/site/assets/css/_schedule-page.scss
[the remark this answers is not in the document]
thecristen: The config is [expecting](https://github.com/sds/scss-lint/blob/master/lib/scss_lint/linter/README.md#placeholderinextend) us to extend a [placeholder](https://sass-lang.com/documentation/style-rules/placeholder-selectors) and use that instead of extending the class. 

We have a handful of placeholder selectors scattered across our code base, though most of them are only used in the file they are defined in.

Some options:
1) Use these two extended classes in the HTML markup (likely the intended usage of such a class and easy enough to do for this selector).
2) Update the scss-lint config to disable that particular check (easy, and we can remove 3 of ~30 instances of `// scss-lint:disable`).
3) Write and use placeholders instead (this will work for this case, but still leaves us high and dry in the case where we want to extend an existing class from, say, Bootstrap -- which admittedly doesn't happen very often).

thecristen:
Well, that's true - it's specific to the route and direction but not the stop. Let me see what I can do...

apps/site/assets/css/_schedule-page.scss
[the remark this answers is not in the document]
digitalcora: It seems like there are [legitimate pitfalls](http://8gramgorilla.com/mastering-sass-extends-and-placeholders/) to extending selectors, so I don't think continuing to disable the linter check is the best option in the long run. I like the idea of using placeholders since that's the recommended alternative, but I'm not sure how much work it would be to replace all our existing instances of `@extend .selector`. If it's non-trivial, maybe we can file that as a separate tech debt task?

thecristen:
@ingridpierre This should be showing first/last trips specific to each stop/route/direction now!! :)

thecristen:
@digitalcora I changed my entire approach and am no longer attaching this information to the schedule note. Instead I wrote a schedules_api endpoint that will return information for the first/last trips. I still need to tweak it and add tests but I think it works for what I'm trying to do....

ingridpierre:
Working correctly for the most part where service was normal. Wanted to look at this during the weekend shuttles and managed to catch some interesting states. The shuttled stops for Alewife through Harvard lost their first and last trip time information wholesale (see screenshot below). Can you help me understand what might be happening here? I'd like to think through a follow up ticket to cover this case with some additional messaging, maybe a link to the alerts tab or something lightweight to explain the missing info. 

Not sure it belongs in this PR though. I was seeing a few errors like: "Failed to load resource: the server responded with a status of 500 (Internal Server Error)" in the console.

image

Weirdly very far away prediction showing:
image

ingridpierre:
Hopefully the image below provides more context for the related errors I saw:
image

thecristen:
@ingridpierre The weirdly far away prediction is definitely a task for another ticket (I just filed one for that), but I'd like to look into the rest. 

Thanks for looking at this at a time of non-normal service; it really highlights how my approach falls apart there! 

**What is happening:** Since the subway / schedule note modal design removes the service picker, fetching and showing first/last trips becomes a bit more ambiguous as we first need a set of schedules to read from. Which schedules do we use? I was taking an upcoming weekday and upcoming weekend date to fetch schedules for those dates and compute the first/last trip times, but this produces strange results when said upcoming date happens to have unusual service.

In the case of the 500 error you cited here, I am getting no schedules for the relevant Saturday and Sunday for that route/stop/direction (because yay shuttles). Back to the drawing board...

ingridpierre:
@thecristen do you want to just merge in the bit that adds the blue note part in place of the schedule, and leave the first/last trip display for a different ticket? This would unblock the flag removal.

thecristen:
An excellent idea! I'm going to archive this branch and make a new PR with the first/last trip display omitted.

PR #355, opened 2020-01-22
Fix preview_opts

phildarnowsky:
Controllers should not send preview_opts. At this higher level, they need to pass the full conn.query_params. Extraction of preview_opts happens at a lower level, in the repo and deeper. Added an integration test that would have caught this. Also includes a fix for live-data-related test regressions that are unrelated but were blocking CI.

### DIFF
--- a/apps/site/lib/site_web/controllers/project_controller.ex
+++ b/apps/site/lib/site_web/controllers/project_controller.ex
@@ -4,8 +4,6 @@ defmodule SiteWeb.ProjectController do
   """
   use SiteWeb, :controller
 
-  import CMS.Helpers, only: [preview_opts: 1]
-
   alias CMS.{Partial.Teaser, Repo}
   alias CMS.Page.{Project, ProjectUpdate}
   alias Plug.Conn
@@ -97,7 +95,7 @@ defmodule SiteWeb.ProjectController do
 
     "/projects"
     |> Path.join(project_alias)
-    |> get_page_fn.(preview_opts(conn.query_params))
+    |> get_page_fn.(conn.query_params)
     |> case do
       %Project{} = project ->
         breadcrumbs = [
--- a/apps/site/test/site_web/controllers/project_controller_test.exs
+++ b/apps/site/test/site_web/controllers/project_controller_test.exs
@@ -7,6 +7,13 @@ defmodule SiteWeb.ProjectControllerTest do
   alias Plug.Conn
 
   describe "project_updates" do
+    test "successfully renders the project update list", %{conn: conn} do
+      path = project_updates_path(conn, :project_updates, "project-name")
+
+      conn = get(conn, path)
+      assert conn.status == 200
+    end
+
     test "renders a list of updates related to a project", %{conn: conn} do
       project = project_factory(0)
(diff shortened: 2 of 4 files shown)


apps/v3_api/test/routes_test.exs
```diff
@@ -6,8 +6,7 @@ defmodule V3Api.RoutesTest do
   @opts ["page[limit]": 1, sort: "long_name"]
 
   describe "all/1" do
-    %JsonApi{data: [%JsonApi.Item{} = route]} = Routes.all(@opts)
-    assert route.id == "80"
+    %JsonApi{data: [%JsonApi.Item{} | _]} = Routes.all(@opts)
```
digitalcora: Given the `page[limit]=1`, shouldn't this have only been returning one item? Did the parameter stop working?

phildarnowsky: no, I just missed it, will remove

phildarnowsky: no, I just missed it, will remove

apps/site/lib/site_web/controllers/project_controller.ex
```diff
@@ -97,7 +95,7 @@ defmodule SiteWeb.ProjectController do
 
     "/projects"
     |> Path.join(project_alias)
-    |> get_page_fn.(preview_opts(conn.query_params))
+    |> get_page_fn.(conn.query_params)
```
digitalcora: It's fine as a quick fix, but this seems contra to the separation-of-concerns idea discussed in #347 of having controllers be solely responsible for extracting and parsing query params. @amaisano would refactoring this be included in the new Asana task you created?

apps/site/test/site_web/controllers/project_controller_test.exs
```diff
@@ -7,6 +7,13 @@ defmodule SiteWeb.ProjectControllerTest do
   alias Plug.Conn
 
   describe "project_updates" do
+    test "successfully renders the project update list", %{conn: conn} do
+      path = project_updates_path(conn, :project_updates, "project-name")
+
+      conn = get(conn, path)
+      assert conn.status == 200
```
digitalcora: Can we go a bit further here? Something like

```elixir
assert html_response(conn, 200) =~ "some string that should be on the page"
```

apps/site/lib/site_web/controllers/project_controller.ex
amaisano: Yes - this will be refactored to use the new `preview_opts` shape with another PR.

apps/site/test/site_web/controllers/project_controller_test.exs
amaisano: I ensured that `CMS.Repo.get_page/2` is natively called via this test, which is all the fix needed. It would have otherwise issued a 500 response for `conn.status` prior to the fix.

digitalcora: Right, I understand this was the minimum required to expose the bug at hand. I was wondering if we couldn't get some extra value out of this test, while we're here, for a small additional effort.

PR #356, opened 2020-01-23
Add deprecation note for Title Card Set paragraph type

amaisano:
#### Summary of changes
**Asana Ticket:** No ticket.

Follow on to mbta/cms#321

This paragraph type is being retained for historical purposes (which is why we're not pulling it), but it is no longer used in any current CMS content (and can no longer be added via the CMS). Use Descriptive Links and MultiColumn layout instead.

Also note that this struct is being hard coded at the bottom of news pages in Elixir land, but that is the only remaining artifact.

### DIFF
--- a/apps/cms/lib/partial/paragraph/title_card_set.ex
+++ b/apps/cms/lib/partial/paragraph/title_card_set.ex
@@ -1,5 +1,6 @@
 defmodule CMS.Partial.Paragraph.TitleCardSet do
   @moduledoc """
+  DEPRECATED. Use %DescriptiveLink{} structs inside %Column{} structs instead.
   Represents a collection of Title Cards in the CMS
   """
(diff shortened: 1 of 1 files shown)


digitalcora:
> This paragraph type is being retained for historical purposes (which is why we're not pulling it), but it is no longer used in any current CMS content (and can no longer be added via the CMS). Use Descriptive Links and MultiColumn layout instead.

Can we put this full explanation in the moduledoc? Otherwise looks fine, of course, not sure what's going on with CI but you shouldn't treat that as a blocker for merge.

digitalcora:
btw, I think Semaphore supports the "standard" of adding `[ci skip]` to the body of the commit message in order to skip the build for that commit. Useful in cases where we only change documentation 🙂

PR #357, opened 2020-01-23
Line diagram predictions

digitalcora:
**Asana ticket:** [Line Diagram | Show live predictions](https://app.asana.com/0/385363666817452/1155489101061691/f)

**Also resolves:** [🐞 Ferry showing scheduled trips as "realtime"](https://app.asana.com/0/385363666817452/1152119852318081/f)

There are some small refactorings sprinkled in here as separate commits, so I'd suggest reviewing them individually rather than the whole changeset 🙂

Notes for @ingridpierre:

* The refresh interval is currently set to 15 seconds, which is the same as most other places we do live data (in some places it's 30 seconds). I think this is a decent balance between being up-to-date and not sucking battery on mobile, but not sure if you had a specific interval in mind.

* Currently if there's an issue fetching the live data (e.g. the user's network drops out), we retain and continue to display whatever data we had previously. Should we clear the live info in this case, so as not to be misleading? Should we have an "error" state that retains the info but displays it differently? If so, what should that look like?

* I set the prediction text to not wrap, which I think normally is what we want, but when the full "Headsign • Train • Track • Status" is present for CR it can really squish the connection icons, or even overflow the stop card on small screens or with long headsigns. Should we allow a split here? (maybe Headsign goes on its own line and everything else on a new line?)

* For ferry lines, Zeroheight says we should show the next scheduled departure, but Katrina indicated [in this bug ticket](https://app.asana.com/0/385363666817452/1152119852318081/f) that we should not show schedules here at all (I assume because, as with commuter rail, a trip could be cancelled and then the "departure" would be misleading). Is that correct or should we show the scheduled time?


apps/predictions/lib/repo.ex
```diff
@@ -1,12 +1,13 @@
...
+  @moduledoc "Repo for V3 API Prediction resources."
   use RepoCache, ttl: :timer.seconds(10), ttl_check: :timer.seconds(2)
   require Logger
   alias Predictions.Parser
   alias Stops.Stop
 
   @default_params [
     "fields[prediction]":
-      "track,status,departure_time,arrival_time,direction_id,schedule_relationship,stop_sequence",
+      "status,departure_time,arrival_time,direction_id,schedule_relationship,stop_sequence",
```
thecristen: Why did you remove the track here (and only here?)?

apps/site/assets/ts/schedule/components/line-diagram/StopPredictions.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import { Headsign, Route } from "../../../__v3api";
+import {
+  timeForCommuterRail,
+  statusForCommuterRail
+} from "../../../helpers/prediction-helpers";
+
+interface Props {
+  headsigns: Headsign[];
+  route: Route | null;
```
thecristen: Since this is only used to determine whether the route is a commuter rail route or not, would it make more sense to have a boolean prop like `isCommuterRail` here instead?

apps/site/assets/ts/schedule/__tests__/SingleStopTest.tsx
```diff
@@ -0,0 +1,172 @@
...
+        color="000"
+        isOrigin={false}
+        isDestination={false}
+        liveData={crLiveData}
+      />
+    );
+
+    expect(wrapper.debug()).toMatchSnapshot();
+  });
+});
```
thecristen: Can we add some tests here to check whether the `.m-schedule-diagram__prediction` text matches an expectation?

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -38,28 +36,45 @@ defmodule SiteWeb.ScheduleController.LineApi do
...
   def update_route_stop_data({data, %RouteStop{id: stop_id} = map}, alerts, date) do
     %{
-      alerts: alerts |> Stop.match(stop_id) |> json_safe_alerts(date),
+      alerts: alerts |> AlertsStop.match(stop_id) |> json_safe_alerts(date),
       route_stop: RouteStop.to_json_safe(map),
       stop_data: Enum.map(data, fn {key, value} -> %{branch: key, type: value} end)
     }
   end
+
+  @spec get_line_data(Route.id_t(), 0 | 1, Route.branch_name(), boolean()) :: [
```
thecristen: May as well use the `direction_id` type here too

apps/predictions/lib/repo.ex
digitalcora: It's explained in the commit dedicated to it: https://github.com/mbta/dotcom/pull/357/commits/20b6dd759160b819e87c229a0e290870e3d57e4f

`track` is not a field that exists on Prediction resources in the API. It may have existed in an earlier version of the API, or maybe someone just got confused by our custom data types which do have a "track" attached directly to predictions. In any case, requesting a nonexistent field from the API still works — it was just unnecessary, and potentially confusing for future readers of the code.

apps/site/assets/ts/schedule/components/line-diagram/StopPredictions.tsx
digitalcora: Maybe. The `if (route.type === 2)` logic does have to live somewhere — do you think it should be at the point where the `StopPredictions` component is invoked?

apps/site/assets/ts/schedule/__tests__/SingleStopTest.tsx
digitalcora: I considered it, but technically this _is_ such an expectation — the snapshot contains all text rendered by the component, and would fail if the text changed. Maybe this is too much of a "scattershot" approach, though? What's your own feeling on when to write a snapshot test vs. standard assertions?

ingridpierre:
Looks like I was overly ambitious about what fits on a line. Seems like div.m-schedule-diagram__predictions needs to move down to its own row instead of sharing width with the connection icons.

**CURRENT:**
image
**DESIRED:**
image

ingridpierre:
Also, we may want to check on super long headsign routes like the 435 and see what we can do to help the wrapping situation, even when the predictions do get their own rows.
image

ingridpierre:
Noticed what some might find to see as a regression (comparing old line diagram predictions to new), looks like we used to be able to show a little extra info about CR shuttle buses. They come through the timetable and the line diagram as "BXXXX" trains (B=shuttle bus). But on this branch they no longer get picked up for display.

image

timetable reference:
image

ingridpierre:
Interesting! Seeing some additional predictions here in your pr compared to the one in prod. Not complaining, just curious about the difference - potentially demo-able improvement as well. Note that the missing predictions are because of shuttling, but your branch accurately shows that we are providing service starting at Harvard.

image

Apart from the comments above, the rest of the functionality in this PR looks great to me. The polling speed seemed fine, though if we are very concerned it would be good to try it on dev green on some older mobile phones on Monday in the office.

digitalcora:
@ingridpierre 

> Noticed what some might find to see as a regression (comparing old line diagram predictions to new), looks like we used to be able to show a little extra info about CR shuttle buses. They come through the timetable and the line diagram as "BXXXX" trains (B=shuttle bus). But on this branch they no longer get picked up for display.

Without looking, I can say this is almost certainly because the shuttle buses are only part of the schedule data and not the predictions, and in this branch we are only showing prediction data (as per our earlier conversation).

ingridpierre:
@digitalcora For bullet four, do not show scheduled or predicted times at all, for the reasons you described (we don't want to show trip info for something that is cancelled but won't also update). I've made a note to update the documentation to reflect that later today.

ingridpierre:
@digitalcora Still talking with the signs folks about conventions for bullet #2, sounds like they're still coming up with fallbacks of their own. As I'm mulling it over with them, does the website know its own connection state? You suggested an error state is possible so I'm assuming we can detect a lost connection.

digitalcora:
> does the website know its own connection state? You suggested an error state is possible so I'm assuming we can detect a lost connection.

We know whether the last polling request was successful or not, which _could_ represent an issue on our end but in most cases would indicate a problem with the user's connection.

digitalcora:
> ETA: There may be time thresholds to consider, so that we don't over-announce data issues on a mostly stable but intermittently shaky connection. Besides knowing connected state, is having a threshold for holding a stale prediction possible?

On a device that is consistently _connected_ in some sense but has a poor-quality connection, we'd likely never see the error state, as browsers will wait a long time (at least 30 seconds) before giving up on a request. If the device is fully disconnected, the waiting doesn't happen and the request fails immediately. So it depends what kind of "shaky" you'd like to account for.

digitalcora:
@ingridpierre The new styles are on this branch if you'd like to check them out. I think the only other design item that's unresolved is this discussion about the error state, if we're going to include that in this PR (I think it could be a separate task that doesn't block the release of the new diagram, though). Let me know if there's something else I'm missing!

ingridpierre:
Hm the stop icons are broken locally for me, this is what i'm seeing as output. Should I try again?
image

ingridpierre:
- Stop icons (the little white circles on the line diagram) are still missing/broken on dev green.

- Really love the style changes, lines are wrapping where they ought to - everything is nicely visually balanced. Much improved.

**Lastly, up to you if you want to declare this as separate work:**
For the error/lost connection, let's keep the "stale" predictions if we had them on screen but style them in a faded "grey color'. This is similar to the style we default to on TNM when a prediction is not available for an active bus or subway route. This very minor visual difference is consistent with how Transit app shows our predictions as well with a lost connection. Until the screens/CTD design team lands on some rules/fallbacks for these states, this will help us keep it moving for this pr.
image

image

apps/site/assets/ts/schedule/__tests__/SingleStopTest.tsx
thecristen: The snapshots don't necessarily validate that your text rendering logic works in the way you expect, but it's up to you! Looks like the coverage checker is happy, anyhow.

apps/site/assets/ts/schedule/components/line-diagram/StopPredictions.tsx
thecristen: Yes.

digitalcora:
> Stop icons (the little white circles on the line diagram) are still missing/broken on dev green.

I've fixed this, it was an oversight when I refactored some CSS.

> Lastly, up to you if you want to declare this as separate work:

I think it should be separate work =) Mostly because I'd like to discuss the design a bit as I think the data situation here on the line diagram is a bit different from those other places you mentioned.

PR #359, opened 2020-01-24
Schedule Finder | Reorganize services (date) picker (polish task)

thecristen:
#### Summary of changes
**Asana Ticket:** [Schedule Finder | Reorganize services (date) picker | Polish](https://app.asana.com/0/385363666817452/1133934151259693/f)

* Reverts work from [this past commit](https://github.com/mbta/dotcom/commit/171281e838cf51942b5c567d1c3ebd4192d91288)
* Adds the new rating dates and description to our service data model
* Adjusts the wording in the service picker, using the rating description if available


apps/services/lib/service.ex
```diff
@@ -46,19 +50,24 @@ defmodule Services.Service do
           start_date: Date.t() | nil,
           type: type,
           typicality: typicality,
-          valid_days: [valid_day]
+          valid_days: [valid_day],
+          rating_start_date: Date.t() | nil,
+          rating_end_date: Date.t() | nil,
+          rating_description: String.t() | nil
```
digitalcora: Given the default-to-empty-string logic below, it doesn't seem like this field _should_ ever be nil (and it looks like the type of the regular `description`, which has the same logic, is just listed as `String.t()`). Of course, we aren't using `@enforce_keys`, so  ¯\\\_(ツ)\_/¯ (it could be a larger tech debt project to make our enforced keys align with our typespecs)

apps/site/assets/ts/__v3api.d.ts
```diff
@@ -242,6 +242,9 @@ export interface Service {
   typicality: ServiceTypicality;
   valid_days: DayInteger[];
   name: string;
+  rating_start_date: string;
+  rating_end_date: string;
```
digitalcora: Based on the logic and typespecs in `Service`, shouldn't these two be `string | null`?

apps/site/assets/ts/helpers/service.ts
```diff
@@ -1,35 +1,14 @@
...
-  `${monthIntegerToString(date.getUTCMonth())} ${date.getUTCDate()}`;
-
-const holidayDate = (service: Service): string => {
-  const note = service.added_dates_notes[service.start_date];
-  return note || `on ${formattedDate(new Date(service.start_date))}`;
-};
+import { Service } from "../__v3api";
+import { shortDate } from "./date";
+
+const formattedDate = (date: Date): string => shortDate(date);
```
digitalcora: Could we replace instances of `formattedDate` with `shortDate` and remove this delegation?

apps/site/assets/ts/helpers/service.ts
```diff
@@ -103,44 +75,46 @@ export const groupServiceByDate = (
...
+  future: "Future Schedules"
+};
 
-    if (serviceDateTime < startDateTime) {
-      return `starts ${formattedDate(startDateObject)}`;
+export const groupedServiceOptLabel = (
+  groupName: ServiceOptGroupName,
+  groupedServices: ServiceByOptGroup[]
+): string => {
+  // TODO: Validate whether all services in a group will have the same rating description and service period. Also check correspondence between service period and rating start/end dates.
```
digitalcora: Could we group the services based on a combination of rating description + service period? Then we wouldn't need to make this assumption, and maybe the service picker would automatically handle a wider range of unusual situations.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
```diff
@@ -194,7 +178,8 @@ export const ServiceSelector = ({
               route: routeId,
               origin: stopId,
               direction: directionId,
-              date: selectedService.end_date
+              date: services.find(service => service.id === selectedServiceId)!
```
digitalcora: Can we avoid the need to not-null this? For example, pick the first service, in the event that there's no selected service? Or do we know for sure this cannot ever be null, and need to get the compiler to understand that?

ingridpierre:
It looks like if I look at a weekday-only route during the weekend (the CT2 for example), the service returns no schedules:
image

Not super sure why this is occurring, that's just my guess. There is an ongoing detour but it shouldn't affect this stop from showing schedules in the outbound direction.

ingridpierre:
Like the above, this is an extant bug in production, but logging it here anyway to check on it while we're here if you like.  Looking at the 504, looks like the weekday schedules isn't populating (but in this case the route runs on weekdays AND weekends, just not on Sunday). Is there something in our logic that prevents schedules from showing when some services aren't present?

image

ingridpierre:
Probably repeating myself from earlier this week, but I'm seeing the duplicated services for subway here, but they shouldn't be a concern given we're replacing the Daily Schedules content for subway in another PR.

thecristen:
@ingridpierre Looking at the 504, I wonder if you saw no schedules because you happened to be look at it on a Sunday? 😱

image

apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
thecristen: Sadly this is a latter case, as Array.find is typed to possibly return undefined

ingridpierre:
@thecristen that seems to be so! eek!

apps/site/assets/ts/__v3api.d.ts
thecristen: Should it? I'm trying to reuse the same logic for start_date and end_date.

apps/site/assets/ts/helpers/service.ts
thecristen: Done

thecristen:
@digitalcora @ingridpierre This is ready for re-review! 

Also I don't _think_ there's any logic in here that would stop schedules from showing if the service is selected. @ingridpierre, are there any routes that don't run on weekdays that we can look at? 🤔

ingridpierre:
@thecristen  Mayyyybe? The shuttles that run on the weekend actually get schedule pages, and you can see them (if you make the grave mistake of clicking our suggested links at the bottom of a Trip Planner itinerary) https://mbta.com/schedules/Shuttle-AlewifeHarvard/line?date=2020-02-08&trip=43146036-AlewifeHarvard1 . 

Here I'm looking at a weekend-only shuttle, but i'm unable to see the schedule I've chosen bc it doesn't run on weekdays. Kinda the reverse of the 504.
image

ingridpierre:
oh wait it does give me the sunday service, but not saturday. BIZARRE.
image

ingridpierre:
Since the bugginess is already in prod, we wouldn't want to address it with this ticket. Pulling down your branch now for final review.

ingridpierre:
Bus is looking good, but looks like some of the changes aren't taking on CR, and also Ferry:
image.

image

apps/site/assets/ts/__v3api.d.ts
thecristen: Yep, these can be null, often on ferries or commuter rail routes 🙃

apps/site/assets/ts/helpers/service.ts
```diff
@@ -1,155 +1,197 @@
...
+  Current: 4,
+  Holiday: 3,
+  Future: 2,
+  Other: 1
 };
 
-type MonthInteger = number | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11;
+export const optGroupComparator = (groupA: string, groupB: string): number => {
+  const a = groupA.split(" ")[0];
+  const b = groupB.split(" ")[0];
```
digitalcora: This seems fragile — is there any way we can accomplish the same goal without relying on the string names of the option groups?

apps/site/assets/ts/helpers/service.ts
```diff
@@ -1,155 +1,197 @@
...
+  ][day - 1];
+
+const daysAreConsecutive = (
+  days: DayInteger[],
+  areConsecutive: boolean
+): boolean => {
+  if (!areConsecutive) return areConsecutive;
+  const [day, nextDay, ...rest] = days;
+  if (rest.length === 0) return day + 1 === nextDay;
+  return daysAreConsecutive([nextDay, ...rest], day + 1 === nextDay);
```
digitalcora: I think we could express the same logic without needing an extra argument on the function:

```js
const daysAreConsecutive = (days: DayInteger[]): boolean => {
  const [day, nextDay, ...rest] = days;
  if (rest.length === 0) return day + 1 === nextDay;
  if (day + 1 !== nextDay) return false;
  return daysAreConsecutive([nextDay, ...rest]);
};
```

apps/site/assets/ts/helpers/service.ts
```diff
@@ -1,155 +1,197 @@
...
+};
+
+export const serviceDays = ({
+  type,
+  valid_days: validDays
+}: Service): string => {
+  if (type === "saturday" || type === "sunday") return "";
```
digitalcora: What's the purpose of this check? It seems like it would be fine to get the service days of a saturday or sunday service, if we ever wanted to (it doesn't appear we ever go down this branch in practice, since the only calls to this function are guarded by `isMultipleWeekday`).

apps/site/assets/ts/schedule/__tests__/ServiceOptGroupTest.tsx
```diff
@@ -2,18 +2,30 @@ import React from "react";
...
       />
     );
     expect(tree.toJSON()).toBeNull();
   });
+
+  it("renders if services present", () => {
+    createReactRoot();
+    const tree = renderer.create(
+      <ServiceOptGroup
+        label={"This label does not appear"}
```
digitalcora: In this case, I think it does appear? 😄

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOptGroup.tsx
```diff
@@ -1,34 +1,67 @@
...
+            startDate === endDate
+              ? shortDate(startDate)
+              : startToEnd(startDate, endDate);
+
+          optionText = `Storm service, ${stormServicePeriod}`;
+        } else if (service.typicality === "holiday_service") {
+          optionText = service.added_dates
+            .map(
+              addedDate =>
+                `${service.added_dates_notes[addedDate]}, ${shortDate(
```
digitalcora: Do we know if `added_dates_notes` will always be there for a holiday service? I seem to recall a recent conversation about how it wasn't present for some holiday.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOptGroup.tsx
```diff
@@ -1,34 +1,67 @@
...
+                  new Date(addedDate)
+                )}`
+            )
+            .join(" ");
+        } else if (label === "Other Schedules") {
```
digitalcora: Can we do this without relying on a comparison against the user-facing title of the group? If we change the wording, for example, the logic here would fail.

apps/site/assets/ts/schedule/__tests__/ServiceOptGroupTest.tsx
thecristen: Whoops, bad copy paste, thanks

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOptGroup.tsx
thecristen: @phildarnowsky Do you know?

thecristen:
Design note: This defaults to an existing service if there is no service on the day this is being viewed -- for example this is what the 19 from Ruggles will look like on a Saturday. 

| Before | After |
| --- | --- |
| IMG_2350 | image |

This is an improvement but I'm wondering if we should be clearer about "hey, there's no service **today**", and how we should try to communicate that. cc @ingridpierre

apps/site/assets/ts/helpers/service.ts
thecristen: You're right about not going down this branch in practice -- this is only used to describe valid days of service during weekdays -- e.g. `Tuesday - Friday` or `Monday, Friday` or other such patterns.

apps/site/assets/ts/helpers/service.ts
thecristen: Good call, I think that argument was an artifact of how this function was invoked previously. I'll adjust that

apps/site/assets/ts/helpers/service.ts
thecristen: I'm not sure how to address this since the whole point of this function is to sort the string names of the option groups?

apps/site/assets/ts/helpers/__tests__/service-test.ts
```diff
@@ -1,111 +1,345 @@
...
+    );
+    const futureService = grouped[futureKey!][0];
+    const name = `Future Schedules (${
+      futureService.rating_description
+    }, starts ${shortDate(new Date(futureService.rating_start_date!))})`;
+    expect(futureKey).toEqual(name);
   });
+});
+
+it("optGroupComparator sorts properly", () => {
```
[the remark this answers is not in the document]
thecristen: I think you missed a closing bracket? 😅 The current setup is this:

```
describe("groupServicesByDateRating"
  it("lists future services as 'future'"
  it("lists other services as 'other'"
  ...
)

it("optGroupComparator sorts properly"
it("serviceDays lists weekday days of service"
```

There aren't as many tests for the other functions so I'd chosen not to wrap them each in their own `describe`.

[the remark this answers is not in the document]
digitalcora: I did miss a closing brace 😑

I still think there is test readability value in having the function name in a `describe` block and test cases in `it` blocks, even if there's only one test case, but wouldn't say it's essential.

apps/site/assets/ts/helpers/service.ts
digitalcora: I think it's sort of the same thing you addressed by adding the `enum` — if the user-facing names of the option groups changed, they would sort in a different order (despite that our intention is e.g. "the one that represents current schedules should be first" and not "the one that starts with the string `Current` should be first"). It'd be great if `optGroupOrder` could now use the `enum`, since we have that.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceSelector.tsx
thecristen: Yes, now I have -- see latest commit. We still need to use the ID for the select onChange event.

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOptGroup.tsx
phildarnowsky: @digitalcora @thecristen as I understood what we agreed on with the data team, `added_dates_notes` should be present for all holidays. Is that how you remember it @paulswartz @handorff ?

ingridpierre:
As a potentially simple addition, maybe we should still display that grey box, but with "There is no scheduled service for *today*" *above* the service picker.

thecristen:
@ingridpierre Like this?

image

apps/site/assets/ts/schedule/components/schedule-finder/ServiceOptGroup.tsx
thecristen: @digitalcora I have updated this bit of logic, so that it'll fall back to a different bit of text if for whatever reason there is no `added_dates_notes` present.

PR #361, opened 2020-01-28
Add conditional display of schedule note to modal

thecristen:
#### Summary of changes
**Asana Ticket:** [Schedule Finder | Show schedule note instead of trips for subway](https://app.asana.com/0/385363666817452/1156654983519748/f)

If there is a schedule note present, which should happen on subway lines, the schedule note will display in place of the service picker + daily schedules listing.

image

It's mostly the same code from #354 so I'm hopeful review will be easy!

### DIFF
--- a/apps/site/assets/css/_schedule-page.scss
+++ b/apps/site/assets/css/_schedule-page.scss
@@ -58,6 +58,19 @@
   padding: $base-spacing * 2 $base-spacing;
 }
 
+.m-schedule-page__schedule-notes--modal {
+  margin-top: $base-spacing;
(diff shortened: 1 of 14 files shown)


digitalcora:
@phildarnowsky Can you take this review? Not sure if I'll have some free time today with trying to wrap up the vehicle tracking work.

phildarnowsky:
@digitalcora /cc @thecristen I got it

ingridpierre:
@thecristen Looking good from where I sit! Thanks for breaking this out so quickly Cristen!

PR #363, opened 2020-01-29
Show real-time vehicles on new line diagram

digitalcora:
Visualizing Green Line bunching in real-time! 🚀

### DIFF
--- a/apps/site/assets/css/_schedule-page-line-diagram.scss
+++ b/apps/site/assets/css/_schedule-page-line-diagram.scss
@@ -175,6 +175,7 @@ $line-width: 8px; // width of the colored line in the diagram
 .m-schedule-diagram__line {
   background-color: currentColor;
   margin: 0 $line-width / 2;
+  position: relative;
   width: $line-width;
 }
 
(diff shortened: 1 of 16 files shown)


ingridpierre:
For commuter rail only, can we add train #s to the tooltip? Looks like we do it on the timetable and the map, but not through this version of the line diagram. See:
image
image
**CURRENT:**
image

**DESIRED:**
image

ingridpierre:
Also, can we left align the tooltip label? Or have some general tooltip rules that update the text alignment based on where the arrow is pointing?  (e.g.: if arrow is left or right, text-align left. If arrow is top or bottom, text-align center.)

digitalcora:
> For commuter rail only, can we add train #s to the tooltip?

Done.

> Also, can we left align the tooltip label here?

Mmmmmaybe. This turned out to be surprisingly non-trivial so I'll get back to you.

ingridpierre:
IE 11, Win 10 - no map or line diagram rendered apart from the header. 😨 
image

digitalcora:
@ingridpierre Cristen noticed this as well, I'm working on a fix at the moment

PR #364, opened 2020-01-29
Special rendering of Silver Line Waterfront trips & route page

phildarnowsky:
Probably best reviewed one commit at a time.

#### Summary of changes
**Asana Ticket:** [🐞Schedule Finder | Silver Line variants are displayed incorrectly](https://app.asana.com/0/555089885850811/1148141637658238)

Per the ticket:

> Silver Line buses like 746 (Silver Line Waterfront) can be seen inside other SL schedules, and 
> found via google (they are not usually directly linked on MBTA.com). 
> When viewing schedule pages and schedule finder results for these vehicles we see poorly styled > results. 

Here we improve the rendering of SL Waterfront results in the places specified in the ticket, as well as a few other places that seemed logical (e.g. the variant selector in the direction picker).

### DIFF
--- a/apps/routes/lib/route.ex
+++ b/apps/routes/lib/route.ex
@@ -1,4 +1,6 @@
 defmodule Routes.Route do
+  @moduledoc false
+
   alias Routes.Repo
 
   defstruct id: "",
@@ -46,6 +48,7 @@ defmodule Routes.Route do
 
   @silver_line ~w(741 742 743 746 749 751)
   @silver_line_set MapSet.new(@silver_line)
+  @silver_line_waterfront "746"
 
   @spec type_atom(t | type_int | String.t()) :: gtfs_route_type
   def type_atom(%__MODULE__{type: type}), do: type_atom(type)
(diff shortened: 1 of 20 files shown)


digitalcora:
All this special-casing of headsigns and destination names is making me wonder whether we shouldn't be addressing the problem in the underlying data. Otherwise, how can we expect anyone else to get this right? We might have hard-coded a bunch of stuff on the web site, but it'll still look weird in e.g. Transit.

paulswartz:
Transit also shows "SL" as the route name, so we can look at switching it in the data. The other information (like the color) is correct in the API.

phildarnowsky:
@paulswartz /cc @digitalcora So I looked into it, and here's what would have to change on the transit data side to implement as much as possible of this feature without special-case code:

* On route 746 itself:
  * Both `direction_destinations` changed to "Silver Line Waterfront"
  * `short_name` changed to "SL"
* On each route pattern for route 746:
  * `headsign` changed to "Silver Line Waterfront"
* On each trip for route 746:
  * `headsign` changed to "Silver Line Waterfront"

There would still need to be a little special-case code on the dotcom side, but not nearly as much. Please advise how big a lift that would be.

ingridpierre:
@phildarnowsky @paulswartz  short name should be changed to SLW. Changing it to SL only makes it seem like a parent route of some sort (like GL is to GL-B)

digitalcora:
@phildarnowsky Should we close this? It seems like most of the relevant work has been offloaded to the API and #394.

PR #366, opened 2020-01-30
Fix new schedules being a blank page in IE11

digitalcora:
**Asana Ticket:** [🐞 New schedules page is blank in IE11](https://app.asana.com/0/385363666817452/1159531751432025/f)

So as not to mess with our existing Webpack config at all, this creates a new config that explicitly compiles `react-async` using the same options. I've tested this using CBT and it allows IE11 to load the line diagram, though unfortunately the live data is still missing (and it's very unclear why).

### DIFF
--- a/apps/site/assets/tsconfig.webpack.json
+++ b/apps/site/assets/tsconfig.webpack.json
@@ -2,7 +2,6 @@
   "extends": "./tsconfig",
   "exclude": [
     "**/__tests__/**/*",
-    "node_modules",
     "vendor/",
     "ts/ts-build/"
   ]
--- a/apps/site/assets/webpack.config.base.js
+++ b/apps/site/assets/webpack.config.base.js
@@ -6,6 +6,20 @@ const webpack = require("webpack");
 const path = require("path");
 const postcssPresetEnv = require("postcss-preset-env");
 
+const babelLoader = {
+  loader: "babel-loader",
+  options: {
+    configFile: "./babel.config.js"
+  }
+};
+
+const tsLoader = {
+  loader: "ts-loader",
+  options: {
+    configFile: "tsconfig.webpack.json"
+  }
+};
+
 module.exports = {
   entry: {
     app: ["./js/app-entry.js"],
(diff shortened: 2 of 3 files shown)


apps/site/assets/tsconfig.webpack.json
```diff
@@ -2,7 +2,6 @@
   "extends": "./tsconfig",
   "exclude": [
     "**/__tests__/**/*",
-    "node_modules",
     "vendor/",
```
thecristen: I'm curious is this might work instead of removing `node_modules` entirely: https://github.com/webpack/webpack/issues/2031#issuecomment-219040479

thecristen: whoops, that's probably syntax for an older webpack version, maybe the newer comments have a better approach

apps/site/assets/webpack.config.base.js
```diff
@@ -50,6 +50,19 @@ module.exports = {
           loader: "babel-loader"
         }
       },
+      {
+        // https://docs.react-async.com/getting-started/installation#transpiling-for-legacy-browsers
+        test: /\/node_modules\/react-async\//,
```
thecristen: The linked page doesn't really explain why, but it directs users to exclude `core-js` from transpilation. Did you try that / could that help here?

apps/site/assets/tsconfig.webpack.json
thecristen: Or with an [include](https://github.com/babel/babel-loader/issues/171#issuecomment-486380160)

digitalcora: Good call on the `include`, I'm trying that approach now as it seems more reliable than regexes. I think we do still need to remove this line though, since it would filter out `node_modules` at the ts-loader level even after it's been included at the Webpack level. This is not otherwise a concern because we're also filtering out `node_modules` in the Webpack config everywhere we use ts-loader.

apps/site/assets/webpack.config.base.js
digitalcora: The example given assumes you're using a single rule to compile your whole app, and want to transpile _all_ of your `node_modules` instead of just react-async — in this case it's necessary to exclude the `core-js` package. Here we're creating a separate rule that only includes react-async, so we don't have to worry about that. But, as a separate tech debt task we should investigate whether we want to transpile all our packages. If nothing else it would solve this "oh I didn't realize it wouldn't work in IE" problem for good.

thecristen:
I'm looking into why this seems to break the Jest test suite -- so far it seems as though the few JS files that get used in some of our tests are not yet transpiled (but should already be transpiled in the test).

PR #368, opened 2020-01-31
Add direction and date to CR timetable ARIA label

phildarnowsky:
The ARIA label for the commuter rail timetable had the route name, which
was useful, but was lacking the direction and date. Added.

#### Summary of changes
**Asana Ticket:** [Timetable | Add direction to table aria label](https://app.asana.com/0/555089885850811/1130340429843892)

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
@@ -22,6 +22,11 @@ defmodule SiteWeb.ScheduleController.TimetableController do
     as: :assign_direction_id
 
   def show(conn, _) do
+    direction_id = conn.assigns[:direction_id]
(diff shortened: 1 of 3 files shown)


apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
```diff
@@ -22,13 +22,20 @@ defmodule SiteWeb.ScheduleController.TimetableController do
     as: :assign_direction_id
 
   def show(conn, _) do
+    direction_id = conn.assigns[:direction_id]
+    direction_name = conn.assigns.route.direction_names[direction_id]
+
+    {:ok, full_date} = Timex.format(conn.assigns.date, "{Mfull} {D}, {YYYY}")
```
digitalcora: Could we call this `formatted_date` or something to distinguish it from a struct? Might be confusing with all the `date` related stuff floating around.

PR #370, opened 2020-01-31
🐞 IE11 Line Diagram | Height of certain elements is wrong

thecristen:
#### Summary of changes
**Asana Ticket:** [🐞 IE11 Line Diagram | Height of certain elements is wrong](https://app.asana.com/0/385363666817452/1159531751432030/f)

Tweaks some of the markup and styles to ensure consistency between how the line diagram looks in IE11 and other browsers.

### DIFF
--- a/apps/site/assets/ts/schedule/components/line-diagram/SingleStop.tsx
+++ b/apps/site/assets/ts/schedule/components/line-diagram/SingleStop.tsx
@@ -143,12 +143,19 @@ const StopBranchLabel = (stop: RouteStop): JSX.Element | null =>
   ) : null;
 
 const StopGraphic = (isOrigin = false, isTerminus = false): JSX.Element => {
-  let yPosition = "-32px";
-  if (isTerminus) {
-    yPosition = isOrigin ? "-66px" : "32px";
+  // this hardcoded position will be shown in IE, otherwise overwritten by CSS
+  let yPosition = "32px";
+  if (isTerminus && isOrigin) {
+    yPosition = "-3px";
   }
   return (
-    <svg viewBox="0 10 10 10" className="m-schedule-diagram__line-stop">
+    <svg
+      viewBox="0 10 10 10"
+      preserveAspectRatio="xMidYMin slice"
+      width="100%"
+      height="10px"
+      className="m-schedule-diagram__line-stop"
+    >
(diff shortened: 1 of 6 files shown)


apps/site/assets/ts/schedule/components/line-diagram/SingleStop.tsx
```diff
@@ -143,12 +143,19 @@ const StopBranchLabel = (stop: RouteStop): JSX.Element | null =>
   ) : null;
 
 const StopGraphic = (isOrigin = false, isTerminus = false): JSX.Element => {
-  let yPosition = "-32px";
-  if (isTerminus) {
-    yPosition = isOrigin ? "-66px" : "32px";
+  // this hardcoded position will be shown in IE, otherwise overwritten by CSS
+  let yPosition = "32px";
+  if (isTerminus && isOrigin) {
```
digitalcora: Could a stop ever be an origin without being a terminus?

apps/site/assets/ts/schedule/components/line-diagram/SingleStop.tsx
```diff
@@ -143,12 +143,19 @@ const StopBranchLabel = (stop: RouteStop): JSX.Element | null =>
...
   return (
-    <svg viewBox="0 10 10 10" className="m-schedule-diagram__line-stop">
+    <svg
+      viewBox="0 10 10 10"
+      preserveAspectRatio="xMidYMin slice"
+      width="100%"
+      height="10px"
+      className="m-schedule-diagram__line-stop"
+    >
       <circle r={`${isTerminus ? "5" : "4"}`} cx="50%" cy={yPosition} />
```
digitalcora: Re: `yPosition` and the `r` attribute, I think we should consistently either have both of the conditionals as ternaries here, or both of them as variable assignments beforehand. IMO the conditions are short enough that ternaries would be perfectly readable so, no strong opinions here 🙂

apps/site/assets/ts/schedule/components/line-diagram/SingleStop.tsx
thecristen: Origin might be a bit of a misnomer, as it's just the very first stop in the diagram. So, a stop could be a terminus without being an "origin".

PR #371, opened 2020-01-31
Fix React key warnings in new line diagram

digitalcora:
A couple of elements in arrays didn't have unique keys, causing warnings from React.

### DIFF
--- a/apps/site/assets/ts/schedule/components/line-diagram/ExpandableBranch.tsx
+++ b/apps/site/assets/ts/schedule/components/line-diagram/ExpandableBranch.tsx
@@ -36,8 +36,8 @@ const ExpandableBranch = ({
               style={{ color: `#${color}` }}
             >
               <>
-                {times(branchData[0].stop_data.length - 1, () => (
-                  <div className="m-schedule-diagram__line" />
+                {times(branchData[0].stop_data.length - 1, index => (
+                  <div key={index} className="m-schedule-diagram__line" />
                 ))}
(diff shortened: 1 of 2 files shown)


apps/site/assets/ts/schedule/components/line-diagram/ExpandableBranch.tsx
```diff
@@ -36,8 +36,8 @@ const ExpandableBranch = ({
               style={{ color: `#${color}` }}
             >
               <>
-                {times(branchData[0].stop_data.length - 1, () => (
-                  <div className="m-schedule-diagram__line" />
+                {times(branchData[0].stop_data.length - 1, index => (
+                  <div key={index} className="m-schedule-diagram__line" />
```
digitalcora: While normally advised-against, this should be safe here because it's purely static content and will never update after the initial render.

apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
```diff
@@ -212,7 +212,12 @@ const LineDiagram = ({
 
               // is an array of 1-2 stops; show as expanded
               return (
-                <div className="m-schedule-diagram__expanded">
+                <div
+                  key={`${stopOrStops[0].route_stop.id}-${
+                    stopOrStops.length
+                  }-stops`}
+                  className="m-schedule-diagram__expanded"
+                >
```
digitalcora: This key is copied from the ExpandableBranch invocation just above.

PR #375, opened 2020-02-05
Schedules | Search

thecristen:
#### Summary of changes
**Asana Ticket:** [Schedules | Search](https://app.asana.com/0/385363666817452/1118913618186304/f)

Adds a search box above the line diagram that will filter the list of stops. It highlights the matching text with a colored background and shows a brief message if there are no matches found.

This PR adds a few new reusable functions to our codebase:

* **`useFilteredList()` our first custom React hook**: captures logic and manages state for filtering a list of objects or strings based on a string query.
* **`<SearchBox />`**: makes a React component out of our existing big search box
* **`<MatchHighlight />`**: captures logic for matching a substring of text and outputting HTML markup to display the text with the matching text highlighted


ingridpierre:
I am so impressed by how well this works - so smooth, can't fault it! 👍

apps/site/assets/ts/hooks/__tests__/useFilteredListTest.tsx
```diff
@@ -0,0 +1,89 @@
+import React from "react";
+import useFilteredList from "../useFilteredList";
+import { mount, ReactWrapper } from "enzyme";
+import stopData from "../../schedule/__tests__/lineDiagramData/simple.json";
```
digitalcora: This fixture data is a bit of a [mystery guest](http://xunitpatterns.com/Obscure%20Test.html#Mystery%20Guest); I think this test would be more maintainable and readable if we instead used a short list of "example objects" tailored to the needs of the test, defined here in the test file. For example:

```js
const objectList = [
  { id: "a", stop: { name: "Origin" } },
  { id: "b", stop: { name: "Stop" } },
  { id: "c", stop: { name: "Destination" } }
];
```

We could also then include edge cases if we'd like to test them, without messing with the data for the line diagram tests; for example an item whose `stop` is null or undefined.

apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
```diff
@@ -168,104 +170,143 @@ const LineDiagram = ({
...
-                if (adjacentStop && !Array.isArray(adjacentStop)) {
-                  willMerge = mergeIndices.includes(
-                    lineDiagram.indexOf(adjacentStop)
+      <SearchBox
+        id="stop-search"
+        labelText="Search for a stop"
+        onChange={setStopQuery}
+        className="m-schedule-diagram__filter"
+      />
+      {stopQuery !== "" ? (
```
digitalcora: The length of this ternary makes it a bit difficult to scan... would it be reasonable to extract each branch to its own private function/component?

apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
```diff
@@ -168,104 +170,143 @@ const LineDiagram = ({
...
-                  treeDirection === "inward" ? bldIndex + 1 : bldIndex - 1;
-                const adjacentStop = branchedLineDiagram[
-                  adjacentIndex
-                ] as LineDiagramStop;
-                if (adjacentStop && !Array.isArray(adjacentStop)) {
-                  willMerge = mergeIndices.includes(
-                    lineDiagram.indexOf(adjacentStop)
+      <SearchBox
+        id="stop-search"
+        labelText="Search for a stop"
```
digitalcora: This text always says "Search for a stop", but the header sometimes calls them "Stations".

apps/site/assets/ts/schedule/components/line-diagram/SingleStop.tsx
```diff
@@ -19,6 +20,7 @@ interface Props {
   isOrigin?: boolean;
   isDestination?: boolean;
   liveData?: LiveData;
+  query?: string;
```
digitalcora: Naming this `searchQuery` I think would make it clearer what it's referring to; since the functionality is part of a different component, it's not obvious what "query" means in this context (it could refer to the `query` part of the page URL, for example).

apps/site/assets/ts/schedule/components/line-diagram/SingleStop.tsx
```diff
@@ -229,7 +232,12 @@ const SingleStop = ({
           <div className="m-schedule-diagram__stop-name">
             <a href={`/stops/${routeStop.id}`}>
               <h4>
-                {MaybeAlert(stopAlerts)} {routeStop.name}{" "}
+                {MaybeAlert(stopAlerts)}
+                {query ? (
```
digitalcora: Based on how `MatchHighlight` is implemented, I don't think this conditional is needed? If there is no query, it just displays the text with no highlight.

apps/site/assets/ts/components/SearchBox.tsx
```diff
@@ -0,0 +1,74 @@
...
+    () => {
+      onChange(query);
+    },
+    [query, onChange]
+  );
+
+  return (
+    <div className={`${className} c-form__input-container`}>
+      {/* eslint jsx-a11y/label-has-associated-control: ["error", { assert: "either" } ] */}
+      {/* eslint jsx-a11y/label-has-for: ["error", { required: { some: ["nesting", "id"] } } ] */}
```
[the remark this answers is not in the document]
thecristen: You're right, these would be made redundant by that change.

apps/site/assets/ts/components/MatchHighlight.tsx
```diff
@@ -0,0 +1,47 @@
...
+  query: string;
+}
+
+// React component for highlighting matching text
+// will wrap the first substring matching the query
+// in a <b></b> and return the HTML result
+const MatchHighlight = ({
+  text,
+  query
+}: Props): DetailedReactHTMLElement<{}, HTMLElement> => {
```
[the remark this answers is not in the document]
thecristen: It's just the type that `React.createElement` happens to return.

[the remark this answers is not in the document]
digitalcora: Hrm. Wish there were better documentation of this stuff =\

apps/site/assets/ts/components/__tests__/SearchBoxTest.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  afterEach(() => {
+    wrapper.unmount();
+  });
+
+  it("calls onChange function on change event", () => {
+    wrapper
+      .find("input#test-search")
+      .simulate("change", { target: { value: "foo" } });
+    wrapper.setProps({}); // updates useEffect
```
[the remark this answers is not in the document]
thecristen: Yeah, its more of a workaround because functional React components don't have state or instances that Enzyme's methods can test well. Additionally, `wrapper.update()` doesn't work for this, [though you'd think it would](https://github.com/airbnb/enzyme/issues/2254). 

As far as I can tell, **_best practices_** around testing functional React components, particularly those using Hooks, have not really coalesced yet... and what little wisdom exists is spread across a wonderfully varied set of libraries and helpers.

That said. your mention of `act()` made me realize that implementing some of these tests with `react-test-renderer` works, and may look slightly less hacky! 

For example this:
```tsx
it("renders properly if multiple matches", () => {
    const wrapper = mount(
      <MatchHighlight
        text="As part of our $8 billion, 5-year capital investment plan, we're renovating stations, modernizing fare collection systems, upgrading services for our buses, subways, and ferries, and improving the accessibility of the entire system."
        query="and"
      />
    );
    wrapper.setProps({}); // actually updates the wrapper post-useEffect
    expect(wrapper.html()).toMatchSnapshot();
    expect(wrapper.find(".u-highlight")).toHaveLength(1);
  });
```
becomes this:
```tsx
  it("highlights only the first match if there are multiple matches", () => {
    act(() => {
      component = create(
        <MatchHighlight
          text="As part of our $8 billion, 5-year capital investment plan, we're renovating stations, modernizing fare collection systems, upgrading services for our buses, subways, and ferries, and improving the accessibility of the entire system."
          query="and"
        />
      );
    });

    const instance = component!.root;
    const highlight = instance.findByType("b");
    expect(highlight.children).toHaveLength(1);
  });
```

But would still necessitate additional/alternate tools to handle changing inputs and whatnot.

apps/site/assets/ts/components/MatchHighlight.tsx
```diff
@@ -0,0 +1,47 @@
...
+  >([text]);
+
+  useEffect(
+    () => {
+      if (query !== "") {
+        const q = new RegExp(query, "i");
+        const matched = text.match(q);
+        if (matched && matched[0]) {
+          // split text on first instance of match only
+          const [precedingTextNode, ...moreTextNodes] = text.split(matched[0]);
```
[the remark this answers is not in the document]
thecristen: Ah, I think I had been approaching this from the standpoint of, well, they become text nodes in the end. Also `moreTexts` (I'd wanted to keep it plural as it could be of any length) sounded weird to me.

apps/site/assets/ts/components/__tests__/SearchBoxTest.tsx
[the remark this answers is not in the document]
digitalcora: > a wonderfully varied set of libraries and helpers

So far, so JavaScript 😓 Makes me wonder how exactly Facebook is testing all their own functional components...

I can appreciate the slightly-less-hacky-ness of the `act` approach, though my personal preference would be to keep all our tests (or as many as possible) aligned on the same framework. This workaround is officially documented by Enzyme, and they've left the issue open which makes me hopeful it will have a "real" resolution eventually. Nice work sleuthing out that information 😃

apps/site/assets/ts/components/MatchHighlight.tsx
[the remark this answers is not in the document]
digitalcora: `moreTexts` sounds fine to me, but so would something generic like `rest` or `others`; this variable is only relevant for the very next line anyway.

apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
thecristen: I'll extract the filtered stop list piece, that should help with readability.

apps/site/assets/ts/schedule/components/line-diagram/SingleStop.tsx
thecristen: Slightly different issue -- the `query` prop expects a string, but here it can potentially be undefined (as `query` is an optional prop on this particular component). I could change the input to `query || ""` but it seems silly to include the unneeded `<MatchHighlight />` -- and allowing `undefined` to be passed in seems meh to me. Any suggestions?

digitalcora: My suggestion would be:

```jsx
<h4>
  {MaybeAlert(stopAlerts)}
  <MatchHighlight text={routeStop.name} query={query} />
</h4>
```

And make `query` an optional property. This keeps all the highlighting-related logic, including the decision about whether to do any highlighting, encapsulated in `MatchHighlight`.

There are other ways, for sure — what I think is non-ideal here is that the "whether to do highlighting" decision is currently _split_ between `MatchHighlight` and its callers (it checks for empty strings, but callers have to check for `undefined`).

digitalcora:
This is looking good, but do we know what's up with the code coverage change? Any important logic branches we're not testing?

digitalcora:
Ah, I can see it... weird, the coverage changes are reported in some Elixir code whose coverage shouldn't have been changed by this PR. We can look into this separately.

PR #377, opened 2020-02-07
Deal with no live data for now.

thecristen:
#### Summary of changes
**Asana Ticket:** [Pull out the code about polling](https://app.asana.com/0/385363666817452/1160944420414808/f)

Deleted code relevant to fetching and setting our live data but left relevant dependencies installed and affected components as is, setting the affected variable to an empty object.

### DIFF
--- a/apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
+++ b/apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
@@ -1,6 +1,5 @@
 import React, { ReactElement, useState } from "react";
 import { useFetch } from "react-async";
-import { useInterval } from "use-interval";
 import {
   LineDiagramStop,
   LineDiagramVehicle,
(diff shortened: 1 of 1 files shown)


apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
```diff
@@ -89,21 +87,7 @@ const LineDiagram = ({
...
-    `/schedules/line_api/realtime?id=${route.id}&direction_id=${directionId}`,
-    {},
-    { json: true, watch: directionId }
-  );
-  const liveData = (maybeLiveData || {}) as LiveDataByStop;
-  useInterval(() => {
-    /* istanbul ignore next */
-    if (!liveDataIsLoading) reloadLiveData();
-  }, 15000);
+  const liveData: LiveDataByStop = {};
```
phildarnowsky: @thecristen maybe I was unclear, but the idea was to remove just the `useInterval` call so that the page will load the `liveData` one time, without repeatedly polling it. @ingridpierre am I remembering our agreement of this morning correctly?

thecristen: Sorry about that, I've updated this accordingly.

PR #381, opened 2020-02-10
Handle missing route in route_to_class function

thecristen:
At least one schedule page 500'ed because this one function was being passed a nil route.


apps/site/lib/site_web/views/helpers.ex
```diff
@@ -181,6 +181,7 @@ defmodule SiteWeb.ViewHelpers do
 
   @doc "Returns a css class: a string with hyphens."
   @spec route_to_class(Routes.Route.t()) :: String.t()
+  def route_to_class(nil), do: nil
```
digitalcora: This should probably be `""` as the return, so existing code that expects it to always be a string will work (and so the typespec remains accurate).

thecristen: Done.

PR #388, opened 2020-02-17
Fix current service validity check

thecristen:
On a holiday when viewing the daily schedules.

Turns this (misleading, though the holiday schedule is still selectable and viewable here):

image

Into this (correct):

image

### DIFF
--- a/apps/site/assets/ts/helpers/service.ts
+++ b/apps/site/assets/ts/helpers/service.ts
@@ -21,6 +21,11 @@ const isInRemovedDates = (
   currentDate: Date = new Date()
 ): boolean => service.removed_dates.includes(dateObjectToString(currentDate));
 
+const isInAddedDates = (
+  service: Service,
+  currentDate: Date = new Date()
+): boolean => service.added_dates.includes(dateObjectToString(currentDate));
+
 const isOnValidDay = (
   service: Service,
   currentDate: Date = new Date()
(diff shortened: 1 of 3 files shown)


apps/site/assets/ts/helpers/service.ts
```diff
@@ -44,6 +49,11 @@ export const isCurrentValidService = (
   service: Service,
   currentDate: Date = new Date()
 ): boolean => {
+  // check against added dates
+  if (isInAddedDates(service, currentDate)) {
+    return true;
+  }
+
   // check against removed dates
```
thecristen: This is the key change. Everything else is mainly testing coverage enhancements.

apps/site/assets/ts/helpers/service.ts
```diff
@@ -91,11 +101,16 @@ export const isInFutureRating = (
...
   currentDate: Date = new Date()
 ): boolean => {
-  if (hasNoRating(service)) {
-    return false;
+  if (hasIncompleteRating(service)) {
+    // check for rating start date
+    return service.rating_start_date
+      ? currentDate <= stringToDateObject(service.rating_start_date!)
+      : false;
   }
```
thecristen: This change lets us list a future service with a start date, for CR or Ferry services that might have a rating start date but no rating end date

PR #389, opened 2020-02-18
Migrate to Stylelint

thecristen:
#### Summary of changes
**Asana Ticket:** [🔧 Tech Debt | Migrate off SCSS-Lint](https://app.asana.com/0/385363666817452/1156631985186489/f)

We're moving to [Stylelint](https://stylelint.io/)! This requires a completely new set of configuration files. I've setup a configuration of rules that should feel most similar to our old SCSS-lint config, though I've also cleaned up a bunch of older files to bring everything up to passing.

A few additional things happen with these changes:

* Node-Sass is replaced with [Dart Sass](https://sass-lang.com/dart-sass), which is currently the primary implementation of Sass.
* SCSS-lint relied on the Ruby implementation of Sass but Stylelint does not! So apparently Dotcom no longer needs Ruby. I've updated the documentation to reflect this.
* Updated NPM scripts
* Updated shell scripts
* Updated this repo's semaphore build settings!

Hoping I didn't miss anything here.


apps/site/assets/css/_checkbox.scss
```diff
@@ -3,7 +3,9 @@
   position: relative;
 }
 
+
 .c-checkbox__input {
+  /* stylelint-disable-next-line property-no-vendor-prefix */
```
digitalcora: As in `_radio.scss`, can we instead use `appearance: none` and let autoprefixer handle it? And in that case, would we still need `opacity: 0` or `width: 0`?

apps/site/assets/css/_schedule-page.scss
```diff
@@ -379,7 +379,7 @@
   border-top: 3px solid $gray-lightest;
   display: flex;
   justify-content: space-around;
-  // scss-lint:disable DuplicateProperty IE11 doesn't support space-evenly
+  // IE11 doesn't support space-evenly
```
digitalcora: This seems like it should be one line up, since it's really describing the `space-around` rule.

apps/site/assets/css/_stop-bubbles.scss
```diff
@@ -373,8 +367,8 @@ $location-line-width: $space-6;
   position: relative;
   top: 0;
   transform: scale(1.2); // vehicle SVGs are currently sized incorrectly
-                         // (circle is rendered inside of SVG and svg is too small);
-                         // this is a temporary fix until we get new assets -KH
+  // (circle is rendered inside of SVG and svg is too small);
+  // this is a temporary fix until we get new assets -KH
```
digitalcora: ditto about moving comments above

apps/site/assets/css/_alerts.scss
```diff
@@ -126,7 +126,7 @@
   position: absolute;
   top: $base-spacing;
 
-  [class*="c-svg__icon"] {
+  [class*='c-svg__icon'] {
```
digitalcora: If the current codebase prefers double quotes, could we configure Stylelint to prefer them and avoid the diff noise of converting the quote style? Or do we have a mix of doubles and singles currently?

apps/site/assets/css/_trip-plan-form.scss
```diff
@@ -73,10 +73,10 @@
...
     top: 0;
     z-index: $z-index-datepicker-content; // required so that a new stacking context is defined;
-                                          // without setting z-index, the footer's stacking context
-                                          // is higher than this sidebar so the datepicker popup
-                                          // is covered by the footer. -kh
-                                          // https://philipwalton.com/articles/what-no-one-told-you-about-z-index/
+    // without setting z-index, the footer's stacking context
+    // is higher than this sidebar so the datepicker popup
+    // is covered by the footer. -kh
+    // https://philipwalton.com/articles/what-no-one-told-you-about-z-index/
```
digitalcora: more disjointed comments

apps/site/assets/package.json
```diff
@@ -107,6 +113,7 @@
     "ts:build": "cd ts/ && tsc",
     "eslint": "cd js && npx eslint --ext .js .",
     "tslint": "cd ts && npx eslint --ext .ts,.tsx --max-warnings=0 .",
+    "stylelint": "npx stylelint css/**/*.scss --ignore-path .stylelintignore",
```
digitalcora: `.stylelintignore` appears to be the default path for the ignore file; do we need to specify it here?

docs/TESTING.md
```diff
@@ -1,8 +1,8 @@
 ## Tests
 
-* `mix test` — Elixir tests
-* `npm run test:js` — JS tests
-* `npm run backstop` — Backstop tests (see section below for details)
+- `mix test` — Elixir tests
+- `npm run test:js` — JS tests
+- `npm run backstop` — Backstop tests (see section below for details)
```
digitalcora: Same comment as in the README re: diff noise

semaphore/stylelint.sh
```diff
@@ -7,6 +7,5 @@ changed=$(git diff --name-only "$fork_point" apps/site/assets/css)
 if [ -z "$changed" ]; then
   echo "No CSS files changed relative to origin/master fork point."
 else
-  gem install scss_lint -v 0.59.0 --no-document
-  scss-lint $changed
+  npx stylelint $changed --config apps/site/assets/.stylelintrc --ignore-path apps/site/assets/.stylelintignore
```
digitalcora: I don't think we need this script anymore, since (thanks to your work in this PR!) all current SCSS should pass lint anyway. We could replace the whole Semaphore step that runs this script with `npm run stylelint`.

apps/site/assets/css/_alerts.scss
thecristen: Somehow, we had a mix of both!

digitalcora: Fantastic 😆 Ignore me, then

semaphore/stylelint.sh
thecristen: True. Depends on whether we want to only check changed files vs linting everything every time. Figured I'd make this consistent with the credo script but we could go either way.

apps/site/assets/package.json
thecristen: It is the default, but for some reason it would never take, at least when I'd try to run the script. 😖

semaphore/stylelint.sh
digitalcora: The Credo script is like that is because we have a _large_ number of extant Credo failures, and all of them are "missing moduledoc" so it's impractical to clean them up all at once. I expect someday we will delete that script and replace it with `mix credo`, since there's otherwise no downside to checking all the code every time (it takes seconds).

apps/site/assets/package.json
digitalcora: Strange. I suppose it can't hurt to have it here.

apps/site/assets/css/_mixins.scss
```diff
@@ -284,25 +283,23 @@
...
       position: relative;
       top: $offset;
       width: $width-height;
     }
   } @else {
-    [class*="c-svg__icon"] {
-      @include center-icon($width-height)
-    }
+    [class*='c-svg__icon'] {
+      @include center-icon($width-height); }
```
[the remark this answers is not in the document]
thecristen: Hmm yeah I don't like this. Looks like the rule controlling newlines before closing braces for @if/@else/@mixin etc blocks is inconsistent with the rule controlling these closing braces elsewhere -- I'll fix this to be consistent.

apps/site/assets/css/_checkbox.scss
thecristen: Given the comments, this looked like a case where we _only_ wanted this for FF, and the opacity setting serves a certain purpose. I was hesitant to modify this, mostly because of the testing effort involved to ensure it works as intended however many years/months ago when this was written.

semaphore/stylelint.sh
thecristen: Point taken: I've deleted this unneeded script!

PR #390, opened 2020-02-18
Fix blank pages with single-direction routes

digitalcora:
**Asana:** [Schedule Finder | Add fallbacks for unidirectional routes with broken schedules (171, 195)](https://app.asana.com/0/385363666817452/734177801773551/f)

Screen Shot 2020-02-14 at 2 52 50 PM

The app has not really supported single-direction routes until now, so this required a significant change to the data modeling: Route direction names can now be null, indicating there is no service in that direction.

On the backend, this must be determined using RoutePatterns, since there is no property of the Route itself that indicates which directions are valid, and RoutePatterns are the only resource that *does* indicate this and can be included in the same request as a Route. This required some changes to start including RoutePatterns in Route requests by default.

On the frontend, when we encounter a single-direction route, we always display data for the sole valid direction and hide UI that would allow the user to change it. This includes overriding the server-side logic that would normally pick a default direction based on the time of day.

Interestingly, the existing `Defaults` plug had some code that seemed intended to support single-direction routes, but it relied on an assign named `headsigns` that was not assigned or even referenced anywhere else in the current code.


apps/v3_api/lib/routes.ex
```diff
@@ -11,7 +11,7 @@ defmodule V3Api.Routes do
   end
 
   def get(id, opts \\ []) do
-    get_json("/routes/#{id}", [], opts)
+    get_json("/routes/#{id}", opts)
```
phildarnowsky: @digitalcora whither removing that one `[]`?

digitalcora: It's my belief this was either a mistake or an artifact of some removed code. Other functions in this module pass their params as the second argument to `get_json` — the third argument is "opts" and allows overriding things like the API key and request timeout. In master there is no code that calls `get` with a second argument, so it's a mystery why this was here. However, we now actually need to pass some request params, so I aligned this with the other functions.

digitalcora:
@phildarnowsky I added some new commits here to address test failures and etc. just as you were commenting, so you might want to take another look? The commit messages should be explanatory.

ingridpierre:
@digitalcora This looked great, sorry I forgot to hit send on my lgtm 👍  ! I tested in the afternoon to verify it properly avoided the logic to switch directions in the AM/PM.

PR #392, opened 2020-02-19
Cache realtime endpoint response

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Polling | Introduce caching for the end points getting polled](https://app.asana.com/0/555089885850811/1160944420414821)

(Also includes https://app.asana.com/0/555089885850811/1160944420414816)

Here we introduce a cache to the line diagram realtime endpoint with a 60 second TTL. ~We also slow down polling of this endpoint to once per 60 seconds.~

### DIFF
--- a/apps/site/assets/package.json
+++ b/apps/site/assets/package.json
@@ -94,7 +94,7 @@
     "stylelint-order": "^4.0.0",
     "stylelint-scss": "^3.14.2",
     "stylelint-selector-bem-pattern": "^2.1.0",
-    "stylelint-selector-no-mergeable": "git+ssh://git@github.com/nkingsy/stylelint-selector-no-mergeable.git",
+    "stylelint-selector-no-mergeable": "git+https://git@github.com/nkingsy/stylelint-selector-no-mergeable.git",
     "svg-inline-loader": "^0.8.0",
     "svgo": "^1.1.1",
     "svgo-loader": "^2.2.0",
--- a/apps/site/lib/site/application.ex
+++ b/apps/site/lib/site/application.ex
@@ -1,4 +1,6 @@
 defmodule Site.Application do
+  @moduledoc false
+
   use Application
 
   # See http://elixir-lang.org/docs/stable/elixir/Application.html
@@ -15,6 +17,7 @@ defmodule Site.Application do
 
     children = [
       # Start the endpoint when the application starts
+      supervisor(ConCache, [[], [name: :line_diagram_realtime_cache, ttl: :timer.seconds(60)]]),
       supervisor(SiteWeb.Endpoint, []),
       supervisor(Site.GreenLine.Supervisor, []),
       supervisor(Site.Stream.Vehicles, []),
--- a/apps/site/lib/site_web/controllers/schedule/line_api.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line_api.ex
@@ -59,12 +59,25 @@ defmodule SiteWeb.ScheduleController.LineApi do
   """
   @spec realtime(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def realtime(conn, %{"id" => route_id, "direction_id" => direction_id}) do
+    cache_key = {route_id, direction_id, conn.assigns.date}
+
+    payload =
+      ConCache.get_or_store(:line_diagram_realtime_cache, cache_key, fn ->
+        do_realtime(route_id, direction_id, conn.assigns.date, conn.assigns.date_time)
+      end)
+
+    conn
+    |> put_resp_content_type("application/json")
+    |> send_resp(200, payload)
+  end
+
+  defp do_realtime(route_id, direction_id, date, now) do
     headsigns_by_stop =
       TransitNearMe.time_data_for_route_by_stop(
         route_id,
         String.to_integer(direction_id),
-        date: conn.assigns.date,
-        now: conn.assigns.date_time
+        date: date,
+        now: now
       )
 
     vehicles_by_stop =
@@ -87,7 +100,7 @@ defmodule SiteWeb.ScheduleController.LineApi do
       end)
       |> Enum.into(%{})
 
-    json(conn, combined_data_by_stop)
+    Jason.encode!(combined_data_by_stop)
   end
 
   @spec update_route_stop_data({any, RouteStop.t()}, any, DateTime.t()) :: map()
--- a/apps/site/mix.exs
+++ b/apps/site/mix.exs
@@ -57,7 +57,8 @@ defmodule Site.Mixfile do
       :util,
       :trip_plan,
       :services,
-      :route_patterns
+      :route_patterns,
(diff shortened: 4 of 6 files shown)


digitalcora:
Commit message:

> With a 60-second cache introduced, there's no point in polling the realtime endpoint every 15 seconds anymore.

I'm not sure this is true. Consider this timeline:

```
+ [00:00] (cache is refreshed)
|
|
|
| [00:59] New user loads page; displayed data is now 59s old
+ [01:00] (cache TTL is reached)
|
|
| [01:58] Displayed data is now 118s old (!)
| [01:59] Second poll occurs; cache miss; displayed data is now fresh
+ [02:00]
```

In other words, though I assume the design goal here is "live data shouldn't be more than a minute behind", in this case the data _started out_ a minute behind and drifted to 2 minutes before being refreshed. Depending on the timing of cache invalidations from the activity of other users, you could get "stuck" in a state where your live data ranges from 1 to 2 minutes out-of-date, but is almost never better than that.

All this to say, I think we should keep the previous interval of 15s. Then even in the worst case, the data will never be more than 75s behind.

phildarnowsky:
@digitalcora right, good point, I'll revert that bit

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -87,7 +82,25 @@ defmodule SiteWeb.ScheduleController.LineApi do
...
+    Jason.encode!(combined_data_by_stop)
+  end
+
+  @doc """
+  Provides predictions and vehicle information for a given route and direction, organized by stop.
+  The line diagram polls this endpoint for its real-time data.
+  """
+  @spec realtime(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def realtime(conn, %{"id" => route_id, "direction_id" => direction_id}) do
+    realtime_key = [route_id, direction_id, conn.assigns.date]
```
digitalcora: `cache_key` I think would be a more intention-revealing name here; and I think we should make it a tuple (since those are generally preferred for fixed-length collections that are not intended to be iterated)

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -87,7 +82,25 @@ defmodule SiteWeb.ScheduleController.LineApi do
...
+
+    payload =
+      ConCache.get_or_store(:line_diagram_realtime_cache, realtime_key, fn ->
+        do_realtime(route_id, direction_id, conn.assigns.date, conn.assigns.date_time)
+      end)
+
+    conn
+    |> put_resp_content_type("application/json")
+    |> send_resp(200, payload)
```
digitalcora: Curious why this was changed from `json(conn, payload)`, as it seems to do the same thing

apps/site/mix.exs
```diff
@@ -127,7 +128,8 @@ defmodule Site.Mixfile do
       {:trip_plan, in_umbrella: true},
       {:services, in_umbrella: true},
       {:route_patterns, in_umbrella: true},
-      {:repo_cache, in_umbrella: true}
+      {:repo_cache, in_umbrella: true},
+      {:con_cache, "~> 0.12.0"}
```
digitalcora: Currently these seem to be organized as "non-umbrella deps" followed by "umbrella deps"; could we put this one up there with the others of its kind?

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -53,18 +53,13 @@ defmodule SiteWeb.ScheduleController.LineApi do
     json(conn, nil)
   end
 
-  @doc """
-  Provides predictions and vehicle information for a given route and direction, organized by stop.
-  The line diagram polls this endpoint for its real-time data.
-  """
-  @spec realtime(Plug.Conn.t(), map()) :: Plug.Conn.t()
-  def realtime(conn, %{"id" => route_id, "direction_id" => direction_id}) do
+  defp do_realtime(route_id, direction_id, date, now) do
```
digitalcora: I think it'd make the data flow easier to follow if we placed "implementation detail" functions below the higher-level functions that call them — do you think that's a good convention?

apps/site/lib/site_web/controllers/schedule/line_api.ex
phildarnowsky: `do_realtime` returns a string, as the work it does includes the JSON encoding of the payload. `json(conn, payload)` automatically encodes `payload`, so if we kept using `json`, we'd end up with a doubly-encoded payload.

apps/site/lib/site_web/controllers/schedule/line_api.ex
phildarnowsky: Yes, I like to adhere to that when possible, will move the function

apps/site/lib/site_web/controllers/schedule/line_api.ex
digitalcora: I see. Could `do_realtime` _not_ encode the data? Or is that required for the caching?

phildarnowsky: It's not required, but my reasoning is that JSON-encoding isn't free either, so if we're caching work let's cache that too.

digitalcora: Sure. My personal inclination is to "let it encode" given the likely-small time contribution of the encoding process (Jason was developed specifically for speed), but I don't think the discussion is worth holding things up.

apps/site/assets/package.json
```diff
@@ -94,7 +94,7 @@
     "stylelint-order": "^4.0.0",
     "stylelint-scss": "^3.14.2",
     "stylelint-selector-bem-pattern": "^2.1.0",
-    "stylelint-selector-no-mergeable": "git+ssh://git@github.com/nkingsy/stylelint-selector-no-mergeable.git",
+    "stylelint-selector-no-mergeable": "git+https://git@github.com/nkingsy/stylelint-selector-no-mergeable.git",
```
paulswartz: You can configure this on your local machine, so that you don't need to change the configuration: `git config --global url."https://github.com/".insteadOf git@github.com:`

apps/site/lib/site/application.ex
```diff
@@ -19,7 +21,8 @@ defmodule Site.Application do
       supervisor(Site.GreenLine.Supervisor, []),
       supervisor(Site.Stream.Vehicles, []),
       supervisor(Site.React, []),
-      supervisor(Site.RealtimeSchedule, [])
+      supervisor(Site.RealtimeSchedule, []),
+      supervisor(ConCache, [[], [name: :line_diagram_realtime_cache, ttl: :timer.seconds(60)]])
```
paulswartz: Since your controller depends on this cache being available, the cache should be started before `SiteWeb.Endpoint`.

PR #393, opened 2020-02-20
Mark as "current" if service in rating date range

thecristen:
Take two on tackling the issue noted in #391 with a hopefully better approach. 

Because we filter out `Weekday (no school)` services (as of #307 ), the frontend has no logical manner in which to determine what the day's valid current service is when it happens to be a Weekday (no school) day (as was/is the case every weekday this week after Monday). 

As happens to be the case, the regular weekday service, whose regular weekday schedules do apply this week, has a service start date next week, e.g. it is not in effect now and thus not considered a current service.

Handling this case deserves a more robust approach, but in the meantime we don't want to show customers "there is no service" when the regular weekday schedules are applicable. This change makes use of the fact that that future weekday service is part of this rating to mark it as "current" and have it show as desired in the ServiceSelector.

### DIFF
--- a/apps/site/assets/ts/helpers/service.ts
+++ b/apps/site/assets/ts/helpers/service.ts
@@ -42,7 +42,21 @@ const isInCurrentService = (
 ): boolean => {
   const serviceStartDate = stringToDateObject(service.start_date);
   const serviceEndDate = stringToDateObject(service.end_date);
-  return currentDate >= serviceStartDate && currentDate <= serviceEndDate;
+  const inServiceDates =
+    currentDate >= serviceStartDate && currentDate <= serviceEndDate;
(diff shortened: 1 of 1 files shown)


thecristen:
Technically correct that is what will show here -- @ingridpierre I was under the impression that the schedules for both would be the same though?

paulswartz:
They're not the same, at least not for all routes. For example, the 9 has extra service which doesn't run during school vacations.

thecristen:
That situation is something we currently account for on dotcom by annotating the relevant weekday schedule trips:

image

paulswartz:
Talked with @thecristen in person: this will be good for now to get through the rest of school vacation. Going forward, we should think about how we can better indicate to riders that a school vacation service is active on a particular date.

PR #394, opened 2020-02-20
Use SL pills instead of round SL icons in one spot

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Change "Silver Line Way-South Station" pill for SLW to regular SL pill](https://app.asana.com/0/555089885850811/1160132048446105)

In the "Daily Schedules" section of the schedule modal, we were using round SL icons, when we should have been using mini-pills in SL color. Fixed.

### DIFF
--- a/apps/site/assets/ts/schedule/components/schedule-finder/TableRow.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/TableRow.tsx
@@ -47,6 +47,25 @@ export const fetchData = (
   );
 };
 
+const RouteIcon = ({
+  route: { type, name, id }
+}: Journey): ReactElement<HTMLElement> | undefined => {
+  if (type !== 3) {
+    return modeIcon(id);
+  }
+
+  const backgroundClass = name.startsWith("SL")
+    ? "u-bg--silver-line"
+    : "u-bg--bus";
+  return (
+    <span
+      className={`c-icon__bus-pill--small schedule-table__scheduled-bus-pill ${backgroundClass}`}
+    >
+      {name}
+    </span>
+  );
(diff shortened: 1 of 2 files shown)


apps/site/assets/ts/schedule/components/schedule-finder/TableRow.tsx
```diff
@@ -55,28 +55,33 @@ const BusTableRow = ({
...
-}): ReactElement<HTMLElement> => (
-  <>
-    {anySchoolTrips && (
-      <td className="schedule-table__td--tiny">
-        {isSchoolTrip && <strong>S</strong>}
+}): ReactElement<HTMLElement> => {
+  const backgroundClass = journey.route.name.startsWith("SL")
+    ? "u-bg--silver-line"
+    : "u-bg--bus";
+  const busPillClasses = `c-icon__bus-pill--small schedule-table__scheduled-bus-pill ${backgroundClass}`;
```
digitalcora: This was a bit confusing for me to follow since these lines only become relevant within the `journey.route.type === 3` branch, and otherwise go unused. Perhaps a better way to extract this would be a separate component (it could just be defined in this file) that contains both this logic and the route-type conditional? e.g.

```jsx
<td className="schedule-table__td">
  <RouteIcon route={journey.route} />
  {breakTextAtSlash(journey.trip.headsign)}
</td>
```

Come to think of it, we already have a `RoutePill` component used elsewhere in `projects` — it doesn't look _identical_ to this logic, but I wonder if there's enough overlap to justify a refactor.

phildarnowsky:
@digitalcora ready for another look

PR #395, opened 2020-02-20
Yank "No scheduled service for today" notice

phildarnowsky:
The logic around when to display this notice and when not to is trickier
than we first anticipated. To avoid misleading riders, we've pulled that
notice for the moment until we can correct that logic.

#### Summary of changes
**Asana Ticket:** [Suppress "No scheduled service today" message](https://app.asana.com/0/555089885850811/1162903717378782)


digitalcora:
Should we also revert e45c8d0773f211c8520d62959c44984e940b281a, since that was itself intended as a quick hack to resolve this issue?

phildarnowsky:
@digitalcora yes let's do

PR #396, opened 2020-02-20
Don't say CR is "On time" with no scheduled time

digitalcora:
This also tweaks the rules for CR status displays: previously "Delayed X min" would override any manually-entered `status` string, but now the latter always takes precedence. This is in line with our published Best Practices for prediction data.

**Asana Ticket:** [CR says "On time" when we don't know the scheduled time](https://app.asana.com/0/385363666817452/1161689987228877/f)


digitalcora:
If it's alright with you, going ahead with this despite not meeting the Jest branch coverage threshold; see #397.

phildarnowsky:
@digitalcora sounds good, run with it. I did approve that other branch (lowering the threshold) as well.

PR #397, opened 2020-02-20
Adjust TS test coverage thresholds

digitalcora:
This fixes the immediate issue of every new branch and statement needing to have 100% coverage or it fails the build, because we were so close to the current thresholds. Separately, we can discuss how to align our code coverage approaches and decide on a long-term solution.

Relevant Asana ticket: [🔧 Decide on a common approach to test coverage](https://app.asana.com/0/385363666817452/1162707544957701/f)


phildarnowsky:
@digitalcora I'm going to merge this because the same problem is biting me on my own branches.

PR #399, opened 2020-02-22
Add cleanup function that turns off channel event listeners when component is unmounted

thecristen:
This is a tiny update mostly aiming to improve the ✨ developer experience ✨ 

This warning shows up on the new schedules page:

image

We don't want a memory leak! `useEffect()`'s first argument can be used to return a function that will fire when the React component is unmounted, so we use that here to turn off the event listeners that are activated by the `setupChannels` call in the `<Map>` component. At least, I think that's how channels here work. :)


digitalcora:
@phildarnowsky Ready for your review.

PR #402, opened 2020-02-24
Re-revert polling

phildarnowsky:
No ticket. Yanking polling again until we figure out the problem with the cache.


phildarnowsky:
@digitalcora I hear you. I was excited to turn this on too.

PR #405, opened 2020-02-25
Turn no-longer-realtime endpoint back on

phildarnowsky:
This reverts commit 5ff14ffd21bb251d5c8691d866c1655de97ad78f. It also
moves the endpoint to a different route. This is so that any browsers
out there that still have the polling code will not be hitting this
endpoint.


digitalcora:
Seems this needs an `npm run format`.

PR #406, opened 2020-02-26
Poll for live data only when page is visible

digitalcora:
**Asana Ticket:** [Line Diagram | Only poll when page is visible](https://app.asana.com/0/555089885850811/1160944420414819/f)

⚠️ Due to the need to restore the old polling code in order to implement this, it should not be merged until we also restore the old `realtime` endpoint with caching in place.

### DIFF
--- a/apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
+++ b/apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
@@ -1,6 +1,6 @@
 import React, { ReactElement, useState } from "react";
-import { useFetch } from "react-async";
 import { useInterval } from "use-interval";
+import { useFetch } from "react-async";
 
 import {
   LineDiagramStop,
@@ -105,9 +105,11 @@ const LineDiagram = ({
     { json: true, watch: directionId }
   );
   const liveData = (maybeLiveData || {}) as LiveDataByStop;
+
   useInterval(() => {
-    /* istanbul ignore next */
-    if (!liveDataIsLoading) reloadLiveData();
+    requestAnimationFrame(() => {
+      if (!document.hidden && !liveDataIsLoading) reloadLiveData();
+    });
   }, 15000);
 
   const handleStopClick = (stop: RouteStop): void => {
--- a/apps/site/assets/ts/jest.config.js
+++ b/apps/site/assets/ts/jest.config.js
@@ -1,5 +1,6 @@
 module.exports = {
   preset: "ts-jest",
+  clearMocks: true,
   collectCoverage: true,
   collectCoverageFrom: [
     "**/*.{ts,tsx}",
--- a/apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
+++ b/apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { mount, ReactWrapper } from "enzyme";
+import { mount, ReactWrapper, shallow, ShallowWrapper } from "enzyme";
 import { cloneDeep, merge } from "lodash";
 import LineDiagram from "../components/line-diagram/LineDiagram";
 import { EnhancedRoute, RouteType } from "../../__v3api";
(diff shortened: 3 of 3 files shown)


apps/site/assets/ts/schedule/components/line-diagram/LineDiagram.tsx
```diff
@@ -100,6 +106,12 @@ const LineDiagram = ({
   );
   const liveData = (maybeLiveData || {}) as LiveDataByStop;
 
+  useInterval(() => {
+    requestAnimationFrame(() => {
+      if (!liveDataIsLoading) reloadLiveData();
+    });
+  }, 15000);
```
paulswartz: Won't this keep running even if the page isn't visible? It won't make an API call, but it'll still spin on the client. Any chance of using the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API)?

digitalcora: > It won't make an API call, but it'll still spin on the client.

I'm not sure what you mean by "spin on the client". The interval function does still run once every 15 seconds regardless of page visibility, but it becomes nearly a no-op when the page is not visible. What is the downside vs. using the page visibility API?

paulswartz: It seems like the interval will continue to run in the background, scheduling the jobs, and then they'll all run when the browser tab becomes visible again? I'll check it out more tomorrow.

digitalcora: I've tested this manually in different browsers, and only a single request occurs when the page becomes visible again, regardless of how long it's been non-visible. Although many function calls could be queued up over time, since JS execution is single-threaded, only one of them arbitrarily gets to execute first when the page becomes visible. That one performs the reload, and `liveDataIsLoading` becomes true, causing all the remaining calls to return without doing anything.

paulswartz: Did you test it with Firefox? Running it locally, I'm seeing the `predictions_and_vehicles` API continue to be hit, even when the tab is not visible.

digitalcora: Firefox does still deliver animation frames in non-visible tabs (AFAIK it is the only modern browser engine that does so), but each call is delayed by an exponentially increasing amount of time. I'm not sure what the initial value is, but I did some additional testing and the delay is about 1 minute after 5 minutes, and about 10 minutes after 10 minutes. Extrapolating, after 15 minutes the delay would be nearly 2 hours. However, it seems to occasionally "reset" the delay even when the page hasn't been activated, so I've only seen it go up to a 10-minute delay.

This seems like a good argument for using `visibilityState` instead and avoiding this weirdness, although it might also be a good excuse for us to check out [SWR](https://swr.now.sh/), a library Cristen discovered after we had already implemented the `react-async` approach. It has polling intervals built in, as well as "only poll when visible" logic. The only thing that makes me nervous is that it's not stable yet (version 0.1).

As the assigned reviewer I will leave it to @phildarnowsky to decide whether it's worth revisiting our choice of library here, or just changing to use `visibilityState`.

paulswartz: FWIW, I was seeing them every 30 seconds the entire time I was testing, but maybe it's only my browser.

phildarnowsky: @digitalcora so if I understand your proposal about `visibilityState`, you would just add a check of same to the callback for `requestAnimationFrame`, and refrain from actually hitting the endpoint unless `visibilityState` were `visible`. If so, I would prefer that solution over adding a separate library, but let's revisit that at such time that we want to add polling elsewhere (possibly the stops page?). That will also give SWR a bit of time to mature, on consideration adding an unproven dependency to a feature that's already given us so much trouble is more risk than I want to take on.

digitalcora:
@phildarnowsky Yes, I think the way to go is actually to combine the two approaches. If we only did this:

```js
if (document.visibilityState !== "hidden" && !liveDataIsLoading) reloadLiveData();
```

Then after the page becomes visible, there could be up to a 15-second delay before the data is refreshed. The `requestAnimationFrame` allows us to have a call queued up that fires immediately.

digitalcora:
> I would prefer that solution over adding a separate library

...and to be clear, the other proposed solution would be to replace both `use-interval` and `react-async` with `swr`. So technically it would be removing a library, on net 😄

I've pushed up the solution I mentioned above for review.

PR #407, opened 2020-02-27
Switch from mailgun to ses

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Ops | Mailgun vs Amazon SES to send emails from dotcom](https://app.asana.com/0/555089885850811/1160349191370777)

Here we implement the switch from Mailgun to SES so we don't get charged for Mailgun.


digitalcora:
About the testing approach: rather than passing in the "perform" function, could we configure it in the application environment? It seems like we'd always want the mailer to be the "mock" version in the test env, and never in others. This would avoid the need to store the function in the `conn` and pass it down through several layers of calls.

mix.exs
```diff
@@ -13,7 +13,7 @@ defmodule DotCom.Mixfile do
       ],
       test_coverage: [tool: ExCoveralls],
       dialyzer: [
-        plt_add_apps: [:mix, :phoenix_live_reload, :laboratory],
+        plt_add_apps: [:mix, :phoenix_live_reload, :laboratory, :ex_aws, :ex_aws_ses],
```
digitalcora: I know you were having issues with Dialyzer at first — is this config change still required? If ExAWS is specified as a dependency in the usual way, it _should_ get picked up automatically.

phildarnowsky: That's what I thought, but this change, though I was cargo culting a bit, is what got Dialyzer to pass. Still, I will double check without it.

phildarnowsky: @digitalcora yup, took them out and dialyzer fails again.

digitalcora: I don't suppose I'll ever understand all the mysteries of Dialyzer ¯\\\_(ツ)_/¯

apps/feedback/test/mailer_test.exs
```diff
@@ -54,7 +54,10 @@ defmodule Feedback.MailerTest do
     end
 
     test "uses the comments of the message for the description" do
-      Mailer.send_heat_ticket(%{@base_message | comments: "major issue to report"}, nil)
+      Mailer.send_heat_ticket(
+        %{@base_message | comments: "major issue to report"},
+        nil
+      )
```
digitalcora: Looks like this change is now formatting-only? Could revert it to minimize diff noise

apps/feedback/test/mailer_test.exs
```diff
@@ -81,7 +84,11 @@ defmodule Feedback.MailerTest do
     end
 
     test "the email does not have leading or trailing spaces" do
-      Mailer.send_heat_ticket(%{@base_message | email: "   fake_email@gmail.com  "}, nil)
+      Mailer.send_heat_ticket(
+        %{@base_message | email: "   fake_email@gmail.com  "},
+        nil
+      )
```
digitalcora: As above

mix.exs
phildarnowsky: I don't think one _can_ ever understand all the mysteries of Dialyzer. Maybe Joe Armstrong did.

apps/feedback/mix.exs
```diff
@@ -21,7 +21,7 @@ defmodule Feedback.Mixfile do
   #
   # Type "mix help compile.app" for more information
   def application do
-    [applications: [:logger, :mailgun, :timex, :briefly]]
+    [applications: [:logger, :mailgun, :timex, :briefly, :ex_aws, :ex_aws_ses, :mail]]
```
digitalcora: It seems odd that we're manually specifying an `applications` key — normally this list is automatically inferred from the dependencies, and if you need a built-in Elixir application that's not started by default (like `logger`), you put it in `extra_applications`.

phildarnowsky: perhaps, but given the hard deadline let's look into this later, once we've confirmed successful transition to SES.

PR #408, opened 2020-02-27
 Add "View schedules" links to destination stops 

digitalcora:
**Asana Ticket:** [Schedules | "View schedules" on terminus stops](https://app.asana.com/0/385363666817452/1160182089454825/f)

This includes a non-essential refactoring that may be best to review separately.


apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
```diff
@@ -55,6 +55,9 @@ let lineDiagramBranchingIn = cloneDeep(lineDiagramBranchingOut).reverse();
 const CRroute = merge(cloneDeep(route), { type: 2 as RouteType });
 lineDiagramBranchingIn.forEach(({ route_stop }) => {
   route_stop.route = CRroute;
+  if (route_stop["is_terminus?"]) {
```
phildarnowsky: @digitalcora what's this all about now?

digitalcora: This code takes a fixture for an "outward" line diagram and attempts to turn it upside-down to create an "inward" line diagram, but in doing so it neglected to change the origins to destinations and vice-versa. Previously this didn't matter since the code was ignoring the `is_beginning?` field of RouteStops and doing its own thing, but now that we are using that field, we need to be more thorough with our fake data.

PR #410, opened 2020-03-02
Additional Schedule Finder refactors

digitalcora:
**Related to:** [Schedules | Add inputs to schedule finder modal](https://app.asana.com/0/385363666817452/1160795456837881/f)

These are a couple of other refactors that were left on Cristen's schedule-finder-inputs branch (see also #404). Since they are mostly independent of the feature work, and in the interest of keeping changes small and easily reviewed, I've separated them out.

### DIFF
--- a/apps/site/assets/css/_schedule-page.scss
+++ b/apps/site/assets/css/_schedule-page.scss
@@ -191,13 +191,6 @@
   margin: 0 $base-spacing;
 }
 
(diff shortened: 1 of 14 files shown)


digitalcora:
I added another small refactor onto this, based on some weirdness I discovered on the way to implementing the schedule finder inputs.

PR #411, opened 2020-03-02
Re-enable realtime line diagram data

phildarnowsky:
This re-enables both the polling of the realtime endpoint and the
caching of same, only this time with the correct options.

#### Summary of changes
**Asana Ticket:** [Restore caching and polling to realtime endpoint](https://app.asana.com/0/555089885850811/1164530949166961)

### DIFF
--- a/apps/site/lib/site/application.ex
+++ b/apps/site/lib/site/application.ex
@@ -17,6 +17,10 @@ defmodule Site.Application do
 
     children = [
       # Start the endpoint when the application starts
+      supervisor(ConCache, [
+        [ttl: :timer.seconds(60), ttl_check: :timer.seconds(5)],
+        [name: :line_diagram_realtime_cache]
+      ]),
       supervisor(SiteWeb.Endpoint, []),
       supervisor(Site.GreenLine.Supervisor, []),
       supervisor(Site.Stream.Vehicles, []),
--- a/apps/site/lib/site_web/controllers/schedule/line_api.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line_api.ex
@@ -50,14 +50,27 @@ defmodule SiteWeb.ScheduleController.LineApi do
   Provides predictions and vehicle information for a given route and direction, organized by stop.
   The line diagram polls this endpoint for its real-time data.
   """
-  @spec predictions_and_vehicles(Plug.Conn.t(), map()) :: Plug.Conn.t()
-  def predictions_and_vehicles(conn, %{"id" => route_id, "direction_id" => direction_id}) do
+  @spec realtime(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def realtime(conn, %{"id" => route_id, "direction_id" => direction_id}) do
+    cache_key = {route_id, direction_id, conn.assigns.date}
+
+    payload =
+      ConCache.get_or_store(:line_diagram_realtime_cache, cache_key, fn ->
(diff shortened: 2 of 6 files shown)


apps/site/lib/site/application.ex
```diff
@@ -17,6 +17,10 @@ defmodule Site.Application do
 
     children = [
       # Start the endpoint when the application starts
+      supervisor(ConCache, [
+        [ttl: :timer.seconds(60), ttl_check: :timer.seconds(15)],
```
digitalcora: In the ConCache README, I see only references to a `global_ttl` option for configuring the cache at the supervisor level, but no plain `ttl` option. Where is this documented?

Also, if the goal is "not more than a minute out-of-date", I think we should reduce the TTL and/or the check interval, since this means an item could be kept in the cache for up to 75 seconds. The TTL check is apparently an optimized process; given the docs recommend 1 second as a good default, I think we can feel very safe setting it to e.g. 5 seconds.

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -52,12 +52,25 @@ defmodule SiteWeb.ScheduleController.LineApi do
   """
   @spec predictions_and_vehicles(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def predictions_and_vehicles(conn, %{"id" => route_id, "direction_id" => direction_id}) do
```
digitalcora: Renaming this back to `realtime` would be nice, I think.

apps/site/lib/site/application.ex
phildarnowsky: https://hexdocs.pm/con_cache/0.12.1/ConCache.html. Remember that we're using an old version of ConCache here, the name got changed at some point since.

"So why not update ConCache?" Every single `Repo` we have is built on it, and given that this feature has a history of exploding in our faces, I'm not inclined to undertake the upgrade now.

phildarnowsky: Will reduce the check interval however.

apps/site/lib/site_web/controllers/schedule/line_api.ex
phildarnowsky: @digitalcora one fine point of that, though, is we should keep the old route hooked up, so that during the transition when the old and new code are both running, it doesn't 404

digitalcora: Sure; can we just redirect it, at the routing level?

phildarnowsky:
@digitalcora this should address your feedback

apps/site/lib/site_web/router.ex
```diff
@@ -137,7 +137,13 @@ defmodule SiteWeb.Router do
...
       "/schedules/line_api/predictions_and_vehicles",
       ScheduleController.LineApi,
-      :predictions_and_vehicles
+      :realtime
+    )
+
+    get(
+      "/schedules/line_api/realtime",
+      ScheduleController.LineApi,
+      :realtime
```
digitalcora: This fits on one line, I think (and maybe the above one does too, now that it's shorter)

phildarnowsky: interesting...I'd have expected my Elixir formatting plugin to catch that

digitalcora: `mix format` generally preserves your chosen line breaks even if it's possible to fit the statement on one line.

PR #413, opened 2020-03-03
Use SWR in place of react-async and use-interval

digitalcora:
It combines the functions of both, does page visibility and debouncing in a much more robust way, and works in IE11 where react-async didn't.


apps/site/assets/package.json
```diff
@@ -23,11 +23,11 @@
...
     "react": "^16.8.1",
-    "react-async": "^10.0.0",
     "react-dom": "^16.8.1",
     "react-leaflet": "^2.2.1",
     "react-leaflet-fullscreen": "^1.0.1",
     "sifter": "^0.5.1",
+    "swr": "^0.1.18",
     "tether": "^1.3.4",
     "turbolinks": "git+https://github.com/paulswartz/turbolinks.git#82b188e",
     "use-interval": "^1.2.1",
```
digitalcora: `use-interval` is not removed here because we have other things that use it (although eventually those other things should be converted to use SWR too).

apps/site/assets/ts/schedule/__tests__/LineDiagramTest.tsx
```diff
@@ -355,72 +344,3 @@ describe("LineDiagram for CR with branches going inward", () => {
...
-    useFetch.mockImplementationOnce(() => ({
-      data: {},
-      isLoading: true,
-      reload: liveDataReload
-    }));
-    mountComponent();
-
-    expect(liveDataReload).not.toHaveBeenCalled();
-  });
-});
```
digitalcora: These tests are simply deleted because they were testing the "glue code" we wrote to pull together `useInterval`, `useFetch`, and some other things. With SWR none of that glue is needed anymore, and the snapshot test combined with the mock data verifies that we are still displaying live data as we should (when it's present).

digitalcora:
@phildarnowsky I've melded this together with a revert of the revert on `master`. Will merge when/if tests pass.

PR #414, opened 2020-03-04
Correct finder api 500s

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Tech Debt | In finder_api, handle failures to look up trip info more gracefully](https://app.asana.com/0/555089885850811/1157508136093385)

This should eliminate one large source of noise in our Sentry logs.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/finder_api.ex
+++ b/apps/site/lib/site_web/controllers/schedule/finder_api.ex
@@ -16,6 +16,8 @@ defmodule SiteWeb.ScheduleController.FinderApi do
 
   import SiteWeb.ScheduleController.ScheduleApi, only: [format_time: 1, fares_for_service: 4]
 
+  require Logger
+
   @type react_keys :: :date | :direction | :is_current
   @type react_strings :: [{react_keys, String.t()}]
   @type converted_values :: {Date.t(), integer, boolean}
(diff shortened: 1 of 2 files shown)


digitalcora:
For reference, in what real-world situation would we not find TripInfo when we were expecting to? That might be a good thing to include in the commit message, as well.

phildarnowsky:
@digitalcora I've actually forgotten what leads to this situation, but it's apparently when the frontend code calls the endpoint with a nil trip ID. Why that comes up, I haven't been able to retrace yet.

digitalcora:
@phildarnowsky I see. Since the issue is with a param not being provided, can we add a guard to the action clause instead? Then the 404 will be automatic. Like `when not is_nil(id)`

phildarnowsky:
@digitalcora sorry, I meant when the `id` is `""` not `nil`--otherwise this wouldn't pattern match. 

One thing about going the guard clause route, is it would send back an entire HTML page, as if it were intended for human consumption rather than an XHR. Because of that I'm inclined to leave this as-is.

phildarnowsky:
@digitalcora ...wait, I think I'm wrong about the `id` being `""` being the cause. Am continuing to investigate.

phildarnowsky:
@digitalcora let's do this. I think to debug this properly I'm going to have to see the original query params, before they get overwritten. That code needs refactoring so as to avoid doing that, but I don't want to untangle the whole plate of spaghetti today. As a workaround let's at least save off the original query params and log them in a Splunk warning if we come up empty on `trip_info`.

phildarnowsky:
@digitalcora this is ready for another look

PR #415, opened 2020-03-05
Add concurrency to realtime cache

phildarnowsky:
These options should prevent the locking related errors for the realtime API under heavy load. They were present originally, but in removing and restoring the `ttl` options repeatedly I accidentally removed these.

### DIFF
--- a/apps/site/lib/site/application.ex
+++ b/apps/site/lib/site/application.ex
@@ -18,7 +18,11 @@ defmodule Site.Application do
     children = [
       # Start the endpoint when the application starts
       supervisor(ConCache, [
-        [ttl: :timer.seconds(60), ttl_check: :timer.seconds(5)],
+        [
+          ttl: :timer.seconds(60),
+          ttl_check: :timer.seconds(5),
+          ets_options: [read_concurrency: true]
+        ],
         [name: :line_diagram_realtime_cache]
       ]),
       supervisor(SiteWeb.Endpoint, []),


digitalcora:
If I'm reading this ConCache documentation correctly...

```elixir
# Returns existing value, or calls function and stores the result.
# If many processes simultaneously invoke this function for the same key, the function will be
# executed only once, with all others reading the value from cache.
ConCache.get_or_store(:my_cache, key, fn() ->
  initial_value
end)
```

...then we don't need to add `write_concurrency`, since multiple processes will never attempt to write at the same time anyway (subsequent calls to `get_or_store` will block until the write is done, then perform a read instead). The current RepoCache configuration backs this up, as it does not have write concurrency enabled. I'd be interested to see a load test comparison with and without.

phildarnowsky:
@digitalcora `RepoCache` does in fact use `write_concurrency`, see `RepoCache.include_defaults`.

With both `read_concurrency` and `write_concurrency` engaged, there is no improvement in performance. I'm going to try it with just `read_concurrency` and if that doesn't help, I'm closing this PR umerged.

digitalcora:
Another thing I just noticed: the README says these have to be specified under a key of `ets_options`, but we're not using that – and strangely, neither is the RepoCache as far as I can tell. Is it possible RepoCache has also been doing this wrong, or is there another undocumented way to provide the options?

phildarnowsky:
@digitalcora regarding your README question: neither, it's another matter of us using an old version of ConCache. v0.12.0, which we're on, accepts two keyword lists of options, and the second is the ETS options.

I've just load-tested with only `read_concurrency` and, while not perfect, it showed a threefold improvement in the error rate, so keeping in mind that my load tests are pretty heavy-duty tortures of the site, I think this version is as good as it's going to get, and hopefully will suffice.

paulswartz:
It looks like the issue is not with the ETS table directly, but with getting the lock from the lock server.
```
exit: ** (exit) exited in: GenServer.call(#PID<0.2317.0>, {:lock, {:get, {"117", [include: "route_patterns"]}}, #Reference<0.3334031122.2508980225.178770>}, 5000)
    ** (EXIT) time out
  File "lib/gen_server.ex", line 989, in GenServer.call/3
  File "lib/con_cache/lock.ex", line 30, in ConCache.Lock.exec/4
  File "lib/repo.ex", line 46, in Routes.Repo.get/1
  File "lib/site_web/controllers/schedule/finder_api.ex", line 100, in SiteWeb.ScheduleController.FinderApi.trip/2
  File "lib/site_web/controllers/schedule/finder_api.ex", line 1, in SiteWeb.ScheduleController.FinderApi.action/2
...
(3 additional frame(s) were not displayed)
```

My guess would be a [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem):
- cache expires
- many requests come in for the cached data
- all of them try to get the lock to update the data, as the cache is expired
- a few of them succeed, but the ones at the end time out

If we're going to send this much traffic through one path, we might need to let the expired cache continue to be served while the data updates behind the scenes (like @digitalcora 's change for the frontend).

digitalcora:
@phildarnowsky I was referring to the README for our version, though I now see that the child-spec syntax is seemingly different from the `start_link` syntax (Cmd-F for `ets_options` and you'll see the part I was looking at). Strange discrepancy...

digitalcora:
@paulswartz That stack trace you posted doesn't seem to have anything to do with the `realtime` API which this fix is targeted at; it occurred in the `FinderApi` which is part of the Schedule Finder modal.

phildarnowsky:
@paulswartz yes, I'd about come to the same conclusion. I think, based on the results of the latest load test, that although it doesn't eliminate the problem entirely, `read_concurrency` improves performance enough that this lock problem will be rare at production volumes.

The SWR idea is interesting, and seems like it would be relatively straightforward to implement--perhaps could make a worthwhile open-source contribution to ConCache too. We'll keep that in our back pocket.

phildarnowsky:
@digitalcora `ConCache.start_link/2` (https://hexdocs.pm/con_cache/0.12.0/ConCache.html#start_link/2) has this signature:

```
start_link(options, [{:name, GenServer.name}]) :: Supervisor.on_start
```

Notice how `name` is apparently a valid and required option in the second `sup_options` argument. We also see below that `write_concurrency` and `read_concurrency` are grouped with `name` under the heading "In addition, following ETS options are supported".

So what I take from that is that that sentence should really read "In addition, following ETS options are supported, *which are specified by passing them as a keyword list as the second argument of `start_link`*."

digitalcora:
> Notice how `name` is apparently a valid and required option in the second `sup_options` argument. We also see below that `write_concurrency` and `read_concurrency` are grouped with `name` under the heading "In addition, following ETS options are supported".

I think these are two different options. One is the name of the cache itself, the other is the name of the ETS table it should use when `named_table` is specified.

[Looking at the ConCache tests for our version](https://github.com/sasa1977/con_cache/blob/0.12.1/test/con_cache_test.exs), it appears that list of options is definitely specified under an `ets_options` key — the tests create different kinds of ETS tables using this, and verify that the behavior of the cache is different when using them.

The only conclusion I can draw is that the documentation for 0.12 simply neglects to mention the `ets_options` key, giving the (false) impression that ETS options are specified at the top level. The tests tell the real story:

```elixir
# Line 202
ConCache.start_link(ets_options: [:named_table, name: :test_name])

# Line 226
ConCache.start_link(ttl_check: 10, ttl: 50, time_size: unquote(time_size))
```

Based on this, I'd suggest testing out the following supervisor definition:

```elixir
supervisor(ConCache, [
  [
    ttl: :timer.seconds(60),
    ttl_check: :timer.seconds(5),
    ets_options: [read_concurrency: true]
  ],
  [name: :line_diagram_realtime_cache]
])
```

digitalcora:
I've tested this locally by adding `:named_table` and `name: foobar` in various places and checking the actual resulting table name in Observer (since it doesn't seem to show concurrency options), and the above definition seems to be the only correct way to do it.

In other words, this works:

```elixir
supervisor(ConCache, [
  [ets_options: [:named_table, name: :foobar]],
  [name: :line_diagram_realtime_cache]
])
```

This doesn't work:

```elixir
supervisor(ConCache, [
  [:named_table, name: :foobar],
  [name: :line_diagram_realtime_cache]
])
```

This also doesn't work (it names the _cache_ foobar, not the table):

```elixir
supervisor(ConCache, [
  [],
  [:named_table, name: :foobar]
])
```

You can also verify this by intentionally misspelling the option name:

```elixir
supervisor(ConCache, [
  [
    ttl: :timer.seconds(60),
    ttl_check: :timer.seconds(5),
    ets_options: [xead_concurrency: true]
  ],
  [name: :line_diagram_realtime_cache]
])
```

In this case the supervisor will crash on startup with `{:invalid_ets_option, {:xead_concurrency, true}}`. Putting the misspelled option anywhere else won't result in a crash.

phildarnowsky:
@digitalcora I should know better than to trust the ConCache docs by now. Will try that.

phildarnowsky:
@digitalcora `read_concurrency` by itself doesn't make any appreciable difference, I'm trying engaging `write_concurrency` as well.

phildarnowsky:
@digitalcora as discussed in Slack, revised to have `read_concurrency` only

PR #416, opened 2020-03-05
Add version checking to realtime endpoint

phildarnowsky:
To prevent the odd remaining herd from thundering through our website.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line_api.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line_api.ex
@@ -51,17 +51,24 @@ defmodule SiteWeb.ScheduleController.LineApi do
   The line diagram polls this endpoint for its real-time data.
   """
   @spec realtime(Plug.Conn.t(), map()) :: Plug.Conn.t()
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -51,17 +51,21 @@ defmodule SiteWeb.ScheduleController.LineApi do
   The line diagram polls this endpoint for its real-time data.
   """
   @spec realtime(Plug.Conn.t(), map()) :: Plug.Conn.t()
-  def realtime(conn, %{"id" => route_id, "direction_id" => direction_id}) do
-    cache_key = {route_id, direction_id, conn.assigns.date}
+  def realtime(conn, %{"id" => route_id, "direction_id" => direction_id} = query_params) do
+    if query_params["v"] == "2" do
```
digitalcora: We should leave a comment here as to why we're doing this, so future generations (or just us, later) will feel confident in removing and/or augmenting it.

PR #418, opened 2020-03-09
Add inputs to schedule finder modal

digitalcora:
**Asana Ticket:** [Schedules | Add inputs to schedule finder modal](https://app.asana.com/0/385363666817452/1160795456837881/f)

Because the Schedule Finder modal now needs an internal state for which direction and origin are selected, independent of their values when it was launched, this required many changes to how state flows through the involved components.

New components are introduced:

* `ScheduleFinderForm` is the blue box with the "Schedule Finder" header that contains the direction and origin inputs; this now appears in both the original ScheduleFinder component and the schedule modal. To account for the differences in how direction and origin changes are handled in each of these contexts, the form has no internal state and the values of the inputs must be puppeteered by the parent component, which is expected to provide callbacks for various events.

* `ScheduleFinderModal` is a wrapper around both the "origin modal" and "schedule modal", which can switch between them as needed and maintain a common state across the transition. This now appears in both the ScheduleFinder component and the LineDiagram, so a modal launched from either location can behave in the same way.

Existing components are changed:

* `ScheduleFinder` is now a simple wrapper for the two new components. Note that if the modal is launched from this "widget", changing the inputs within the modal will also change the corresponding inputs in the widget. Making this also work for modals launched from the line diagram would be difficult, as the widget and line diagram are in two separate React roots.

* The `openState` prop of the `Modal` component is removed in favor of parent components conditionally including the modal in their output. This enables the behavior where Schedule Finder can be launched with specific initial values derived from the state of its parent, but the values can be changed freely while open, independent of the parent.

This also includes a couple of separate changes: a bump to Enzyme and a new patch version of React to fix some testing issues, and a change to the details of our synthetic `Green` route. These are explained in their own commit messages.


ingridpierre:
Looks as though this pr has introduced a smidge of horizontal scrolling to this modal, can jiggle the elements around.

digitalcora:
@ingridpierre Good catch, I assume this is in Chrome?

ingridpierre:
@digitalcora chrome, ff, and safari

digitalcora:
@ingridpierre Should be fixed in the latest commit.

ingridpierre:
On my older Android phone in Chrome:
-  if I open the schedule modal using the finder
- then reselect a stop using the the modal origin input (choose a stop very low on the list)
- when it returns me to the schedule finder modal it seems to remember my scroll position on the previous modal, and pushes my view lower down than I expected

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleFinderModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+            ? "Choose Origin Stop"
+            : `Schedules to ${route.direction_names[direction]}`
+      }}
+      className={mode === "origin" ? "schedule-finder__origin-modal" : ""}
+      closeModal={closeModal}
+    >
+      {() => (
+        <>
+          {mode === "origin" && (
+            <OriginModalContent
```
phildarnowsky: @digitalcora I'm thinking here and below, let's extract the TSX markup into functions, so that the markup doesn't drown out the control-flow logic. I mean something like:

```
const originModalContent = () => 
  <OriginModalContent 
    handleChangeOrigin={handleChangeOrigin}
    ...
/>

const scheduleModalContent = () =>
  <ScheduleModalContent
    handleChangeDirection={handleChangeDirection}
    ...
  />

{mode === "origin" && originModalContent()}
{mode === "schedule" && origin !== null && scheduleModalContent()}
```

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleFinderModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+      ariaLabel={{
+        label:
+          mode === "origin"
+            ? "Choose Origin Stop"
+            : `Schedules to ${route.direction_names[direction]}`
+      }}
+      className={mode === "origin" ? "schedule-finder__origin-modal" : ""}
+      closeModal={closeModal}
+    >
+      {() => (
```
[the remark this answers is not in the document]
digitalcora: The existing `Modal` logic expects its `children` to be a function that returns a JSX element, rather than a JSX element. I'm not sure exactly why this is, but I'd prefer not to mess with it unless we know it's causing issues.

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleFinderModal.tsx
digitalcora: As-written, this wouldn't compile because the `ScheduleModalContent` component needs to have an origin defined, but in the context of the larger `ScheduleFinderModal` the origin is allowed to be null. But beyond that, I think this could have some unwanted side effects because we'd be "rendering" both of the components no matter what the modal mode is, instead of just the one that corresponds to the current mode (and having the other one not exist at all).

phildarnowsky: @digitalcora right, that's why they're anonymous functions which return React elements, rather than React elements themselves--we only call the one, if any, that's appropriate.

digitalcora: I see, but that doesn't resolve the issue of required values possibly not existing.

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleFinderModal.tsx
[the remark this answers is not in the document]
digitalcora: ...I may try just removing this aspect of it and seeing if everything still works, though, because this particular modal is the only user of `Modal` and doesn't seem to use the argument that is passed to the `children` function. I wonder if this is a leftover of a different approach to modals.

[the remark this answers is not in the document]
digitalcora: Seems like this works fine. I've made this change now.

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleFinderModal.tsx
phildarnowsky: true, in this case I think it would be OK to use the `!` modifier for the possibly-null values within the function bodies to force the issue, as we're assuring that at the callsite.

digitalcora:
@ingridpierre The scroll issue should be fixed on this branch; I tested it in the Chrome device simulator, but let me know if you'd like another deploy to green.

ingridpierre:
@digitalcora fabulous! If you don't mind can you put it on green again so I can use it on my mobile(s)?

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleFinderModal.tsx
digitalcora: I've resolved this per our in-person conversation, hopefully it's more readable now!

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleFinderForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+      {originError && (
+        <div className="error-container">
+          <span role="alert">Please provide an origin</span>
+        </div>
+      )}
+
+      <div className="schedule-finder__inputs">
+        <label className="schedule-finder__label">
+          Choose a direction
+          <SelectContainer>
```
phildarnowsky: @digitalcora so this is a new usage to me, I assume that the elements inside the `SelectContainer` tag get bound to the `children` of the `SelectContainer` component itself? is that automatic or am I overlooking an explicit binding somewhere?

ingridpierre:
@phildarnowsky i'll take a look at IE and Edge right now.

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleFinderForm.tsx
digitalcora: The former. Anything you put inside the "tag" of a React component is passed to it as a `children` prop.

digitalcora:
@ingridpierre I've checked this out in IE11 on CrossBrowserTesting and it all works and looks fine. I did find a visual bug within the Schedule Finder that seems unrelated to this change, as it also happens on `master`, so I wrote that up as a separate ticket.

digitalcora:
@phildarnowsky Just awaiting your approval or further comments, then, I think!

PR #421, opened 2020-03-10
Use arrival time instead of departure time for predictions

beeuhtricks:
#### Summary of changes
**Asana Ticket:** [Prefer arrival time over departure time for predictions](https://app.asana.com/0/385363666817452/935149940976868/f)

When parsing the attributes from a JSON:API response for predictions, the `arrival_time` attribute now comes first in the record returned by the parsing function. Also fixed a failing parser test that expected `departure_time` to come first.


digitalcora:
Looking at the results on Semaphore:

* There is a known flaky test failure in `Vehicles.StreamTest`. It's safe to ignore this, and it looks like all the other Elixir tests have passed.
* There is a Credo failure because we're requiring new/changed Elixir modules to have docs. If you're comfortable doing so, adding a `@moduledoc` with a concise summary of the module's purpose (just a sentence is fine) will satisfy the check. Or, you can use `@moduledoc false` to silence it.
* There are some Backstop failures, which I think we'd expect since this changes how times are displayed across the site. Technically it's possible to review and approve these without having Backstop set up locally, but this might be a good opportunity to do that anyway. The [testing docs](https://github.com/mbta/dotcom/blob/master/docs/TESTING.md#backstop-tests) on how to do this should be current. Once that's done, `npm run backstop` will generate the new images and open a comparison in your browser, then after looking at them, `npm run backstop:approve` will approve the changes.

beeuhtricks:
Thanks @digitalcora, I'll get right on those.

paulswartz:
General comment: this works in the case where you're looking at a whole trip, but doesn't work if you're looking at a subset of a trip.

Example trip:

| Stop | Arrival | Departure |
| ---| ---- | --- |
| A | null | 1:05 |
| B | 1:10 | 1:15 |
| C | 1:20 | 1:25 |
| D | 1:30 | null |

If you look at the full trip, this works: each stop has the departure time, except for D which has the arrival time. However, if you're only going from B to C, you get the wrong time for C: it should show you 1:20 but instead shows 1:25. An example page where we only show partial trips is the schedle page https://www.mbta.com/schedules/Red/schedule?origin=place-andrw&destination=place-knncl where the Trip Information is a partial trip.

paulswartz:
I recall most of my comment :) We likely do still need to know which type of time was used, in order to distinguish the "Arriving" vs "Boarding" case, but that might be outside the scope of this particular PR.

digitalcora:
> I recall most of my comment :) We likely do still need to know which type of time was used, in order to distinguish the "Arriving" vs "Boarding" case, but that might be outside the scope of this particular PR.

It is indeed likely out of scope. In the current Dotcom data-modeling of predictions, there is only a single `time` field that is populated from either the arrival or departure time when the data is initially ingested — and we don't retain the data about _which_ time we chose. Changing this model would be a significant undertaking.

phildarnowsky:
@paulswartz could you expand on the arriving vs. boarding matter? Also, it seems like the logic in your recalled comment is correct, what's the problem with it?

paulswartz:
@phildarnowsky re: arriving vs boarding, here's what it says in the best practices and the body of the Asana ticket:

> **If both departure_time and arrival_time are present:**
> - When the predicted time drops below 45 seconds in the future, you can optionally display "Arriving."
>
> **Else (departure_time is present, but arrival_time is not)**
> - In this case, you should not use the word "Arriving" and should instead display "Boarding" when the predicted time drops below 30 seconds in the future.

It might be out of scope for this particular PR, but it doesn't seem out of scope for the ticket. It would be a change to keep track of which time you're using, or some other way of distinguishing the two cases.

digitalcora:
I'll defer to Katrina on the scoping question. We had estimated this as a 1-point task, and if the scope is going to expand beyond just "we should prefer arrival time over departure time", I think we're going to have to re-estimate and re-prioritize it, or simply create a new task and mark this one as completed. In any case I think we should move forward with this particular change, since the new behavior is significantly _more_ correct than the current behavior.

PR #422, opened 2020-03-11
Remove width restriction

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Bug | Schedules | Hours of Operation missing spacing](https://app.asana.com/0/385363666817452/1165381204331765)
I removed the restriction of width for class `hours-directions` because that was causing the spacing to be nonexistent.
I tested this with different widths but let me know if you still find issues.

### DIFF
--- a/apps/site/assets/css/_route.scss
+++ b/apps/site/assets/css/_route.scss
@@ -10,10 +10,6 @@
 
 .hours-directions {
   margin-bottom: 0;
-
-  @include media-breakpoint-up(md) {
-    width: 75%;
-  }
 }
 
 .hours-direction-name {
--- a/apps/site/assets/css/_route.scss
+++ b/apps/site/assets/css/_route.scss
@@ -10,17 +10,7 @@
 
 .hours-directions {
   margin-bottom: 0;
-}
-
-.hours-direction-name {
-  display: inline-block;
-  width: 40%;
-}
-
-.hours-time {
-  display: inline-block;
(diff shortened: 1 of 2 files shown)


phildarnowsky:
@alicia-mbta FYI it looks like CI failed on this branch. However, I had a look, and one problem is a known intermittent bug in the tests, the other seems to be a glitch with the CI. system. Just to be certain I've kicked off another CI run here though.

phildarnowsky:
@alicia-mbta This looks good to me, and the second attempt at CI passed. However I'd also like @ingridpierre to weigh in in case there are unforeseen consequences of removing this rule that she might know of, so I'm marking this "needs design review." So I'm gonna approve it but please don't merge yet.

alicia-mbta:
Oh thanks @phildarnowsky, I wasn't sure.

ingridpierre:
@alicia-mbta It looks improved on most breakpoints, but we might need some additional support for the MD & SM breakpoints to keep the info closer together. At the worst it looks like this (image below).  It's important to keep gaps at a minimum based for users with Dyslexia or zoom magnification who have trouble reading across gaps like this.

image

alicia-mbta:
I just added more changes. This is what it looks like now:
Screen Shot 2020-03-11 at 16 23 16

ingridpierre:
lgtm! 👍 thanks for the extra polish on this one @alicia-mbta

alicia-mbta:
Thanks!

PR #423, opened 2020-03-11
Add active row highlight for CR timetables and lighten header

beeuhtricks:
----

#### Summary of changes
**Asana Ticket:** [🆕 CR Timetable | Highlight Timetable Row on Hover](https://app.asana.com/0/385363666817452/937511019039317/f)

Commuter Rail timetable rows now highlight in blue when they are moused over. The header color was changed as well, as per conversation with @ingridpierre.


ingridpierre:
@beeuhtricks can we disable the highlight on just these first few cells in the corner of the header?
image
image
etc.

beeuhtricks:
This should fix the header highlight problem, and Backstop still seems ok.

ingridpierre:
👍beauteous! lgtm!

PR #424, opened 2020-03-12
Add PR checklist to PR template

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Ops | Write a PR checklist](https://app.asana.com/0/555089885850811/1162467761333249)

Or, "Some Things We Learned The Hard Way". These are some things that have proven to be gotchas in the past. In an effort to prevent making the same mistakes again, let's systematically check these things whenever we open a PR.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,3 +2,14 @@
 **Asana Ticket:** [TICKET_NAME](TICKET_LINK)
 
 [Please include a brief description of what was changed]
+
+---
+
+Before getting review, please check the following:
+
+* [ ] Does frontend functionality render and work correctly in IE?
+* [ ] Have we load-tested any new pages or internal API endpoints that will receive significant traffic?
+* [ ] Are interactive elements accessible to screen readers?
+* [ ] Have you checked for tech debt you can address in the area you're working in?
+* [ ] If this change involves routes, does it work correctly with pertinent "unusual" routes such as the combined Green Line, Silver Line, Foxboro commuter rail, and single-direction bus routes like the 170?
+* [ ] Are the changes organized into self-contained commits with descriptive and well-formatted commit messages?


phildarnowsky:
@beeuhtricks @alicia-mbta though of course you won't know the specific things we've tripped over in the past here, I'd be interested in hearing any things you can think of that have given you trouble in the past on other sites and which might apply to dotcom as well.

digitalcora:
FYI for PRs that don't change any code, you can add the text `[ci skip]` to the body of your commit message, and Semaphore will recognize that and not bother running anything.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,3 +2,19 @@
 **Asana Ticket:** [TICKET_NAME](TICKET_LINK)
 
 [Please include a brief description of what was changed]
+
+---
+
+Before getting review, please check the following:
+
+* [ ] Does frontend functionality render and work correctly in IE?
+* [ ] Has every endpoint that will be polled from the frontend been load-tested?
```
digitalcora: More generally, I would say "Have we load-tested any new pages or internal API endpoints that will receive significant traffic?" We don't introduce new pages or endpoints that often, and I think it'd be nice to consider this question whenever we do. The site does serve a large number of visitors with a small number of instances, so it needs to scale well in general.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,3 +2,19 @@
...
+* [ ] Do form elements have ARIA labels?
```
digitalcora: I'd suggest expanding this to "Are interactive elements accessible to screen readers?" In most cases, making elements accessible will simply involve using appropriate semantic markup (for example `<label>` tags wrapping a form control), rather than adding ARIA attributes.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,3 +2,19 @@
...
+* [ ] Have you checked for tech debt you can address in the area you're working in?
+* [ ] Have you considered the following cases that often require special handling:
```
digitalcora: All these items seem to fall under the category of "special _routes_" specifically, so maybe we could condense this to:

```
* [ ] If this change involves routes, does it work correctly with "unusual"
      routes such as the combined Green Line, Silver Line, Foxboro commuter
      rail, and single-direction bus routes like the 170?
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,3 +2,19 @@
...
+  * [ ] Green Line
+  * [ ] Foxboro CR
+  * [ ] Silver Line
+  * [ ] Unidirectional bus routes
+  * [ ] Ferry
+  * [ ] Red Line on the branches
```
digitalcora: What kind of special handling does the Red Line require? I didn't think it was too different from most other routes; at least, I don't remember personally having to add special handling for it.

phildarnowsky: Hmmmm, not sure what I had in mind here, I'll take the RL out unless I can think of a specific case that broke on it.

.github/PULL_REQUEST_TEMPLATE.md
phildarnowsky: @digitalcora sounds good, I will also add a note that this list is not meant to be exhaustive (and I imagine we'll remember/discover a couple of other weird routes from time to time).

digitalcora:
One addition I think would be good:

* Are the changes organized into self-contained commits with descriptive and well-formatted commit messages? (see e.g. [this guide](https://github.com/danjdrennan/git-style-guide#commits))

This is more of a "check before merging" than a "check before approving" thing (since neatening up the change history usually comes _after_ approval), but I think it will be nice to have the reminder there.

phildarnowsky:
@digitalcora latest version addresses all your feedback.

PR #425, opened 2020-03-12
Add reCAPTCHA protection to feedback form

digitalcora:
**Asana Ticket:** [Address spam on feedback form](https://app.asana.com/0/385363666817452/1165373168335001/f)

### DIFF
--- a/apps/site/lib/site_web/controllers/customer_support_controller.ex
+++ b/apps/site/lib/site_web/controllers/customer_support_controller.ex
@@ -27,15 +27,17 @@ defmodule SiteWeb.CustomerSupportController do
     )
   end
 
-  def submit(conn, %{"support" => data}) do
-    case do_validation(data) do
+  def submit(conn, %{"support" => form_data, "g-recaptcha-response" => recaptcha_response}) do
+    params = Map.put(form_data, "recaptcha_response", recaptcha_response)
+
+    case do_validation(params) do
       [] ->
-        do_submit(conn, data)
+        do_submit(conn, params)
 
       errors ->
         conn
         |> put_status(400)
-        |> render_form(%{errors: errors, comments: Map.get(data, "comments")})
+        |> render_form(%{errors: errors, comments: Map.get(params, "comments")})
     end
   end
 
@@ -98,10 +100,11 @@ defmodule SiteWeb.CustomerSupportController do
           &validate_photos/1,
           &validate_name/1,
           &validate_email/1,
-          &validate_privacy/1
+          &validate_privacy/1,
+          &validate_recaptcha/1
         ]
       else
-        [&validate_comments/1, &validate_service/1, &validate_photos/1]
+        [&validate_comments/1, &validate_service/1, &validate_photos/1, &validate_recaptcha/1]
       end
 
     Site.Validation.validate(validators, params)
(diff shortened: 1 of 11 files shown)


apps/site/mix.exs
```diff
@@ -112,6 +113,8 @@ defmodule Site.Mixfile do
       {:poolboy, "~> 1.5"},
       {:wallaby, "~> 0.22", runtime: false, only: :test},
       {:con_cache, "~> 0.12.0"},
+      {:recaptcha,
+       github: "samueljseay/recaptcha", ref: "8ea13f63990ca18725ac006d30e55d42c3a58457"},
```
phildarnowsky: @digitalcora I consider it undesirable as a rule to point to some random SHA in a dep, but I'm guessing we need something not yet included in a release? If that's so could you please leave a `FIXME` comment here, explaining what that not-yet-included stuff is, and noting that we should switch to a proper release of `recaptcha` when that's possible?

apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -151,6 +154,24 @@ defmodule SiteWeb.CustomerSupportController do
...
+    :timeout_or_duplicate,
+    # https://github.com/samueljseay/recaptcha/issues/58
+    :"invalid-input-response"
+  ]
+
+  @spec validate_recaptcha(map) :: :ok | String.t()
+  defp validate_recaptcha(%{"recaptcha_response" => response}) do
+    case Recaptcha.verify(response) do
+      {:ok, _} -> :ok
+      {:error, [error]} when error in @expected_recaptcha_errors -> "recaptcha"
```
phildarnowsky: @digitalcora let's log the original error at level `warn`, because I'd like to set up a Splunk alert in case we start getting repeated errors here.

digitalcora: Okay, although I'm not sure of the usefulness, since this case covers just the errors that are part of the normal operation of a properly-implemented reCAPTCHA (hence _expected_ errors).

apps/site/mix.exs
digitalcora: I consider it undesirable as well, but this is unfortunately the only way we can make the controller logic testable, since the ability to artificially generate an error response from the `MockClient` doesn't exist in any released version of the package. I can add a comment explaining this.

apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -167,8 +167,12 @@ defmodule SiteWeb.CustomerSupportController do
   @spec validate_recaptcha(map) :: :ok | String.t()
   defp validate_recaptcha(%{"recaptcha_response" => response}) do
     case Recaptcha.verify(response) do
-      {:ok, _} -> :ok
-      {:error, [error]} when error in @expected_recaptcha_errors -> "recaptcha"
+      {:ok, _} ->
+        :ok
+
+      {:error, [error]} when error in @expected_recaptcha_errors ->
+        _ = Logger.warn("reCAPTCHA failed validation: #{error}")
```
phildarnowsky: @digitalcora let's reword the error to `"recaptcha failed_validation=#{error}"`, as that will expedite certain analyses in Splunk

PR #426, opened 2020-03-13
Change "Storm service" copy to "Reduced service"

phildarnowsky:
Currently services with a `typicality` of `unplanned_disruption` are
labelled "Storm service" in the schedule finder, on the assumption that
we would reserve that typicality for winter storms. Given the
possibility of reduced service due to the COVID pandemic, however, here
we change our copy to the more general "Reduced service", which is still
accurate in case of a storm as well.

#### Summary of changes
**Asana Ticket:** [Change "Storm Service" hardcoded copy in schedule selector to "Reduced Service"](https://app.asana.com/0/555089885850811/1166409415592881)

### DIFF
--- a/apps/site/assets/ts/schedule/__tests__/__snapshots__/ServiceOptGroupTest.tsx.snap
+++ b/apps/site/assets/ts/schedule/__tests__/__snapshots__/ServiceOptGroupTest.tsx.snap
@@ -32,7 +32,7 @@ exports[`ServiceOptGroup renders 1`] = `
   <option
     value="BUS319-storm"
   >
-    Storm service, Jul 15 to Jul 15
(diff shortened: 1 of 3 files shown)


paulswartz:
Do we want to do anything with the storm icons? https://github.com/mbta/dotcom/search?q=Storm&unscoped_q=Storm

phildarnowsky:
@digitalcora /cc @paulswartz thanks for the speedy review, I'm not familiar with those icons either but I will have a look after I merge this.

phildarnowsky:
@digitalcora @paulswartz looks like those icons are only used in Liquid replacements, i.e. the rewriting system that allows content authors to include icons and other rich elements in the content they write.

PR #431, opened 2020-03-16
Fix Elixir tests for spring 2020 rating

phildarnowsky:
Some tweaks to fix Elixir tests. Backstop ones still broken and need re-recording.

### DIFF
--- a/apps/site/test/site_web/controllers/schedule/line_test.exs
+++ b/apps/site/test/site_web/controllers/schedule/line_test.exs
@@ -11,12 +11,14 @@ defmodule SiteWeb.ScheduleController.LineTest do
 
   @deps %SiteWeb.ScheduleController.Line.Dependencies{}
 
+  @base_end_date ~D[2020-12-08]
+
   @thirtynine_services [
     %Service{
       added_dates: [],
       added_dates_notes: %{},
       description: "Weekday schedule",
-      end_date: ~D[2020-03-13],
+      end_date: Date.add(@base_end_date, 5),
       id: "BUS120-3-Wdy-02",
       name: "Weekday",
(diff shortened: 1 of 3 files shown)


apps/site/test/site_web/controllers/schedule_controller_test.exs
```diff
@@ -23,27 +23,27 @@ defmodule SiteWeb.ScheduleControllerTest do
...
       conn =
         get(
           conn,
-          trip_view_path(conn, :show, "1", origin: "2167", destination: "82", direction_id: "1")
+          trip_view_path(conn, :show, "1", origin: "66", destination: "82", direction_id: "1")
         )
 
       html_response(conn, 200)
-      assert conn.assigns.origin.id == "2167"
+      assert conn.assigns.origin.id == "66"
```
paulswartz: It'll be easier to update this in the future if you extract it to a `@modulevar`

apps/site/test/site_web/controllers/schedule/line_test.exs
```diff
@@ -199,7 +199,7 @@ defmodule SiteWeb.ScheduleController.LineTest do
       added_dates: [],
       added_dates_notes: %{},
       description: "Weekday schedule",
-      end_date: ~D[2020-03-13],
+      end_date: ~D[2020-12-13],
```
paulswartz: It'll be easier to update this in the future if you extract it to a `@modulevar`, possibly combined with `Date.add`.

phildarnowsky:
@paulswartz I'm hoping by the next time this comes up (i.e. presumably the next rating), we won't be using live API data anymore, but it will only take a second to extract those values so why not.

PR #432, opened 2020-03-17
Alerts should use url field

alicia-mbta:
**Asana Ticket:** [Alerts | Website should use URL field](https://app.asana.com/0/385363666817452/1116522101105763)
#### Summary of changes
URL coming from Alerts wasn't being taken into account. Now it's being displayed in the header of each alert, in the format [Alert Message] [URL]
<br/>
This contains an unrelated fix for spacing in the Trip Planner, made in passing:
image

### DIFF
--- a/apps/site/assets/ts/components/Alerts.tsx
+++ b/apps/site/assets/ts/components/Alerts.tsx
@@ -147,6 +147,18 @@ const alertDescription = (alert: AlertType): ReactElement<HTMLElement> => (
 const Alert = ({ alert }: { alert: AlertType }): ReactElement<HTMLElement> => {
   const [expanded, toggleExpanded] = useState(false);
   const onClick = (): void => toggleExpanded(!expanded);
+  // remove [http:// | https:// | www.] from alert URL:
+  let strippedAlertUrl = alert.url.replace(/(https?:\/\/)?(www\.)?/i, "");
+
+  // capitalize 'mbta' (special case):
+  strippedAlertUrl = strippedAlertUrl.replace(/mbta/gi, "MBTA");
+
+  const headerContent = alert.url
+    ? `${alert.header}<span>&nbsp;</span><a href="${
+        alert.url
+      }" target="_blank">${strippedAlertUrl}</a>`
+    : alert.header;
+
   return (
     <li
       id={`alert-${alert.id}`}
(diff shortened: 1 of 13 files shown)


apps/site/lib/site_web/templates/alert/_item.html.eex
```diff
@@ -18,6 +18,9 @@
       </div>
       <div>
         <%= replace_urls_with_links(@alert.header) %>
+        <%= if @alert.url != nil do %>
+          <%= replace_urls_with_links(@alert.url) %>
```
phildarnowsky: @alicia-mbta will `replace_urls_with_links` do the right thing in case of an `@alert.url` of `""`? if not, let's check for that in the conditional on line 21 as well.

apps/site/lib/site_web/templates/alert/_banner.html.eex
```diff
@@ -13,6 +13,9 @@
       </div>
       <div>
         <%= replace_urls_with_links(@alert.header) %>
+        <%= if @alert.url != nil do %>
```
phildarnowsky: @alicia-mbta same question regarding a potential `@alert.url` of `""` as before

apps/site/assets/ts/components/Alerts.tsx
```diff
@@ -147,6 +147,10 @@ const alertDescription = (alert: AlertType): ReactElement<HTMLElement> => (
 const Alert = ({ alert }: { alert: AlertType }): ReactElement<HTMLElement> => {
   const [expanded, toggleExpanded] = useState(false);
   const onClick = (): void => toggleExpanded(!expanded);
+  // remove [http:// | https:// | www] from alert URL:
```
phildarnowsky: @alicia-mbta it looks like this commit has a lot of code that should have been in the previous commit, in addition to the space fix mentioned in the commit message. 

I would do the following:

* split this commit into two, one containing the space fix, one containing all URL-related fixes. if you haven't split a commit before, this is a good guide: https://emmanuelbernard.com/blog/2014/04/14/split-a-commit-in-two-with-git/
* take the new commit with URL fixes, and squash it into https://github.com/mbta/dotcom/pull/432/commits/9345bfca5e53d9e45c29858b554c17cf8c845417
* take the following commit, https://github.com/mbta/dotcom/pull/432/commits/66d61ed92bab58c39c994d609c8a4132fb915257, which has all the tests for URL fixes, and squash that into https://github.com/mbta/dotcom/pull/432/commits/9345bfca5e53d9e45c29858b554c17cf8c845417 as well--tests should go in the same commit as the code they exercise

You should end up with two commits, one containing everything URL-related (code and tests both), the other with just this space fix.

I remember you said you were pretty familiar with rebasing but let me know if you want to pair a bit to get help in rearranging this PR.

ingridpierre:
@alicia-mbta One improvement we'd like to make:

Instead of the following ''https://www.mbta.com/south-shore-garages", change how the link displays to be: "**MBTA.com/south-shore-garages**"? Visual change only, the actual url path should remain the same.
image

alicia-mbta:
Capitalization of 'mbta' is now included in this branch.

ingridpierre:
URLs now looks trim and elegant in all the places! I think the PIOs will really appreciate the change. lgtm as well 👍 @alicia-mbta

PR #433, opened 2020-03-17
Update location of CrossBrowserTesting AuthKey

phildarnowsky:
They moved it.

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -47,8 +47,8 @@ The easiest way to install this is as a global NPM package:
 
     npm install -g cbt_tunnels
 
-Then click on the "Local Connection" indicator in the CBT menu bar, copy the
-**Authkey**, and start the tunnel using this command:
+Then go to https://app.crossbrowsertesting.com/account, copy the **Authkey**,
+and start the tunnel using this command:
 
     cbt_tunnels --username web-tools@mbtace.com --authkey <AUTHKEY>
 


digitalcora:
As the person who wrote this just a couple weeks back (in #429), when I saw this I thought I had the worst timing ever, for CBT to remove this UI right after... but looking at it today, it's still there for me? Maybe I'm in a parallel universe 😄

I guess the Account page is a more reliable place to find the auth key after all. Shame it requires navigating away from the main test dashboard.

Screen Shot 2020-03-23 at 10 33 53 AM

digitalcora:
As the person who wrote this just a couple weeks back (in #429), when I saw this I thought I had the worst timing ever, for CBT to remove this UI right after... but looking at it today, it's still there for me? Maybe I'm in a parallel universe 😄

I guess the Account page is a more reliable place to find the auth key after all. Shame it requires navigating away from the main test dashboard.

Screen Shot 2020-03-23 at 10 33 53 AM

beeuhtricks:
The A/B test strikes again? Here's mine, for comparison: 
Screen Shot 2020-03-23 at 11 35 57 AM

beeuhtricks:
AHA! 

> You can download the correct version of the standalone binary for your system by logging into your CBT account **(on any browser other than Chrome)** and clicking along the top where it says “Local Connection”.


Emphasis mine.

digitalcora:
Ah, I see. With Chrome they want you to install their Chrome Extension, so they direct you away from the manual command-line workflow. I forgot that as a Firefox user I _do_ live in a parallel universe!

phildarnowsky:
@digitalcora @beeuhtricks in any case, I see I have a little revision to do here.

digitalcora:
@phildarnowsky I think it's good as-is; the Account page should reliably have the auth key on it, no matter what browser you're using.

beeuhtricks:
Yeah, and I would just rather not use the Chrome extension anyways.

phildarnowsky:
@beeuhtricks @digitalcora could one of you please approve this puppy when you have a minute?

PR #434, opened 2020-03-18
Reduce height of glitched line diagram div

beeuhtricks:
----

#### Summary of changes
**Asana Ticket:** [Line diagram visual glitch on Providence line](https://app.asana.com/0/385363666817452/1161660843757401/f)

I used the calc function to take the computed height and reduce it by 15px. This fixes the visual glitch.

### DIFF
--- a/apps/site/assets/css/_schedule-page-line-diagram.scss
+++ b/apps/site/assets/css/_schedule-page-line-diagram.scss
@@ -424,6 +424,10 @@ $line-width: 8px; // width of the colored line in the diagram
   .m-schedule-diagram__lines--merging .m-schedule-diagram__line:not(:first-child) {
     height: calc(#{$base-spacing * 2} + 10px);
   }
+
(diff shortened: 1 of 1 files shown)


ingridpierre:
Hmm, sadly this fixed  Providence, but re-broke Franklin Line (inbound direction only) :
image
image

This Foxboro pilot continues to create headaches for us. I'm not sure why these issues interact in this way @phildarnowsky may have more context than I do on that front.

If no good solution for both routes' diagrams can be reached, I'll see what Katrina says about priorities here.

beeuhtricks:
@ingridpierre it's possible there are better ways of fixing this, but the problem I keep running into is that those divs have a computed height that is larger than their containing parent and I can't figure out why. It is likely that root-causing this and fixing that would solve the issue.

phildarnowsky:
Going to close this one for neatness.

PR #435, opened 2020-03-19
Use variant ID in line page URL

phildarnowsky:
#### Summary of changes
**Asana Ticket:** [Schedules | Variant parameters in URL](https://app.asana.com/0/555089885850811/1162914752771485)

If you pick a non-default variant on the line page, the URL updates a new `variant` query param accordingly, and that param allows deep-linking to a given variant of a route. This is much the same with what we do about direction IDs in the same URL.


phildarnowsky:
Have also asked @aayesha to QA in dev-green.

PR #437, opened 2020-03-20
Don't send Phoenix.NotAcceptable errors to Sentry

phildarnowsky:
We get a lot of Phoenix.NotAcceptable errors due to script kiddies
and/or poorly behaving apps. They take up a significant fraction of our
monthly Sentry error allowance. Sentry allows us to filter out arbitrary
errors, so we filter those out now.

#### Summary of changes
**Asana Ticket:** [Investigate Phoenix.NotAcceptableErrors](https://app.asana.com/0/555089885850811/1166771847899433)

### DIFF
--- a/apps/site/config/config.exs
+++ b/apps/site/config/config.exs
@@ -47,7 +47,8 @@ config :sentry,
   enable_source_code_context: false,
   root_source_code_path: File.cwd!(),
   included_environments: [:prod],
-  json_library: Poison
+  json_library: Poison,
+  filter: Site.SentryFilter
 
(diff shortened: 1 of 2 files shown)


alicia-mbta:
Just to check I'm understanding this correctly: this filter will not send the specific NotAcceptableError to Sentry but otherwise all other exceptions will be sent?

phildarnowsky:
@alicia-mbta exactly. Documentation is available here: https://hexdocs.pm/sentry/7.0.4/Sentry.EventFilter.html

PR #438, opened 2020-03-20
Log warning for prediction with nil trip

phildarnowsky:
Although all predictions should have a trip ID, some are getting fetched
by the realtime schedule endpoint without one. To help the data team
debug this, we've added a bit of logging for when that happens.

#### Summary of changes
**Asana Ticket:** [Log a warning to Splunk whenever the realtime schedule endpoint encounters a prediction without a trip.](https://app.asana.com/0/555089885850811/1167515632160323)


beeuhtricks:
Looks good Phil!

PR #440, opened 2020-03-25
Add params to URL in Schedule Finder

alicia-mbta:
In the Schedule finder in the "Schedule and Maps" tab for Schedules the ability to update the URL with parameters `direction` and `origin` has been added.

#### Summary of changes
**Asana Ticket:** [Schedules | Schedule Finder works with URL parameters](https://app.asana.com/0/555089885850811/1162467761296251)


alicia-mbta:
Hi @phildarnowsky and @ingridpierre, this is still incomplete because there is one test that is failing but in the meantime you can take a look at the code and behavior of this branch to make sure it's working properly. Thanks.

apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
```diff
@@ -36,9 +37,51 @@ const ScheduleFinder = ({
...
+    setSelectedOrigin(origin);
+    if (selectedDirection !== null && origin !== null) {
+      updateURL(selectedDirection, origin);
+    }
+  };
+
+  const getCurrentScheduleFinderValues = (): void => {
+    let params = window.location.search;
+
+    if (params.startsWith("?")) {
```
phildarnowsky: @alicia-mbta since we're using `use-query-params` we can and should use its capabilities for parsing params, rather than rolling our own. See https://github.com/pbeshai/use-query-params.

apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
```diff
@@ -36,9 +41,31 @@ const ScheduleFinder = ({
   const [selectedDirection, setSelectedDirection] = useState(directionId);
   const [selectedOrigin, setSelectedOrigin] = useState<SelectedOrigin>(null);
 
+  const [query] = useQueryParams({
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    direction_id: StringParam,
```
phildarnowsky: @alicia-mbta as we discussed in chat the other day, please adjust this `direction_id` as well as that in `apps/site/assets/ts/schedule/components/direction/reducer.ts`so that they're scoped and don't collide. That is, we should call the param here `schedule_finder[direction_id]` and the other `schedule_direction[direction_id]`.

Then, for consistency, we should also rename the `origin` param here to `schedule_finder[origin]`, and likewise `variant` in the other file to `schedule_direction[variant]`.

And of course the corresponding backend code should get adjusted to match.

phildarnowsky:
@alicia-mbta let me experiment a little with this, I want to try something

apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
```diff
@@ -82,16 +82,21 @@ const ScheduleFinder = ({
 
   React.useEffect(
     () => {
-      const newDirection = query.direction_id === "0" ? 0 : 1;
+      const scheduleFinder = {
```
phildarnowsky: @alicia-mbta so this is not quite what I meant. The concern here is not what the variable names are going to be in the Typescript code, but rather what the keys of the query parameters are going to be. As it stands we are using the key `direction_id` for two separate things, one in the schedule finder, one in the direction pickers. This is going to lead to bugs if we let it stand.

My solution is to change the names of these params in the query string, so that they're called "schedule_finder[direction_id]" and "schedule_direction[direction_id]" respectively. For the sake of consistency, we should also change the name of the `origin` param here to `schedule_finder[origin]`, and the name of the `variant` param in the direction picker to `schedule_direction[variant]`.

When you change the names of the query params in the frontend, of course, you'll need to change them in the backend to match. Because of the way Elixir parses query params, if you use the names I suggest, you'll end up with nested query params on the Elixir side, which is what we want in this case.

So what I suggest is that you revert this commit and try changing the query parameter names as I describe. If you want we could pair later in the day on this, in case what I'm saying is not clear.

alicia-mbta: Made some more changes, please review them when you can because I'm still unsure of whether I'm still missing something or not.
Also investigating why `npm run format:check` is failing.

phildarnowsky: one thing about formatting, is you can also do `npm run format` and that'll automatically format your JS/TS code to match the standard.

phildarnowsky:
@alicia-mbta looks good to me so far, let me know when that test passes, or if you need help getting it to pass.

ingridpierre:
I can see parameters generated when I interact with the schedule finder, however noticing a few things that I'm wondering if we missed in the requirements discussion of this ticket. Let me know if the following are included here and I'm missing something, or if we should determine the other requirements as separate but additional work to make this feature complete.

1. When I copy and paste the URL with parameters for both origin and direction I should see the Schedule Finder opened and updated with the parameters provided. Right now I only return to the base line page. (Unless I'm missing something?)
2 . If only the direction OR origin parameter is provided, when a user visits that URL, the schedule finder widget should update to reflect that selection automatically.
2. When I click "schedules from here" the URL should also update with the parameters using the direction set on the page and the origin as the stop I have clicked the link from.

1 is the biggest thing as the ultimate goal of this is to be able to share a link with parameters for customer support sending information or helping people bookmark links to schedule finder results.

ingridpierre:
Revisiting the parameters seems to work well for all the modes except subway. So far not able to revisit links generated on subway lines like: /schedules/Blue/line?schedule_finder%5Bdirection_id%5D=0&schedule_finder%5Borigin%5D=place-wondl

alicia-mbta:
@ingridpierre are you saying it's not working for subway at all or that you're not able to test? I'm checking now and I'm unable to _but_ because the widget that appears is Schedule Note instead of the two dropdowns (in which case no matter what the URL parameters are, they are ignored). Isn't this the right behavior?

ingridpierre:
@beeuhtricks All good in IE11, thanks for helping me get that on dev!

ingridpierre:
Looks like this pr changed some expected behavior. Normally when reselecting an origin, the stop picker modal replaces the finder modal like so:

image

In this PR the modal is still visible below:
image

ingridpierre:
@alicia-mbta  believe the URL parameters for the schedule finder should be removed when i click it closed. Playing around with it now, it's a bit unexpected to close ut but return to an old state when I refresh.

Users typically refresh on the line page when they don't trust our vehicle info is live, so having the finder pop up again would be frustrating.

phildarnowsky:
@alicia-mbta regarding the commit message, two things:
  * The title of a commit message should be no more than 50 characters, and the lines of the body no more than 72 characters.
  * Given the amount of stuff in this one commit, more detail would be helpful. Specifically, I would like to see a high-level overview of the major changes. Introducing Redux would be one, and causing the URL parameters to change would be another...anything else?

apps/site/assets/ts/schedule/ScheduleLoader.tsx
```diff
@@ -0,0 +1,241 @@
...
+import { DirectionId } from "../__v3api";
+import mapStateToProps, {
+  getCurrentState,
+  storeHandler
+} from "./ScheduleStore";
+
+interface Props {
+  schedulePageData: SchedulePageData;
+  component: string;
+  updateURL: Function;
```
phildarnowsky, line 20: @alicia-mbta let's take advantage of the type system and write a more specific function type here, as described in https://www.typescriptlang.org/docs/handbook/functions.html#function-types

apps/site/assets/ts/schedule/ScheduleStore.tsx
```diff
@@ -0,0 +1,57 @@
+import { createStore } from "redux";
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+const storeReducer = (
+  state: any,
```
phildarnowsky, line 6: @alicia-mbta we should be able to write a more specific type than `any` in this file. It's best to avoid using `any` whenever possible because you lose the benefit of having a static type system when you use it.

apps/site/lib/site_web/controllers/schedule/line.ex
```diff
@@ -39,16 +39,33 @@ defmodule SiteWeb.ScheduleController.Line do
...
-        %Conn{assigns: %{route: %Route{} = route, direction_id: direction_id}} = conn,
+        %Conn{
+          assigns: %{
+            route: %Route{} = route,
+            direction_id: direction_id
+          }
+        } = conn,
         args
       ) do
+    schedule_direction = conn.query_params["schedule_direction"] || %{}
```
phildarnowsky, line 50: @alicia-mbta a somewhat more idiomatic way to write this is via [Map.get/3](https://hexdocs.pm/elixir/Map.html#get/3), something like:

`schedule_direction = Map.get(conn.query_params, "schedule_direction", %{})`

That third argument is the default to use if the key is not found in the map.

apps/site/lib/site_web/controllers/schedule/line.ex
```diff
@@ -39,16 +39,33 @@ defmodule SiteWeb.ScheduleController.Line do
...
+
+    direction_id_value =
+      case schedule_direction["direction_id"] do
```
phildarnowsky, line 53: @alicia-mbta something has occurred to me, and this one's on me, I should have thought of this earlier. We need to make this backwards-compatible so that if someone has a bookmark saved using the old `direction_id` and `variant` parameters, it should honor those.

apps/site/assets/ts/schedule/ScheduleLoader.tsx
```diff
@@ -0,0 +1,241 @@
...
+    schedule_note: scheduleNote,
+    today,
+    line_diagram: lineDiagram,
+    shape_map: shapesById,
+    variant: initialSelectedRoutePatternId
+  } = schedulePageData;
+
+  const currentState = getCurrentState();
+
+  if (!!currentState && Object.keys(currentState).length !== 0) {
```
phildarnowsky, line 129: @alicia-mbta can you write `currentState.keys`? I think that works, not sure without trying it.

apps/site/assets/ts/schedule/ScheduleLoader.tsx
```diff
@@ -0,0 +1,241 @@
...
+import ScheduleFinderModal from "./components/schedule-finder/ScheduleFinderModal";
+import { DirectionId } from "../__v3api";
+import mapStateToProps, {
+  getCurrentState,
+  storeHandler
+} from "./ScheduleStore";
+
+interface Props {
+  schedulePageData: SchedulePageData;
+  component: string;
```
phildarnowsky, line 19: @alicia-mbta this would be a good place to use a [string literal type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types). In this case it would be something like:

```
type ComponentType = "MAIN" | "SCHEDULE_NOTE" | "SCHEDULE_FINDER" | "SCHEDULE_DIRECTION"
```

and then you can define the `component` field here as having type `ComponentType`.

apps/site/assets/ts/schedule/ScheduleLoader.tsx
```diff
@@ -0,0 +1,241 @@
...
+    let modalOpen = false;
+
+    if (query["schedule_finder[direction_id]"] !== undefined) {
+      newDirection = query["schedule_finder[direction_id]"] === "0" ? 0 : 1;
+    }
+    if (query["schedule_finder[origin]"] !== undefined) {
+      newOrigin = query["schedule_finder[origin]"];
+    }
+
+    const button = document.getElementById("modal-close")!;
```
phildarnowsky, line 92: @alicia-mbta so basically here you're looking for the modal-close button as a way of determining if the modal is open?

apps/site/assets/ts/schedule/schedule-loader.tsx
```diff
@@ -27,40 +32,60 @@ const renderMap = ({
...
-      <ScheduleFinder
-        directionId={directionId}
-        route={route}
-        stops={stops}
-        services={services}
-        routePatternsByDirection={routePatternsByDirection}
-        today={today}
-        scheduleNote={null}
-      />,
+      <Provider store={store}>
```
phildarnowsky, line 81: @alicia-mbta not too familiar with `Provider`s, what's this about?

phildarnowsky:
@alicia-mbta as a reminder I will be out tomorrow but I will take a look online from time to time as I can, to see if you have any questions about my feedback.

apps/site/assets/ts/schedule/schedule-loader.tsx
thecristen: @phildarnowsky it's a redux thing! it's what enables the `ScheduleLoader` to connect to the store

alicia-mbta:
Hi @ingridpierre and @phildarnowsky, this is ready for retest and review. Thanks.

apps/site/lib/site_web/controllers/schedule/line.ex
ryan-mahoney: @phildarnowsky @alicia-mbta are you thinking to achieve this via a redirect?

ingridpierre:
@alicia-mbta Looks like everything I reported above has been addressed. It's great and adds wonderful functionality for folks, especially while folks may need to continue to refresh realtime information. 👍

apps/site/lib/site_web/controllers/schedule/line.ex
alicia-mbta: Yes, @phildarnowsky has explained this to me. The changes for that are in file `dotcom/apps/site/lib/site_web/controllers/schedule/line.ex`

apps/site/lib/site_web/controllers/schedule/line.ex
```diff
@@ -38,18 +40,86 @@ defmodule SiteWeb.ScheduleController.Line do
     Util.log_duration(__MODULE__, :do_call, [conn, opts])
   end
 
+  @spec get_updated_format_for_url_params(Plug.Conn.t()) :: Enum.t()
+  def get_updated_format_for_url_params(conn) do
+    new_params =
```
ryan-mahoney, line 45: This code could be a lot more DRY, for example:

```
defp replace_map_key(input, from, to) do
  if Map.has_key?(input, from) do
    input
    |> Map.put(to, Map.get(input, from))
    |> Map.delete(from)
  else
    input
  end
end

new_params = conn.query_params
  |> replace_map_key("direction_id", "schedule_finder[direction_id]")
  |> replace_map_key("origin", "schedule_finder[origin]")
  |> replace_map_key("variant", "schedule_direction[variant]")
```

apps/site/lib/site_web/controllers/schedule/line.ex
```diff
@@ -38,18 +40,86 @@ defmodule SiteWeb.ScheduleController.Line do
...
+    # check if URL has parameters like direction_id, origin or variant
+    # If so, they will get changed to:
+    # - "schedule_finder[direction_id]" and "schedule_finder[origin]" for direction_id and origin
+    # - "schedule_direction[variant]" for variant
+    # e.g. /schedules/31/line?direction_id=1&origin=525 will be changed to
+    # /schedules/31/line?schedule_finder%5Bdirection_id%5D=1&schedule_finder%5Borigin%5D=525
+
+    new_params = get_updated_format_for_url_params(conn)
+
+    if !Map.equal?(new_params, conn.query_params) do
```
alicia-mbta, line 93: What am I missing here? why is this comparison giving an error in Dialyzer?

ryan-mahoney: @alicia-mbta I'll take a look after dinner

phildarnowsky:
@ryan-mahoney this is where I'm leaving off today, a lot of tests are still failing and it got less mechanical. Thinking I may well have introduced a legit regression, we'll see tomorrow.

apps/site/lib/site_web/controllers/schedule/green.ex
```diff
@@ -259,13 +259,21 @@ defmodule SiteWeb.ScheduleController.Green do
...
+             direction_id: direction_id
+           }
+         } = conn,
          _
        )
        when not is_nil(origin) and not is_nil(destination) do
     {stops, _map} = conn.assigns.stops_on_routes
 
+    stops = Enum.reverse(stops)
+    Util.ListHelpers.find_first(stops, origin, destination)
```
ryan-mahoney, line 278: @phildarnowsky confused by this line, looks like this is being called but the return value discarded?

phildarnowsky: @ryan-mahoney I probably had an `IO.inspect` in there, this line can go

phildarnowsky:
@ryan-mahoney CI green, going to take a short break and start exercising this manually.

phildarnowsky:
@ryan-mahoney hooray, coverage is there--clocking out for now if you want to start the review process.

apps/site/test/site_web/controllers/schedule/line_test.exs
```diff
@@ -517,15 +517,15 @@ defmodule SiteWeb.ScheduleController.LineTest do
...
+            {"Canton Junction", 9},
+            {"Stoughton", 7},
+            {"Wickford Junction", 0}
           ] do
         assert stops |> Enum.at(idx) |> elem(1) == id
       end
     end
 
-    test "CR-Providence inbound has correct number of bubbles" do
+    test "CR-Providence inboung has correct number of bubbles" do
```
ryan-mahoney: not a big deal, but this should be `inbound` :)

phildarnowsky:
@ryan-mahoney subway direction switching is working for me, can you try it again?

phildarnowsky:
@ryan-mahoney try this on for size

phildarnowsky:
@ryan-mahoney fixed the subway origin problem.

PR #441, opened 2020-03-26
Redirect /style-guide to zeroheight

beeuhtricks:
#### Summary of changes
**Asana Ticket:** [Retire /style-guide, 301 redirect to zeroheight](https://app.asana.com/0/385363666817452/1160793451440400/f)

The style_guide_controller now redirects requests to corresponding pages on the new zeroheight design site. I also updated the test for this controller to verify that the redirects are happening correctly. A note about the zeroheight slugs, for my reviewers: they seem to function equally well with or without the page title appended to the end with hyphens, i.e. https://zeroheight.com/2fedee66c/p/43fa10 functions just as well as https://zeroheight.com/2fedee66c/p/43fa10-about, which the former redirects to. I elected to include the title but am looking for feedback in that area as to what best practice should be.

---

(I've filled out this template to the best of my ability, this doesn't contain front-end changes so some of them are kind of n/a)

### DIFF
--- a/apps/site/lib/site_web/controllers/style_guide_controller.ex
+++ b/apps/site/lib/site_web/controllers/style_guide_controller.ex
@@ -1,23 +1,25 @@
 defmodule SiteWeb.StyleGuideController do
+  @moduledoc """
+  Defines the behavior of style guide requests.
+  """
   use SiteWeb, :controller
 
   def index(conn, %{"section" => "content"}) do
-    redirect(conn, to: "/cms/content-style-guide")
+    conn
+    |> put_status(301)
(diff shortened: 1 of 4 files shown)


ingridpierre:
Interesting note about the title suffix, let's actually not include them. The shorter URL seems preferable and if both are equal, we probably want flexibility to change the page title without accidentally updating a URL suffix.

This is looking good to me, when you follow up with Aayesha (she's off today) can you ask her to update mbta.com/dsg vanity URL as well?

beeuhtricks:
Gotcha @phildarnowsky. Will squash 👍

PR #442, opened 2020-04-03
hide /org section from search engines

ryan-mahoney:
#### Summary of changes
**Asana Ticket:** No Ticket

Prevent search engines from indexing pages under the `/org` directory.

---


ryan-mahoney:
@phildarnowsky I had originally added a function to `apps/site/lib/site_web/views/helpers.ex` with the function below, but I thought it might be overkill for this purpose. Let me know what you think -- thanks!

```
  @doc """
  Determine if a requested path begins with /org/
  """
  @spec is_org_page(String.t()) :: boolean
  def is_org_page("/org/" <> _), do: true
  def is_org_page(_), do: false
```

phildarnowsky:
@ryan-mahoney yeah overkill IMHO

PR #444, opened 2020-04-06
Show all vehicles in given direction in map

phildarnowsky:
On the line page, if you look at a branching subway line such as the RL
northbound or the GL in either direction, you'd see only markers for
one branch. Corrected.

#### Summary of changes
**Asana Ticket:** [Bug | Map missing Southbound Braintree trains](https://app.asana.com/0/555089885850811/1169381171498584)

---

### DIFF
--- a/apps/site/assets/ts/schedule/__tests__/MapTest.tsx
+++ b/apps/site/assets/ts/schedule/__tests__/MapTest.tsx
@@ -55,7 +55,9 @@ const data: MapData = {
 
 describe("Schedule Map", () => {
   it("renders", () => {
-    const wrapper = mount(<Map data={data} channel="vehicles:Red:0" />);
+    const wrapper = mount(
+      <Map data={data} channel="vehicles:Red:0" shapeIds={[]} />
+    );
     expect(() => wrapper.render()).not.toThrow();
   });
 });
@@ -77,7 +79,7 @@ describe("reducer", () => {
 
   it("resets markers", () => {
     const result = reducer(
-      { markers: data.markers, shapeId: "1", channel: "vehicle:1:1" },
+      { markers: data.markers, channel: "vehicle:1:1", shapeIds: ["1"] },
       {
         action: {
           event: "reset",
@@ -95,7 +97,7 @@ describe("reducer", () => {
 
   it("adds vehicles", () => {
     const result = reducer(
-      { markers: data.markers, channel: "vehicle:1:1", shapeId: "1" },
+      { markers: data.markers, channel: "vehicle:1:1", shapeIds: ["1"] },
       {
         action: { event: "add", data: [{ marker: newMarker }] },
         channel: "vehicle:1:1"
@@ -108,7 +110,7 @@ describe("reducer", () => {
 
   it("updates markers", () => {
     const result = reducer(
-      { markers: data.markers, channel: "vehicle:1:1", shapeId: "1" },
+      { markers: data.markers, channel: "vehicle:1:1", shapeIds: ["1"] },
       {
         action: {
           event: "update",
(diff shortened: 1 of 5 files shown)


phildarnowsky:
@beeuhtricks I haven't checked this in IE yet, I am downloading a VM to do that with in the absence of Crossbrowsertesting, but it's a big download so it'll be a while. Otherwise this is ready for review.

phildarnowsky:
@beeuhtricks IE confirmed buttery, or as buttery as IE gets.

apps/site/assets/ts/schedule/__tests__/MapTest.tsx
```diff
@@ -77,7 +79,7 @@ describe("reducer", () => {
 
   it("resets markers", () => {
     const result = reducer(
-      { markers: data.markers, shapeId: "1", channel: "vehicle:1:1" },
+      { markers: data.markers, shapeIds: ["1"], channel: "vehicle:1:1" },
```
beeuhtricks, line 82: Small thing: this shapeIds key appears to come last in the object in every other test in this file, making this one inconsistent in appearance.

apps/site/assets/ts/schedule/schedule-loader.tsx
```diff
@@ -8,15 +8,23 @@ import { SchedulePageData } from "./components/__schedule";
...
 
-const renderMap = (): void => {
+const renderMap = ({
+  route_patterns: routePatternsByDirection,
+  direction_id: directionId
+}: SchedulePageData): void => {
+  const routePatterns = routePatternsByDirection[directionId];
+  const shapeIds = routePatterns.map(routePattern => routePattern.shape_id);
   const mapDataEl = document.getElementById("js-map-data");
   if (!mapDataEl) return;
```
beeuhtricks, line 18: So, I'm not sure of the details of calling .render() on line 61 of MapTest.tsx, but this line appears to return from the function immediately if the map data element is not found, while that test only checks for thrown exceptions. Is there another test covering that?

apps/site/assets/ts/schedule/__tests__/MapTest.tsx
```diff
@@ -95,7 +97,7 @@ describe("reducer", () => {
 
   it("adds vehicles", () => {
     const result = reducer(
-      { markers: data.markers, channel: "vehicle:1:1", shapeId: "1" },
+      { markers: data.markers, channel: "vehicle:1:1", shapeIds: ["1"] },
```
beeuhtricks, line 100: This may be out of scope, but shapeIds in this PR (and section of the codebase) appear to be strings containing integers. Is there a reason they aren't just regular integers, or a specific Shape ID type?

phildarnowsky: Generally they're not integers, one example (a particularly complicated one) of a non-integral ID is `933_0010-70105-70102-2`. You can see more examples by going to e.g. https://api-v3.mbta.com/shapes?filter[route]=Red.

A ShapeID type would be a good idea but I don't see it as in scope of this PR. Since you bring it up, though, it would be good for us to make a practice out of introducing intention-revealing types in new TS code.

apps/site/assets/ts/schedule/schedule-loader.tsx
phildarnowsky: Yes, that should be covered by `ScheduleDirectionTest.tsx`, lines 281 and on. The map data element should be missing when we're using a static map image rather than the dynamic map we're used to, which IIRC we do only for ferry.

phildarnowsky:
@beeuhtricks this should address all your feedback

apps/site/assets/ts/schedule/schedule-loader.tsx
beeuhtricks: Ok, thanks

apps/site/assets/ts/schedule/__tests__/MapTest.tsx
beeuhtricks: Ah yeah, ok. I missed that. Definitely out of scope.

PR #446, opened 2020-04-15
further specify style for expanded branch segment

thecristen:
#### Summary of changes
**Asana Ticket:** [🐞 Line diagram visual glitch on Providence line](https://app.asana.com/0/385363666817452/1161660843757401/f)

For short branches that are by default expanded, there needed to be additional styling to achieve the desired appearance. This has now been implemented!

image
---

### DIFF
--- a/apps/site/assets/css/_schedule-page-line-diagram.scss
+++ b/apps/site/assets/css/_schedule-page-line-diagram.scss
@@ -11,9 +11,10 @@ $line-width: 8px; // width of the colored line in the diagram
   }
 }
 
-// Adds a curved line from branch to trunk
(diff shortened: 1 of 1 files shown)


ingridpierre:
@thecristen  This looks great! Last time the fix for Stoughton broke Franklin/Foxboro - but just checked that too and it's perfect now.

Found a bug that's probably unrelated, but wondering if you can reproduce: clicking view schedule from the CR line diagram, I can't then change the direction from the top of the modal (clicking dropdown options doesn't do anything). This doesn't happen in prod so I'm wondering if it was introduced somewhere else?

PR #447, opened 2020-04-16
Redirect users away from mbtace.com

phildarnowsky:
We want all users, including search engine robots, to use the canonical
mbta.com hostname. The exception is mTicket, which uses a special
hostname that we detect to flag the request for special handling; there
is a carveout for that special hostname.

#### Summary of changes
**Asana Ticket:** [mbtace.com being indexed/viewed by public](https://app.asana.com/0/555089885850811/1169245762828233)

---

### DIFF
--- a/apps/site/lib/site_web/endpoint.ex
+++ b/apps/site/lib/site_web/endpoint.ex
@@ -1,4 +1,6 @@
 defmodule SiteWeb.Endpoint do
+  @moduledoc false
+
   use Phoenix.Endpoint, otp_app: :site
 
   @doc """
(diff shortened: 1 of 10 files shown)


ryan-mahoney:
One of those rare cases where `Plug` is a good solution :D

I was thinking this ticket could be satisfied by some changes to the `nginx.conf`, but this is probably more maintainable.

phildarnowsky:
@ryan-mahoney yeah my preference is to keep it in the code, kind of have everything together in one place where all the devs can see and touch it.

PR #448, opened 2020-04-21
Consider local IPs to be canonical hostnames

phildarnowsky:
The health checker runs on the same hosts as the application servers and
uses a local IP address to access the /_health endpoint. Whitelist local
IP addresses as well.

No ticket.

---

### DIFF
--- a/apps/site/lib/site_web/plugs/canonical_hostname.ex
+++ b/apps/site/lib/site_web/plugs/canonical_hostname.ex
@@ -23,7 +23,9 @@ defmodule SiteWeb.Plugs.CanonicalHostname do
   def call(%Plug.Conn{host: requested_hostname} = conn, _) do
     canonical_hostname = SiteWeb.Endpoint.config(:url)[:host]
 
-    if requested_hostname != canonical_hostname do
+    if requested_hostname == canonical_hostname or is_private_ip(requested_hostname) do
+      conn
+    else
       rewritten_url =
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/plugs/canonical_hostname.ex
```diff
@@ -32,8 +34,18 @@ defmodule SiteWeb.Plugs.CanonicalHostname do
       |> put_status(:moved_permanently)
       |> redirect(external: rewritten_url)
       |> halt()
-    else
-      conn
+    end
+  end
+
+  # The health checker uses a local IP hostname
+  defp is_local_ip(hostname) do
```
thecristen, line 41: Will this approach work ok in both prod and dev, and are the addresses something that should be specified in the config?

phildarnowsky: It should work on both, I just double-checked Splunk and the `/_health` endpoint is being hit from a local IP there as well, so I assume that it's requesting the local IP as the host.

The specific addresses it uses seem arbitrary so I don't want to specify them, rather just to check that it's some private IP or another.

phildarnowsky:
@thecristen technically I should have said this works with _private_ rather than local IPs, will update.

phildarnowsky:
@thecristen I think as a consequence of the Github troubles yesterday CI is not running on this PR, I will close and open a fresh one.

PR #449, opened 2020-04-22
Consider private IPs to be canonical hostnames

phildarnowsky:
The health checker runs on the same hosts as the application servers and
uses a private IP address to access the /_health endpoint. Whitelist local
IP addresses as well.

No ticket.

---


phildarnowsky:
@thecristen hm, still not running CI, going to try a fresh branch entirely

PR #453, opened 2020-04-22
Revert canonical hostname stuff

phildarnowsky:
Due to an apparent bug in Phoenix these fixes don't actually work. Reverting to make master deployable again until this can be further investigated.

---


phildarnowsky:
@thecristen some days I win, some days the computer wins. Today the computer won.

PR #454, opened 2020-04-23
Add ability to make timetable top header sticky

thecristen:
#### Summary of changes
**Asana Ticket:** [CR Timetable | Header difficult to track](https://app.asana.com/0/385363666817452/950408339997961/f)

Depending on zoom levels / window size, it was possible to lose sight of the top header row while viewing the commute rail timetable. This somewhat hacky change aims to remedy that. As is, **this won't work in IE or Edge**, so those browsers should not see any change to this page.

A few new helpers were also added:

* `isMSBrowser()` for detecting the browser with JavaScript
* `full-width()` Sass mixin

---

### DIFF
--- a/apps/site/assets/css/_mixins.scss
+++ b/apps/site/assets/css/_mixins.scss
@@ -354,3 +354,11 @@
     /* stylelint-enable declaration-no-important */
   }
 }
+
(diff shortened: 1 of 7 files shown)


thecristen:
Gonna rework this.

PR #455, opened 2020-04-23
Redirect users away from canonical hostnames

phildarnowsky:
This un-reverts 1713c9a53c0060275350af9fb78c05d5e1c856f6 and
3d479fcba1aa86dfe6d35bd43ddf8c8c3de7c5e3. It also includes one fix, to
work around an apparent Phoenix bug whereby
SiteWeb.Endpoint.config(:url)[:host] is not set to the canonical
hostname as it should be. Instead we reach directly for System.get_env.

**Asana Ticket:** [mbtace.com being indexed/viewed by public](https://app.asana.com/0/555089885850811/1169245762828233)

---

### DIFF
--- a/apps/site/lib/site_web/endpoint.ex
+++ b/apps/site/lib/site_web/endpoint.ex
@@ -1,4 +1,6 @@
 defmodule SiteWeb.Endpoint do
+  @moduledoc false
+
   use Phoenix.Endpoint, otp_app: :site
 
   @doc """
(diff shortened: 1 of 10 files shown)


phildarnowsky:
@thecristen I tried it in dev-green and I'll double check in dev as well; if it works in those two it _should_ work in prod. Just to be safe I'm going to be ready to roll this back quickly when I do deploy.

PR #457, opened 2020-04-23
Better handle reduced services in the service picker

thecristen:
#### Summary of changes
**Asana Ticket:** [Update reduced schedules to use the service description, hide validity dates from UI](https://app.asana.com/0/385363666817452/1169855995557655/f)

Prompted by the new abundance of reduced service, this PR cleans up the service picker by implementing the following changes:

* Uses the `service.description`text for services of type `unplanned_disruption`
* Allows services of type `unplanned_disruption` to be listed as "Current"
* Adds logic to omit from the dropdown the service of type `typical_service` that was replaced by the unplanned disruption, if applicable.

| original | updated |
| ------------- | ------------- |
| image | image |

image

---

### DIFF
--- a/apps/site/assets/ts/helpers/service.ts
+++ b/apps/site/assets/ts/helpers/service.ts
@@ -114,41 +114,11 @@ export const isInFutureRating = (
   return currentDate <= ratingStartDate && currentDate <= ratingEndDate;
 };
 
-// sometimes a current unplanned_disruption supercedes typical_service.
-// in this case both will be present in services, with most of the same
-// properties. when we don't want to show both services, prefer the unplanned
-export const omitReplacedServices = (services: Service[]): Service[] => {
-  const unplannedServices = services.filter(
-    (s: Service) => s.typicality === "unplanned_disruption"
-  );
-
-  if (unplannedServices.length) {
-    // find the corresponding planned services by these properties
-    const matchers: (keyof Service)[] = [
-      "type",
-      "rating_description",
-      "valid_days"
-    ];
-
-    const matchedServices = services
-      .filter(s => s.typicality === "typical_service")
-      .filter((ts: Service) =>
-        unplannedServices.find((us: Service) =>
-          matchers.every(m => _.isEqual(ts[m], us[m]))
-        )
-      );
-
-    return _.difference(services, matchedServices);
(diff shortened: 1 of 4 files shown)


ingridpierre:
(Copied from Slack for reference) 
While I instinctually like what you’ve done for the most part,  its a little different than I was expecting. In particular:

- I wonder if we can also show the non-reduced weekday schedule?
- I wonder what happens to the picker on the weekend? Or when there’s weekend and weekday reduced services?
- I worry the parenthetical of (reduced service), will clash with the near future design change to append “(Now)” to the current service
- Re: the above point, “Reduced Weekday Service (Now)” might read better?

thecristen:
@ingridpierre To answer some of these questions: 

>    I wonder if we can also show the non-reduced weekday schedule?

We totally can. Would it make sense to categorize this particular case as "Other" (otherwise it'll show up as "Future" since its service date is in the future)?

> I worry the parenthetical of (reduced service), will clash with the near future design change to append “(Now)” to the current service ... “Reduced Weekday Service (Now)” might read better?

Yeah, don't know. Right now it's just reading the `service.description` and displaying it as-is. If we don't have a clear choice now, maybe we can adjust this text later?

>    I wonder what happens to the picker on the weekend? 

It'll pick the current weekend schedule if available, just like it does currently. You can test it by adding the date_time parameter to the URL, e.g. `?date_time=2020-04-25T11:11`

> Or when there’s weekend and weekday reduced services?

I would think it'd follow the same pattern we saw here with the weekday. Do you know if there's any routes with weekend reduced services we can validate this on?

thecristen:
@ingridpierre I've reverted the last commit so it will show the non-reduced schedules, let me know if this works better!

ingridpierre:
@thecristen there is a ticket in the backlog to address the fact that the original spring rating gets moved out of current. We probably shouldn't worry about it for this ticket. Here it is fyi:

https://app.asana.com/0/385363666817452/1170658788930619/f

Either way I'll take a look on Monday, thanks for the heads up!

ingridpierre:
@thecristen This is now looking good to me for the purpose of the original scope of the ticket! Thanks for adding back the regular schedules here. 👍

PR #458, opened 2020-04-28
Specify width to avoid overflow

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Line Diagram | IE11 stop content width overflows](https://app.asana.com/0/385363666817452/1171464082559329)

Content overflows for IE in Line Diagram when the number of connections is too big. It needs to wrap.

Before:
image

After:
Screen Shot 2020-04-28 at 08 36 33

This has been checked in IE, Chrome, Firefox and Safari.


ryan-mahoney:
@alicia-mbta LGTM, have you verified that this change doesn't have any side-effect in Chrome/FF?

alicia-mbta:
I checked Chrome, Firefox and Safari and didn't notice anything strange. I can double check and will also tag Ingrid.

ingridpierre:
@alicia-mbta looks good to me! 👍 Also, no issues observed in other browsers, also checked safari iOS.

PR #459, opened 2020-04-29
Add sticky table header to the CR timetable

thecristen:
#### Summary of changes
**Asana Ticket:** [CR Timetable | Header difficult to track](https://app.asana.com/0/385363666817452/950408339997961/f)

---


apps/site/assets/js/timetable-stick.js
```diff
@@ -0,0 +1,124 @@
...
+const trainLabels = Array.from(
+  document.querySelector(".js-timetable-train-labels").children
+);
+const tableEl = document.getElementById("timetable");
+const tableBodyEl = document.querySelector(".m-timetable tbody");
+const stickyHeaderEl = document.createElement("div");
+stickyHeaderEl.setAttribute(
+  "style",
+  "border: 1px solid #b0b5c0; overflow: hidden; position: fixed; top: 0;"
+);
```
alicia-mbta, line 15: Could we use `stickyHeaderEl.className = [class name]` to "extract" this styling or it doesn't really make sense here?

ingridpierre:
@thecristen Interesting update! This approach definitely is working much smoother for me. A couple things I'd like to try:

- Can we have the leftmost floating column with the list of stops display above or on top of the sticky header? the fact that the header labels train columns that are covered feels odd.

- Can we keep the same font size for train #s in the sticky header? The slightly larger font size almost creates an animated effect that is a little distracting when the table header goes sticky.

? - If the sticky header is another faux header element, will we need to hide it from screenreaders (so they can traverse the more semantic, basic table below)?

- Sticky header is still showing up for me and flashing while i scroll past the table extents:

thecristen:
@ingridpierre So.. the leftmost column _should_ be showing above... I'm curious what browser/device you're experiencing this in?

As to the other notes, yes:

- [ ] I can make the font size consistent
- [ ] The screenreader question is a great point; I think it'll be safer to hide it so I'll go do that.
- [ ] I am also seeing the occasional weirdness when scrolling past the table, so I'll look into that

apps/site/assets/js/timetable-stick.js
thecristen: Ha, that was my initial approach!  I ultimately decided that since I don't anticipate using this styling anywhere else, I may as well inline it and leave our spreadsheets alone. 🤷‍♀️

alicia-mbta: Ok sounds good.

ingridpierre:
@thecristen that was in Chrome, but I cleared my cache and it works now :-/ sorry!

thecristen:
This is ready for re-review.

ingridpierre:
This is _gawjus_. Changes look great 👍

alicia-mbta:
Wow you're doing almost everything programmatically (that is so cool), I guess because of the need to listen to the events.
I didn't know you could do it like that so I learned so much from this PR.
I pulled your latest changes and it looks great.

thecristen:
[My initial approach](https://github.com/mbta/dotcom/pull/454) was more focused on achieving the desired effects with CSS, but it was more complicated and not as nice of a result... so I'm glad I decided to try a more programmatic approach. Thanks y'all!

PR #460, opened 2020-04-29
Highlight current service with "Now" label

thecristen:
#### Summary of changes
**Asana Ticket:** [Services Picker | Highlight current service with "Now" label](https://app.asana.com/0/385363666817452/1170321197668881/f)

If there is a current service, this change adds a (Now) indicator to the relevant option in the services picker. 

---

### DIFF
--- a/apps/site/assets/css/_select-custom.scss
+++ b/apps/site/assets/css/_select-custom.scss
@@ -4,6 +4,7 @@
   border: 0;
   overflow: hidden;
   padding-right: $base-spacing * 2;
+  text-overflow: ellipsis;
   white-space: nowrap;
   width: 100%;
 
--- a/apps/site/assets/ts/schedule/__tests__/ServiceOptGroupTest.tsx
+++ b/apps/site/assets/ts/schedule/__tests__/ServiceOptGroupTest.tsx
@@ -13,6 +13,7 @@ describe("ServiceOptGroup", () => {
           label={"Group label"}
           services={services}
           multipleWeekdays={false}
+          now=""
         />
       )
       .toJSON();
@@ -22,7 +23,12 @@ describe("ServiceOptGroup", () => {
   it("returns null if no services", () => {
     createReactRoot();
     const tree = renderer.create(
-      <ServiceOptGroup label={""} services={[]} multipleWeekdays={false} />
+      <ServiceOptGroup
+        label={""}
+        services={[]}
+        multipleWeekdays={false}
+        now=""
+      />
     );
     expect(tree.toJSON()).toBeNull();
   });
--- a/apps/site/assets/ts/schedule/__tests__/__snapshots__/ServiceSelectorTest.tsx.snap
+++ b/apps/site/assets/ts/schedule/__tests__/__snapshots__/ServiceSelectorTest.tsx.snap
@@ -44,6 +44,7 @@ Array [
             value="BUS319-P-Sa-02"
           >
             Saturday schedule, Test
+             (Now)
           </option>
           <option
             value="BUS319-Q-Su-02"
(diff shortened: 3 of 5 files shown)


ingridpierre:
**@thecristen One quick thing first:  looks like the description text in the API is using sentence case. To match let's update to say "now" instead of "Now".**

-----
**Less quick thing:**
Open to your thoughts on this because functionally this is pretty complete and works well for all modes, but I'm wondering if theres an opportunity to add some polish here. 

My main concern is the reduced service description is so long that things are getting cut off in the select input (it mostly fits in the actual picker/dropdown after the click though). Do you think it's worth filing a ticket with the trc team to shorten the description? See below images for what I mean:

**414w device minimum for full label:**
image

**at 320w:**
image

**at 320w with optimized text:**
image

I ask because the other reason it's a struggle to fit is not just that the description is long, but because we currently set the input text size to 16px to avoid how iOS Safari zooms in when you interact with inputs with small text. So the text size is relatively gigantic to start here. Any thoughts on how we might approach an adjustment?

thecristen:
> Do you think it's worth filing a ticket with the trc team to shorten the description?

@ingridpierre  Nah, if only because this text truncating problem also manifests elsewhere, such as in the schedule finder inputs since direction and stop names can be somewhat verbose. It probably needs a more comprehensive design tweak. :-\ If anything we could at least handle the cut off text a bit more gracefully by truncating with an ellipsis (knowing that the user should be able to read the complete option when opening the `<select>`).

Another idea: Add the "Now" label to the start rather than end of the text, so at least it's always visible.

That said I do like how "Weekday reduced service` reads, but then the verbiage in the selector becomes a mix of `__ schedule` and `__ service`, which might be awkward.

ingridpierre:
If we can make the ellipsis appear dynamically according to what fits, rather than just arbitrarily cut it off after x characters, I'd be into that. If not or if it will take too much time, let's skip the polish for now and this will be a good feature to merge as soon as we can, given that it's solving a more immediate need for clarity on the site.

Took a look at moving now to the front, but I wasn't pleased with how it broke up scannability of the options. What you have now I think gets us closer to what we'd want in a longer term, non-emergency situation.

thecristen:
Added the bonus polish to enable truncating long text in our custom `<select>` elements with an ellipsis. This is implemented by the browser and is based on and responsive to the size of its container.

ingridpierre:
@thecristen currently I'm unable to observe your changes on green.dev. This is what I'm seeing in chrome (see screenshot). I'm not able to see the ellipsis, and the capitalization of Now seems to have returned?

Unfortunately, since reduced schedules are not available on green, it's making it tough to review #461 as well.

image

> Added the bonus polish to enable truncating long text in our custom `<select>` elements with an ellipsis. This is implemented by the browser and is based on and responsive to the size of its container.

paulswartz:
We've updated dev-green with the new GTFS which includes the reduced service schedules again.

ingridpierre:
@thecristen It looks like I can see the truncation working in Safari and FF, but not in Chrome.

I'd love to get this deployed in ahead of any GTFS changes showing the second spring rating, it will help reduce a bit of confusion in a universe with buggy/multiple services.


----
It just occurred to me that another place we do something like this is the transit-near-me page. When you click a stop to filter the results by stop, it shows the shortened stop name:
image

Example with the super long bus stop named "VFW Pkwy opp Wonderland Market Place"
http://localhost:4001/transit-near-me?_utf8=%E2%9C%93&location%5Baddress%5D=wonderland+boston+ma&location%5Blatitude%5D=&location%5Blongitude%5D=

ingridpierre:
@thecristen let's call this one good! 👍

PR #461, opened 2020-04-29
Get all services in the current rating into the current optgroup

thecristen:
#### Summary of changes

**Asana Ticket:** [Services Picker | Get all services in the current rating into the current optgroup](https://app.asana.com/0/385363666817452/1170658788930619/f)

Made a few adjustments to the service picker logic to achieve the following:

* Let the upcoming service in the current rating be categorized in the UI as "current"
* Order the service choices to align with the mockup

---

### DIFF
--- a/apps/site/assets/ts/helpers/__tests__/service-test.ts
+++ b/apps/site/assets/ts/helpers/__tests__/service-test.ts
@@ -7,7 +7,6 @@ import {
   isCurrentValidService,
   isInFutureRating,
   optGroupComparator,
-  hasIncompleteRating,
   isInFutureService
 } from "../service";
 import { Service, DayInteger } from "../../__v3api";
@@ -193,21 +192,6 @@ describe("groupServicesByDateRating", () => {
     });
   });
 
-  it("lists only typical_service services in the relevant rating as 'current'", () => {
-    const currentKeys = Object.keys(grouped).filter(
-      key => key.split(" ")[0] === ServiceGroupNames.CURRENT.split(" ")[0]
-    );
-    currentKeys.forEach(currentKey => {
-      const currentServices = grouped[currentKey!];
-      currentServices.forEach(service => {
-        expect(service.typicality).toEqual("typical_service");
-        expect(isInCurrentRating(service, testDate)).toBe(
-          !hasIncompleteRating(service)
-        );
-      });
-    });
-  });
-
   it("lists only holiday_service services as 'holiday'", () => {
     grouped[ServiceGroupNames.HOLIDAY].forEach(service => {
       expect(service.typicality).toEqual("holiday_service");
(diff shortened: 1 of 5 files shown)


ingridpierre:
Nice! I appreciate the sensible ordering with weekday first, glad to see that added here.

A couple things: 
Looks like we also need to account for the longer rating on the Commuter Rail:
image

Also, some considerations for Ferry, I suppose these should all be "current" as well though. They change annually around July? Probably doesn't happen on a fixed date, but we should hopefully leverage current v. future schedules correctly in that case.

image

thecristen:
Let's see if this works a bit better now for the CR and Ferry.

ingridpierre:
Ferry looks good, but CR is showing something new and different from last week that I'm not understanding well.

Here, see the duplicates - I think this is different than the known issue with [two spring ratings] but I don't know what's going on.(https://app.asana.com/0/385363666817452/1170658788930623/f). 

image

thecristen:
Looks like those last two current Saturday/Sunday schedules have a typicality of `planned_disruption`. Taking a closer look it looks like I might not be handling some _future services within the current rating_ correctly, so I'll see how I can revise my approach accordingly.

thecristen:
@ingridpierre Ready for rereview, some random observations:

76 was showing multiple similar services in the current rating, so we chose the first one (based on start date). I'm defining "similar" here as having the same `valid_days`, `typicality`, `type`,`rating_start_date`, `rating_end_date`, `name`, and `description`.

(now) was incorrect for the situation where there's no active service, e.g. Ferry. This is now corrected.

The default selected service on the Ferry is Sunday's, but that's because it's technically "the next available service" based on its start date (since it starts the day before the weekday one). I don't like it but maybe we can tackle that another time.

ingridpierre:
Just making sure I have this right, can you verify the below is expected? All of these examples are potentially fine, but I would have to significantly update the design documentation because it's quite different from what I was expecting.

To summarize, the current design says that all modes will have some opt group called "Current Schedules (ends on [rating end date])".  But as we've run into different issues by mode, only bus has taken on this approach exactly. CR and Ferry don't show a rating end date, and are either grouped in current or future by some other logic. Can you help me understand what the logic is doing differently should we be notating important differences/exceptions in this component?

-----
Examples to verify I'm seeing what you are:

1) **Bus looks great, shows the rating end dates and everything grouped correctly.**
image


2) Commuter rail has no more duplicates, but the **weekday schedule is showing into the future** there is no date indicating the end of the current rating:
image

3) Ferry has no rating end date, all services are in the future, nothing current:

image


4) Interlined buses like the 62/76 are extremely complicated (especially here where they don't both run on the same days of the week). I think what's here still helps someone know what service is happening, but I might file a ticket for us to investigate interlined bus schedules as a whole. These routes are currently suppressed on the main schedules page /schedules list, so it's less pressing. That being said it does exist and get indexed by google,  thank you so much for addressing the issue here so far! I'll have to on making deeper sense of this at a later time.
image

thecristen:
Yeah! Those all are what I see. Aaaand now I realize there's some less than ideal logic here that was a holdover of earlier iterations. I'd been handling services with/without a rating_end_date separately but not similarly enough, thus producing different results for bus vs CR. Whoops.

Just pushed a change to _always_ use the rating dates to determine "future" service (even when there is a null rating_end_date, e.g. for ferry and commuter rail). Sorry to keep you coming back to this but I think this finally gets it? 😬

image

PR #463, opened 2020-05-01
Add test_vcr ExVCR helper

phildarnowsky:
We currently have roughly 3,000 tests, many of which use live data from
the API and/or CMS, which practice we are trying to put an end to. The
ExVCR package allows us to record interactions with remote services for
later use as test fixtures. As such, it's a straightforward way to
eliminate live data use from a test.

However, since we have so many tests, it's important to reduce the
amount of boilerplate involved in converting tests to use ExVCR. This
commit adds a `test_vcr` macro meant to be a dropin replacement for the
normal ExUnit `test` macro, which automatically adds ExVCR use to a test
with no explicit configuration needed.

No ticket. This is something I've had going on the side for a while, and I want to get it merged before the summer rating hits dev, so that we can use it for that.

---


apps/exvcr_helpers/README.md
```diff
@@ -0,0 +1,21 @@
+# ExVCRHelpers
+
+**TODO: Add description**
```
thecristen, line 3: Are you going to add a description?

phildarnowsky:
@thecristen I added a description and a smidgen of instructions. A test happened to break due to a data change over the weekend too, so I threw that in as well.

PR #464, opened 2020-05-04
Prevent error when timetable element not present

thecristen:
Quick Javascript fix for handling when `document.querySelector(".js-timetable-train-labels")` is `undefined`.


thecristen:
@phildarnowsky Yup.

PR #467, opened 2020-05-05
Fix test broken by change to stops live data

phildarnowsky:
No ticket. Another live data fix nailed down with ExVCR.

---


apps/stops/fixture/vcr_cassettes/elixir_stops_apitest__parses_parent_id_and_child_ids.json
```diff
@@ -0,0 +1,242 @@
...
+      "type": "ok"
+    }
+  },
+  {
+    "request": {
+      "body": "",
+      "headers": {
+        "accept-encoding": "gzip",
+        "accept": "application/vnd.api+json",
+        "x-api-key": "1c05fc0118144698888cc5a1b08d3a56"
```
paulswartz, line 48: This shouldn't be checked in. I'm going to disable this key and I'll share a different one with you.

Are there other API keys in the other cassettes?

phildarnowsky: @paulswartz it's supposed to filter that header out, and in fact in other places in this same cassette it does so--this is very odd. Possibly a bug in ExVCR?

phildarnowsky:
@paulswartz corrected the pertinent lines manually. I'm concerned about ExVCR not consistently blanking out that key, will open a ticket to look into that.

alicia-mbta:
A couple notes:
- From my investigation so far, it seems that ExVCR expects a 'default' configuration, so I got the filtering of the headers to work by adding `config :exvcr, filter_request_headers: ["x-api-key"]` in file `dotcom/config/config.exs`. My understanding is that this `config.exs` file is for the whole application so I'm not sure if this change is an issue or not.

- This change successfully filtered the "x-api-key" values in the cassettes but I still can't get the vcr test in file `dotcom/apps/site/test/site_web/controllers/fare_controller_test.exs` to pass. I'm going to continue digging.

PR #468, opened 2020-05-07
Normalized addresses in trip planner

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🗺️ Trip Planner | Use normalized, not raw, addresses in output](https://app.asana.com/0/555089885850811/1158824251239697)

To make sure "from" and "to" locations contain real values, a small pre-processing is done to properly show said locations in a more normalized way.

Before:
image
<br/>
After:
image

### DIFF
--- a/apps/site/lib/site_web/controllers/trip_plan_controller.ex
+++ b/apps/site/lib/site_web/controllers/trip_plan_controller.ex
@@ -1,4 +1,7 @@
 defmodule SiteWeb.TripPlanController do
+  @moduledoc """
+  Page for the Trip Planner. Manages the itineraries.
+  """
   use SiteWeb, :controller
   alias Routes.Route
   alias Site.TripPlan.{Query, RelatedLink, ItineraryRow, ItineraryRowList}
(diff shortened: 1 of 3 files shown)


apps/site/lib/site_web/controllers/trip_plan_controller.ex
```diff
@@ -59,7 +62,38 @@ defmodule SiteWeb.TripPlanController do
...
+    # same with normalized_to
+    last_mode = List.last(itinerary.legs).mode
+
+    normalized_to =
+      if Map.has_key?(last_mode, :steps) do
+        List.last(last_mode.steps).street_name
+      else
+        Map.get(plan, "to")
+      end
+
```
alicia-mbta, line 95: I can probably create a function here in order to DRY.

ingridpierre:
**Paraphrasing our conversation over Meet:**

We think ultimately this ticket is blocked by a deeper investigation and probably some other work around how trip planner resolves addresses coming from different places: trip planner user input, algolia/google places autocomplete suggestions, and whatever ends up coming from OpenTrip Planner.

Taking a look we found many cases where this change seems to adds more confusion rather than clarity, and highlighted some existing bugs. We'll need to put our heads together to solve this one a bit more reliably.

alicia-mbta:
Hi @ingridpierre, would you mind taking another look at this? I took a different approach recommended by Paul S. and I think the results look better now. Coreanos' address gets identified so that's a start.

paulswartz:
This works, although it isn't quite what I meant. I think the Javascript address library on the client returns a nicely formatted name; can we use that when submitting an address? That might keep us from having to make duplicate geocoding requests.

arkadyan:
Closing for now until we are ready to come back to it.

PR #471, opened 2020-05-12
Update bus connection tooltips

thecristen:
**Asana Ticket:** [🐞Stations/Stops | Tooltips for bus connections on map are wrong](https://app.asana.com/0/385363666817452/1147672918652027/f)

Because `<StopCard />` already has a `routesWithDirection` prop to render direction-specific routes, this change lets the `<StopMap />` make use of it. This is done by generating the requisite data from the stop controller and passing it through via `conn.stop_page_data` to `<StopPage />` to `<StopMapContainer />` to `<StopMap />` to `<MapTooltip />` (which will finally pass it to `<StopCard />`).

~The controller code is cobbled together with useful bits of code from elsewhere.~ I expose the `Nearby.merge_routes` function so that I can use it to get directions associated with the stop's routes. ~Much of the rest of the code involves route sorting in order to preserve the current sorting done in the map tooltip (subways first in a certain order, then bus with Silver Lines and CT* buses first, then CR...), otherwise the connections order seem very arbitrary.~ 

This also adds a new attribute from the API into Dotcom's routes - `sort_order` - and uses it to sort the resulting routes.

Note this change will impact stops where there are buses going in only one direction - e.g. it will not change the displayed tooltips at terminals or those big stations where the bus line passes through both ways, as it stands right now.

### DIFF
--- a/apps/routes/lib/route.ex
+++ b/apps/routes/lib/route.ex
@@ -8,6 +8,7 @@ defmodule Routes.Route do
             name: "",
             long_name: "",
             color: "",
+            sort_order: 99_999,
             direction_names: %{0 => "Outbound", 1 => "Inbound"},
             direction_destinations: :unknown,
             description: :unknown,
@@ -20,6 +21,7 @@ defmodule Routes.Route do
           name: String.t(),
           long_name: String.t(),
           color: String.t(),
+          sort_order: non_neg_integer,
           direction_names: %{0 => String.t() | nil, 1 => String.t() | nil},
           direction_destinations: %{0 => String.t(), 1 => String.t()} | :unknown,
           description: gtfs_route_desc,
@@ -207,6 +209,7 @@ defmodule Routes.Route do
         name: name,
         long_name: long_name,
         color: color,
+        sort_order: sort_order,
         direction_names: direction_names,
         direction_destinations: direction_destinations,
         description: description,
@@ -226,6 +229,7 @@ defmodule Routes.Route do
       name: name,
       long_name: long_name,
       color: color,
+      sort_order: sort_order,
       direction_names: %{
         "0" => direction_names[0],
         "1" => direction_names[1]
--- a/apps/site/assets/ts/stop/__tests__/StopCardTest.tsx
+++ b/apps/site/assets/ts/stop/__tests__/StopCardTest.tsx
@@ -1,0 +1,93 @@
+import React from "react";
+import { mount, ReactWrapper } from "enzyme";
+import StopCard from "../components/StopCard";
+import { EnhancedRoute, DirectionId } from "../../__v3api";
+import stopData from "./stopData.json";
+import {
+  StopPageData,
+  TypedRoutes,
+  RouteWithDirection
+} from "../components/__stop";
+
+const { stop, routes } = JSON.parse(JSON.stringify(stopData)) as StopPageData;
+// mimic the transformation done for the StopMapContainerProp
+const typedRoutes = routes.reduce(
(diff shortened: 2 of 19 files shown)


ingridpierre:
Beauteous! Looks like the tooltip is a doing a better job at sorting than the old Transfers Nearby. Compare the nicely sorted tooltip you've got at / stops/place-dudly :
image

...to the corresponding one for it on this other stop's Suggested Transfers Nearby: 
http://localhost:4001/stops/64002

image

Possible to add the proper sorting to Suggested Transfers stop cards so they match? Or should we call that a separate ticket?

apps/routes/lib/route.ex
```diff
@@ -8,6 +8,7 @@ defmodule Routes.Route do
             name: "",
             long_name: "",
             color: "",
+            sort_order: :unknown,
```
arkadyan, line 11: It looks like this isn't currently an allowed value according to the typespec. But does the API always return a `sort_order ` value for every route? Maybe we can just not specify a default value here?

apps/stops/lib/nearby.ex
```diff
@@ -110,7 +110,7 @@ defmodule Stops.Nearby do
   def do_routes_for_stop_direction(_, _), do: :error
 
   @spec merge_routes(String.t(), fun()) :: {:ok | :error, [route_with_direction] | :timeout}
-  defp merge_routes(stop_id, routes_fn) do
+  def merge_routes(stop_id, routes_fn) do
```
arkadyan, line 113: If this is now a public function used elsewhere it would probably be good to have tests covering it.

apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -444,4 +455,8 @@ defmodule SiteWeb.StopController do
 
     %{conn | assigns: assigns}
   end
+
+  defp route_sorter(%{route: route}) do
+    {Routes.Group.sorter({Route.type_atom(route), nil}), route.sort_order}
```
arkadyan, line 460: Tiny thing: `Routes.Group` has an alias, so this could just be `Group.sorter`.

arkadyan: There's an existing `grouped_routes` function in this module that calls `Group.sorter`. Would we want to update that to use this better sorter as well?

apps/site/assets/ts/stop/__tests__/StopCardTest.tsx
```diff
@@ -0,0 +1,93 @@
+import React from "react";
```
arkadyan, line 1: ❗ Thank you for adding this! Testing components seems pretty important.

apps/routes/lib/route.ex
thecristen: I'll supply a better default value...

apps/site/lib/site_web/controllers/stop_controller.ex
thecristen: I tried, it was a bit weird given that that function also groups the results. Ultimately I decided I might not need to have a new `route_sorter` function now that I have a `sort_order` attribute to use!

thecristen:
Made a few changes. @ingridpierre, it looks like I'm able to sort out the suggested transfers too, so feel free to take another peek!

ingridpierre:
@thecristen looks great! 👍 thanks!

arkadyan:
Rebasing should hopefully fix that `Stops.ApiTest` test.

PR #473, opened 2020-05-14
Get services for the merged Green Line

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Schedule Finder | No scheduled trips found from the merged Green line page](https://app.asana.com/0/555089885850811/1174356662552522)

The logic to retrieve the services for the merged Green line was not receiving the parameter in the right format.
Also tests were added and a pre-processing of trips.<br/>
**Before**:
image

**After**:
image

### DIFF
--- a/apps/services/lib/repo.ex
+++ b/apps/services/lib/repo.ex
@@ -14,8 +14,7 @@ defmodule Services.Repo do
   end
 
   def by_route_id("Green", params) do
-    GreenLine.branch_ids()
-    |> Enum.join(",")
+    "Green-B,Green-C,Green-D,Green-E"
     |> by_route_id(params)
   end
 
--- a/apps/services/lib/repo.ex
+++ b/apps/services/lib/repo.ex
@@ -1,5 +1,7 @@
 defmodule Services.Repo do
-  @moduledoc false
+  @moduledoc """
+  Retrieves services for a route.
+  """
 
   use RepoCache, ttl: :timer.hours(1)
   alias Services.Service
@@ -9,6 +11,12 @@ defmodule Services.Repo do
 
   def by_route_id([route_id] = route, params) when is_list(route) do
     by_route_id(route_id, params)
+  end
+
+  def by_route_id("Green", params) do
+    GreenLine.branch_ids()
+    |> Enum.join(",")
+    |> by_route_id(params)
   end
 
   def by_route_id(route_id, params) when is_binary(route_id) do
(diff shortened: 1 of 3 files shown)


apps/services/lib/repo.ex
```diff
@@ -11,6 +13,11 @@ defmodule Services.Repo do
     by_route_id(route_id, params)
   end
 
+  def by_route_id("Green", params) do
+    expanded_green_routes = Enum.join(GreenLine.branch_ids(), ",")
```
ryan-mahoney, line 17: @alicia-mbta could this be:
```
GreenLine.branch_ids()
|> Enum.join(",")
|> by_route_id(params)
```

apps/site/lib/site_web/channels/vehicle_channel.ex
```diff
@@ -42,6 +45,14 @@ defmodule SiteWeb.VehicleChannel do
     stop_name = get_stop_name(vehicle.stop_id)
     trip = Schedules.Repo.trip(vehicle.trip_id)
 
+    # Sometimes trip is nil, so to avoid errors, we create a "polished" trip:
+    preprocessed_trip =
+      if trip == nil do
```
arkadyan, line 50: This is totally fine, but just to mention another option: `nil` is falsy, so you can also do something like `trip || %{shape_id: nil}`.

apps/site/lib/site_web/channels/vehicle_channel.ex
```diff
@@ -51,14 +62,14 @@ defmodule SiteWeb.VehicleChannel do
           id: vehicle.id,
           icon: "vehicle-bordered-expanded",
           rotation_angle: vehicle.bearing,
-          shape_id: trip.shape_id,
+          shape_id: preprocessed_trip.shape_id,
```
arkadyan, line 65: Following up on my comment above, using an inverse of the same trick you can also do something like `trip && trip.shape_id` here and avoid having to make the `preprocessed_trip` in the first place.

apps/site/lib/site_web/channels/vehicle_channel.ex
```diff
@@ -51,7 +54,7 @@ defmodule SiteWeb.VehicleChannel do
           id: vehicle.id,
           icon: "vehicle-bordered-expanded",
           rotation_angle: vehicle.bearing,
-          shape_id: trip.shape_id,
+          shape_id: (trip && trip.shape_id) || nil,
```
arkadyan, line 57: `trip && trip.shape_id` will return `nil` if `trip` is `nil`, so the ` || nil` is unecessary.

    iex(1)> trip = nil
    nil
    iex(2)> trip && trip.shape_id
    nil

PR #476, opened 2020-05-18
Map & trip details accordion with total itinerary times

thecristen:
#### Summary of changes
**Asana Ticket:** [🎫 Style update | Map & trip details accordion](https://app.asana.com/0/385363666817452/1162903717356817/f) and [🎫 Style update | Add total itinerary time to summary section](https://app.asana.com/0/385363666817452/1169236529454065/f)

Updated itinerary styles to accommodate the upcoming fares display!

---

### DIFF
--- a/apps/site/assets/css/_trip-plan-results.scss
+++ b/apps/site/assets/css/_trip-plan-results.scss
@@ -1,4 +1,4 @@
-.m-trip-plan-results__itinerary-container {
+.m-trip-plan-results__itinerary {
   margin-bottom: $base-spacing * 1.5;
 
   h2 {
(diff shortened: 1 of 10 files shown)


arkadyan:
I don't know if there something screwed up with my machine, but when I try to run the branch and get trip plans I'm seeing a messed up page layout.

Screen Shot 2020-05-19 at 11 19 15

apps/site/assets/css/_trip-plan-results.scss
```diff
@@ -1,27 +1,34 @@
 .m-trip-plan-results__itinerary-container {
   margin-bottom: $base-spacing * 1.5;
 
+  h2 {
+    font-size: 1.2em;
+    margin-top: 0;
+  }
+
   .c-expandable-block__header {
```
arkadyan, line 9: Is `c-expandable-block__header` maybe not so relevant now that this is more of a block above an accordian? I wonder if it makes sense to rename it—you'd have to reproduce some of the styling, but you'd also no longer have to override other styles (like the cursor).

apps/site/assets/ts/components/Accordion.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { ReactElement } from "react";
+
+// A simple accordion implementation, based on site_web/templates/partial/_accordion_ui
+interface Props {
+  id: string;
+  title: string;
+  children: ReactElement<HTMLElement>;
+}
+
+const Accordion = (props: Props): ReactElement<HTMLElement> => {
```
arkadyan, line 10: Could you please add a quick test for this component, just rendering it to a snapshot?

apps/site/lib/site_web/templates/trip_plan/_itinerary_tab.html.eex
```diff
@@ -1,17 +1,27 @@
...
       <%= Timex.format!(@itinerary.start, "{h12}:{m} {AM}") %> - <%= Timex.format!(@itinerary.stop, "{h12}:{m} {AM}") %>
     </span>
+    <span>
+      <% duration =  Timex.diff(@itinerary.stop, @itinerary.start, :minutes) %>
+      <%= if duration >= 60 do %>
+        <%= duration |> div(60) %> hr
+        <%= duration |> rem(60) %> min
+      <% else %>
+        <%= duration |> rem(60) %> min
+      <% end %>
```
arkadyan, line 18: If you wanted, this section feels like a candidate for refactoring into a function in `TripPlanView`, maybe called `format_duration/2` or something along those lines. It would allow you to test that function and would make this template simpler and more readable.

thecristen:
Ahh I was getting that page layout at the end of yesterday too and decided to sleep on it, was hoping it was just a fluke! I can't for the life of me figure out what could be causing it - maybe we can figure it out together @arkadyan

ingridpierre:
The map in the accordion is rendering a bit strangely. Cleared cache, tried different trips and I seem to always see the map looking like this:
image

thecristen:
Thanks for flagging that; I was able to identify the issue (needed to trigger an extra render) and have pushed a fix for the map.

ingridpierre:
@thecristen looks great! 👍

PR #477, opened 2020-05-19
Get additional teasers from mode

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [⏱️Schedules | Display news entries for the specific route AND its parent mode](https://app.asana.com/0/555089885850811/1169031641902619)

Besides retrieving news/teasers from the specific line, we also want to retrieve the ones tagged with its mode to make content editors' lives easier. The changes in this PR retrieve those, removes duplicates and sorts them by date descending as requested.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/cms.ex
+++ b/apps/site/lib/site_web/controllers/schedule/cms.ex
@@ -34,9 +34,26 @@ defmodule SiteWeb.ScheduleController.CMS do
     end
 
     news_fn = fn ->
-      [route_id: route.id, type: [:news_entry], sidebar: 1]
-      |> Repo.teasers()
-      |> Enum.map(&set_utm_params(&1, route))
+      teasers =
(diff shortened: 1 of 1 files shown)


apps/site/lib/site_web/controllers/schedule/cms.ex
```diff
@@ -19,6 +19,35 @@ defmodule SiteWeb.ScheduleController.CMS do
   @impl Plug
   def init([]), do: []
 
+  @spec get_mode_from_route_description(Route.gtfs_route_desc()) :: String.t() | nil
+  defp get_mode_from_route_description(description) do
```
arkadyan, line 23: This is a real nitpick, but could you please move this function lower? In Elixir, I'm a big adherent of the convention of putting public functions at the top of the file and the private functions below. Personally, I find it helps make the file easier to read: I can see the general things the module does quickly, and only dig down into the details if I'm interested.

PR #479, opened 2020-05-20
Get highest and lowest fare

alicia-mbta:
#### Summary of changes
Asana ticket: [🎫 Trip Planner | Determine highest and lowest fares in Fares app](https://app.asana.com/0/555089885850811/1175251002332392)

- Moved and renamed apps/site/lib/site/base_fare.ex -> apps/fares/lib/min_max_fare.ex
- Added the logic and tests to obtain the highest fare analogously as it is done for the lowest fare

### DIFF
--- a/apps/fares/lib/min_max_fare.ex
+++ b/apps/fares/lib/min_max_fare.ex
@@ -1,8 +1,6 @@
-defmodule Site.BaseFare do
+defmodule Fares.MinMaxFare do
   @moduledoc """
-  Calculates the "base fare" for a particular trip.
-  The base fare is a regular priced, one-way fare for the given mode.
-  If there are multiple fare media, the lowest priced is chosen.
+  Calculates the lowest and highest fare for a particular trip i.e. a regular priced, non-discounted, one-way fare for the given mode.
   Commuter rail and ferry fares distinguish between the possible sets of stops.
   Bus fares for express buses do not distinguish between the local and express portions;
   the express fare is always returned.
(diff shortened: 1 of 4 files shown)


apps/fares/lib/min_max_fare.ex
```diff
@@ -17,22 +15,32 @@ defmodule Site.BaseFare do
...
-  def base_fare(route, nil, origin_id, destination_id, fare_fn) do
-    base_fare(route, @default_trip, origin_id, destination_id, fare_fn)
+  def lowest_fare(route, nil, origin_id, destination_id, fare_fn) do
+    lowest_fare(route, @default_trip, origin_id, destination_id, fare_fn)
   end
 
-  def base_fare(route, trip, origin_id, destination_id, fare_fn) do
+  def lowest_fare(route, trip, origin_id, destination_id, fare_fn) do
+    get_fares(route, trip, origin_id, destination_id, fare_fn)
+    |> Enum.min_by(& &1.cents, fn -> nil end)
```
arkadyan, line 35:
```suggestion
    route
    |> get_fares(trip, origin_id, destination_id, fare_fn)
    |> Enum.min_by(& &1.cents, fn -> nil end)
```

This is a real minor style thing, but in Elixir there's a general convention of starting pipe chains with data, not with a function call. In practice that generally just means pulling out the first argument to your first function in the chain.

alicia-mbta: Yes, after looking at the code it seemed like this could be a simple enough solution. I didn't add the tests because I wasn't sure this was a good approach. I'll do that now and will also address your change requests.

alicia-mbta:
> @alicia-mbta I think this is looking terrific! It would be nice to get a few tests covering the `highest_fare` function.
> 
> It's nice to see that this update is so relatively straightforward in the end. Is it feeling like a good approach to you?

Yes, after looking at the code it seemed like this could be a simple enough solution. I didn't add the tests (yet) because I wasn't sure this was a good approach. I'll do that now and will also address your change requests.

PR #481, opened 2020-05-21
Add empty state for <UpcomingDepartures />

thecristen:
#### Summary of changes
**Asana Ticket:** [🚍Schedule Finder | Upcoming Departures | Add empty state when no predictions available](https://app.asana.com/0/555089885850811/1173752461829233)

---

### DIFF
--- a/apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
@@ -135,6 +135,22 @@ const TableRow = ({
   );
 };
 
+const UpcomingDeparturesHeader = (
+  <div className="schedule-table__upcoming-departures-header">
+    <h3>Upcoming Departures</h3>
+    <span className="schedule-table__live-clock" aria-hidden>
+      <span className="icon-realtime animate notranslate">
+        <span
+          className="c-svg__icon-live-clock"
+          // eslint-disable-next-line react/no-danger
+          dangerouslySetInnerHTML={{ __html: liveClockSvg }}
+        />
+        <span className="icon-realtime-text">live</span>
+      </span>
+    </span>
+  </div>
+);
+
 export const UpcomingDepartures = ({
   state,
   input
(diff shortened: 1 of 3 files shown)


apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
```diff
@@ -135,30 +135,33 @@ const TableRow = ({
...
-                dangerouslySetInnerHTML={{ __html: liveClockSvg }}
-              />
-              <span className="icon-realtime-text">live</span>
-            </span>
-          </span>
-        </div>
+  return (
+    <>
+      {UpcomingDeparturesHeader}
+      {journeys !== null && hasPredictions(journeys) ? (
```
alicia-mbta, line 164: Is `UpcomingDeparturesHeader` supposed to be always shown or only when `journeys !== null && hasPredictions(journeys)` as in the previous version of the file? i.e. should lines 163 and 164 be swapped?

thecristen: The design called for it being always visible!

alicia-mbta: Disregard then 😅

ingridpierre:
The routes and stops that I would expect to show an empty state are not all behaving as hoped.

Stops that we never have predictions for like non-MBTA operated 700s _do_ show the empty state. Routes with shuttled stops like Lechmere, do show the empty state. Some non-running school variants [like this one on the 39](http://localhost:4001/schedules/39/line?from=search-route&query=39&schedule_direction%5Bdirection_id%5D=0&schedule_direction%5Borigin%5D=117&schedule_direction%5Bvariant%5D=39-_-0) do show the empty state.

Some stops I found that aren’t showing the empty state:

**Some routes with no service on weekends, therefore no service during reduced schedules**
CT2, CT3, Route 4, Silver Line Waterfront ([example url](https://www.mbta.com/schedules/708/line?schedule_direction%5Bdirection_id%5D=0&schedule_direction%5Borigin%5D=place-andrw))
image

**Stops on some variants without service:**
For example, Route 16 (variant isn’t running), empty state does not show. 

image

thecristen:
> Stops that we never have predictions for like non-MBTA operated 700s _do_ show the empty state. Routes with shuttled stops like Lechmere, do show the empty state. Some non-running school variants like this one on the 39 do show the empty state.

@ingridpierre , can you clarify which of these situations are desired vs undesired? I'm guessing the first one is problematic, and the second and third are not?

ALSO: I figured out what might be part of the problem -- the whole component doesn't show when there's no service today (hence you're seeing the "There are no scheduled trips for [date]" message).

Basically the logic goes like this right now: 

- Is there service today?
  - No: Show _no schedule trips_ message
  - Yes: Show upcoming departures component
    - (inside upcoming departures) Are there departures?
       - No: Show _no realtime departures_ message
       - Yes: Show table with live departure data

I'm inclined to think this logic makes sense because if there's no service today then there shouldn't be any realtime departure data, but if it is producing surprising results we should definitely revisit it.

But that only explains the routes with no service -- I'll take another look at the 16.

ingridpierre:
> Stops that we never have predictions for like non-MBTA operated 700s _do_ show the empty state. Routes with shuttled stops like Lechmere, do show the empty state. Some non-running school variants like this one on the 39 do show the empty state.

These are all good things, empty state should and does show.

ingridpierre:
> I'm inclined to think this logic makes sense because if there's no service today then there shouldn't be any realtime departure data,

You're 100% right, thanks for that helpful explanation. I think I was focused on "this is the empty state feature" last night, but we don't need two empty states if there is no service to speak of!

ingridpierre:
@thecristen lgtm 👍 can't reproduce rte 16 bug.

PR #482, opened 2020-05-22
Display arrival times for a chosen destination

arkadyan:
#### Summary of changes
**Asana Ticket:** [Schedule Finder | Arrivals](https://app.asana.com/0/555089885850811/1158222420890001)

Added the ability to select a destination stop in the schedule finder to see arrival times.

---


arkadyan:
@ingridpierre Could you please take a look at this next week and let me know if it is generally acting the way you were hoping for? I just kicked off a deploy to dev-green.

ingridpierre:
Thanks for the deploy, it's helping me see a lot of issues in the original design and how it was described. While it looks and works as I had hoped in desktop, on small mobile devices my concerns about the way I had decided to implement the input are growing.

I had tried to avoid needing a third input at the top of the modal, but we may need to switch to something closer to what we had previously on the old schedules tab, similar to: 
image

I'm also concerned with the way all the columns + train numbers interact on commuter rail lines as well. I'm glad you've got this staged so I can really take another look and see if there are some optimizations we might need to make before this feature is really ready for release. 

Screen Shot 2020-05-27 at 12 35 43 AM

The documentation is also scant on a feature I mistakenly thought was supported in the old schedule tab. The list of stops in the accordion should end at the selected arrival, and the fares info and total stops and trip time should reflect the selected arrival as well.

image

**Overall though, I feel I owe this a deeper look as the mobile experience was not well considered in my original design. I would also be happy to chat this one through in person just to reorient about possible approaches that build on the great work that you/cristen have already done here.**

arkadyan:
Closing for now to give time for further design thinking.

PR #483, opened 2020-05-22
Update human labels for ferry routes

amaisano:
#### Summary of changes
**Asana Ticket:** [Update display text for ferry fare cards](https://app.asana.com/0/399597520748778/1175281228582685)

Commuter Ferry + Inner Harbor Ferry updated

image

---


apps/fares/lib/format.ex
```diff
@@ -83,10 +83,10 @@ defmodule Fares.Format do
   def name(:local_bus), do: "Local Bus"
   def name(:inner_express_bus), do: "Inner Express Bus"
   def name(:outer_express_bus), do: "Outer Express Bus"
-  def name(:ferry_inner_harbor), do: "Inner Harbor Ferry"
+  def name(:ferry_inner_harbor), do: "Charlestown Ferry"
```
alicia-mbta, line 86: Would it make sense to rename the atoms as well not only the strings?

amaisano: I think the atoms should resemble the GTFS system route names more than human label. That being said, not sure what those are off the top of my head but I assumed the current atoms lined up with GTFS.

alicia-mbta: Ah ok disregard then

aayesha:
Sorry for the late review. Looks good!

PR #484, opened 2020-05-26
Add highest and lowest fares to each leg of an itinerary

arkadyan:
#### Summary of changes
**Asana Ticket:** [🎫 Trip Planner | Compile fare information for an itinerary](https://app.asana.com/0/555089885850811/1175370216290833)

Add highest and lowest fares to each leg of an itinerary. Not used anywhere yet, but will be in follow-up work.

---

### DIFF
--- a/apps/site/lib/site_web/controllers/trip_plan_controller.ex
+++ b/apps/site/lib/site_web/controllers/trip_plan_controller.ex
@@ -1,9 +1,13 @@
 defmodule SiteWeb.TripPlanController do
+  @moduledoc """
+  Controller for trip plans.
+  """
+
   use SiteWeb, :controller
   alias Routes.Route
   alias Site.TripPlan.{Query, RelatedLink, ItineraryRow, ItineraryRowList}
   alias Site.TripPlan.Map, as: TripPlanMap
-  alias TripPlan.{Itinerary, Leg}
+  alias TripPlan.{Itinerary, Leg, PersonalDetail, TransitDetail}
 
   plug(:require_google_maps)
   plug(:assign_initial_map)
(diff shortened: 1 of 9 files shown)


arkadyan:
@alicia-mbta I've added a test and typespecs, and piped the data into the view for later use.

@thecristen pining you just in case you are interested in looking over this as well.

apps/site/lib/site_web/controllers/trip_plan_controller.ex
```diff
@@ -56,6 +65,37 @@ defmodule SiteWeb.TripPlanController do
...
+    leg
+  end
+
+  defp leg_with_fares(%Leg{mode: %TransitDetail{}} = leg) do
+    route = Routes.Repo.get(leg.mode.route_id)
+    trip = Schedules.Repo.trip(leg.mode.trip_id)
+    origin_id = leg.from.stop_id
+    destination_id = leg.to.stop_id
+    lowest_fare = Fares.MinMaxFare.lowest_fare(route, trip, origin_id, destination_id)
+    highest_fare = Fares.MinMaxFare.highest_fare(route, trip, origin_id, destination_id)
```
alicia-mbta, line 88: It looks great to me :smiley:
The only 'nitpicky' thing I was wondering was if the `MinMaxFare` file I created early on should be renamed to something else for consistency because in it `min` and `max` refer to `highest` and `lowest`. But this is not really a big deal.

arkadyan: 👍 sure, that's a good idea.

apps/site/test/site_web/controllers/schedule_controller_test.exs
```diff
@@ -204,19 +204,6 @@ defmodule SiteWeb.ScheduleControllerTest do
...
-          conn,
-          trip_view_path(conn, :show, "Green", origin: "place-north", vehicle_tooltips: %{})
-        )
-
-      for frequency <- conn.assigns.frequency_table do
-        assert frequency.min_headway > 0
-        assert frequency.max_headway > 0
-      end
-    end
-
```
alicia-mbta, line 219: Thanks for investigating this

PR #485, opened 2020-05-26
Update Trip Planner fares URLs for bus and subway

thecristen:
#### Summary of changes
**Asana Ticket:** [🎫 Trip Planner | Update fare information URLs](https://app.asana.com/0/555089885850811/1176023923687794)

Turns out these correct URLs are also [hardcoded](https://github.com/mbta/dotcom/blob/c917e4799930b7636a178f01cb1fabc2d8216a97/apps/site/lib/site_web/views/schedule_view.ex#L426) in a [number](https://github.com/mbta/dotcom/blob/98a19925be72ff28a9cffd797b774d1d2ca046bc/apps/site/lib/site_web/views/fare_view.ex#L133) of other [places](https://github.com/mbta/dotcom/blob/8999a512499c2b695aceb44b07458b4553ede8da/apps/site/lib/site_web/templates/schedule/_trip_info.html.eex#L54).

---

### DIFF
--- a/apps/site/lib/site/trip_plan/related_link.ex
+++ b/apps/site/lib/site/trip_plan/related_link.ex
@@ -146,7 +146,8 @@ defmodule Site.TripPlan.RelatedLink do
   end
 
   # if there are multiple fare links, show fare overview link
-  defp simplify_fare_text(fare_links) when Kernel.length(fare_links) > 1, do: [fare_overview_link()]
+  defp simplify_fare_text(fare_links) when Kernel.length(fare_links) > 1,
+    do: [fare_overview_link()]
 
   defp simplify_fare_text([fare_link]) do
     # if there's only one fare link, change the text to "View fare information"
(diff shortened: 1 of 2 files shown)


apps/site/lib/site/trip_plan/related_link.ex
```diff
@@ -123,16 +123,8 @@ defmodule Site.TripPlan.RelatedLink do
...
-    "commuter rail"
-  end
-
-  defp fare_link_text(:ferry) do
-    "ferry"
-  end
-
-  defp fare_link_text(_) do
-    "bus/subway"
+  defp fare_link_text(type) when type in [:commuter_rail, :ferry, :bus, :subway] do
```
arkadyan, line 126: Do we even need the `when` clause at this point?

thecristen:
Updated to show a single link to /fares instead of links for each mode, as discussed earlier.

apps/site/lib/site/trip_plan/related_link.ex
thecristen: This was me being paranoid about whether a different type somehow snuck in here, but it looks like that won't happen given the [mode types coming in](https://github.com/mbta/dotcom/blob/383405b80994b0f6a347f390360cc63c3b029b29/apps/trip_plan/lib/trip_plan/api/open_trip_planner/parser.ex#L14) through OTP... can remove this

PR #487, opened 2020-05-28
Show highest one-way fare in accordion for Trip Planner

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Base Fares | One-way/base fare trips and total price](https://app.asana.com/0/555089885850811/1162903717356809)

This ticket adds the logic to include an estimation of the fares for the itineraries (one way and round trip):<br/>
image

### DIFF
--- a/apps/site/assets/css/_trip-plan-results.scss
+++ b/apps/site/assets/css/_trip-plan-results.scss
@@ -24,10 +24,16 @@
     background-color: $brand-primary-lightest-contrast;
     border: 1px solid $brand-primary;
     border-bottom: 0;
+    display: flex;
     font-size: $base-spacing;
+    justify-content: space-between;
     line-height: 1.5;
     margin-top: 0;
     padding: $base-spacing;
+
+    @include media-breakpoint-down(sm) {
+      display: block;
+    }
   }
 
   &-accessible {
@@ -80,5 +86,25 @@
     background-color: $gray-lightest;
     margin-bottom: $base-spacing;
     padding: $base-spacing;
+  }
+
+  &-fares {
+    min-width: 40%;
+    padding-left: $base-spacing;
+    padding-top: 0;
+
+    @include media-breakpoint-down(sm) {
+      padding-left: 0;
+      padding-top: $base-spacing;
+    }
+  }
+
+  &-fare-title {
+    font-size: $font-size-h5;
+    font-weight: bold;
+  }
+
+  &-fare {
+    padding-left: $base-spacing;
   }
 }
--- a/apps/site/lib/site_web/views/trip_plan_view.ex
+++ b/apps/site/lib/site_web/views/trip_plan_view.ex
@@ -1,11 +1,14 @@
 defmodule SiteWeb.TripPlanView do
+  @moduledoc "Contains the logic for the Trip Planner"
   use SiteWeb, :view
   require Routes.Route
   alias Site.React
   alias Site.TripPlan.{Query, ItineraryRow}
+  alias TripPlan.{Leg}
   alias Routes.Route
   alias Phoenix.{HTML, HTML.Form}
   alias SiteWeb.PartialView.SvgIconWithCircle
+  alias Fares.{Format}
 
   import Schedules.Repo, only: [end_of_rating: 0]
 
@@ -551,6 +554,15 @@ defmodule SiteWeb.TripPlanView do
 
       access_html = i |> accessibility_icon() |> HTML.safe_to_string()
 
+      fare = get_highest_one_way_fare(i)
+
+      fares_html =
+        "_itinerary_fares.html"
+        |> render_to_string(
+          one_way_total: Format.price(fare),
+          round_trip_total: Format.price(fare * 2)
+        )
+
       html =
         "_itinerary.html"
         |> render_to_string(
@@ -568,7 +580,8 @@ defmodule SiteWeb.TripPlanView do
         tab_html: tab_html,
         id: index,
         map: itinerary_map(map),
-        access_html: access_html
+        access_html: access_html,
+        fares_html: fares_html
       }
     end
   end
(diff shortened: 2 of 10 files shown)


apps/site/assets/css/_trip-plan-results.scss
```diff
@@ -81,4 +84,24 @@
     margin-bottom: $base-spacing;
     padding: $base-spacing;
   }
+
+  &-fares {
+    min-width: 40%;
+    padding-left: 0;
+    padding-top: $base-spacing;
+
+    @include media-breakpoint-up(xsm) {
```
thecristen, line 93: I'm curious about how well this is working -- I was able to find a state where the container width is too small for adding the left padding but also too big to put the fares below the other content -- see screenshot at 396px. You might need to add some logic to enforce these to be full width in the narrowest viewports

image

apps/site/assets/css/_trip-plan-results.scss
```diff
@@ -24,7 +24,10 @@
     background-color: $brand-primary-lightest-contrast;
     border: 1px solid $brand-primary;
     border-bottom: 0;
+    display: flex;
+    flex-wrap: wrap;
```
thecristen, line 28: I wonder if it makes sense to set these flexbox properties only for wider viewports, so we don't get wrapping in the mobile view

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -589,4 +601,22 @@ defmodule SiteWeb.TripPlanView do
...
+    itinerary.legs
+    |> Enum.filter(fn leg -> leg.description != "WALK" end)
+    |> Enum.reduce(0, fn leg, acc ->
+      highest_one_way_fare =
+        leg
+        |> Kernel.get_in([
+          Access.key(:mode, %{}),
+          Access.key(:fares, %{}),
+          Access.key(:highest_one_way_fare, %{}),
+          Access.key(:cents, 0)
```
thecristen, line 616: TIL about `Access`,  I'm guessing you need these to ensure the lookup never returns `nil`? Very cool!

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -589,4 +601,22 @@ defmodule SiteWeb.TripPlanView do
       }
     )
   end
+
+  @spec get_highest_one_way_fare(TripPlan.Itinerary.t()) :: integer
+  def get_highest_one_way_fare(itinerary) do
+    itinerary.legs
+    |> Enum.filter(fn leg -> leg.description != "WALK" end)
```
thecristen, line 608: I think searching the description for "WALK" might be a bit brittle so let's get to the transit legs some other way. The [Leg.transit?](https://github.com/mbta/dotcom/blob/eaac9fdb9b32f0c55f64689ec23de01de283a513/apps/trip_plan/lib/trip_plan/leg.ex#L57) method might do the trick

apps/site/test/site_web/views/trip_plan_view_test.exs
```diff
@@ -522,4 +522,62 @@ closest arrival to 12:00 AM, Thursday, January 1st."
...
+    end
+
+    test "gets the highest one-way fare" do
+      itinerary = %TripPlan.Itinerary{
+        start: nil,
+        stop: nil,
+        legs: [
+          %{
+            description: "BUS",
+            mode: %{
```
thecristen, line 552: Might make sense to add a non-transit leg to this test itinerary just to validate that the mode filtering doesn't produce unexpected results.

ingridpierre:
@alicia-mbta is this behind a flag? The work in this sprint probably should be deployed all at once since the prices will look pretty high in some cases.

apps/site/lib/site_web/views/trip_plan_view.ex
alicia-mbta: Yes, I've seen it used instead of the dot notation to access deeply nested structures and since it has a default field, we make sure nothing breaks.

apps/site/assets/css/_trip-plan-results.scss
alicia-mbta: Yeah, I was hesitant between 395 and 398 pixels...Are you saying that introducing `xsm` is not really a good approach? I can certainly remove this new grid breakpoint but if I do that, I'm going to need your CSS expertise, how do I detect the wrapping?

thecristen: Good questions... I'd avoid adding new breakpoints, though I could see why this new breakpoint would do well here. But probably isn't practical to try to detect _exactly_ where to transition from showing these sections side-by-side to stacked, since the left section width will vary based on itinerary. 

I think for this reason I'd remove the `flex-wrap` property... I also think we should try using the `sm` breakpoint, even though it might look spacious for many itineraries. If we add `display: block` to the smaller breakpoint it'll unset the flexboxing and let the sides stack. Hope this helps!

apps/site/lib/site_web/views/trip_plan_view.ex
arkadyan: That's cool, I didn't know this technique either. Another option that's similar is `Map.get/3` which takes an optional default value instead of the `nil` default.

apps/site/assets/css/_trip-plan-results.scss
alicia-mbta: I hadn't looked at it that way, thanks a lot!

thecristen:
@ingridpierre It'll go into a feature branch, which'll get merged once the related tasks are also done. So you can see the change in this branch right now without any flagging 🤭

apps/site/assets/css/_trip-plan-results.scss
```diff
@@ -24,10 +24,16 @@
...
     border: 1px solid $brand-primary;
     border-bottom: 0;
+    display: flex;
     font-size: $base-spacing;
+    justify-content: space-between;
     line-height: 1.5;
     margin-top: 0;
     padding: $base-spacing;
+
+    @include media-breakpoint-up(sm) {
```
thecristen, line 34: It's backwards! media-breakpoint-**down** will make it apply to sm and smaller :)

alicia-mbta: :woman_facepalming:

ingridpierre:
@alicia-mbta lgtm 👍

PR #489, opened 2020-05-28
Display recommended monthly pass in the trip plan results summary

arkadyan:
#### Summary of changes
**Asana Ticket:** [Base Fares | Recommend Monthly Passes](https://app.asana.com/0/555089885850811/1162903717356811)

Display recommended monthly pass in the trip plan results summary box. Add a new `Fare.Month` module to support finding monthly passes.

Targets the `feature/base-fares` branch.

Depends on  https://github.com/mbta/dotcom/pull/487

Screen Shot 2020-05-28 at 16 37 15

---

### DIFF
--- a/apps/fares/lib/format.ex
+++ b/apps/fares/lib/format.ex
@@ -99,7 +99,8 @@ defmodule Fares.Format do
   def name(:premium_ride), do: "Premium Ride"
   def name(:invalid), do: "Invalid Fare"
 
-  @spec full_name(Fare.t()) :: String.t() | iolist
+  @spec full_name(Fare.t() | nil) :: String.t() | iolist
+  def full_name(nil), do: "Free"
   def full_name(%Fare{mode: :subway, duration: :month}), do: "Monthly LinkPass"
   def full_name(%Fare{mode: :commuter_rail, duration: :weekend}), do: "Weekend Pass"
   def full_name(%Fare{duration: :week}), do: "7-Day Pass"
--- a/apps/fares/lib/format.ex
+++ b/apps/fares/lib/format.ex
@@ -1,4 +1,7 @@
 defmodule Fares.Format do
+  @moduledoc """
+  Formatting functions for fare data.
+  """
   alias Fares.{Fare, Summary}
 
   @type mode_type :: :bus_subway | :commuter_rail | :ferry
--- a/apps/fares/lib/month.ex
+++ b/apps/fares/lib/month.ex
@@ -11,8 +11,8 @@ defmodule Fares.Month do
   @type fare_fn :: (Keyword.t() -> [Fare.t()])
 
   @spec lowest_pass(
-          Route.t(),
-          Trip.t() | nil,
+          Route.t() | Route.id_t(),
+          Trip.t() | Trip.id_t() | nil,
           Stop.id_t(),
           Stop.id_t(),
           fare_fn()
@@ -21,6 +21,16 @@ defmodule Fares.Month do
   def lowest_pass(route, trip, origin_id, destination_id, fare_fn \\ &Repo.all/1)
   def lowest_pass(nil, _, _, _, _), do: nil
 
+  def lowest_pass(route_id, trip, origin_id, destination_id, fare_fn) when is_binary(route_id) do
+    route = Routes.Repo.get(route_id)
+    lowest_pass(route, trip, origin_id, destination_id, fare_fn)
+  end
+
+  def lowest_pass(route, trip_id, origin_id, destination_id, fare_fn) when is_binary(trip_id) do
+    trip = Schedules.Repo.trip(trip_id)
+    lowest_pass(route, trip, origin_id, destination_id, fare_fn)
+  end
+
   def lowest_pass(route, trip, origin_id, destination_id, fare_fn) do
     route
     |> get_fares(trip, origin_id, destination_id, fare_fn)
@@ -28,8 +38,8 @@ defmodule Fares.Month do
   end
 
   @spec highest_pass(
-          Route.t(),
-          Trip.t() | nil,
+          Route.t() | Route.id_t(),
+          Trip.t() | Trip.id_t() | nil,
           Stop.id_t(),
           Stop.id_t(),
           fare_fn()
@@ -37,6 +47,16 @@ defmodule Fares.Month do
           Fare.t() | nil
   def highest_pass(route, trip, origin_id, destination_id, fare_fn \\ &Repo.all/1)
   def highest_pass(nil, _, _, _, _), do: nil
+
+  def highest_pass(route_id, trip, origin_id, destination_id, fare_fn) when is_binary(route_id) do
+    route = Routes.Repo.get(route_id)
+    highest_pass(route, trip, origin_id, destination_id, fare_fn)
+  end
+
+  def highest_pass(route, trip_id, origin_id, destination_id, fare_fn) when is_binary(trip_id) do
+    trip = Schedules.Repo.trip(trip_id)
+    highest_pass(route, trip, origin_id, destination_id, fare_fn)
+  end
 
   def highest_pass(route, trip, origin_id, destination_id, fare_fn) do
     route
--- a/apps/fares/lib/month.ex
+++ b/apps/fares/lib/month.ex
@@ -10,7 +10,7 @@ defmodule Fares.Month do
 
   @type fare_fn :: (Keyword.t() -> [Fare.t()])
 
-  @spec lowest_pass(
+  @spec recommended_pass(
           Route.t() | Route.id_t(),
           Trip.t() | Trip.id_t() | nil,
           Stop.id_t(),
@@ -18,26 +18,28 @@ defmodule Fares.Month do
           fare_fn()
         ) ::
           Fare.t() | nil
-  def lowest_pass(route, trip, origin_id, destination_id, fare_fn \\ &Repo.all/1)
-  def lowest_pass(nil, _, _, _, _), do: nil
+  def recommended_pass(route, trip, origin_id, destination_id, fare_fn \\ &Repo.all/1)
+  def recommended_pass(nil, _, _, _, _), do: nil
 
-  def lowest_pass(route_id, trip, origin_id, destination_id, fare_fn) when is_binary(route_id) do
+  def recommended_pass(route_id, trip, origin_id, destination_id, fare_fn)
+      when is_binary(route_id) do
     route = Routes.Repo.get(route_id)
-    lowest_pass(route, trip, origin_id, destination_id, fare_fn)
+    recommended_pass(route, trip, origin_id, destination_id, fare_fn)
   end
 
-  def lowest_pass(route, trip_id, origin_id, destination_id, fare_fn) when is_binary(trip_id) do
+  def recommended_pass(route, trip_id, origin_id, destination_id, fare_fn)
+      when is_binary(trip_id) do
     trip = Schedules.Repo.trip(trip_id)
-    lowest_pass(route, trip, origin_id, destination_id, fare_fn)
+    recommended_pass(route, trip, origin_id, destination_id, fare_fn)
   end
 
-  def lowest_pass(route, trip, origin_id, destination_id, fare_fn) do
+  def recommended_pass(route, trip, origin_id, destination_id, fare_fn) do
     route
     |> get_fares(trip, origin_id, destination_id, fare_fn)
     |> Enum.min_by(& &1.cents, fn -> nil end)
   end
 
-  @spec highest_pass(
+  @spec base_pass(
           Route.t() | Route.id_t(),
           Trip.t() | Trip.id_t() | nil,
           Stop.id_t(),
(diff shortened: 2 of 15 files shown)


ingridpierre:
I'd like more even vertical spacing between the access label, the base fare estimate section, and the monthly pass fare. It will add a little to the height, but that's expected. This applies to mobile and desktop views. (There's some difference in indentation in the design, I prefer to keep the slight indentation that's already in this PR).

**Current vertical spacing:**
image
**Desired vertical spacing:**
image

ingridpierre:
Below it should should be "Inner Express Bus Pass" instead of "CharlieCard or CharlieTicket". Similar goes for any Outer Express itineraries.

image

ingridpierre:
To add to the above, instead of "CharlieCard or CharlieTicket: $90.00"  should say "Monthly LinkPass"

ingridpierre:
@arkadyan It's important to note that we are currently unable to see ferry for this work (or all the fare calculator work), we'll need to spin up a follow up ticket to check on those trips once some service to ferries is restored.

If we had some mock data/trips would that help?  I'm not sure what's possible here for testing purposes.

apps/site/lib/site_web/controllers/trip_plan_controller.ex
```diff
@@ -96,6 +102,49 @@ defmodule SiteWeb.TripPlanController do
...
 
+  @spec highest_highest_month_pass(Itinerary.t()) :: Fare.t() | nil
+  defp highest_highest_month_pass(%Itinerary{legs: legs}) do
+    legs
+    |> Enum.map(&highest_month_pass/1)
+    |> max_by_cents()
+  end
+
+  @spec highest_lowest_month_pass(Itinerary.t()) :: Fare.t() | nil
+  defp highest_lowest_month_pass(%Itinerary{legs: legs}) do
```
thecristen, line 113: I don't really get the naming of these two functions... help? 😅 
I'm also starting to wonder whether we should migrate away from using "highest" and "lowest" to something more meaningful.

arkadyan:
@ingridpierre This is how the spacing is looking now, is this looking better?

Screen Shot 2020-06-01 at 10 22 19

ingridpierre:
@arkadyan Yes looking good, thanks!

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -620,4 +621,13 @@ defmodule SiteWeb.TripPlanView do
...
   end
+
+  @spec monthly_pass(Fare.t()) :: String.t()
+  def monthly_pass(fare) do
+    "#{zone(fare)}#{Format.media(fare)}: #{Format.price(fare)}"
+  end
+
+  @spec zone(Fare.t()) :: String.t()
+  defp zone(%Fare{mode: :commuter_rail, name: {:zone, zone}}), do: "Commuter Rail Zone #{zone} "
+  defp zone(_), do: ""
```
thecristen: Consider adding these formatting functions to `Fares.Format` (and perhaps leverage some more of its functionality)

arkadyan:
@ingridpierre Thanks, I fixed the media names. I can upload to dev-green once it passes CI.

That's a great point about ferries. We have some tests using ferry data and I could fake some things, but I think it's still best to check for real by eye so I'll make a ticket for that.

apps/site/lib/site_web/controllers/trip_plan_controller.ex
arkadyan: Yeah, I've been struggling with all these names the whole time. It's tricky because you can't necessarily go by media, they can mean different things for different modes and contexts. Maybe we can use the "base" and "recommended" terms we present to the users. I was trying to avoid those because we used to use "base" in the codebase to mean what we are now calling "recommended" which is obviously confusing. But we removed that and if we decide that it's not going to confuse any of us, maybe that route is best? So these functions could be something like `highest_base_monthly_pass` and `highest_recommended_monthly_pass`. Or do you have any other recommendations?

I'm looking forward to the day when these fare differences just go away!

arkadyan: @thecristen Or maybe another approach would be to just change the first "highest". I don't know what the best alternative would be, but maybe instead these could be something like `overarching_highest_month_pass` and `overarching_lowest_month_pass`?

apps/fares/lib/month.ex
```diff
@@ -0,0 +1,119 @@
...
+
+  @spec lowest_pass(
+          Route.t() | Route.id_t(),
+          Trip.t() | Trip.id_t() | nil,
+          Stop.id_t(),
+          Stop.id_t(),
+          fare_fn()
+        ) ::
+          Fare.t() | nil
+  def lowest_pass(route, trip, origin_id, destination_id, fare_fn \\ &Repo.all/1)
```
thecristen, line 21: These lowest and highest pass functions (which look good generally speaking) are almost identical save for `Enum.max_by` vs `Enum.min_by`, so I'm wondering if the abstraction is in the wrong place. Maybe we need one method that handles most of the work (`passes_for_route(route, trip, origin_id, destination_id, fare_fn)` maybe, or just expose `get_fares`)

apps/fares/test/month_test.exs
```diff
@@ -0,0 +1,335 @@
...
+
+      assert Month.lowest_pass(route, nil, origin_id, destination_id, fare_fn) == nil
+    end
+  end
+
+  describe "ferry" do
+    test "returns the fares that are not discounted for the correct ferry trip" do
+      route = %Route{type: 4}
+      origin_id = "Boat-Charlestown"
+      destination_id = "Boat-Long-South"
```
thecristen, line 301: Not sure it matters, but our current ferry IDs are Boat-F1 and Boat-F4

arkadyan: Are those the IDs for the routes or for the stops (or both)?

apps/site/lib/site_web/views/trip_plan_view.ex
arkadyan: I think this got simplified and more specialized with a recent change. At this point, we just want to add "Commuter Rail" to the beginning, but not do anything for any other modes.

apps/site/lib/site_web/controllers/trip_plan_controller.ex
thecristen: Yeah, I see the struggle! I think introducing some consistency between the backend and frontend would make the _most_ sense. I'd go more conceptual rather than using the function name to explain the implementation. IMO `highest_base_monthly_pass` and `highest_recommended_monthly_pass` wouldn't make sense here are there will only be one recommended monthly pass, and I'm not sure "base" is a relevant concept for monthly passes anyway (which have no reduced fare counterpart AFAIK).

We hadn't discussed it much, but my overarching understanding is:

**For a trip:**
Base = sum of all the regular one-way fares involved
Reduced  = same, but with reduced fare instead of full fare values
(eventually these will incorporate transfers, which will lower the resulting values)

**For monthly passes:**
All monthly passes relevant to the trip (there may be more than one) will be shown in the accordion - includes regular (base?) fares as well as reduced fares
One of these passes will be highlighted as the Recommended one (the highest one, maybe?).

Sorry I don't have a simpler answer at the moment but hopefully this aids our thinking. This discussion all applies to the functions in the `Month` and `OneWay` modules too.

apps/site/lib/site_web/controllers/trip_plan_controller.ex
```diff
@@ -96,6 +102,49 @@ defmodule SiteWeb.TripPlanController do
...
+       }) do
+    Month.lowest_pass(route_id, trip_id, origin_id, destination_id)
+  end
+
+  @spec max_by_cents([Fare.t() | nil]) :: Fare.t() | nil
+  defp max_by_cents(fares), do: Enum.max_by(fares, &cents_for_max/1, fn -> nil end)
+
+  @spec cents_for_max(Fare.t() | nil) :: non_neg_integer
+  defp cents_for_max(nil), do: 0
+  defp cents_for_max(%Fare{cents: cents}), do: cents
```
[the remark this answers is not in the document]
arkadyan: The reason I made this just a little local private helper function is that I think it's only really useful in this context where `Enum.max` needs a value to compare, but you might have `nil` instead of a Fare struct. Generally, I don't think you'd want to say that an undefined fare is 0 cents.

We could also consider `max_by_cents` moving into `Fare`. It didn't seem worth it for this single usage, but if we ever need that functionality in another location I'd probably suggest that.

apps/fares/test/month_test.exs
thecristen: Ahh right nevermind, you're talking stops and I'm talking routes. Sorry!

apps/site/lib/site_web/controllers/trip_plan_controller.ex
arkadyan: I think it's actually a little more complicated than that for passes. 

Each leg of your trip will have potential passes for it. Let's take an example where you take the subway to the commuter rail. There is one monthly pass for the subway (Monthly LinkPass). It's good on the subway and bus but it doesn't get you on the CR. For the CR there are two monthly passes available, the CharlieTicket or the mTicket. The CR pass does grant you access to the subway. 

So this is the reason for the two `highest`s in `highest_highest_month_pass`. The first (inner) one is because we need to figure out the highest or base fare for each mode. The Monthly LinkPass in the case of the subway, and the CharlieTicket in the case of CR. And then the second `highest` is because we need to say, ok, given all these legs, what pass grants you access to everything? So, in this case, we want to display the CR CharlieTicket monthly pass, not the subway Monthly LinkPass.

apps/fares/lib/month.ex
arkadyan: Could you please elaborate on this more? I'm not sure I'm following quite yet.

apps/site/lib/site_web/controllers/trip_plan_controller.ex
thecristen: Thanks for the explanation! I think that's not too different than what I was thinking. My thoughts were more: first get _every_ pass for every mode (e.g. [LinkPass, CR CharlieTicket,  CR mTicket]) and then pick the recommended one via a view function (e.g. CR CharlieTicket). It'd be clearer and we can use the list of all passes in the accordion later.

apps/fares/lib/month.ex
thecristen: Was thinking:

`def get_passes(nil, _, _, _, _), do: nil`
`def get_passes(route_id, trip, origin_id, destination_id, fare_fn) when is_binary(route_id) do`
etc etc etc... 
---eventually it calls `get_fares(route, trip, origin_id, destination_id, fare_fn)` and returns `[Fare.t()] | nil`

then those passes serve as the argument for `lowest_pass(passes)` and `highest_pass(passes)`, which are smaller single-purpose functions that return the relevant pass (or nil)

Would that work for this use case? Hopefully I'm not interpreting the intent incorrectly!

apps/site/lib/site_web/controllers/trip_plan_controller.ex
arkadyan: Ah that’s an interesting thought. So with that approach we’d need to figure out which passes to display or not display later? With my current approach I was thinking we’d always display the highest and display the lowest only if it was different. I’m trying to think about what the logic would be with a list. Maybe going through the covered modes and see what covers everything?

thecristen: Maybe! If that's too complicated, we might want to put that off until a future PR. 

But all this was to say that your original function names still confused me and I wouldn't have been sure how to leverage them if I were to pick up related tasks :/

apps/fares/lib/month.ex
arkadyan: Still trying to see if I'm with you yet or not.

When you called `lowest_pass(passes)`, would you be passing in the passes for all the legs? I think this would be a problem because you don't necessarily want the absolute lowest. In the case of the subway to CR trip, for all legs, you'd have 3 passes: the subway Monthly LinkPass, the CR CharlieTicket, and the CR mTicket. What we want to display in the accordion is the CR CharlieTicket and the CR mTicket. But the lowest of all these is the subway Monthly LinkPass. 

Or am I misunderstanding the idea? Would you just be passing in the passes for one leg or something different altogether?

arkadyan: I definitely think the confusing names are still adding to the overall confusion in all of this!

thecristen: Ah great point about the lowest_pass, indeed part of my misunderstanding (which you did interpret correctly) was caused by the naming! That definitely makes my logic flawed.

As for naming, why are we annotating anything as "lowest" then? Definitely seems more confusing than useful here. I'd think we want mTicket in the display not because it's a "lowest" anything, but because it's another applicable pass for the trip (if there were ever a third CR pass, we'd want all three listed no?).

arkadyan:
There is a bug in this branch where you get a 500 error mapping a trip from Lechmere to North Station (likely because of the shuttle buses). https://sentry.io/organizations/mbtace/issues/1702818304/?project=194344&referrer=slack

arkadyan:
Fixed the error condition.

arkadyan:
@thecristen Made those name changes. I think that ends up covering everything you touched on, but please let me know if there's anything else.

apps/fares/lib/format.ex
```diff
@@ -96,7 +99,8 @@ defmodule Fares.Format do
   def name(:premium_ride), do: "Premium Ride"
   def name(:invalid), do: "Invalid Fare"
 
-  @spec full_name(Fare.t()) :: String.t() | iolist
+  @spec full_name(Fare.t() | nil) :: String.t() | iolist
+  def full_name(nil), do: "Free"
```
thecristen, line 103: Why is this Free?

arkadyan: Oh yeah sorry, I should have mentioned that. This is for a shuttle bus leg which doesn't have a fare because it is free. (And crashed before I added this case for.) There's a larger ticket which this is a first little step towards as well: https://app.asana.com/0/1134952491130998/1178333054166489

PR #490, opened 2020-05-30
Obtain reduced fares

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Fare Calculator Accordion | Data for reduced fares](https://app.asana.com/0/555089885850811/1177449860049777)

This PR contains the logic to retrieve the reduced fares per mode. Backend changes, no visual impact.

### DIFF
--- a/apps/trip_plan/lib/trip_plan/transit_detail.ex
+++ b/apps/trip_plan/lib/trip_plan/transit_detail.ex
@@ -16,6 +16,7 @@ defmodule TripPlan.TransitDetail do
 
   @type fares :: %{
           highest_one_way_fare: Fare.t(),
-          lowest_one_way_fare: Fare.t()
+          lowest_one_way_fare: Fare.t(),
+          reduced_one_way_fare: Fare.t()
         }
 end
--- a/apps/fares/lib/one_way.ex
+++ b/apps/fares/lib/one_way.ex
@@ -1,6 +1,6 @@
 defmodule Fares.OneWay do
   @moduledoc """
-  Calculates the lowest and highest fare for a particular trip i.e. a regular priced, non-discounted, one-way fare for the given mode.
+  Calculates the lowest, highest and reduced one-way fares for a particular trip for the given mode.
   Commuter rail and ferry fares distinguish between the possible sets of stops.
   Bus fares for express buses do not distinguish between the local and express portions;
   the express fare is always returned.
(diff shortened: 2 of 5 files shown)


apps/trip_plan/lib/trip_plan/transit_detail.ex
```diff
@@ -16,6 +16,7 @@ defmodule TripPlan.TransitDetail do
 
   @type fares :: %{
           highest_one_way_fare: Fare.t(),
-          lowest_one_way_fare: Fare.t()
+          lowest_one_way_fare: Fare.t(),
+          reduced_one_way_fares: [Fare.t()]
```
arkadyan, line 20: Are there any cases for which there are multiple reduced fares for a given route or trip? If I'm interpreting the designs correctly it seems like they expect to only display one or none.

If that's the case it might be easier for the front-end if we provide a single value here instead of a list. If there is never more than one result anyway, maybe it's as simple as adding a `List.first()` after line 65 in `highest_lowest_fare.ex` or something. If there are sometimes multiple results, maybe we need to clarify with Ingrid?

alicia-mbta: I was also confused about that...from what I saw, type `reduced` that can have these values:
`@type reduced :: nil | :student | :senior_disabled | :any`
I saw that although the fares for students and senior citizens is the same price, it is returned as separate objects (one for each), so querying without the 'reduced' filter returns more than one value.
If it makes more sense to return just a unique value symbolizing a "one-way reduced fare" I can do that, just let me know.

arkadyan: Interesting. @ingridpierre Do you have thoughts on how you want these displayed?

ingridpierre: The reduced fare is the same regardless of the different ways you get it. One unified reduced fare only is correct.

alicia-mbta: Sounds good, I'll make those changes.

apps/fares/lib/highest_lowest_fare.ex
```diff
@@ -54,9 +55,20 @@ defmodule Fares.HighestLowestFare do
   def highest_fare(route, trip, origin_id, destination_id, fare_fn) do
     route
     |> get_fares(trip, origin_id, destination_id, fare_fn)
+    |> Enum.filter(fn fare -> fare.reduced == nil end)
     |> Enum.max_by(& &1.cents, fn -> nil end)
   end
 
+  def reduced_fare(route, trip, origin_id, destination_id, fare_fn \\ &Repo.all/1) do
```
arkadyan, line 62: One small thing I missed in the first pass: could you please add a typespec for this new function?

PR #491, opened 2020-06-01
Add transfer note to trip planner itinerary

thecristen:
#### Summary of changes
**Asana Ticket:** [Base Fares | Add transfers note (not subway)](https://app.asana.com/0/555089885850811/1169236529454061)

Adds the "Total may be less with transfers" note to the itinerary fare summary. This note is added when the itinerary contains two consecutive transit legs which fit the following:

- Subway to subway **(to be removed upon completion of the Base Fares | Add transfers logic (subway) ticket)**
- Subway to local bus
- Local bus to subway
- Local bus to local bus
- Inner express bus to subway
- Outer express bus to subway
- Inner express bus to local bus
- Outer express bus to local bus

If none of the above are present, then no transfer note will be shown.

---

### DIFF
--- a/apps/site/lib/site_web/views/trip_plan_view.ex
+++ b/apps/site/lib/site_web/views/trip_plan_view.ex
@@ -4,7 +4,7 @@ defmodule SiteWeb.TripPlanView do
   require Routes.Route
   alias Site.React
   alias Site.TripPlan.{Query, ItineraryRow}
-  alias TripPlan.{Leg}
+  alias TripPlan.{Leg, Itinerary}
   alias Routes.Route
   alias Phoenix.{HTML, HTML.Form}
   alias SiteWeb.PartialView.SvgIconWithCircle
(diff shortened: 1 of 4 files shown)


apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -523,6 +523,47 @@ defmodule SiteWeb.TripPlanView do
     SiteWeb.ViewHelpers.mode_name(type)
   end
 
+  @spec transfer_note(Itinerary.t()) :: String.t() | nil
+  def transfer_note(itinerary) do
+    itinerary
+    |> Itinerary.route_ids()
+    |> Enum.chunk_every(2, 1, :discard)
+    |> Enum.find(&satisfies_transfer_note_conditions?(&1))
```
arkadyan, line 531: Based on our conversation this morning it sounds like we now want to display the note in all cases. I don't know if you want to remove this logic in this PR, or just wait and do it later. Either way seems fine to me.

apps/site/test/site_web/views/trip_plan_view_test.exs
```diff
@@ -450,6 +451,110 @@ closest arrival to 12:00 AM, Thursday, January 1st."
...
+    @bus_leg leg_for_route.("77")
+    @subway_leg leg_for_route.("Red")
+    @cr_leg leg_for_route.("CR-Lowell")
+    @ferry_leg leg_for_route.("Boat-F4")
+    @innerxp_leg leg_for_route.("326")
+    @outerxp_leg leg_for_route.("505")
+    @sl_rapid_leg leg_for_route.("741")
+    @sl_bus_leg leg_for_route.("751")
+
+    test "shows note for subway-bus transfer" do
```
arkadyan, line 467: I know they might be going away, but these tests are great!

apps/site/lib/site_web/views/trip_plan_view.ex
thecristen: I don't think I'd say **all**; for example for single-route trips it wouldn't make sense. We'll leave it for later..

thecristen:
Note: We can re-review the logic and design when we are closer to merging `feature/base-fares` into `master` (by which point the unrelated failing tests might be fixed too).

apps/site/lib/site_web/views/trip_plan_view.ex
arkadyan: Good point

PR #495, opened 2020-06-03
Add free subway transfer logic to base fare estimates

thecristen:
#### Summary of changes
**Asana Ticket:** [Base Fares | Add transfers logic (subway)](https://app.asana.com/0/555089885850811/1169236529454063)

Slightly adjusts the implementation of adding the transfers note, and also introduces logic for identifying subway-subway transfers and accounting for this free transfer when computing the highest one way fare.

Some caveats: there doesn't seem to be an iron-clad way to figure out whether the OTP-generated itinerary contains a real _behind the fare gate_ free subway transfer.

---

### DIFF
--- a/apps/fares/lib/transfer.ex
+++ b/apps/fares/lib/transfer.ex
@@ -71,8 +71,8 @@ defmodule Fares.Transfer do
   def is_free_transfer(leg_pair) when has_nil(leg_pair), do: nil
 
   def is_free_transfer([%{:to => nil} | _]), do: nil
-  def is_free_transfer([%{:to => %{:stop_id => xfer_stop}} | _] = leg_pair) do
 
+  def is_free_transfer([%{:to => %{:stop_id => xfer_stop}} | _] = leg_pair) do
     xfer_lines = lines_for_xfer_stop(xfer_stop)
 
     if xfer_lines do
--- a/apps/fares/lib/transfer.ex
+++ b/apps/fares/lib/transfer.ex
@@ -1,0 +1,71 @@
+defmodule Fares.Transfer do
+  @moduledoc """
+    Tools for handling logic around transfers between transit legs and modes.
+    The MBTA allows transfers between services depending on the fare media used
+    and the amount paid.
+
+    This logic may be superseded by the upcoming fares work.
+  """
+  require Fares
+  alias Routes.{Repo, Route}
+
+  defguard has_nil(pair)
+           when is_nil(elem(pair, 0)) or
+                  is_nil(elem(pair, 1))
+
+  @type fare_atom :: Route.gtfs_route_type() | :inner_express_bus | :outer_express_bus
+
+  # Paying a single-ride fare for the first may get you a transfer to the second
+  # (can't be certain, as it depends on media used)!
+  @single_ride_valid_transfers [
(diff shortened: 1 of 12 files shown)


apps/fares/lib/transfer.ex
```diff
@@ -0,0 +1,120 @@
...
+    %{lines: ["Orange", "Green-C", "Green-E"], stop: "place-haecl"},
+    %{lines: ["Green-B", "Green-C", "Green-D"], stop: "place-kencl"},
+    %{lines: ["Green-B", "Green-C", "Green-D", "Green-E"], stop: "place-armnl"}
+  ]
+
+  @doc """
+  Takes a pair of routes and returns true if there might be a transfer between
+  the two, based on the list in @single_ride_valid_transfers
+  """
+  @spec is_maybe_transfer([Route.id_t()]) :: boolean | nil
```
arkadyan, line 50: Nit: the [convention](https://hexdocs.pm/elixir/master/naming-conventions.html#trailing-question-mark-foo) in Elixir is to name functions that return booleans with a trailing question mark. (Same thing for `is_free_transfer`.)

apps/fares/lib/transfer.ex
```diff
@@ -0,0 +1,120 @@
...
+
+  defp lines_for_xfer_stop(stop_id) do
+    Enum.find(@underground_xfers, %{}, fn %{:stop => stop} ->
+      stop == stop_id
+    end)
+    |> Map.get(:lines)
+  end
+
+  @spec to_fare_atom(fare_atom | Route.id_t() | Route.t()) :: fare_atom
+  defp to_fare_atom(route_or_atom) do
```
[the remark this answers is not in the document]
thecristen: Ha, just tried this refactor and it went slightly awry! But I will change the name of the ScheduleView function to better indicate the difference.

ingridpierre:
Planning a trip from South Station to Wonderland (for example),  subway transfer logic isn't coming through. Prices appear to still be added 3x.
image

thecristen:
Thank you!! I'd forgotten about the concept of parent stations and platforms sometimes having their own stop ids. Update forthcoming...

thecristen:
Okay I don't know why CI keeps failing me (because it passes on my machine, _classic_), but please go ahead and re-review. 

It should now properly handle transfers at large stations, and should be adding the fare properly but many-route multimodal trips like that pictured above! @arkadyan I think I've addressed all your comments but I've refactored a fair bit so this could use another look.

apps/fares/lib/transfer.ex
```diff
@@ -0,0 +1,106 @@
...
+  # (can't be certain, as it depends on media used)!
+  @single_ride_transfers %{
+    :bus => [:subway, :bus],
+    :subway => [:subway, :bus],
+    :inner_express_bus => [:subway, :bus],
+    :outer_express_bus => [:subway, :bus]
+  }
+
+  @doc "Searches a list of legs for evidence of an in-station subway transfer."
+  @spec is_subway_transfer?([Leg.id_t()]) :: boolean
```
arkadyan, line 24: Here's one thing that dialyzer is complaining about: `Leg` doesn't have an `id_t` type.

PR #499, opened 2020-06-07
Logic for the Fare Calculator in the Trip Planner (2nd accordion)

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Fare Calculator Accordion | One-Way/Round Trip fares](https://app.asana.com/0/555089885850811/1162903717356813)

This second accordion in the Trip Planner will contain all the calculated fares for the specific itineraries. Said fares will be presented in a table with the information broken down by mode and also by fare type:

**General view**:<br/>
image

**Small screen**:<br/>
image

### DIFF
--- a/apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
+++ b/apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
@@ -1,0 +1,60 @@
+
+<div class="h2 m-trip-plan-farecalc__title">Fare Calculator</div>
+
+<div role="table" aria-label="A summary of the different fares by media for each itinerary" class="m-trip-plan-farecalc__table">
+  <div role="columnheader" class="m-trip-plan-farecalc__table-cell m-trip-plan-farecalc__subheader">One-Way Fare</div>
+  <div role="columnheader" class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">Base</div>
+  <div role="columnheader" class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">Recommended</div>
+  <div role="columnheader" class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">Reduced</div>
+
+  <%= for {_mode_key, fare_values} <- @fares do %>
+
+    <div role="cell" class="m-trip-plan-farecalc__table-cell m-trip-plan-farecalc__mode-name">
+      <%= SiteWeb.TripPlanView.format_mode(fare_values.mode) %>
+    </div>
+    <div role="cell" class="m-trip-plan-farecalc__table-cell">
+      <span class="m-trip-plan-farecalc__label">Base: </span>
+      <%= fare_values|>SiteWeb.TripPlanView.get_fare_by_type(:highest_one_way_fare)|> fare_cents()|>Format.price() %>
+    </div>
+    <div role="cell" class="m-trip-plan-farecalc__table-cell">
+      <span class="m-trip-plan-farecalc__label">Recommended: </span>
+      <% recommended = fare_values|>SiteWeb.TripPlanView.get_fare_by_type(:lowest_one_way_fare) %>
+      <% media = recommended |> SiteWeb.TripPlanView.filter_media() %>
+      <%= recommended |> fare_cents()|>Format.price() %> <%= SiteWeb.TripPlanView.format_media(media) %>
+    </div>
+    <div role="cell" class="m-trip-plan-farecalc__table-cell">
+      <span class="m-trip-plan-farecalc__label">Reduced: </span>
+      <%= fare_values|>SiteWeb.TripPlanView.get_fare_by_type(:reduced_one_way_fare)|> fare_cents()|>Format.price() %>
+    </div>
+
+  <% end %>
+
+  <div role="cell" class="m-trip-plan-farecalc__table-cell m-trip-plan-farecalc__transfer-note">
+    <% transfer_note = transfer_note(@itinerary) %>
+    <%= if transfer_note == nil do %>
+      Total
+    <% else %>
+      <%= transfer_note %>
+    <% end %>
+  </div>
+
+  <div role="cell" class="m-trip-plan-farecalc__table-cell">
+    <span class="m-trip-plan-farecalc__label">Base: </span>
+    <span>
+      <strong><%=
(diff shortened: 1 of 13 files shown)


apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -558,7 +558,7 @@ defmodule SiteWeb.TripPlanView do
     HTML.Tag.content_tag(
       :span,
       [
-        "Total may be less with ",
+        "Total, may be less with ",
```
arkadyan, line 561: Do we actually want the comma in the accordion, but not in the summary above? Maybe a question for Katrina? (I am probably not the person to ask about grammar questions…)

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -645,21 +652,16 @@ defmodule SiteWeb.TripPlanView do
     )
   end
 
-  @spec get_highest_one_way_fare(TripPlan.Itinerary.t()) :: integer
-  def get_highest_one_way_fare(itinerary) do
+  @spec get_one_way_total_by_type(TripPlan.Itinerary.t(), atom) :: integer
```
arkadyan, line 655: The `atom` type is fine, but if you wanted you could also define a new type to describe the options since they are so limited. Maybe something like:

    @type fare_type :: :highest_one_way_fare | :lowest_one_way_fare | :reduced_one_way_fare

apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
```diff
@@ -0,0 +1,61 @@
+
+<div class="h2 m-trip-plan-farecalc-title">Fare Calculator</div>
```
arkadyan, line 2: I think some of these class names could have the "element" portion separated out per our [BEM](http://getbem.com/naming/) standard. I.e. maybe this would be `m-trip-plan-farecalc__title`, and a couple examples from below would be `m-trip-plan-farecalc__table` and `m-trip-plan-farecalc__table-cell`.

apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
```diff
@@ -0,0 +1,61 @@
...
+  <div class="m-trip-plan-farecalc-table-cell m-trip-plan-farecalc-subheader">One-Way Fare</div>
+  <div class="m-trip-plan-farecalc-table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc-header">Base</div>
+  <div class="m-trip-plan-farecalc-table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc-header">Recommended</div>
+  <div class="m-trip-plan-farecalc-table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc-header">Reduced</div>
+
+  <%= for {_fare_key, fare_values} <- @fares do %>
+
+    <% mode = fare_values.mode %>
+
+    <div class="m-trip-plan-farecalc-table-cell bold">
```
arkadyan, line 14: This is a bit of a style question, but to me, it feels like the `bold` here and below aren't as semantically informative if they could be. Like maybe instead if we use a modifier, this one could be something like, `m-trip-plan-farecalc__table-cell--title`, and the one below might be, `m-trip-plan-farecalc__table-cell--total`.

apps/site/lib/site_web/views/trip_plan_view.ex
alicia-mbta: Ah, good point. I added the comma here because that's what was in the design but I will double check with Katrina.

apps/site/lib/site_web/views/trip_plan_view.ex
alicia-mbta: :thumbsup: Yes, makes sense, I didn't think of that.

apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
thecristen: Those examples are great. These conventions are also described [here](https://github.com/mbta/wiki/blob/4185fb315937c1370374a88f0555c32b6d44d090/website/development/guidelines/css_guidelines.md).

apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
```diff
@@ -0,0 +1,61 @@
+
+<div class="h2 m-trip-plan-farecalc__title">Fare Calculator</div>
+
+<div class="m-trip-plan-farecalc__table">
+  <div class="m-trip-plan-farecalc__table-cell m-trip-plan-farecalc__subheader">One-Way Fare</div>
```
thecristen, line 5: Not urgent, but thinking about accessibility here...  this should probably all be a `<table>`.

ingridpierre:
Looking really good so far, still reviewing however.

One thing for CR one-way fares is we don't recommend cash for CR. There are some fare surcharges if you pay with cash on board (not at an FVM).

Let's update from "CharlieTicket, Cash, or mTicket App" in the CR one way recommended column to "CharlieTicket or mTicket" which is what we say on other fares pages.

ingridpierre:
More language tweaks under one-way fares: 
-  "Local Bus" rather than "Bus (Local Bus)".
- "Inner Express Bus" instead of "Bus (Inner Express Bus)"
- "Outer Express Bus"  "Bus (Outer Express Bus)"

ingridpierre:
Here's a good corner case, involving two CR trains. In this case, I planned a trip from Swampscott to Providence:

image

I think the base fare is taking into account the CR trips happening in 2 different zone fares, but **the fare summary only shows zone 8 (south station - providence).** (north station - swampscott is zone 3). Can we show subsequent CR trains as different rows in the fare summary if they take place in different zones?

ingridpierre:
At medium screen sizes, right before the mobile breakpoint pretty much, the table wrapping makes it a bit hard to scan across the row. As much as I hate asking to vertically align text, is there a way we can vertically align the text to make this a little easier to scan:
image

apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
alicia-mbta: `<table>` wasn't quite working but I added roles for the divs instead.

arkadyan:
Looks good to merge when Ingrid gives the thumbs up.

ingridpierre:
@alicia-mbta the changes look great 👍 thanks!

arkadyan:
Closed during feature branch management. Recreate as https://github.com/mbta/dotcom/pull/513

arkadyan:
I mixed up the target feature branch. Recreated again as https://github.com/mbta/dotcom/pull/515

PR #518, opened 2020-06-12
👩‍👩‍👧‍👦 Crowding | Map Tooltip

thecristen:
#### Summary of changes
**Asana Ticket:** [👩‍👩‍👧‍👦 Crowding | Map Tooltip](https://app.asana.com/0/385363666817452/1178899641126298/f)

Just like you saw in the demo.
Also added some tests I should have added with my previous update.

---

### DIFF
--- a/apps/site/assets/ts/leaflet/__tests__/MapTest.tsx
+++ b/apps/site/assets/ts/leaflet/__tests__/MapTest.tsx
@@ -3,6 +3,7 @@ import { mount } from "enzyme";
 import Map from "../components/Map";
 import { MapData, MapMarker } from "../components/__mapdata";
 import getBounds from "../bounds";
+import { Marker } from "react-leaflet";
 
 /* eslint-disable @typescript-eslint/camelcase */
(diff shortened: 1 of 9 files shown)


apps/site/assets/ts/leaflet/__tests__/MapTest.tsx
```diff
@@ -73,4 +74,18 @@ it("it renders using the default center position", () => {
...
+    ...data,
+    markers: [{ ...marker, vehicle_crowding: "some_crowding" }]
+  };
+  const div = document.createElement("div");
+  document.body.appendChild(div);
+  const wrapper = mount(<Map mapData={dataWithCrowding} />, {
+    attachTo: div
+  });
+
+  expect(wrapper.find(Marker)).toHaveLength(1);
```
arkadyan, line 89: That's great, I didn't know you could just pass in a component instead of a selector!

ingridpierre:
@thecristen looks fab! 👍

PR #520, opened 2020-06-15
Trip Plan Language | Transfer

thecristen:
#### Summary of changes
**Asana Ticket:** [Trip Plan Language | Transfer](https://app.asana.com/0/385363666817452/1179317340719423/f)

Changes "Depart onto Transfer" to either "Transfer" or "Depart" depending on whether the subsequent leg is transit or not, respectively.

---

### DIFF
--- a/apps/site/lib/site/trip_plan/itinerary_row_list.ex
+++ b/apps/site/lib/site/trip_plan/itinerary_row_list.ex
@@ -49,9 +49,9 @@ defmodule Site.TripPlan.ItineraryRowList do
         ]
   defp get_rows(itinerary, deps, opts, alerts) do
     rows =
-      for leg <- itinerary do
+      for {leg, index} <- Enum.with_index(itinerary.legs) do
         leg
-        |> ItineraryRow.from_leg(deps)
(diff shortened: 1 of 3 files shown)


apps/site/lib/site/trip_plan/itinerary_row_list.ex
```diff
@@ -49,9 +49,9 @@ defmodule Site.TripPlan.ItineraryRowList do
         ]
   defp get_rows(itinerary, deps, opts, alerts) do
     rows =
-      for leg <- itinerary do
+      for {leg, index} <- Enum.with_index(itinerary.legs) do
         leg
-        |> ItineraryRow.from_leg(deps)
+        |> ItineraryRow.from_leg(deps, Enum.at(itinerary.legs, index + 1))
```
alicia-mbta, line 54: Is there always going to be a `index + 1` element? or will we run into 'out of bounds' scenarios at some point?

thecristen: No and yes! `Enum.at` for an out of bounds index should return `nil` in that case.

ingridpierre:
👍 lgtm!

ingridpierre:
lgtm! 👍

PR #521, opened 2020-06-15
Display shuttle fares properly

arkadyan:
#### Summary of changes
**Asana Ticket:** [Fare Calculator Accordion | Shuttle Fare (free)](https://app.asana.com/0/555089885850811/1178333054166489)

Display shuttle fares properly

In the summary, the Monthly Pass is called "Shuttle: None".

Include a row for the shuttle fare in the accordion section.

Use "None" in recommended and reduced columns.

Display fare name as "Shuttle".

Screen Shot 2020-06-15 at 12 26 28

Screen Shot 2020-06-15 at 12 26 51

---

### DIFF
--- a/apps/fares/lib/format.ex
+++ b/apps/fares/lib/format.ex
@@ -94,7 +94,8 @@ defmodule Fares.Format do
   def name({:zone, zone}), do: "Zone #{zone}"
   def name({:interzone, zone}), do: "Interzone #{zone}"
   def name(:foxboro), do: "Foxboro Special Event"
-  def name(:free_fare), do: "Free Fare for SL1 Trips from Airport Stops"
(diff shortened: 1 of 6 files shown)


ingridpierre:
In the shuttle only case where we show a shuttle:

**CURRENT**
image

Instead of base: [blank]; Reduced: n/a; Valid On [blank]

Show "None" for Base, Reduced. And a blank for valid on or remove the column entirely.

Also, I think we have n/a as a value for a few other cases (like when something gets no reduced fares, or no reduced commuter rail passes), but it seems like "None" is more clear and universal.

arkadyan:
@ingridpierre Added "None" to both columns. This addresses other conditions "n/a" used to appear as well.

Screen Shot 2020-06-17 at 11 17 42

ingridpierre:
👍 lgtm!

PR #524, opened 2020-06-16
Fix error when bad data is sent to FinderApi.trip/2

arkadyan:
#### Summary of changes
**Asana Ticket:** [🐞 Elixir.Phoenix.ActionClauseError: no function clause matching in SiteWeb.ScheduleController.FinderApi.trip/2](https://app.asana.com/0/385363666817452/1180647194087413)

Fix error when bad data is sent to FinderApi.trip/2.

---


apps/site/lib/site_web/controllers/schedule/finder_api.ex
```diff
@@ -138,6 +138,13 @@ defmodule SiteWeb.ScheduleController.FinderApi do
     end
   end
 
+  def trip(conn, _) do
+    conn
+    |> put_resp_content_type("application/json")
+    |> put_status(400)
+    |> json(%{error: "Bad Request"})
```
thecristen, line 145: Maybe consider reusing [one of these functions](https://github.com/mbta/dotcom/blob/master/apps/site/lib/site_web/controllers/places_controller.ex#L79), or at least using the `:bad_request` atom?

arkadyan: Oh, that's great, thanks! I'm going to start with the atom just to get this out quickly and then come back with a new PR standardizing those helper functions with tests.

PR #526, opened 2020-06-17
Add castore dependency

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Results on the search page can't be clicked](https://app.asana.com/0/555089885850811/1179421898312873)

Re-enabling Algolia tracking and adding CaStore dependency to check if that's a viable solution.


paulswartz:
You also need to configure HTTPoison/Hackney to use this store: see the link I posted in Asana.

apps/v3_api/mix.exs
```diff
@@ -52,7 +52,8 @@ defmodule V3Api.Mixfile do
       {:json_api, in_umbrella: true},
       {:util, in_umbrella: true},
       {:sentry, "~> 7.0"},
-      {:exvcr_helpers, in_umbrella: true, only: :test}
+      {:exvcr_helpers, in_umbrella: true, only: :test},
+      {:castore, "~> 0.1.0"}
```
paulswartz, line 56: This should be part of the `algolia` app, since that's where it's being used.

PR #529, opened 2020-06-18
Avoid redirect for trips

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Elixir.Phoenix.ActionClauseError: no function clause matching in SiteWeb.ScheduleController.FinderApi.trip/2](https://app.asana.com/0/555089885850811/1180647194087413)

Scenario: obtain trip information for a trip e.g. for ferry `https://www.mbta.com/schedules/finder_api/trip?id=Boat-F1-1000-Long-Sunday&route=Boat-F1&date=2020-08-23&direction=0&stop=Boat-Hingham`

When there is no trip information, a redirection is made with a new URL containing parameter `origin` (e.g. `https://www.mbta.com/schedules/finder_api/trip?origin=Boat-Hingham`). The `trip` function doesn't expect such parameter so it is sending errors to Sentry. This is a possible solution to just return an empty trip instead.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/trip_info.ex
+++ b/apps/site/lib/site_web/controllers/schedule/trip_info.ex
@@ -7,9 +7,7 @@ defmodule SiteWeb.ScheduleController.TripInfo do
   """
   @behaviour Plug
   alias Plug.Conn
-  import Plug.Conn, only: [assign: 3, halt: 1]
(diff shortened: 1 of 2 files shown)


alicia-mbta:
Yes, the check is already there in https://github.com/mbta/dotcom/blob/master/apps/site/lib/site_web/controllers/schedule/finder_api.ex#L121

PR #530, opened 2020-06-18
Remove realtime endpoint version checking

arkadyan:
Reverts mbta/dotcom#416

No ticket. This check was added to ensure that there weren't problems with old versions of the client-side code calling the new version of the server-side code. I think we are well past the point of needing to worry about that.


alicia-mbta:
Just to make sure, is this to remove an unnecessary check for v2 since that is already the default version?

arkadyan:
@alicia-mbta Exactly. I think the idea was if you loaded the page right before the "v2" deploy, your page might later ask for an update and the server post update would respond with something the code in your browser suddenly wasn't expecting. But at this point, I don't think we need to worry about that condition any longer.

PR #531, opened 2020-06-18
Adjust ferry test data to match upcoming limited schedule

arkadyan:
No ticket, fixing breaking tests.

---

### DIFF
--- a/apps/site/test/site_web/controllers/fare_controller_test.exs
+++ b/apps/site/test/site_web/controllers/fare_controller_test.exs
@@ -31,21 +31,21 @@ defmodule SiteWeb.FareControllerTest do
       assert response =~ "Hingham"
     end
 
-    # Commenting out this test temporarily.
-    # As of Summer 2020 the limited service does not include this ferry.
-    # test "renders Georges Island ferry when present in the data", %{conn: conn} do
-    #   conn =
-    #     get(
-    #       conn,
-    #       fare_path(conn, :show, :ferry, origin: "Boat-George", destination: "Boat-Logan")
-    #     )
+    @tag skip:
+           "Commenting out this test temporarily. As of Summer 2020 the limited service does not include this ferry."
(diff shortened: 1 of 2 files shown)


apps/stops/test/route_stop_test.exs
```diff
@@ -188,18 +188,20 @@ defmodule Stops.RouteStopTest do
...
-      stops = make_stops(~w(Boat-Long Boat-Logan Boat-Hull Boat-Rowes Boat-Hingham))
-      route = %Route{id: "boat", type: 4}
-      actual = list_from_shapes([primary, other], stops, route, 1)
-
-      assert_stop_ids(
-        actual,
-        ~w(Boat-Long Boat-Logan Boat-Hull Boat-Rowes Boat-Hingham)
-      )
-    end
+    # Commenting out this test temporarily.
```
thecristen, line 191: Would it make sense to use the [skip tag](https://hexdocs.pm/ex_unit/ExUnit.Case.html#module-known-tags) here? (e.g. `@tag skip: "As of Summer 2020 the limited service does not include multiple shapes for this ferry."`)

apps/site/test/site_web/controllers/fare_controller_test.exs
```diff
@@ -18,28 +18,34 @@ defmodule SiteWeb.FareControllerTest do
...
     end
 
-    test "renders Georges Island ferry when present in the data", %{conn: conn} do
-      conn =
-        get(
-          conn,
-          fare_path(conn, :show, :ferry, origin: "Boat-George", destination: "Boat-Logan")
-        )
+    # Commenting out this test temporarily.
+    # As of Summer 2020 the limited service does not include this ferry.
```
thecristen, line 35: Same comment about the skip tag here

apps/stops/test/route_stop_test.exs
arkadyan: Oh that is better for sure, I didn't know about that.

PR #532, opened 2020-06-19
Add fare calculator widget to fares CMS pages: the prequel

thecristen:
#### Summary of changes
**Asana Ticket:** [🎫 Trip Planner | Add fare calculator widget to fares CMS pages](https://app.asana.com/0/385363666817452/1162903717356815/f)

This change, in tandem with [this change to the CMS](https://github.com/mbta/cms/pull/339), will enable content authors to add the existing trip planner widget to their pages, with the ability to customize text if desired. The initial use case is for adding a Fare Calculator to the sidebar of certain fares pages (example below):

image

image

---

### DIFF
--- a/apps/cms/lib/partial/paragraph/trip_plan_widget.ex
+++ b/apps/cms/lib/partial/paragraph/trip_plan_widget.ex
@@ -1,0 +1,32 @@
+defmodule CMS.Partial.Paragraph.TripPlanWidget do
+  @moduledoc """
+  Represents a widget on the page leading to the Trip Planner.
+  Is also used for the Fare Calculator widget.
+  """
+  import CMS.Helpers,
+    only: [
+      field_value: 2
+    ]
+
+  @type t :: %__MODULE__{
+          title: String.t() | nil,
+          text: String.t() | nil,
+          button_text: String.t() | nil,
+          right_rail: boolean
+        }
+
+  defstruct title: "Plan a trip",
+            text: nil,
+            button_text: "Get trip suggestions",
+            right_rail: true
+
+  @spec from_api(map, Keyword.t()) :: t
+  def from_api(data, _preview_opts \\ []) do
+    %__MODULE__{
+      title: field_value(data, "field_widget_title"),
+      text: field_value(data, "field_widget_text"),
+      button_text: field_value(data, "field_widget_submit_text"),
+      right_rail: field_value(data, "field_right_rail")
+    }
+  end
+end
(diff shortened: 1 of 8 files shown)


apps/cms/lib/partial/paragraph/trip_plan_widget.ex
```diff
@@ -0,0 +1,32 @@
...
+    ]
+
+  @type t :: %__MODULE__{
+          title: String.t() | nil,
+          text: String.t() | nil,
+          button_text: String.t() | nil,
+          right_rail: boolean
+        }
+
+  defstruct title: "Plan a trip",
```
amaisano, line 18: Defaults can be setup in Drupal, too, if you want to save hard coding it (admins would see this default text initially and change if the wanted to).

apps/site/lib/site_web/views/paragraph_view.ex
```diff
@@ -99,6 +100,26 @@ defmodule SiteWeb.CMS.ParagraphView do
     )
   end
 
+  def render_paragraph_content(%TripPlanWidget{} = paragraph, conn) do
```
amaisano, line 103: Have you considered just using the default render_paragraph_content clause, and then sending the values to the trip planner widget via template? These custom clauses were made because something about the paragraph needed to exist "above" the paragraph type template, but it seems here you're just extracting field values to send to a template.

`render_paragraph_content` -> paragraph_type_template -> render tripplanwidget w/content->field vals

thecristen: This is a good point, I will look into this!

thecristen:
I made an example custom trip plan widget based fare calculator widget in the sandbox CMS and have placed it on the dev green Ferry fares page: https://green.dev.mbtace.com/fares/ferry-fares - @ingridpierre feel free to review based on appearance/interaction rather than content 😉 

image

ingridpierre:
@thecristen I think these inputs should include Algolia autocomplete suggestions like they do on the trip planner or the /stops page. As-is, the server-side (?) location results don't tend to suggest things that we definitely should be suggesting on these pages. For example,  typing "Hull" >> gets you "Hull, UK".

ingridpierre:
Works beautifully, the Algolia inputs make a ton of difference here. 👍 

If I understand, we'll next have to let the content team know how to use it and when it gets merged right? [Would it make sense to submit a ticket to webcontent?](
 https://form.asana.com/?hash=d045af16e90907762b08e142dc0809e2abf58ecfa5bad4e5ea02caa156496d75&id=1117064241406883)

Not sure if the next steps or how to add the widget in the CMS are already documented somewhere?

thecristen:
@ingridpierre Yep, already reviewed and merged (as of earlier today) on the CMS/content side, so here goes!

PR #534, opened 2020-06-23
Include crowding data in data from time_data_for_route_by_stop

arkadyan:
#### Summary of changes
**Asana Ticket:** [👩‍👩‍👧‍👦 Crowding | Get crowding data given a prediction](https://app.asana.com/0/555089885850811/1180639549267087)

Include crowding data in data from time_data_for_route_by_stop

Adjust client-side to match new data structure.

Remove an unused function.

---

### DIFF
--- a/apps/site/assets/ts/__v3api.d.ts
+++ b/apps/site/assets/ts/__v3api.d.ts
@@ -1,3 +1,5 @@
+import { CrowdingType } from "./schedule/components/__schedule";
+
 export interface Direction {
   direction_id: DirectionId;
   headsigns: Headsign[];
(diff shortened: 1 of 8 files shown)


arkadyan:
@alicia-mbta In addition to the standard review, does this data structure look useful for the work you've started?

alicia-mbta:
@arkadyan I think it looks pretty good!

I have a question, should interface `Prediction` also contain crowding information? I'm referring to https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/schedule/components/__trips.d.ts#L36
In [`TripDetails.tsx`](https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/schedule/components/schedule-finder/TripDetails.tsx), `TripInfo` is the only structure available and we have to show crowding there too.

arkadyan:
@alicia-mbta Ugh, ok. If I'm following things correctly, it looks like the data from `/schedules/finder_api/trip` that `TripDetails` uses is completely disconnected from the data from `/schedules/line_api/realtime` that I've worked on here and is used by `LineDiagram`. 

What if I make a new ticket for the `/schedules/finder_api/trip` data? That would at least let this bit get merged and unblock some of the other ticket(s) presumably? We should also see what we can do to unify/clarify all these data sources, but that might be an ongoing project.

alicia-mbta:
Yes, that sounds reasonable. I'll take a look too to see if there are ways to reconcile the data.

arkadyan:
Add a follow-up ticket here: https://app.asana.com/0/555089885850811/1181808239433833

PR #535, opened 2020-06-23
Cross out skipped/cancelled stops (all modes)

thecristen:
#### Summary of changes
**Asana Ticket:** [🚍Schedule Finder | Cross out skipped/cancelled stops (all modes)](https://app.asana.com/0/385363666817452/1173665781696615/f)

Uses the existing but neglected `schedule_relationship` field on the prediction to identify skipped or cancelled stops. Also adds a short message for screen reader users: `This trip skips {schedule.stop.name}.`

I'm having trouble testing it on other modes, but the example below is from the 201 bus:

image

---

### DIFF
--- a/apps/site/assets/ts/schedule/components/schedule-finder/TripStop.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/TripStop.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from "react";
-import { TripDeparture } from "../__trips";
+import { TripDeparture, Prediction } from "../__trips";
 import { breakTextAtSlash } from "../../../helpers/text";
+import { alertIcon } from "../../../helpers/icon";
 
 interface Props {
   departure: TripDeparture;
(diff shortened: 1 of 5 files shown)


apps/site/assets/ts/schedule/components/schedule-finder/TripStop.tsx
```diff
@@ -41,6 +41,9 @@ const formattedDepartureTimes = (
 
   return (
     <span className={skippedOrCancelled ? "strikethrough" : ""}>
+      {skippedOrCancelled && (
+        <span className="sr-only">This trip skips {schedule.stop.name}.</span>
```
arkadyan, line 45: 👍

apps/site/assets/ts/schedule/components/schedule-finder/TripStop.tsx
```diff
@@ -14,24 +14,39 @@ const formattedDepartureTimes = (
...
-
-    return <>{schedule.time}</>;
+  if (
+    routeType === 2 &&
+    delay &&
+    delay >= 300 &&
+    prediction &&
+    prediction.time
+  ) {
+    return (
```
arkadyan, line 31: Do we want these times to also be crossed out if skipped or canceled? Or maybe just have an additional condition so this branch is never shown if the stop is skipped or canceled? Maybe a question for @ingridpierre?

thecristen: I can try to look at this this weekend during some expected diversions on Rockport, Kingston, and Middleborough -- this will be somewhat hard to test otherwise (we need to observe an upcoming _and_ delayed commuter rail train that will pass some skipped/cancelled stops). 👯‍♀️

arkadyan: @ingridpierre What do you think would be the best behavior here?

Personally I think the code is straightforward enough that I'm not so worried about seeing it live with actual data. We can always put together a simple test case to prove the results are what we want.

ingridpierre: Paraphrasing my convo with Cristen in Slack. This particular case had me thinking we're potentially crossing out the wrong element. Since we cross out arrival times for CR when they are adjusted, not when they are skipped, you're right to flag that this may need a rethink.

I'm now of the mind that it might make more sense instead to cross out the stop name also/instead, and also apply the sr-only skipped stop message to that. It may flow more naturally on a screenreader as well that way. I had chosen to cross out arrivals initially to mirror what was done for the schedules tab, but that pattern doesn't hold here and it's worth trying it out something else.

Cristen has said she'll test it out and we can take a look.

thecristen:
Updated! This simplifies the logic a fair bit. It looks like this:

image

ingridpierre:
@thecristen sorry to make you tweak without a visual comp to follow. This looks better overall to me, just want to tweak:

1) add a little more space between the alert icon and the crossed out stop
2) smaller alert icon overall to be closer to the size of the text.

ingridpierre:
It looks good where I can see it, for example on the 201 you gave.

The SL4 outbound detour at Washington St @ Herald isn't showing the skipped stop the way I would have expected though. Any thoughts about why we might see this discrepancy?

image

thecristen:
@ingridpierre something in the data. Looking up the trips in the API, it seems the `schedule_relationship` is not marked as skipped or cancelled for that stop (no idea why).

PR #542, opened 2020-06-24
Add crowding information to the Line Diagram

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [👩‍👩‍👧‍👦 Crowding | Schedule Pages, Line Diagram](https://app.asana.com/0/555089885850811/1177064032714383)

For **bus** routes that contain crowding information, it will now be displayed with the corresponding color legend:<br/>
image


arkadyan:
Temporarily blocked by https://app.asana.com/0/555089885850811/1180639549267087 having to be reverted.

ingridpierre:
Looks like they gray used in these icons is a little too light. We have quite a few grays defined in several places, so I've had a tough time keeping them documented myself. If there's an existing grey in our website colors that's a little darker or closer to the mockup - let's use that.

ingridpierre:
I'm not seeing crowding info next to predictions on routes that should be showing it. Here for example on the 111 and 77 which are both routes that were added recently. Do we have a configuration somewhere we need to add routes to the list or should that happen automatically? See:

image
image

alicia-mbta:
Those two routes had not been included in the list. It's done now in #564 (which needs to be merged before you look at this again).

apps/site/assets/css/_schedule-page-line-diagram.scss
```diff
@@ -141,6 +141,14 @@ $line-width: 8px; // width of the colored line in the diagram
   }
 }
 
+.m-schedule-table-crowding {
+  margin-left: .5rem;
+}
+
+.m-schedule-diagram__prediction-crowding span {
+  vertical-align: text-top;
```
thecristen, line 149: You might want to try using the `icon-size-inline` mixin instead to get the desired alignment, like `@include icon-size-inline(1.125em, .21em);` but with better values

apps/site/assets/css/_schedule-page-line-diagram.scss
```diff
@@ -141,6 +141,14 @@ $line-width: 8px; // width of the colored line in the diagram
   }
 }
 
+.m-schedule-table-crowding {
```
thecristen, line 144: Is this class needed, as opposed to putting this rule on `.m-schedule-diagram__prediction-crowding`?

alicia-mbta: You're right, it's a bit superfluous.

ingridpierre:
👍 lgtm!

PR #544, opened 2020-06-25
Add new routes with crowding data

arkadyan:
#### Summary of changes
**Asana Ticket:** [9 new routes with crowding data](https://app.asana.com/0/555089885850811/1181885620584410)

Add new routes with crowding data

---

### DIFF
--- a/apps/routes/lib/route.ex
+++ b/apps/routes/lib/route.ex
@@ -51,7 +51,7 @@ defmodule Routes.Route do
   @silver_line ~w(741 742 743 746 749 751)
   @silver_line_set MapSet.new(@silver_line)
 
-  @routes_with_occupancy ~w(1 15 16 22 23 31 32 109 110)
+  @routes_with_occupancy ~w(1 15 16 21 22 23 31 32 57 66 70 71 73 86 104 109 110)
   @routes_with_occupancy_set MapSet.new(@routes_with_occupancy)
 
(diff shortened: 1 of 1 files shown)


paulswartz:
What is this used for? Downstream applications shouldn't need to care about which routes have been approved for crowding information.

arkadyan:
@paulswartz We want to shift the design in some cases based on whether a route you are looking at might have crowding data. E.g., If you are looking at a route with crowding data, we will display a column in a table for a crowding icon. If a specific vehicle isn't serving up data we can simply leave that column blank, but the column is still there. On the other hand, on a route without crowding data that will never show that icon, it is preferable to remove the column altogether.

arkadyan:
Holding for 6/26

paulswartz:
> @paulswartz We want to shift the design in some cases based on whether a route you are looking at might have crowding data. E.g., If you are looking at a route with crowding data, we will display a column in a table for a crowding icon. If a specific vehicle isn't serving up data we can simply leave that column blank, but the column is still there. On the other hand, on a route without crowding data that will never show that icon, it is preferable to remove the column altogether.

Can you base it on whether there's any crowding data for the vehicles being displayed, rather than the route? That's similar to what Skate is doing.

arkadyan:
@paulswartz That's an interesting idea that I can't remember if we've thought about. I'll make a ticket for the team to consider.

PR #550, opened 2020-06-29
Update test details to match the current schedule

arkadyan:
No ticket, tests were broken.

Update test details to match the current schedule

---


alicia-mbta:
Just curious, why has this data changed? Because of changes in this particular Worcester train?

arkadyan:
Yes, there was a GTFS change late last week to the Worcester schedule.

PR #554, opened 2020-06-30
Adjust Trip Planner form initialization to handle more than one form per page

thecristen:
#### Summary of changes
**Asana Ticket:** [Adjust Trip Planner form initialization to handle more than one form per page](https://app.asana.com/0/555089885850811/1182407113686713/f)

Introduced as a pre-requisite for #532, where CMS pages will add a right-railed paragraph to a page twice, but the nice autocomplete / dropdown suggesting popular places only worked as expected in the first form instance.

The meat of the change:

* Add `containerEl` option to `TripPlannerLocControls`
* Add `getById()` helper to get DOM elements within a `containerEl` (or fallback to `document.getElementById`, which is what was used before)
* Adjust the `AlgoliaAutocomplete` classes to use the this new option and helper method
* Add some tests for `TripPlannerLocControls`

---


arkadyan:
Actually, I suppose `AlgoliaAutocompleteWithGeo` and `AlgoliaAutocomplete` don't necessarily need to change as long as TripPlannerLocControls never repeats `id`s it gives them.

apps/site/assets/js/algolia-autocomplete.js
```diff
@@ -6,23 +6,22 @@ import * as QueryHelpers from "../ts/helpers/query";
 /* eslint-disable class-methods-use-this */
 
 export default class AlgoliaAutocomplete {
-  constructor({ id, selectors, indices, parent }) {
+  constructor({ id, selectors, indices, parent, containerEl = null }) {
```
arkadyan, line 9: It would be good to test that `AlgoliaAutocomplete` can optionally accept `containerEl` as a property, as well as testing how that affects its behavior. (As a unit test, that is. I see that it is triggered indirectly via `trip-plan_test.js` as well.)

apps/site/assets/js/algolia-autocomplete.js
```diff
@@ -31,6 +30,12 @@ export default class AlgoliaAutocomplete {
     this.bind();
   }
 
+  getById(id) {
```
arkadyan, line 33: It would be good to test this method as well.

apps/site/assets/js/test/trip-plan_test.js
```diff
@@ -95,4 +96,117 @@ describe("trip-plan", () => {
       assert.equal($from_lng.val(), 4);
     });
   });
+
+  describe("TripPlannerLocControls", () => {
```
arkadyan, line 100: Would it make sense for this to live in a `trip-planner-location-controls_test.js` in order to match the structure of the implementation file?

thecristen:
I definitely considered refactoring these so that we don't end up with multiple instances of the same ID on the page, and I do think that'd be a good thing to do in a future refactor.

As for now, I didn't want to get too far into editing templates, and there were so many places where selectors were passed in as arguments, and places that relied on the selector returning one DOM element (as opposed to multiple elements or a NodeList) that I wasn't comfortable making the more substantive changes across all of the classes.

apps/site/assets/js/test/trip-plan_test.js
thecristen: It would! But `TripPlannerTimeControls` is also tested in this file instead of a standalone one, so, chose to follow that.

arkadyan:
Ok, limiting changes for now, especially to get unblocked makes sense. When you get a chance, could you please write up what needs to be done in a further refactor in a ticket?

thecristen:
@arkadyan I've added the requested tests! 

To use a particular function (Element.closest) I needed to update `mocha-jsdom`, so that's included in its own commit.

apps/site/assets/js/test/trip-plan_test.js
```diff
@@ -136,6 +144,25 @@ describe("trip-plan", () => {
         `);
       });
 
+      it("gets element with getById", () => {
```
arkadyan, line 147: To me as an outsider, I think this could include more about the query selector in the description, like maybe appending something along the lines of, "from within the container element".

PR #556, opened 2020-07-06
Show crowding information in Upcoming Departures

alicia-mbta:
#### Summary of changes
**Asana Tickets:**
[👩‍👩‍👧‍👦 Crowding | Upcoming Departures](https://app.asana.com/0/555089885850811/1180063111128636)
[👩‍👩‍👧‍👦 Crowding | Upcoming Departures, Trip Details](https://app.asana.com/0/555089885850811/1180063111128640)

Information about crowding in vehicles is now displayed in the Upcoming Departures when clicking on a specific stop in the Line Diagram:
image
image


apps/site/assets/ts/schedule/components/schedule-finder/TripDetails.tsx
```diff
@@ -66,9 +67,18 @@ export const TripDetails = ({
...
 
+  const crowding = tripInfo.vehicle ? tripInfo.vehicle.crowding : null;
+
   return (
     <table className="trip-details-table">
       <thead>
+        {crowding && (
+          <tr>
+            <th>
+              <CrowdingPill crowding={crowding} />
```
arkadyan, line 78: Could you please add a quick test for rendering this (i.e. mounting `TripDetails` for a trip with crowding data and checking for the pill)?

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
```diff
@@ -25,19 +28,23 @@ interface Props {
   input: UserInput;
 }
 
+interface AccordionProps {
+  state: State;
+  journey: EnhancedJourney;
+  contentComponent: () => ReactElement<HTMLElement>;
+}
+
+type fetchAction =
```
arkadyan, line 37: Could this please be capitalized to distinguish it as a type?

```suggestion
type FetchAction =
```

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
```diff
@@ -48,17 +55,28 @@ export const RoutePillSmall = ({
...
-  const { trip, route, realtime } = journey;
+  const { trip, route, realtime, tripInfo } = journey;
+
+  // Only display the crowding information if the trip ID of the vehicle matches the trip ID of the prediction being displayed.
+  let showCrowding = false;
+  let crowding = "crowding_unavailable";
+  let crowdingText = "";
+
+  if (tripInfo.vehicle) {
+    showCrowding = tripInfo.vehicle.trip_id === trip.id;
```
arkadyan, line 72: I'm wondering if it's possible to simplify these conditionals and ideally not have to use `let`s. Below there's another check for `showCrowding && crowdingText !== ""`. Am I correct in thinking that `crowdingText` would be `""` in the case where the `vehicle` in `tripInfo` has `null` for `crowding`? If I've got that right, would it be possible to change this line to be something along the lines of, `showCrowding = tripInfo.vehicle.trip_id === trip.id && tripInfo.vehicle.crowding;`? That might allow you to just use `crowdingDescriptions` safely too?

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
```diff
@@ -152,37 +218,77 @@ const UpcomingDeparturesHeader = (
...
+    journeys
+      .filter(journey => journey.realtime.prediction !== null)
+      .map(async journey => {
+        // Sometimes using the route.id from the journey is desired over the route.id
+        // from input (e.g. so we can get trips for "Green-D" instead of "Green")
+        const adjustedRoute = journey.route.id;
+
+        const tripId = journey.trip.id;
+
+        const tripInfo = await await fetch(
```
arkadyan, line 240: It seems like `TableRow` is already requesting this data. Does that mean that we are now requesting it twice? If I'm getting that right that doesn't seem like something we'd want.

I'm trying to follow the changes in the diff syntax… is this data required for something at this level that can't be done inside of `TableRow`?

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
```diff
@@ -114,24 +150,54 @@ const CrTableRow = ({
...
     <Accordion
-      input={getAdjustedInput(input, journey)}
+      state={state}
+      journey={journey}
+      contentComponent={contentComponent}
+    />
+  );
+};
+
+const Accordion = ({
```
arkadyan, line 180: I'm seeing some more things I didn't follow completely at first that are confusing me. What's the need for copying this `Accordion` instead of continuing to use the shared version? That feels like a red flag.

And relatedly, I like the idea of moving `AccordionRow` to its own file given that it's used in multiple places, but can it be the entire `Accordion` component that is moved into its own file (with `AccordionRow` as an internal sub-component)? That feels like a better level of abstraction for what these components that use the Accordion should need to care about.

ingridpierre:
This style wasn't apparent to me before, but the black border on the "not crowded" pill is a little heavy, let's change to one of our other greys: #B0B5C0

image

---
Similar to my comment in #542 , the grey for people icons that are "empty" should be #B0B5C0 as well.

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
alicia-mbta: Before this PR: for upcoming departures, the trip details call (which contains the crowding information) is made when expanding the row.
Since we need the crowding information before even expanding in order to show it in the row header, the big change in this PR is to make the above calls at the time when we retrieve the list of upcoming departures.
If you look at the network calls in the browser's devtools' network tab, expanding the rows will not make extra calls.
The trip information is already available and it's simply displayed.
This behavior only applies to Upcoming Departures, the behavior hasn't changed for the service selector (bottom part in the same modal).
Let me know if this makes sense or there's something still unclear.

arkadyan: Thanks for the explanation. I think I was confusing the other `TableRow` component for the one that is local in this file, but makes more sense now.

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
alicia-mbta: Because of the changes in this PR, although they look the same, we have now two _different_ `Accordion`s (from a behavior standpoint): one that just displays the data (i.e. the one in Upcoming Departures) and one that fetches the data when it's being expanded (i.e. the one in the Daily Schedule).
That is why the piece that I extracted was just the (accordion) row, because some props change.
I can take another look and see how I could extract the complete accordion.

arkadyan: Ok. Yeah, conceptually it seems like you could have an accordion component that didn't have any knowledge of the contents nested inside it (or concern with loading the data for those contents), but I understand that's not directly related to this work. If you think it might be helpful to try to walk through something like that together let me know, or if you think it would be better to consider that a follow-up item to think about later that seems fine too.

alicia-mbta: Indeed I'd like to do some refactoring with rows and accordions but ideally in a separate ticket if that makes sense.

arkadyan: Sounds like a plan.

ingridpierre:
Noticing an unexpected behavior. Clicking on the crowding icon in the accordion row both turns on the tooltip and opens the accordion. Is it possible to separate these actions so that clicking the icon shows/hides the tooltip only, and clicking anywhere else in the row will open the accordion?

ingridpierre:
Does this check whether the trip associated with the prediction is the same as the trip associated with the vehicle? Like in the description for [this ticket](https://app.asana.com/0/385363666817452/1177064032714383/f).

I'm seeing a couple far away predictions that have crowding info so I want to make sure we're not showing something that we don't want to.

alicia-mbta:
> Noticing an unexpected behavior. Clicking on the crowding icon in the accordion row both turns on the tooltip and opens the accordion. Is it possible to separate these actions so that clicking the icon shows/hides the tooltip only, and clicking anywhere else in the row will open the accordion?
> 
> I don't think this behavior is a blocker necessarily, but we should probably see if this is interfering with how a screenreader is interpreting the state of the accordion/tooltip in a potentially confusing way.

I'll see if I can get the tooltip to 'toggle' instead.

alicia-mbta:
> Does this check whether the trip associated with the prediction is the same as the trip associated with the vehicle? Like in the description for [this ticket](https://app.asana.com/0/385363666817452/1177064032714383/f).
> 
> I'm seeing a couple far away predictions that have crowding info so I want to make sure we're not showing something that we don't want to.

Yes, crowding information shows only if the id's of trip and vehicle are the same. If you're still seeing inconsistencies can you point me to any bus route where you're seeing this happening?

ingridpierre:
So far haven't seen any more far away predictions with crowding info, so I think we're good on that!

This lgtm 👍  visually, thanks @alicia-mbta

PR #557, opened 2020-07-07
Update tests to reflect changes in the Green Line C branch

alicia-mbta:
No ticket.

Update in tests to reflect changes in the Green Line C branch:
https://www.mbta.com/news/2020-07-02/reminder-building-better-t-green-line-c-branch-track-improvements-and-intersection

### DIFF
--- a/apps/site/test/site_web/controllers/schedule_controller_test.exs
+++ b/apps/site/test/site_web/controllers/schedule_controller_test.exs
@@ -205,7 +205,7 @@ defmodule SiteWeb.ScheduleControllerTest do
     end
 
     test "assigns schedules, frequency table, and origin for green line", %{conn: conn} do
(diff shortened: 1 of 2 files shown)


apps/site/test/site_web/controllers/schedule_controller_test.exs
```diff
@@ -204,6 +204,8 @@ defmodule SiteWeb.ScheduleControllerTest do
       refute :schedules in Map.keys(conn.assigns)
     end
 
+    @tag skip:
+           "Skipping this test temporarily. In Summer 2020 the Green Line C branch is undergoing improvements"
     test "assigns schedules, frequency table, and origin for green line", %{conn: conn} do
```
thecristen, line 209: Can this test be edited rather than skipped altogether? Maybe using `Green-D` instead?

alicia-mbta: Yes, that makes more sense 👍

PR #559, opened 2020-07-09
Hide realtime status if we don't have a stop name

arkadyan:
#### Summary of changes
**Asana Ticket:** [Polish | Map tooltip looks broken when trip assignment is missing](https://app.asana.com/0/555089885850811/1180626064629737)

Hide realtime status if we don't have a stop name in vehicle tooltips on the line diagram and map.

Screen Shot 2020-07-09 at 11 35 47

Screen Shot 2020-07-09 at 11 57 41

---


apps/stops/lib/stop.ex
```diff
@@ -35,7 +35,7 @@ defmodule Stops.Stop do
           id: id_t,
           parent_id: id_t | nil,
           child_ids: [id_t],
-          name: String.t(),
+          name: String.t() | nil,
```
thecristen, line 38: Not sure if this change is needed, given we appear to be working with an empty string in this case.

apps/site/assets/ts/schedule/components/line-diagram/VehicleIcons.tsx
```diff
@@ -30,10 +30,13 @@ const statusDescriptions = {
 
 const tooltipText = (
   routeType: RouteType | null,
-  stopName: string,
+  stopName: string | null,
   vehicle: LineDiagramVehicle
 ): string => {
-  const status = `${statusDescriptions[vehicle.status]} ${stopName}`;
+  const status =
+    stopName !== null && stopName.length
```
thecristen, line 37: Would `stopName && stopName.length` work here?

arkadyan: Yep! Updated

apps/stops/lib/stop.ex
arkadyan: I can't tell if we are ever getting `nil` in this instance, but it seemed possible and even if not if you look up at the defstruct `name` defaults to `nil`, so this seems more correct.

ingridpierre:
Functionally works as expected, I think our min width on this tooltip causes a strange appearance though. Is there a reason we can't just have a max width for long tooltips and let the smaller ones (in cases like this, esp. without the crowding pill) be small?

image

arkadyan:
@ingridpierre I removed a 320px min width we had set. Does this look better?

Screen Shot 2020-07-10 at 15 47 33

Screen Shot 2020-07-10 at 15 48 26

PR #560, opened 2020-07-09
Show route level alerts in the colored banner

thecristen:
#### Summary of changes
**Asana Ticket:** [🚍 Stop/Station/TNM | Show route level alerts in the colored banner](https://app.asana.com/0/385363666817452/1173665781696617/f)

* Extracted the header from the two `RouteCard` components (stop and transit near me) into a separate `RouteCardHeader` component, also refactoring it to show route level alerts as an icon linking to the route's alert tab.
* We already had the route-level alert information in the transit near me page - now we have such information (simplified) in the stop page data as well

Also tackled a bit of existing tech debt -- there was a lot of redundancy between the two RouteCard components (stop and transit near me), and I extracted/slightly refactored a few of the helper functions used in both. 

A few other things happened, isolated in specific commits:

- Added a CSS component for creating the special link-inside-link structure used here
- Added a new variable, `$mbta-line-colors`, as a Sass map to store MBTA line colors
- Added a Sass function to output colors from a given MBTA line
- Added a Sass mixin to use aforementioned function to adjust colors of SVG icons

---


ingridpierre:
I think we have in a few places (like the main commuter rail schedule page headers) something that help us break long commuter rail names after the slash. 

In this screenshot you can see how the break to wrap is a little awkward. I'd love it if it wrapped instead after "Framingham/":
image

ingridpierre:
Any idea why the detoured stop "Washington St @ Herald St" shows the alert icon, but the line diagram doesn't? Would like to fix this discrepancy either as part of this ticket or as separate follow-up.

**Alert on [stop page](http://localhost:4001/stops/49003):**
image
**No alert on [line diagram](http://localhost:4001/schedules/751/line):**

image

ingridpierre:
Route 23 detour is not showing in the alert banner:

thecristen:
If I had to guess it probably has to do with the particular priority/effect of alerts we're choosing to show, which might be different in different parts of the site.

For the changes I've made here to the stop page, it should be showing _route_ alerts with Priority 7 or higher in the colored banners. So the 23 screenshot could be correct if the detour alert is a lower priority, maybe?

For the line diagram, I'm happy to look into that, as we should probably be intentional on which priority/effect/etc we want to use..

ingridpierre:
@thecristen To see if I have this straight I've updated the documentation here:
https://zeroheight.com/2fedee66c/p/66d65d-alert-icons

For this ticket, we want to show alert icons in the colored banner and on the line diagram if the alert has a **severity >= 7 OR if the priority is high**. This is intended to show more alert icons in these places than if we were to rely on priority alone. 

Let me know if that makes sense with your updated approach here and what we discussed together (or not).

thecristen:
@ingridpierre Makes perfect sense here. 👍 I've updated the line diagram and colored banner on the stop and TNM pages to use this logic.

ingridpierre:
@thecristen As far, as I can tell using the alerts data in production, this is working really nicely! It also helped me uncover some potential bugs and missing alerts we may need to forward to the PIOs. So great work all around :)

👍  this lgtm!

PR #565, opened 2020-07-14
Remove filtering of bus routes for crowding purposes

alicia-mbta:
#### Summary of changes
**Asana Tickets:**
[👩‍👩‍👧‍👦 Crowding | Hide crowding column if no vehicles have crowding data](https://app.asana.com/0/555089885850811/1181994766435225)
[👩‍👩‍👧‍👦 Crowding | Schedule Pages, Line Diagram](https://app.asana.com/0/555089885850811/1177064032714383)
[👩‍👩‍👧‍👦 Crowding | Upcoming Departures](https://app.asana.com/0/555089885850811/1180063111128636)

Get rid of the list of bus routes from a crowding standpoint.

### DIFF
--- a/apps/routes/lib/route.ex
+++ b/apps/routes/lib/route.ex
@@ -49,9 +49,6 @@ defmodule Routes.Route do
   @silver_line ~w(741 742 743 746 749 751)
   @silver_line_set MapSet.new(@silver_line)
 
-  @routes_with_occupancy ~w(1 15 16 21 22 23 31 32 57 66 70 71 73 86 104 109 110)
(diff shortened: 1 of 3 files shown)


alicia-mbta:
At the risk of having been slightly bigger than desired (but having good intentions), I think all the crowding work should have been done in a single PR instead of having 3 parallel ones since they are dependent on each other and need to be merged at once to have consistent crowding information on dotcom.

To answer both questions, this PR is just to remove the restrictions in terms of the list of bus routes that have crowding information available. I was filtering already on the frontend (#542) but didn't remove the backend filter. I put the 3 tickets above because this change in a way affects crowding in general but let me know if it's confusing.

PR #566, opened 2020-07-15
Ensure that the map api doesn't 500 if given bad params

arkadyan:
#### Summary of changes
**Asana Ticket:** [🐞 Elixir.Phoenix.ActionClauseError: no function clause matching in SiteWeb.ScheduleController.MapApi.show/2](https://app.asana.com/0/555089885850811/1181695474176953)

Ensure that the map api doesn't 500 if given bad params

---

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/map_api.ex
+++ b/apps/site/lib/site_web/controllers/schedule/map_api.ex
@@ -6,6 +6,7 @@ defmodule SiteWeb.ScheduleController.MapApi do
 
   alias Leaflet.MapData
   alias Routes.Route
+  alias SiteWeb.ControllerHelpers
   alias SiteWeb.ScheduleController.Line.Helpers, as: LineHelpers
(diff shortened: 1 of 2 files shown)


thecristen:
Did you investigate why the bad params were given in the first place? Given the second argument as cited in the ticket, `%{"id" => "Mattapan&direction_id=0&"}`, it looks like the params are not being parsed correctly into a map.

arkadyan:
@thecristen You can trigger it by removing parts of the URL by hand—I presume something like that is what happened. This hasn't happened since (except by me), so it doesn't seem like the symptom of anything systemic.

thecristen:
> by removing parts of the URL by hand—I presume something like that is what happened

@arkadyan I'm not sure I'm convinced, as it's also called from `ScheduleDirection` in `fetchMapData` - it seems unlikely someone is manually editing a URL there. Anyhow, this change is still big progress, so I'll let it go..

PR #567, opened 2020-07-15
Do not show crowding if none of the Upcoming Departures for buses has that info

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [👩‍👩‍👧‍👦 Crowding | Hide crowding column if no vehicles have crowding data](https://app.asana.com/0/555089885850811/1181994766435225)

The column containing the crowding information will not be shown if there's not at least one vehicle with those values:

With crowding:
image

Without crowding:
image

### DIFF
--- a/apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement, useEffect, useReducer, useState } from "react";
+import { get, isEmpty } from "lodash";
 import {
   timeForCommuterRail,
   trackForCommuterRail,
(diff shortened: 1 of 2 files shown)


apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
```diff
@@ -274,7 +294,8 @@ export const upcomingDeparturesTable = (
                 Destinations
               </th>
               <th scope="col" colSpan={2} className="schedule-table__cell">
-                {isABusRoute(journeysWithTripInfo[0].route) ? (
+                {isABusRoute(journeysWithTripInfo[0].route) &&
```
arkadyan, line 297: It looks like the `isABusRoute` check is combined with the `someCrowdingInfoExists`  everywhere it's used. Would it makes sense to combine it, so `isABusRoute` is just part of what it means for `someCrowdingInfoExists`?

ingridpierre:
👍 lgtm!

apps/site/assets/ts/schedule/components/schedule-finder/UpcomingDepartures.tsx
alicia-mbta: It does make sense to combine that, I'll change it.

PR #568, opened 2020-07-16
🚍 Stops/Station/TNM | CR | Cross out skipped/cancelled stops

thecristen:
#### Summary of changes
**Asana Ticket:** [🚍 Stops/Station/TNM | CR | Cross out skipped/cancelled stops](https://app.asana.com/0/385363666817452/1175248604991146/f)

* Adds `schedule_relationship` to the predictions used on stop pages and transit near me
* Shows predictions with `schedule_relationship` of skipped or canceled as "Canceled" with strikethrough
* Shows the absence of a prediction as "Scheduled" instead of "On time". 
* Adjusts the `$text-muted` color to use a darker gray to ensure good contrast.

Stops

image

Transit Near Me

image

---

### DIFF
--- a/apps/site/assets/ts/components/Headsign.tsx
+++ b/apps/site/assets/ts/components/Headsign.tsx
@@ -118,9 +118,10 @@ const HeadsignComponent = (props: Props): ReactElement<HTMLElement> => {
         {routeType === 2 && renderTrainName(`Train ${headsign.train_number}`)}
       </div>
       <div className="m-tnm-sidebar__schedules">
-        {headsign.times.map((time, idx) =>
-          renderTime(time, headsign.name, routeType, idx)
-        )}
+        {headsign.times.map((time, idx) => {
+          if (routeType === 2 && idx > 0) return null; // limit to 1 headsign
+          return renderTime(time, headsign.name, routeType, idx);
+        })}
       </div>
     </div>
   );
--- a/apps/site/assets/ts/components/Headsign.tsx
+++ b/apps/site/assets/ts/components/Headsign.tsx
@@ -58,9 +58,9 @@ const renderTimeCommuterRail = (
     >
       {timeForCommuterRail(
         data,
-        `m-tnm-sidebar__time-number ${
+        `${
           status === "Canceled" ? "strikethrough" : ""
-        }`
+        } m-tnm-sidebar__time-number`
       )}
       <div className="m-tnm-sidebar__status">
         {`${status}${trackForCommuterRail(data)}`}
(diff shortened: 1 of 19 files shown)


ingridpierre:
How hard would it be to change the TNM logic for CR to only show one schedule at a time as part of this ticket? Showing two on a row is a regression similar to (but different from): https://app.asana.com/0/385363666817452/1184312880988849/f 

Notably, Train 512 is not the train number for both the listed scheduled trains here:
image

apps/site/assets/ts/components/Headsign.tsx
```diff
@@ -45,22 +45,29 @@ const renderTrainName = (trainName: string): ReactElement<HTMLElement> => (
...
+  return (
+    <div
+      className={`m-tnm-sidebar__time m-tnm-sidebar__time--commuter-rail ${modifier} ${
+        status === "Scheduled" ? "text-muted" : ""
+      }`}
+    >
+      {timeForCommuterRail(
+        data,
+        `m-tnm-sidebar__time-number ${
+          status === "Canceled" ? "strikethrough" : ""
```
arkadyan, line 62: I think a quick test for this visual behavior could be useful to prevent regressions.

thecristen:
Added the tests and adjusted the TNM logic for CR to only show one schedule at a time.
Saw some more incorrect CR train numbers this morning (including even a `Train null`!), so will file a new ticket for that.

ingridpierre:
Currently I'm seeing some unexpected displays here.

**Two disconnected train times, stacked on a departed status:**
image

----------
**Possible regression on late train styling?**
CURRENT 
Late trains >5 mins late are not appearing styled as they once were. Here is what I saw:
image
image

**DESIRED** appearance of late trains;
image

thecristen:
> Two disconnected train times, stacked on a departed status

This was actually a visual regression, thanks for catching it! I am fixing this in a moment.

> Late train styling

I'm not sure that's a regression -- how does it compare to what's shown on master? Our commuter status doesn't have any "Late" styling as far as I can tell, so maybe we'll have to add that (in a future ticket).

ingridpierre:
Looks like it is a regression, but I'm not sure how long it has been happening. On this ticket from the original TNM project we have delayed status styled differently in the route cards: https://app.asana.com/0/961607479498016/1100182096176852/f

**Delayed/Early**

- show the original scheduled time with a strikethrough
- show the new time
- show the Delayed/Early train status


--------


There is definitely a regression on the line diagram. Here's an old screenshot of production many months ago.
image

ingridpierre:
@thecristen Sorry for the goose chase on statuses/delayed/Late trips.

I'm still not able to see a canceled trip, but I do like the styling so far. The grey we're using on the 'Scheduled', but not predicted trips is a bit different than the other places we're using grey for schedules in the cards. Compare the grey for this scheduled bus:

image
image

thecristen:
@ingridpierre I see!
Looking at that scheduled bus time, I'm finding it has insufficient contrast (3.95 or 4.01 depending on which of my two tools you ask..) – so I am going to adjust it to use `$gray` instead of `$gray-light` (which will also make it match the 'Scheduled').

PR #571, opened 2020-07-21
Hide monthly pass sections for free Logan trips

arkadyan:
#### Summary of changes
**Asana Ticket:** [Trip Planner | SL Trips from Airport to South Station - Monthly Pass Fare](https://app.asana.com/0/555089885850811/1183525584398249)

Hide monthly pass sections for free Logan trips

This is specifically just for Silver Line only (no transfers) trips
starting at one of the Logan terminals.

Screen Shot 2020-07-21 at 11 40 37

---

### DIFF
--- a/apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
+++ b/apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
@@ -66,7 +66,7 @@
 </div>
 
 
-<%= if @show_monthly_passes? do %>
+<%= if show_monthly_passes?(@itinerary) do %>
   <div class="m-trip-plan-farecalc__table">
     <div class="m-trip-plan-farecalc__table-cell m-trip-plan-farecalc__subheader">
       Monthly Passes
--- a/apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
+++ b/apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
@@ -66,25 +66,27 @@
 </div>
 
 
-<div class="m-trip-plan-farecalc__table">
-  <div class="m-trip-plan-farecalc__table-cell m-trip-plan-farecalc__subheader">
-    Monthly Passes
-  </div>
-  <div class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">
-    Base
-  </div>
-  <div class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">
-    Reduced
-  </div>
-  <div class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">
-    Valid On
-  </div>
+<%= if @show_monthly_passes? do %>
+  <div class="m-trip-plan-farecalc__table">
+    <div class="m-trip-plan-farecalc__table-cell m-trip-plan-farecalc__subheader">
+      Monthly Passes
+    </div>
+    <div class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">
+      Base
+    </div>
+    <div class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">
+      Reduced
+    </div>
+    <div class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">
+      Valid On
+    </div>
 
-  <%= for pass <- Enum.uniq([@itinerary.passes.base_month_pass, @itinerary.passes.recommended_month_pass]) do %>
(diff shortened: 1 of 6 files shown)


apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
```diff
@@ -66,25 +66,27 @@
...
-  <div class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">
-    Base
-  </div>
-  <div class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">
-    Reduced
-  </div>
-  <div class="m-trip-plan-farecalc__table-cell m-trip-plan-results__itinerary-fare-title m-trip-plan-farecalc__header">
-    Valid On
-  </div>
+<%= if @show_monthly_passes? do %>
```
thecristen, line 69: It's a view function, so wouldn't it be accessible from this template doing something like so:

```suggestion
<%= if SiteWeb.TripPlanView.show_monthly_passes?(@itinerary) do %>
```

Then you wouldn't need to compute `@show_monthly_passes?` to pass it in via that [render_to_string](https://github.com/mbta/dotcom/pull/571/commits/b45efcde94d20773f2157ed14d476d8b5e99564b#diff-5306341842cc891dd7f7f0c58435c22fR617) call.

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -598,19 +599,27 @@ defmodule SiteWeb.TripPlanView do
...
       fares = get_calculated_fares(i)
 
       fare_calculator_html =
         "_fare_calculator.html"
-        |> render_to_string(itinerary: i, fares: fares, conn: conn)
+        |> render_to_string(
+          itinerary: i,
+          fares: fares,
+          conn: conn,
+          show_monthly_passes?: show_monthly_passes?
```
thecristen, line 621: See my comments elsewhere about possibly not needing this line addition.

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -786,4 +795,19 @@ defmodule SiteWeb.TripPlanView do
...
+  # Hide monthly pass sections in the case of a Silver Line trip with no transfers from Logain Airport.
+  # Public solely for testing.
+  @spec show_monthly_passes?(Itinerary.t()) :: boolean()
+  def show_monthly_passes?(itinerary), do: !sl_only_trip_from_airport?(itinerary)
+
+  @spec sl_only_trip_from_airport?(Itinerary.t()) :: boolean()
+  defp sl_only_trip_from_airport?(itinerary) do
+    transit_legs = Itinerary.transit_legs(itinerary)
+    first_leg = List.first(transit_legs)
+    route_id = first_leg.mode.route_id
```
thecristen, line 808: This will error if the itinerary happens to have no transit legs - please guard against this.

apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
thecristen, line 69: You might not need to compute and pass in a `show_monthly_passes?` value if you use the view function in the template.

```suggestion
<%= if SiteWeb.TripPlanView.show_monthly_passes?(@itinerary) do %>
```

thecristen:
Sorry about the redundant comments @arkadyan -- the page state here was being all weird on me.

apps/site/lib/site_web/views/trip_plan_view.ex
arkadyan: Good catch! Updated the implementation and added a test.

apps/site/lib/site_web/templates/trip_plan/_fare_calculator.html.eex
arkadyan: Great point! I updated both usages.

arkadyan:
Thanks, @thecristen. I made all those updates.

apps/site/test/site_web/views/trip_plan_view_test.exs
```diff
@@ -1435,4 +1435,442 @@ closest arrival to 12:00 AM, Thursday, January 1st."
...
+              latitude: 42.352271,
+              longitude: -71.055242,
+              name: "South Station",
+              stop_id: "74617"
+            },
+            long_name: nil,
+            mode: %TripPlan.PersonalDetail{
+              distance: 0.0,
+              steps: [
+                %TripPlan.PersonalDetail.Step{
```
thecristen, line 1818: It's not a problem, but generally speaking I don't think you need quite this much detail in your test data? For example `%TripPlan.PersonalDetail.Step{}` will just use the defstruct values. [An example of taking advantage of this for a test](https://github.com/mbta/dotcom/blob/eaac9fdb9b32f0c55f64689ec23de01de283a513/apps/trip_plan/test/itinerary_test.exs#L88).

arkadyan:
@ingridpierre This is a visual change, but there's nothing new added with it. Are you interested in looking it over?

ingridpierre:
@arkadyan This one is fine, with Cristen's approval!

PR #573, opened 2020-07-21
🚍 Stop/Station/TNM | Departure Information

thecristen:
#### Summary of changes
**Asana Ticket:** [🚍 Stop/Station/TNM | Departure Information](https://app.asana.com/0/385363666817452/1173665781696619/f)

Adds text to the white part of the card to describe any alerts related to diversions/shuttles/stop closures. 
(Note this is distinct from the recent work to show the alert icon in the route name header. The header icon indicates the presence of only alerts that are either high-priority or high severity).

---


apps/site/assets/ts/tnm/components/RouteCard.tsx
```diff
@@ -44,12 +47,23 @@ const RouteCard = ({
     return null;
   }
 
+  const hasAlert = route.route.alerts.length > 0;
```
alicia-mbta, line 50: Would it make sense to double check that `route.route.alerts` exists? You could use lodash's `get`:
`get(route, "route.alerts", <some default value>)`

apps/site/assets/ts/leaflet/__tests__/MapTooltipTest.tsx
```diff
@@ -17,7 +17,7 @@ const routes: EnhancedRoute[] = [
     direction_names: { "0": "South", "1": "North" }, // eslint-disable-line @typescript-eslint/camelcase
     direction_destinations: { "0": "Forest Hills", "1": "Oak Grove" }, // eslint-disable-line @typescript-eslint/camelcase
     description: "rapid_transit",
-    alert_count: 0, // eslint-disable-line @typescript-eslint/camelcase
+    alerts: [], // eslint-disable-line @typescript-eslint/camelcase
```
alicia-mbta, line 20: You can probably remove that linting annotation since you're not using camel case anymore.

PR #574, opened 2020-07-22
Add a script for automating production deploys

arkadyan:
#### Summary of changes
**Asana Ticket:** [🔧 Use a script to automate production deploys](https://app.asana.com/0/555089885850811/1185680283158786)

Add a script for automating production deploys

Add python as a dependency.
Add boto3 and docopt as python requirements.

---


arkadyan:
@thecristen and @alicia-mbta, I asked Ian for a review knowing he has experience with these types of python scripts as well as our deploy process, but I'm of course interested in any feedback you have as well.

arkadyan:
I forgot to mention that I successfully ran our recent deploy using this.

alicia-mbta:
This is great! Does the deployment happen just by running this new script or would there be other steps involved?
On that note, could you add a sentence or so to the `README` indicating that?

bin/deploy
```diff
@@ -0,0 +1,31 @@
...
+Arguments:
+    <sha>  The hash of a commit that has been deployed to dotcom-dev
+"""
+
+from docopt import docopt
+import boto3
+
+
+def main(args):
+  environment_name = "dotcom-prod"
```
ianwestcott, line 18: Consider making this a positional argument; I think it's worth requiring the user to explicitly specify the environment that they want to deploy to, so as to make the action more deliberate. Also then it can be used for dev-green as well.

arkadyan:
@alicia-mbta Great point, I added a bunch of documentation.

ianwestcott:
One other thought: As written, the Python script could be replicated with a single AWS CLI call:

```
aws elasticbeanstalk update-environment --environment-name "$dotcom-prod" --version-label "$sha"
```

Granted, you have the option to add a lot of sugar to the Python script – you could use [`describe_application_versions()`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticbeanstalk.html#ElasticBeanstalk.Client.describe_application_versions) to pull a list of most recent versions and choose which one you want to deploy, for example. Or poll to watch the update happen. But if you don't want to do any of that, you could rewrite this as minimal a shell script, FWIW.

ianwestcott:
Oh yeah! That’s exciting 😄

bin/deploy
arkadyan: 👍 that makes sense.

PR #575, opened 2020-07-23
Improve message for additional service schedules

arkadyan:
#### Summary of changes
**Asana Ticket:** [🐞 Schedule Finder | Missing trip details investigation](https://app.asana.com/0/555089885850811/1158218699153321)

Improve message for additional service schedules

Screen Shot 2020-07-23 at 10 16 46

---

### DIFF
--- a/apps/site/assets/ts/schedule/components/schedule-finder/TripDetails.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/TripDetails.tsx
@@ -55,14 +55,12 @@ export const TripDetails = ({
     return <Loading />;
   }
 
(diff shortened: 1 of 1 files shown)


arkadyan:
@thecristen Ah that's interesting. The entire table includes the trip summary (length, fare) that I don't know if we'd want to hide, but maybe at least hiding the `Stops` and `Arrival` table headers would be better? What do you think, @ingridpierre?

paulswartz:
One note on the text: while we don't have a schedule, the real-time data does have predicted arrival times even for added service.

arkadyan:
Dropping in favor of more fundamental changes.

PR #576, opened 2020-07-24
Include URL in list of fields to parse

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Alerts | URL field still doesn't work](https://app.asana.com/0/555089885850811/1176599342634997)

Even though the alerts for dotcom included a URL for more information, it was not 'passed down' and hence not displayed.
Content team members had to manually include said URL in the description.

Also included here a fix to anticipate URLs being null.

Before:
image

After:
image

### DIFF
--- a/apps/site/assets/ts/components/Alerts.tsx
+++ b/apps/site/assets/ts/components/Alerts.tsx
@@ -147,8 +147,11 @@ const alertDescription = (alert: AlertType): ReactElement<HTMLElement> => (
 const Alert = ({ alert }: { alert: AlertType }): ReactElement<HTMLElement> => {
   const [expanded, toggleExpanded] = useState(false);
   const onClick = (): void => toggleExpanded(!expanded);
+
+  const alertUrl = alert.url ? alert.url : "";
+
   // remove [http:// | https:// | www.] from alert URL:
-  let strippedAlertUrl = alert.url.replace(/(https?:\/\/)?(www\.)?/i, "");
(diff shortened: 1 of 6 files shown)


apps/site/assets/ts/components/Alerts.tsx
```diff
@@ -147,8 +147,11 @@ const alertDescription = (alert: AlertType): ReactElement<HTMLElement> => (
 const Alert = ({ alert }: { alert: AlertType }): ReactElement<HTMLElement> => {
```
arkadyan, line 147: Does `AlertType` need to be updated to indicate that `url` could be `null`?

apps/site/assets/ts/components/Alerts.tsx
```diff
@@ -147,8 +147,11 @@ const alertDescription = (alert: AlertType): ReactElement<HTMLElement> => (
...
   const [expanded, toggleExpanded] = useState(false);
   const onClick = (): void => toggleExpanded(!expanded);
+
+  const alertUrl = alert.url ? alert.url : "";
+
   // remove [http:// | https:// | www.] from alert URL:
-  let strippedAlertUrl = alert.url.replace(/(https?:\/\/)?(www\.)?/i, "");
+  let strippedAlertUrl = alertUrl.replace(/(https?:\/\/)?(www\.)?/i, "");
```
arkadyan, line 154: It might already be part of a snapshot, but since this seems important would you mind please adding a test explicitly showing that this URL is displayed properly?

PR #578, opened 2020-07-29
Make Upcoming Departures only show prediction times

arkadyan:
#### Summary of changes
**Asana Ticket:** [Schedule Finder | Make Upcoming Departures only show prediction times, Daily Schedule only schedule times](https://app.asana.com/0/555089885850811/1158218699153321)

Make Upcoming Departures only show prediction times, Daily Schedule only schedule times

I tried to keep things structured in isolated commits. My approach was to completely separate out Upcoming Departures from the Daily Schedule, copying the necessary components so that nothing is shared between these two (at least as an intermediate step). After that I did the work of making Upcoming Departures show prediction times and Daily Schedule show schedule times.

Also mixed in there I added a new useFetch hook in order to add type safety to API data responses, as well as some general refactoring to help me figure everything out.

Potential future steps would be to explore separating / simplifying API endpoints to only return the data that is actually needed for each section. It would also be nice to provide a single endpoint to fetch all trips in Upcoming Departures at once, rather than making a number of requests concurrently as we do now. Then we can explore if there are places to re-introduce shared components safely on the client side.

---


apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/TableRow.tsx
```diff
@@ -1,50 +1,39 @@
...
-        throw new Error(response.statusText);
-      })
-      .then(json => dispatch({ type: "FETCH_COMPLETE", payload: json }))
-      // @ts-ignore
-      .catch(() => dispatch({ type: "FETCH_ERROR" }))
+  { route, direction, date, origin }: UserInput
+): (() => Promise<Response>) => () =>
+  window.fetch &&
+  window.fetch(
+    `/schedules/finder_api/trip?id=${tripId}&route=${route}&date=${date}&direction=${direction}&stop=${origin}`
```
alicia-mbta, line 31: I'm not totally sure but could you use the newly created `fetchTripInfo` here?

apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/TableRow.tsx
```diff
@@ -1,50 +1,39 @@
...
 }
 
-type fetchAction =
-  | { type: "FETCH_COMPLETE"; payload: TripInfo }
-  | { type: "FETCH_ERROR" }
-  | { type: "FETCH_STARTED" };
-
-export const fetchData = (
+// Exported solely for testing
+export const fetchTripInfo = (
```
alicia-mbta, line 25: I'm not totally sure if it's doable but could you use this newly created `fetchTripInfo` in `UpcomingDepartures.tsx`?

apps/site/assets/ts/schedule/components/schedule-finder/upcoming-departures/TripStop.tsx
```diff
@@ -0,0 +1,77 @@
...
+  showFare: boolean;
+  routeType: number;
+}
+
+const formattedDepartureTimes = (
+  departure: TripDepartureWithPrediction,
+  routeType: number
+): ReactElement<HTMLElement> => {
+  const { schedule, prediction, delay } = departure;
+  if (routeType === 2) {
```
alicia-mbta, line 19: We have `isACommuterRailRoute` for better readability.

arkadyan: Good point. I missed that at first because that function took a Route not a RouteType which is all we have here, but I updated it to take either.

apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/TableRow.tsx
arkadyan: Yeah, I do want to get there. The tricky thing about doing it right now is that the data from upcoming departures is requested individually mapping over a list, not just from one request. So we'll either have to make these helpers work with that, or maybe better yet make a route on the server that lets the client request data for that whole list in a single request. But it was enough that I didn't want to take it on in this PR.

alicia-mbta: Sure, thanks for clarifying :smiley:

PR #579, opened 2020-07-29
Remove old fare finder section of the site

arkadyan:
#### Summary of changes
**Asana Ticket:** [🎫 Fare Finder | Remove old tool](https://app.asana.com/0/555089885850811/1169236529454107)

Remove old fare finder section of the site

---


apps/site/lib/site_web/views/mode_view.ex
```diff
@@ -26,20 +26,6 @@ defmodule SiteWeb.ModeView do
...
-          "table of fare zones.",
-          to: cms_static_page_path(SiteWeb.Endpoint, "/fares/commuter-rail-fares/zones")
-        )
-      ]
-    end
-  end
-
   def fares_note(_mode) do
     ""
   end
```
thecristen, line 31: Probably meant to delete this

arkadyan: This was actually intentional, temporarily. The CMS content feeding dev-green still appears to be using this. Once we can remove that then this can go.

arkadyan: Added a comment documenting this.

apps/site/lib/site_web/router.ex
```diff
@@ -56,7 +56,7 @@ defmodule SiteWeb.Router do
     get("/alerts/commuter_rail", Redirector, to: "/alerts/commuter-rail")
     get("/fares/charlie_card", Redirector, to: "/fares/charliecard")
     get("/fares/charlie-card", Redirector, to: "/fares/charliecard")
-    get("/fares/commuter_rail", Redirector, to: "/fares/commuter-rail")
+    get("/fares/commuter_rail", Redirector, to: "/trip-planner")
```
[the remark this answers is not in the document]
arkadyan: Good catch, I added those additional 2 redirects. I was waffling at first with the existing commuter rail one, but I agree with you so I changed that too.

apps/site/lib/site_web/views/mode_view.ex
```diff
@@ -25,6 +25,7 @@ defmodule SiteWeb.ModeView do
 
   def get_route_group(route_type, route_groups), do: route_groups[route_type]
 
+  # TODO: Remove this once it is no longer being used in the content on any environment -- MSS 20200730
```
thecristen, line 28: Looks like this is making credo fail 😬

arkadyan: Thanks! I missed that, but fixed it now (as well as a broken test).

PR #583, opened 2020-08-05
Drop subway predictions in the past

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [RTR Subway Predictions | Drop subway predictions in the past](https://app.asana.com/0/555089885850811/1184691535835665)

Drop subway predictions from the prediction repo if the predicted time is earlier than the current time. This will only affect subway predictions—we want to continue showing commuter rail trains that have recently departed.

### DIFF
--- a/apps/predictions/lib/repo.ex
+++ b/apps/predictions/lib/repo.ex
@@ -1,6 +1,10 @@
 defmodule Predictions.Repo do
+  @moduledoc "Predictions repo module"
+
   use RepoCache, ttl: :timer.seconds(10), ttl_check: :timer.seconds(2)
   require Logger
+  require Routes.Route
+  alias Routes.Route
   alias Predictions.Parser
   alias Stops.Stop
 
@@ -137,11 +141,7 @@ defmodule Predictions.Repo do
          },
          min_time
        ) do
-    case DateTime.compare(prediction_time, min_time) do
-      :gt -> true
-      :eq -> true
-      :lt -> false
-    end
+    Util.time_is_greater_or_equal?(prediction_time, min_time)
   end
 
   def load_from_other_repos([]) do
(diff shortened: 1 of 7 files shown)


apps/util/lib/util.ex
```diff
@@ -25,6 +27,15 @@ defmodule Util do
     now() |> Timex.to_date()
   end
 
+  @spec time_is_greater_or_equal(DateTime.t(), DateTime.t()) :: boolean
+  def time_is_greater_or_equal(time, ref_time) do
```
arkadyan, line 31: This is nice. Could you please add a `?` to the end of the function name to match [Elixir naming conventions](https://hexdocs.pm/elixir/master/naming-conventions.html#trailing-question-mark-foo)?

apps/predictions/lib/repo.ex
```diff
@@ -188,20 +188,32 @@ defmodule Predictions.Repo do
...
-    ]
+    # For subway, drop predictions if the predicted time is earlier than the current time:
+    prediction_in_the_past =
+      if time == nil do
+        false
+      else
+        !Util.time_is_greater_or_equal(Util.to_local_time(time), Util.to_local_time(Util.now()))
+      end
+
+    if Route.subway?(route.type, route_id) && prediction_in_the_past do
```
arkadyan, line 199: One red flag for me about doing this here is that this function is now doing two things. In addition to converting to a struct, it's also filtering (which isn't obvious where it is called). Would it be possible to add a new filter step below `filter_by_min_time`? At a glance I think your logic here would probably fit into the same structure as `filter_by_min_time`.

apps/predictions/test/repo_test.exs
```diff
@@ -274,4 +274,40 @@ defmodule Predictions.RepoTest do
 
     assert log =~ "Discarding prediction"
   end
+
+  test "drops subway prediction if it is in the past" do
```
arkadyan, line 278: This test is great. Just because it is important functionality, could you please also add a test showing that past predictions are _not_ dropped for the commuter rail?

PR #585, opened 2020-08-07
🔧 Refactor implementation of ScheduleDirectionMenu event listeners

thecristen:
#### Summary of changes
**Asana Ticket:** [🔧 Refactor implementation of ScheduleDirectionMenu event listeners](https://app.asana.com/0/385363666817452/1188077420952242/f)

Currently every click anywhere on the schedule page causes unnecessary rerendering of many components. This change fixes that.

---

### DIFF
--- a/apps/site/assets/ts/schedule/components/direction/ScheduleDirectionMenu.tsx
+++ b/apps/site/assets/ts/schedule/components/direction/ScheduleDirectionMenu.tsx
@@ -18,29 +18,6 @@ interface Props {
   dispatch: Dispatch<MenuAction>;
 }
 
-/* istanbul ignore next */
-const externalCloseEvent = (dispatch: Dispatch<MenuAction>): void => {
-  document.addEventListener(
-    "click",
-    (event: Event): void => {
-      if (!event.target) return;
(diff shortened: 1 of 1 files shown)


apps/site/assets/ts/schedule/components/direction/ScheduleDirectionMenu.tsx
```diff
@@ -53,10 +30,35 @@ const ScheduleDirectionMenu = ({
...
+  useEffect(
+    () => {
+      if (!menuOpen) return () => {};
+      // if the menu is open, add event listeners to enable closing the menu by
+      // clicking outside the menu or with keypress
+      const clickToClose = (event: Event): void => {
+        if (!event.target) return;
+        const element = event.target as HTMLElement;
+        if (element.closest(".js-m-schedule-click-boundary")) return;
+        dispatch(closeRoutePatternMenuAction());
```
arkadyan, line 42: Potential future refactor: we could pass a `closeMenu` function into `ScheduleDirectionMenu`. That way it wouldn't have to know anything about `dispatch` or `closeRoutePatternMenuAction` which aren't really its concern.

PR #586, opened 2020-08-10
Pin right rail on the right side for suspended routes

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🔥 Schedules | Fix right rail on suspended routes](https://app.asana.com/0/555089885850811/1187736907832137)

Temporary fix for the upcoming rating.


alicia-mbta:
Instructions to test:
In `apps/site/lib/site_web/templates/schedule/_line.html.eex`, replace all occurrences of `@branches` with `branches` (without the `at` sign) and add the following in line 4:
`<% branches = [] %>`

PR #592, opened 2020-08-17
Fix for no fetching journeys when selecting a different service

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Schedule Finder | Service Picker doesn't update schedule](https://app.asana.com/0/555089885850811/1187746923806232)

Riders may be falsely led to believe that all days have the same schedule because the schedules were not refreshing properly.

### DIFF
--- a/apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/DailySchedule.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/DailySchedule.tsx
@@ -136,23 +136,22 @@ export const DailySchedule = ({
     defaultSelectedService
   );
 
+  const getJourneysForSelectedService = (service: Service): void => {
+    fetch({
+      fetcher: fetchJourneys(routeId, stopId, service, directionId, false),
+      parser: parseResults
+    });
+  };
+
   useEffect(
     () => {
       /* istanbul ignore next */
       if (selectedService && isNotStarted(fetchState)) {
-        fetch({
-          fetcher: fetchJourneys(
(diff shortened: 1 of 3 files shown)


apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/DailySchedule.tsx
```diff
@@ -136,23 +136,22 @@ export const DailySchedule = ({
...
+    fetch({
+      fetcher: fetchJourneys(routeId, stopId, service, directionId, false),
+      parser: parseResults
+    });
+  };
+
   useEffect(
     () => {
       /* istanbul ignore next */
       if (selectedService && isNotStarted(fetchState)) {
```
arkadyan, line 149:
```suggestion
      if (selectedService && !isLoading(fetchState)) {
```

Would changing `isNotStarted` to `!isLoading` allow it to re-fetch properly?

alicia-mbta: I tried that; that change causes an infinite loop because once the data is retrieved, status changes to "with data", which evaluates to true for `isLoading(fetchState)` and hence fetching again.

arkadyan: Ah, ok, thanks. That makes me wonder if `fetchState` shouldn't have been in that dependency list. But what you've done looks great, this is all just idle curiosity on my part.

PR #593, opened 2020-08-19
Display messages and Free Service in fare calculator

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Trip Planner | SL Trips from Airport are Free](https://app.asana.com/0/555089885850811/1179732762379245)

The fare calculator now includes some messages, links and the language "Free Service" when the itinerary includes a leg with the SL1 that relates to a trip from/to Logan Airport:
image


apps/fares/lib/format.ex
```diff
@@ -87,8 +87,11 @@ defmodule Fares.Format do
   def name({:zone, zone}), do: "Zone #{zone}"
   def name({:interzone, zone}), do: "Interzone #{zone}"
   def name(:foxboro), do: "Foxboro Special Event"
-  # A free fare might be an SL1 trip from airport stops or shuttle bus service
+  # A free fare might be a shuttle bus service
   def name(:free_fare), do: "Free Fare"
```
arkadyan, line 91: Quick point before looking at the rest: I believe we want to change the wording to "Free Service" in all cases, not just airport trips.

ingridpierre: Right now it says "Shuttle: $0.00" when I look up a shuttle trip, will that stay the same or change also to Free Service? I like the specificity of "Shuttle: $0.00".

alicia-mbta: @arkadyan, I changed the logic a bit so the string is the only thing that changes from `"Free Fare"` to `"Free Service"`.
@ingridpierre, which A and B points were you entering? there should be a distinction between `"Shuttle"` and SL1's `"Free Service"`.

apps/fares/lib/fares.ex
```diff
@@ -22,6 +22,17 @@ defmodule Fares do
   @foxboro_reverse_commute ~w(741 743 745 750 752 754 756)
   @foxboro_reverse_commute_set MapSet.new(@foxboro_reverse_commute)
 
+  @free_sl1_fare %Fares.Fare{
```
arkadyan, line 25: Why the need for this new type? I thought trips from the airport already correctly had a free fare associated with them?

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -575,9 +575,33 @@ defmodule SiteWeb.TripPlanView do
     |> Map.put(:src, map_src)
   end
 
+  @spec readjust_itinerary_with_sl1(TripPlan.Itinerary.t()) :: TripPlan.Itinerary.t()
+  def readjust_itinerary_with_sl1(itinerary) do
```
arkadyan, line 579: The preceding question might make this irrelevant, but if not, why does this readjustment need to be done in the view layer? Could this logic be incorporated such that the correct fare is determined in the first place?

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -728,7 +761,7 @@ defmodule SiteWeb.TripPlanView do
       else
         name =
           if leg.name && leg.name =~ "Shuttle",
-            do: "Shuttle",
+            do: Format.name(:shuttle),
```
arkadyan, line 764: 👍

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -604,7 +635,8 @@ defmodule SiteWeb.TripPlanView do
         |> render_to_string(
           itinerary: i,
           one_way_total: Format.price(one_way_total_fare),
-          round_trip_total: Format.price(one_way_total_fare * 2)
+          round_trip_total: Format.price(one_way_total_fare * 2),
+          itinerary_includes_sl1: itinerary_includes_sl1
```
arkadyan, line 639: Instead of having to pass this as a variable, could `itinerary_includes_sl1` be a view function that could be called by the templates, passing in the itinerary value they already have? 

Also, per convention, this should end with a `?`, please: `itinerary_includes_sl1?`.

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -586,6 +610,13 @@ defmodule SiteWeb.TripPlanView do
             conn.assigns.itinerary_row_lists,
             Stream.iterate(1, &(&1 + 1))
           ]) do
+      {itinerary_includes_sl1, i} =
+        if Enum.any?(it, fn leg -> Leg.is_sl1?(leg) end) do
```
arkadyan, line 614: I think this logic might not be sufficient. We want the origin or destination trip to be on the SL, but I think the origin or destination needs to be the airport specifically. If I map a trip from South Station to the World Trade Center I don't think we want to show the airport messages.

apps/fares/lib/format.ex
ingridpierre: @alicia-mbta I think we're good then! I didn't see any shuttle fares change. I think I was just confused as to what this change would affect.

PR #596, opened 2020-08-20
💈 special line diagram styling for stops diversion-related alerts

thecristen:
Adds 'barberpole' styling to line, and adds additional alert text in place of predictions

#### Summary of changes
**Asana Ticket:** [🚍 Line Diagram | "Barberpole" Treatment](https://app.asana.com/0/385363666817452/1173665781696621/f)

The line page should now reflect alerts (of any severity) of the following effects: `shuttle`, `stop_closure`, `station_closure`, `detour`, `suspension` by inserting additional text in place of predictions and showing a striped pattern on the line diagram.

<details>
<summary>📷 <b>click here for screenshots (IE11)</b>:</summary>

Screen Shot 2020-08-20 at 11 15 56
Screen Shot 2020-08-20 at 11 16 05
Screen Shot 2020-08-20 at 11 16 43

</details>

---


arkadyan:
Cute branch name

ingridpierre:
From the changes I can observe, there are some nice improvements here to the placement of the stop dots and vehicle locations generally, especially for terminus stops and "arrived" vehicles.

----

**A couple small details:**

There used to be a bit more left padding inside the stop card separating the line diagram from the outer border, visually the vehicles "touch" the left border now.

image

----
 It looks like the stop dot stroke "breaks out" of the line slightly. I think this is because strokes are drawn centered on svg boundaries instead of inside or outside. Should the main line thickness be increased to 9px or something to contain the stop shape again? I don't mind this necessarily, just wondering if a slight adjustment is possible.
Braintree 10 min.

----

A bug/regression in this PR but also in prod: the column for crowding info should go away when there is no information to display. This screenshot is of the Subway which never has vehicle crowding data, but also applies to some bus routes I observed.
image

ingridpierre:
@thecristen On review, especially on these long disruptions, the black alert icon beside the alert effect text looks confusingly duplicative. Let's remove it in this case.

ingridpierre:
Wondering if we might be able to tell why some ongoing detours show, but others do not.

For example, this SL1 Detour is displayed correctly:

ingridpierre:
Wondering

ingridpierre:
So far, the functionality and styling for this looks great! I would love to play with this one overnight(s) though to catch a few more evening alerts situations and double-check everything.

Some of the issues we encountered with the display (too many stops highlighted or too few) are happening because of inconsistencies in the alert data provided by PIOs. 

When you have a chance to make some of these changes it would be good to stage this on dev so we can forward links to broken alerts to the PIOs to fix up as we deploy.

ingridpierre:
For the 8, it looks like we need to apply the barberpole treatment on the segment just ahead of the detour. Right now it looks like you can still stop at Ruggles St @ Cabot St.

image


image

ingridpierre:
This was also reported by Josh F. but the B line shuttle that doesn't start til tomorrow is display as active on the line diagram:

image
image

thecristen:
> For the 8, 17, 47 detour, it looks like we need to apply the barberpole treatment on the segment just before the first detoured stop. Right now it looks like you can still stop at Ruggles St @ Cabot St.

@ingridpierre I'm not sure about this proposed solution -- applying that styling to the segment before the detoured stop makes sense, but it also likely wouldn't make sense before a **shuttle** stop - so we would need some special logic there to capture that (and make sure to capture all relevant cases). I'll think on how to accomplish that. I'm also wondering if a styling specifically applied to stops would more clearly convey the alert.

ALSO - Thanks for highlighting the Green-B issue again, I will add a check for alerts that are not rather than in the future - will add that.

thecristen:
By similar logic I'm guessing this example should **not** extend the line to Eliot?

image

ingridpierre:
> By similar logic I'm guessing this example should **not** extend the line to Eliot?
> 

Correct, I'm making a note to include some of this update styling by effect in the documentation as well.

ingridpierre:
Seeing another case here for the D line, where multiple shuttle messages are being shown, where I believe we do have realtime departures to display.
image
image

thecristen:
Made some more updates, will need more review. Thanks!

* Added some logic on the backend to use the alert effect to determine which line segments should show as affected, in accordance with the new zeroheight documentation. This replaces the corresponding weaker front-end logic in 616bb49720f33f52e4d393562fa1b7df09f9b1d7 and is further refined in  07ccd1aee90620c537777bb8c98d18b496723b1d

* Removed `suspension` effect from this as (a) it seems to be used to describe a whole route rather than a stop and (b) it isn't indicated in the design anyways.

* Some tiny fixes - correcting the location of the branch collapse toggle, styling the curve at a merge stop, showing predictions only if they are actually available first

ingridpierre:
Can confirm that the drawing of the barberpole has updated and looks good for shuttles. However, this shuttle isn't active for the D line and is still being highlighted:

Screen Shot 2020-09-08 at 9 24 46 AM

ingridpierre:
D line detour is showing a highlight even though the shuttle is not  currently active.

Also, it looks like for shuttles we're not yet drawing the effect as described in the docs: only between the stop dots for the affected stops/informed entity. (see below: Green Line - Shuttle)

Current: 
image

Desired (when shuttle running):
image

thecristen:
There's been lots of updates since this was initially started so I'm moving discussion and review over to #618 and closing this PR.

PR #597, opened 2020-08-20
Reduce the price of CharlieTicket/cash fares

thecristen:
#### Summary of changes
**Asana Ticket:** [Reduce the price of CharlieTicket/cash fares](https://app.asana.com/0/385363666817452/1189067113839099/f)

This resurrects an earlier approach (checking and comparing the dates to switch files at a certain time) that @ryan-mahoney had removed... tagging him just in case he has any wisdom to share for why I **shouldn't** do it this way.

The alternate approach would be to edit the relevant lines in the current file (probably renaming it as well), and get this on a feature branch that we will merge in on the appropriate date. Happy to go either way!


apps/fares/lib/fare_info.ex
```diff
@@ -6,15 +6,24 @@ defmodule Fares.FareInfo do
...
+    now
+    |> filename()
     |> fare_data()
     |> Enum.flat_map(&mapper/1)
     |> Enum.concat(free_fare())
     |> split_reduced_fares()
   end
 
+  @spec filename(integer) :: Path.t()
+  defp filename(now) when now >= @september_1_2020, do: "priv/fares-sept1.csv"
```
arkadyan, line 23: It doesn't really matter, but I might even consider including the year in the csv file name. My impression is that this data changes pretty infrequently.

PR #600, opened 2020-08-24
SL Trips from Airport are free

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Trip Planner | SL Trips from Airport are Free](https://app.asana.com/0/555089885850811/1189794962366354)

Any subway legs that come after the Silver Line when leaving the airport will be marked as part of the free service.
Example from Logan with several transfers:
image

### DIFF
--- a/apps/site/lib/site_web/views/trip_plan_view.ex
+++ b/apps/site/lib/site_web/views/trip_plan_view.ex
@@ -14,8 +14,6 @@ defmodule SiteWeb.TripPlanView do
 
   @meters_per_mile 1609.34
 
-  @type fare_type :: :highest_one_way_fare | :lowest_one_way_fare | :reduced_one_way_fare
-
   @type fare_calculation :: %{
           mode: Route.gtfs_route_type(),
           # e.g. :commuter_rail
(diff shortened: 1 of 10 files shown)


ingridpierre:
👍 lgtm!

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -607,20 +607,30 @@ defmodule SiteWeb.TripPlanView do
           round_trip_total: Format.price(one_way_total_fare * 2)
         )
 
-      fares = get_calculated_fares(i)
+      # If Logan airport is the origin, all subsequent subway trips from there should be free
+      first_transit_leg = i |> Itinerary.transit_legs() |> List.first()
+
+      readjusted_itinerary =
+        if stop_is_silver_line_airport?([first_transit_leg], :from) do
```
arkadyan, line 614: It looks like `readjust_itinerary_with_free_fares` could use some tests. But also, could this logic be done when fares are initially calculated in the controller? It seems like it would be better if the view didn't have to update the data.

alicia-mbta: You're right. I also "relocated" some functions that (I felt like) belonged to other modules (e.g. Leg, Fares) rather than to the view. Let me know if these latest changes make sense.

apps/site/test/site_web/views/trip_plan_view_test.exs
```diff
@@ -1132,14 +1132,14 @@ closest arrival to 12:00 AM, Thursday, January 1st."
 
     test "gets fare by type" do
       non_transit_leg = @itinerary.legs |> List.first()
-      assert get_fare_by_type(non_transit_leg, :highest_one_way_fare) == nil
-      assert get_fare_by_type(non_transit_leg, :lowest_one_way_fare) == nil
-      assert get_fare_by_type(non_transit_leg, :reduced_one_way_fare) == nil
+      assert Fares.get_fare_by_type(non_transit_leg, :highest_one_way_fare) == nil
```
arkadyan, line 1135: Could you please also just move these tests to `fares_test.exs` to match the implementation?

PR #602, opened 2020-08-27
Combine CharlieCard and Ticket/Cash media prices after Sep. 1, 2020

amaisano:
#### Summary of changes
**Asana Ticket:** [Fare Change | Fare cards updates for 9/1 transition](https://app.asana.com/0/399597520748778/1186724915893667)

After the changeover date, adjust fare matches so charlie_card, charlie_ticket, and cash all have equal prices on subway/bus single trips (one-way). This affects the descriptive output around the site, such as "with a CharlieCard, CharlieTicket, or Cash."

I tried to make this as small as possible, rather than creating branches for each modes' existing `mapper/1` fare mapping. I recommend yanking this and the other sep_1_2020 logic after the changeover date.

---

### DIFF
--- a/apps/fares/lib/fare_info.ex
+++ b/apps/fares/lib/fare_info.ex
@@ -7,6 +7,7 @@ defmodule Fares.FareInfo do
 
   @doc "Load fare info from a CSV file."
   @september_1_2020 1_598_918_400
+  @september_1_2020_modes ["subway", "local_bus", "inner_express_bus", "outer_express_bus"]
 
   @spec fare_info() :: [Fare.t()]
   @spec fare_info(integer) :: [Fare.t()]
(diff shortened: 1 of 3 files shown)


apps/fares/lib/fare_info.ex
```diff
@@ -24,6 +25,29 @@ defmodule Fares.FareInfo do
 
   defp filename(_), do: "priv/fares-july1.csv"
 
+  @doc "Combines paper and plastic fare into a single price for certain modes"
+  @spec mapper([String.t()], integer) :: [Fare.t()]
+  def mapper([mode | _] = data, now)
```
arkadyan, line 30: This looks good. Could you please include a test or two for it?

apps/fares/lib/fare_info.ex
```diff
@@ -6,15 +6,16 @@ defmodule Fares.FareInfo do
   alias Fares.Fare
 
   @doc "Load fare info from a CSV file."
-  @september_1_2020 1_598_918_400
+  @september_1_2020 1_598_932_800
```
thecristen, line 9: 👍

PR #604, opened 2020-08-28
Remove "Recommended" column from Fare Calculator

arkadyan:
#### Summary of changes
**Asana Ticket:** [🎫 Trip Planner | Eliminate CharlieCard - CharlieTicket/Cash differential](https://app.asana.com/0/555089885850811/1187029333962715)

Remove "Recommended" column from Fare Calculator after the fare transition on 9/1.

~~Depends on https://github.com/mbta/dotcom/pull/602~~

Screen Shot 2020-08-28 at 11 09 23

---

### DIFF
--- a/apps/fares/lib/fare_info.ex
+++ b/apps/fares/lib/fare_info.ex
@@ -438,6 +438,11 @@ defmodule Fares.FareInfo do
     ]
   end
 
+  @spec charging_september_2020_fares?() :: boolean()
+  @spec charging_september_2020_fares?(integer()) :: boolean()
+  def charging_september_2020_fares?(now \\ System.system_time(:second)),
+    do: now >= @september_1_2020
+
   defp fare_data(filename) do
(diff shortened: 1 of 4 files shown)


amaisano:
Is this rebased after my fare mapper work? I was wondering if by merging two fare matches into one, this column would automatically go away?

arkadyan:
@amaisano The column does not automatically go away.

arkadyan:
@thecristen Well this section already uses empty divs that way, that seems to be how it's designed. I also started with using css more, but the design calls for the columns of the first table to line up with the ones on the second.

thecristen:
> @thecristen Well this section already uses empty divs that way, that seems to be how it's designed. I also started with using css more, but the design calls for the columns of the first table to line up with the ones on the second.

Oh... I probably disliked the markup structure then too! Otherwise looks 👍🏼

PR #605, opened 2020-08-28
Fix for photo galleries not working in IE11

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Photo galleries on IE11 only showing first image](https://app.asana.com/0/555089885850811/1190988214208156)

In IE11 (works fine in other browsers), the photo galleries do nothing: clicking on the thumbnails should display the selected image in the bigger area.


alicia-mbta:
I'd say more like a hack than detective work...but hey, IE is another hack at this point :joy:

PR #606, opened 2020-08-31
Use Date sorting to properly sort dates

arkadyan:
Use Date sorting to properly sort dates

No ticket, tests were failing because we were using the wrong service.

---


arkadyan:
Hold on this, will require a different approach for Elixir 1.8

arkadyan:
@thecristen Sorry about that. Converted to Elixir 1.8-friendly code and it's passing now.

arkadyan:
If anyone is interested, more info about this can be found in [this blog post](https://medium.com/@steven.cole.elliott/sorting-dates-in-elixir-59592a0c33a5).

PR #608, opened 2020-09-01
Remove fares date check

arkadyan:
#### Summary of changes
**Asana Ticket:** [Remove date check](https://app.asana.com/0/1189067113834097/1191202344036921)

Remove fares date check

It is now always true.

---


thecristen:
Looks like a test is failing because of the updated fares, I think we'll need to update to use `[:charlie_card, :charlie_ticket, :cash]` where needed. The changes looks great locally!

apps/fares/lib/fare_info.ex
```diff
@@ -5,30 +5,22 @@ defmodule Fares.FareInfo do
...
-  defp filename(now) when now >= @september_1_2020, do: "priv/fares-sept1.csv"
-
-  defp filename(_), do: "priv/fares-july1.csv"
-
   @doc "Combines paper and plastic fare into a single price for certain modes"
-  @spec mapper([String.t()], integer) :: [Fare.t()]
-  def mapper([mode | _] = data, now)
-      when mode in @september_1_2020_modes and now >= @september_1_2020 do
+  @spec maybe_combine_and_map([String.t()]) :: [Fare.t()]
+  def maybe_combine_and_map([mode | _] = data)
```
amaisano, line 22: There are some very simple steps I can code in here to avoid this clause. Just go into the mappers and manually combine/merge these things for these modes. Mind if I do that real quick?

amaisano: I'll just go and do this later in the week as a cleanup.

PR #612, opened 2020-09-04
🐞 Regression/Bug | Crowding column not collapsed properly on routes w/o crowding

thecristen:
#### Summary of changes
**Asana Ticket:** [🐞 Regression/Bug | Crowding column not collapsed properly on routes w/o crowding](https://app.asana.com/0/385363666817452/1190437676035289/f)

Introduced a CSS class to add to the parent element to handle hiding all the crowding icons when needed.

---


alicia-mbta (approved):
Thanks for the broken down commits, so clean 👌

ingridpierre:
👍  Looks great!

PR #614, opened 2020-09-08
Sidebar with static information

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [☎️ Make a new sidebar for static information](https://app.asana.com/0/555089885850811/1188535657153252)

The page has been rearranged. Content is now displayed on the right side and there is a flag that can be set to initially set the blocks as expanded when loading the page.

Before:
image

After:
image

### DIFF
--- a/apps/site/assets/css/_customer-support.scss
+++ b/apps/site/assets/css/_customer-support.scss
@@ -169,3 +169,40 @@ button .waiting {
   top: -$base-spacing / 2;
   width: $base-spacing;
 }
+
+.c-expandable-block__link {
+  color: inherit;
+  text-decoration: none;
+  &:hover {
+    text-decoration: inherit;
+  }
+  &:link {
+    text-decoration: inherit;
+  }
+}
+
+
+a:first-of-type > h3 {
+  margin-top: 0;
+}
+
+.c-expandable-block__header-container {
+  display: flex;
+  justify-content: space-between;
+}
+
+.c-expandable-block-caret {
+  @include caret($color: $brand-primary);
+  align-items: flex-end;
+  display: flex;
+
+  [aria-expanded='false'] &::after {
+    content: $fa-var-angle-down;
+    display: flex;
+  }
+
+  [aria-expanded='true'] &::after {
+    content: $fa-var-angle-up;
+    display: flex;
+  }
+}
--- a/apps/site/lib/site_web/controllers/customer_support_controller.ex
+++ b/apps/site/lib/site_web/controllers/customer_support_controller.ex
@@ -5,6 +5,44 @@ defmodule SiteWeb.CustomerSupportController do
 
   @allowed_attachment_types ~w(image/bmp image/gif image/jpeg image/png image/tiff image/webp)
 
+  @content_blocks [
+    %{
+      header: %{text: "Call Us", iconSvgText: nil},
+      id: "call_us",
+      initially_expanded: true
+    },
+    %{
+      header: %{text: "Lost and Found", iconSvgText: nil},
+      id: "lost_and_found",
+      initially_expanded: true
+    },
+    %{
+      header: %{text: "Get Service Updates", iconSvgText: nil},
+      id: "service_updates",
+      initially_expanded: true
+    },
+    %{
+      header: %{text: "Transit Police", iconSvgText: nil},
+      id: "transit_police",
+      initially_expanded: false
+    },
+    %{
+      header: %{text: "Request Public Records", iconSvgText: nil},
+      id: "request_public_records",
+      initially_expanded: false
+    },
+    %{
+      header: %{text: "Write to Us", iconSvgText: nil},
+      id: "write_to_us",
+      initially_expanded: false
+    },
+    %{
+      header: %{text: "Report Fraud, Waste, or Abuse", iconSvgText: nil},
+      id: "report",
+      initially_expanded: false
+    }
+  ]
+
   plug(Turbolinks.Plug.NoCache)
   plug(:set_service_options)
   plug(:assign_ip)
@@ -39,6 +77,29 @@ defmodule SiteWeb.CustomerSupportController do
         |> put_status(400)
         |> render_form(%{errors: errors, comments: Map.get(params, "comments")})
     end
+  end
+
+  @spec render_expandable_blocks(Plug.Conn.t(), list) :: [Phoenix.HTML.safe()]
+  def render_expandable_blocks(assigns, content_blocks \\ @content_blocks) do
+    content_blocks
+    |> Enum.map(fn block ->
+      view_template = "_#{block.id}.html"
+
+      try do
+        SiteWeb.PartialView.render(
+          "_expandable_block.html",
+          Map.merge(assigns, %{
+            header: block.header,
+            id: block.id,
+            initially_expanded: block.initially_expanded,
+            view_template: view_template
+          })
+        )
+      rescue
+        # We still want to render the page, so we just return empty content:
+        _ -> ""
+      end
+    end)
   end
 
   @spec do_submit(Plug.Conn.t(), map) :: Plug.Conn.t()
--- a/apps/site/lib/site_web/templates/customer_support/_call_us.html.eex
+++ b/apps/site/lib/site_web/templates/customer_support/_call_us.html.eex
@@ -1,0 +1,14 @@
+<p>
+  Monday - Friday: 6:30 AM - 8 PM<br>
+  Saturday - Sunday: 8 AM - 4 PM
+</p>
+
+<p>
+  <strong>Main Hotline:</strong> <%= tel_link "617-222-3200" %><br>
+  <strong>Toll Free:</strong> <%= tel_link "800-392-6100" %><br>
+  <strong>TTY:</strong> <%= tel_link "617-222-5146" %>
+</p>
+
+<p>
+  <strong>Elevator/Escalator Hotline:</strong> <%= tel_link "617-222-2828" %><br>
+</p>
--- a/apps/site/lib/site_web/templates/customer_support/index.html.eex
+++ b/apps/site/lib/site_web/templates/customer_support/index.html.eex
@@ -1,125 +1,32 @@
 <div class="container">
   <div class="page-section">
     <h1>Contact Us</h1>
-    <div class="row">
-      <div class="col-xs-12 support-thank-you <%= if @show_form, do: "hidden-xs-up" %>" tabindex="-1">
-        <div class="support-success" id="<%= if !show_error_message(@conn), do: "support-result" %>">
-          <%= fa "check-circle" %>
-          Thank you for your feedback.
+
+    <div class="col-xs-12 col-md-8">
+
+      <div class="row">
+        <div class="col-xs-12 support-thank-you <%= if @show_form, do: "hidden-xs-up" %>" tabindex="-1">
+          <div class="support-success" id="<%= if !show_error_message(@conn), do: "support-result" %>">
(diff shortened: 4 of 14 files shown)


apps/site/lib/site_web/templates/customer_support/_call_us.html.eex
```diff
@@ -0,0 +1,21 @@
...
+  <strong>Toll Free:</strong> <%= tel_link "800-392-6100" %><br>
+  <strong>TTY:</strong> <%= tel_link "617-222-5146" %>
+</p>
+
+<p>
+  <strong>Elevator/Escalator Hotline:</strong> <%= tel_link "617-222-2828" %><br>
+</p>
+
+<p>
+  <strong>Lost and Found:</strong>
```
thecristen, line 17: It seems like (in the mockups attached on the ticket) Lost and Found might have been updated to be its own section, and also that additional headers have been added to this content (Hours, Support)... these changes don't appear to be assigned to other tickets so you should probably include them there.

apps/site/assets/css/_customer-support.scss
```diff
@@ -169,3 +169,29 @@ button .waiting {
...
+
+
+a[aria-expanded='false'] > div > .c-expandable-block-caret {
+  @include caret($color: $brand-primary);
+  &::after {
+    content: $fa-var-angle-down;
+  }
+}
+
+a[aria-expanded='true'] > div > .c-expandable-block-caret {
```
thecristen, line 192: You probably don't need the `> div >` here

apps/site/assets/js/support-form.js
```diff
@@ -407,3 +408,17 @@ function showExpandedForm($) {
 function hideExpandedForm($) {
   $(".support-form-expanded").hide();
 }
+
+// Click on the blocks that have the 'expand' class (that serves as a flag) to expand them
+export function setupInitialStateOfExpandableBlocks($) {
```
thecristen, line 413: Are you sure this function is needed? The Bootstrap-included Javascript should essentially do this on page load, as long as the trigger link has `[data-toggle="collapse"]` (which it does!)

apps/site/assets/css/_customer-support.scss
thecristen: Have you considered reusing the `.c-expandable-block__header-caret` class in the template instead of using `.c-expandable-block-caret`? Then you wouldn't need to `@include caret` again and can do something like this for rendering the caret:

```scss
.c-expandable-block__header-caret {
  [aria-expanded='false'] &::after {
    content: $fa-var-angle-down;
  }

  [aria-expanded='true'] &::after {
    content: $fa-var-angle-up;
  }
}
```

apps/site/assets/css/_customer-support.scss
```diff
@@ -169,3 +169,29 @@ button .waiting {
...
+.c-expandable-block__link {
+  color: inherit;
+  text-decoration: none;
+  &:hover {
+    text-decoration: inherit;
+  }
+  &:link {
+    text-decoration: inherit;
+  }
+}
```
thecristen, line 182: You might need to add some padding or other layout tweak so the caret doesn't get pushed out :\ 

image

apps/site/lib/site_web/templates/partial/_expandable_block.html.eex
```diff
@@ -0,0 +1,34 @@
...
+<% header_id = "header-#{@id}" %>
+<% panel_id = "panel-#{@id}" %>
+
+<div id="<%= container_id %>">
+  <h2 class="c-expandable-block__header">
+    <a class="c-expandable-block__link"
+      href="#<%= panel_id %>"
+      tabIndex="0"
+      id="<%= header_id %>"
+      aria-expanded="false"
```
thecristen, line 12: This should be true if `@show_expanded` (and false otherwise)

apps/site/lib/site_web/templates/customer_support/index.html.eex
```diff
@@ -1,125 +1,84 @@
...
+    <div class="col-xs-12 col-md-4">
+
+      <%= SiteWeb.PartialView.ExpandableBlock.render_expandable_block(
+          Map.merge(assigns, %{
+            header: %{text: "Call Us", iconSvgText: nil},
+            id: "call_us",
+            show_expanded: true,
+            view_template: "_call_us.html"
+      })) %>
+
```
thecristen, line 37: Part of me wonders if these repetitious calls can be handled a little more elegantly on the Elixir side (with pattern matching perhaps?!), but this way is also perfectly great.

thecristen (changes requested):
Am noting a few details here. I think the approach is 👍 !

apps/site/lib/site_web/templates/partial/_expandable_block.html.eex
```diff
@@ -0,0 +1,34 @@
...
+        <%= content_tag(:span, svg(@header.iconSvgText), class: "notranslate c-expandable-block__header-icon", aria: [hidden: "true"]) %>
+      <% end %>
+      <div class="c-expandable-block__header-text">
+        <%= @header.text %>
+        <span class="c-expandable-block-caret"></span>
+      </div>
+    </a>
+  </h2>
+  <div
+    class="c-expandable-block__panel collapse<%= if @show_expanded do %> expand<% end %>"
```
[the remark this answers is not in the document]
alicia-mbta: I tried that at first and the corresponding blocks are expanded initially but then that messes up with the carets :/
image

That is why I took the jQuery approach to simulate clicks when the page loads. Do you think this can be achieved only with CSS?

apps/site/lib/site_web/views/partial/expandable_block.ex
```diff
@@ -0,0 +1,7 @@
+defmodule SiteWeb.PartialView.ExpandableBlock do
```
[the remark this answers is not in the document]
alicia-mbta: I did that initially but then I created the separate module in case it could be reused in the future. However it's so small that we could just use `SiteWeb.PartialView.render` indeed, which is more straightforward. 👍

apps/site/lib/site_web/templates/customer_support/_call_us.html.eex
alicia-mbta: You're right, I didn't notice that. Will be creating a separate template for Lost and Found.

apps/site/assets/css/_customer-support.scss
alicia-mbta: Yes, this is simpler, thanks!

apps/site/lib/site_web/templates/partial/_expandable_block.html.eex
alicia-mbta: `@show_expanded` is the flag to expand or leave the block collapsed on page load, not a boolean to control the status of the block (perhaps it should be renamed to `@show_expanded_on_load` then?). `aria-expanded="false"` sets all blocks as collapsed initially, then the new jQuery function triggers the click to expand the ones we want. This was the only way I could get the carets to be in sync regardless of whether the block was expanded or collapsed. I don't think this is the best approach so I'm very open to other suggestions!

apps/site/lib/site_web/templates/customer_support/index.html.eex
alicia-mbta: I thought so as well, too repetitive - I'll see if I can optimize this.

ingridpierre:
Would like the top of the sidebar to be level with the top of the form here:

Untitled

--------
Some breakpoints have awkward line breaks that bump the arrows/carets out of bounds

image

We rely on this probably too much, but can we add the .h3 class to the h2 element here so it looks more like "Get service Updates" instead of Transit Police:
image

--------
Can we change "Report Fraud, Waste, Abuse" to "Report Fraud, Waste, or Abuse"

--------
Possible to change the focus boundary so you can click anywhere in the box +background, not just the text and caret?
image

--------
Other accordions on the site put the focus on the content after you click to expand, which I think is a good behavior for this element. I think right now the focus remains on the header when you click. This example is from the fares accordion on schedules:
image

apps/site/lib/site_web/templates/partial/_expandable_block.html.eex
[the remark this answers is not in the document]
thecristen: I think the caret part would be addressed by setting the aria-expanded to match the @show_expanded state on load.

Was going to write this in the @show_expanded comment but basically: the aria-expanded state should always reflect whether the block is collapsed or not. Then the caret will be pointed appropriately (since your CSS uses aria-expanded state to choose which caret to show). This is handled on click fine, but if it's set to false on load when @show_expanded happens to be true, then there's an inconsistency at the start.

alicia-mbta:
This is ready for re-review 🙏

apps/site/assets/js/support-form.js
```diff
@@ -407,3 +408,9 @@ function showExpandedForm($) {
 function hideExpandedForm($) {
   $(".support-form-expanded").hide();
 }
+
+function focusOnContentWhenExpanding($) {
+  $("div[id^='panel-']").on("shown.bs.collapse", function() {
+    $(this).focus();
+  });
+}
```
thecristen, line 416: Did you want to try making use of the event listener that uses a `.js-focus-on-expand` class? Looks like it works the same as you did here. https://github.com/mbta/dotcom/blob/eaac9fdb9b32f0c55f64689ec23de01de283a513/apps/site/assets/js/collapse.js#L20-L21

apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -41,6 +79,22 @@ defmodule SiteWeb.CustomerSupportController do
...
+  def render_expandable_blocks(assigns, content_blocks \\ @content_blocks) do
+    content_blocks
+    |> Enum.map(fn block ->
+      SiteWeb.PartialView.render(
+        "_expandable_block.html",
+        Map.merge(assigns, %{
+          header: block.header,
+          id: block.id,
+          initially_expanded: block.initially_expanded,
+          view_template: "_#{block.id}.html"
```
thecristen, line 92: This is so cool!

One thing I wonder is what would happen if someone passed in a block.id where the corresponding view_template doesn't exist. Would that fail gracefully? 🤔

apps/site/assets/js/support-form.js
alicia-mbta: Arg, I didn't think to check that. Agreed, no need to duplicate.

apps/site/lib/site_web/controllers/customer_support_controller.ex
alicia-mbta: Good point! I made some changes so at least the page can still be viewed but the blocks that don't have a template are just not rendered.

ingridpierre:
Small things, but otherwise all previous comments look addressed 🙌 :

- The **"View phone numbers"** link needs the .c-call-to-action styling too.
- Conversely, where we say **email mbta.rao@mbta.com** with your..", the email address should not have the c-call-to-action class since it's just a regular inline link.
-  Where we have the hours we've been using the wrong kind of dashes:

> Monday — Friday: 6:30 AM — 8 PM
> Saturday — Sunday: 8 AM — 4 PM
> 
should be:
> Monday - Friday: 6:30 AM -8 PM
> Saturday - Sunday: 8 AM - 4 PM
>

thecristen (approved):
Looking good for me! 👍

thecristen:
@alicia-mbta please change the target branch from `master` to `feature/☎️-customer-support`!

PR #618, opened 2020-09-09
(revision) 💈 special line diagram styling for stops diversion-related alerts

thecristen:
Replaces #596 as I wanted a chance to rebase on the latest master and clean up the changed commits list a bit. See #596 for prior review and progress.

Needs dev review from https://github.com/mbta/dotcom/commit/2f0e79952928b48f7952b1c6c6d704d5cc10aa0d onward.

Ongoing design review -- @ingridpierre, the `BadBooleanError` you cited last week should be fixed now.

### DIFF
--- a/apps/site/assets/ts/models/alert.ts
+++ b/apps/site/assets/ts/models/alert.ts
@@ -45,12 +45,12 @@ export const isCurrentAlert = (
 ): boolean => {
   if (!alert.active_period) return false;
   const dateRanges = alert.active_period.map(ap => activePeriodToDates(ap));
-  const isInARange = dateRanges.find(
+  const isInARange = dateRanges.some(
     (range): boolean => {
       const [start, end] = range;
       if (!start) return false; // end might be null for ongoing alerts
       return currentDate >= start && (end ? currentDate <= end : true);
     }
   );
-  return isCurrentLifecycle(alert) && !!isInARange;
+  return isCurrentLifecycle(alert) && isInARange;
 };
--- a/apps/site/assets/ts/models/alert.ts
+++ b/apps/site/assets/ts/models/alert.ts
@@ -9,5 +9,4 @@ export const isDiversion = ({ effect }: Alert): boolean =>
   effect === "shuttle" ||
   effect === "stop_closure" ||
   effect === "station_closure" ||
-  effect === "detour" ||
-  effect === "suspension";
+  effect === "detour";
--- a/apps/site/assets/ts/models/alert.ts
+++ b/apps/site/assets/ts/models/alert.ts
@@ -6,10 +6,8 @@ export const isHighSeverityOrHighPriority = ({
 }: Alert): boolean => priority === "high" || severity >= 7;
 
 export const isDiversion = ({ effect }: Alert): boolean =>
-  [
-    "detour",
-    "shuttle",
-    "stop_closure",
-    "station_closure",
-    "suspension"
-  ].includes(effect);
+  effect === "shuttle" ||
+  effect === "stop_closure" ||
+  effect === "station_closure" ||
+  effect === "detour" ||
+  effect === "suspension";
--- a/apps/site/assets/ts/models/alert.ts
+++ b/apps/site/assets/ts/models/alert.ts
@@ -1,4 +1,4 @@
-import { Alert } from "../__v3api";
+import { Alert, TimePeriodPairs } from "../__v3api";
 
 export const isHighSeverityOrHighPriority = ({
   priority,
@@ -10,3 +10,47 @@ export const isDiversion = ({ effect }: Alert): boolean =>
   effect === "stop_closure" ||
   effect === "station_closure" ||
   effect === "detour";
+
+const withLeadingZero = (n: string): string => `0${n}`.slice(-2);
+
+const activePeriodToDates = (
+  activePeriod: TimePeriodPairs
+): (Date | null)[] => {
+  const datePattern = /^(\d{4})-(\d{1,2})-(\d{1,2})\s(\d{1,2}):(\d{2})$/;
+
+  return activePeriod.map(
+    (d: string): Date | null => {
+      const match = datePattern.exec(d);
+      if (match) {
+        const [, year, rawMonth, rawDay, rawHour, min] = match;
+        return new Date(
+          `${year}-${withLeadingZero(rawMonth)}-${withLeadingZero(
+            rawDay
+          )}T${withLeadingZero(rawHour)}:${min}:00`
+        );
+      }
+      return null;
+    }
+  );
+};
+
+const isCurrentLifecycle = ({ lifecycle }: Alert): boolean =>
+  lifecycle === "new" ||
+  lifecycle === "ongoing" ||
+  lifecycle === "ongoing_upcoming";
+
+export const isCurrentAlert = (
+  alert: Alert,
+  currentDate: Date = new Date()
+): boolean => {
+  if (!alert.active_period) return false;
+  const dateRanges = alert.active_period.map(ap => activePeriodToDates(ap));
+  const isInARange = dateRanges.find(
+    (range): boolean => {
+      const [start, end] = range;
+      if (!start) return false; // end might be null for ongoing alerts
+      return currentDate >= start && (end ? currentDate <= end : true);
+    }
+  );
+  return isCurrentLifecycle(alert) && !!isInARange;
+};
--- a/apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
+++ b/apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
@@ -93,7 +93,6 @@ const StopCard = (props: StopCardProps): ReactElement<HTMLElement> => {
           {hasDiversion
             ? stopAlerts.filter(isDiversion).map(alert => (
                 <div key={alert.id} className="m-schedule-diagram__alert">
-                  {alertIcon("c-svg__icon-alerts-triangle")}
                   {effectNameForAlert(alert)}
                 </div>
               ))
--- a/apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
+++ b/apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
@@ -6,11 +6,15 @@ import { isMergeStop, diagramWidth } from "./line-diagram-helpers";
 import StopConnections from "./StopConnections";
 import StopPredictions from "./StopPredictions";
 import { alertIcon } from "../../../helpers/icon";
-import { isHighSeverityOrHighPriority } from "../../../models/alert";
+import {
+  isHighSeverityOrHighPriority,
+  isDiversion
+} from "../../../models/alert";
 import { Alert, Route } from "../../../__v3api";
 import MatchHighlight from "../../../components/MatchHighlight";
 import StopFeatures from "./StopFeatures";
 import { StopRefContext } from "./LineDiagramWithStops";
+import { effectNameForAlert } from "../../../components/Alerts";
 
 interface StopCardProps {
   stop: LineDiagramStop;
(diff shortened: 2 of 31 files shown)


arkadyan:
You've probably already seen this, but looks like there are a few dialyzer complaints.

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -151,15 +151,14 @@ defmodule SiteWeb.ScheduleController.LineApi do
...
-  # make adjustment to the diagram
-  # based on the nature of the disruption.
-  # e.g. for detour we modify the stop LEADING to it
-  # but for shuttle we DON'T modify the FINAL stop
-  # note: there might be more logic than that.
+  # for each list of disrupted stops, make adjustment to the diagram based on
+  # the nature of the disruption. e.g. for detour or stop/station closure we
+  # modify the stop preceding it but for shuttle we don't style the final stop
   defp shift_indices_by_disruption_effect(indices, effects) do
+    effect_avoids_stop? = effects -- effects -- [:stop_closure, :station_closure, :detour]
```
arkadyan: `effect_avoids_stop?` isn't actually a boolean as the result of this, which is causing the `and` problem below. Would something like this be equivalent but result in a boolean value?

```suggestion
    effect_avoids_stop? = Enum.any?(effects, &Enum.member?([:stop_closure, :station_closure, :detour], &1))
```

Alternately, you could also do something like `effects -- effects -- [:stop_closure, :station_closure, :detour] != []`.

apps/site/assets/ts/models/alert.ts
```diff
@@ -10,3 +10,47 @@ export const isDiversion = ({ effect }: Alert): boolean =>
...
+          )}T${withLeadingZero(rawHour)}:${min}:00`
+        );
+      }
+      return null;
+    }
+  );
+};
+
+const isCurrentLifecycle = ({ lifecycle }: Alert): boolean =>
+  lifecycle === "new" ||
```
arkadyan, line 38: This is good. `["new", "ongoing", "ongoing_upcoming"].includes(lifecycle)` also works.

apps/site/assets/ts/models/alert.ts
```diff
@@ -10,3 +10,47 @@ export const isDiversion = ({ effect }: Alert): boolean =>
...
+  lifecycle === "ongoing" ||
+  lifecycle === "ongoing_upcoming";
+
+export const isCurrentAlert = (
+  alert: Alert,
+  currentDate: Date = new Date()
+): boolean => {
+  if (!alert.active_period) return false;
+  const dateRanges = alert.active_period.map(ap => activePeriodToDates(ap));
+  const isInARange = dateRanges.find(
```
arkadyan, line 48: Using `dateRanges.some` instead of `dateRanges.find` here would give you a boolean and save you the `!!` later.

apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
```diff
@@ -91,23 +101,20 @@ const StopCard = (props: StopCardProps): ReactElement<HTMLElement> => {
 
         <div className="m-schedule-diagram__stop-details">
           {StopConnections(routeStop.connections)}
-          {!isDestination &&
-          liveData &&
-          liveData.headsigns.length &&
-          liveData.headsigns.some(hasPredictionTime) ? (
+          {showPrediction ? (
```
arkadyan, line 104: I think there's already a test for conditionally showing the diversion, is showing the prediction also already covered?

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -157,14 +174,15 @@ defmodule SiteWeb.ScheduleController.LineApi do
...
   defp shift_indices_by_disruption_effect(indices, effects) do
-    effect_avoids_stop? = effects -- effects -- [:stop_closure, :station_closure, :detour]
+    effects_avoiding_stop = effects -- effects -- [:stop_closure, :station_closure, :detour]
 
     cond do
       :shuttle in effects ->
         List.delete_at(indices, -1)
 
-      effect_avoids_stop? && List.first(indices) > 0 ->
+      Kernel.length(effects_avoiding_stop) > 0 && List.first(indices) > 0 ->
```
arkadyan, line 185: This change addresses my previous comment. I think you can probably switch back to `and` if you want.

arkadyan (commented):
Nice! Add a few comments, but mostly looking good on the code side.

apps/site/lib/site_web/controllers/schedule/line_api.ex
arkadyan: (See later comment where I see you updated this further…)

thecristen: I like your proposed change better so will update it again :)

apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
thecristen: Maybe not, will check into it!

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -126,10 +132,99 @@ defmodule SiteWeb.ScheduleController.LineApi do
...
       route_stop: RouteStop.to_json_safe(map),
-      stop_data: Enum.map(data, fn {key, value} -> %{branch: key, type: value} end)
+      stop_data:
+        Enum.map(data, fn {key, value} -> %{branch: key, type: value, has_disruption?: false} end)
     }
   end
 
+  @spec maybe_add_stop_disruptions([line_diagram_stop]) :: [line_diagram_stop]
+  defp maybe_add_stop_disruptions(stops_list) do
+    if Enum.any?(stops_list, &stop_has_disruption?(&1)) do
```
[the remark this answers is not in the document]
thecristen: This has been removed.

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -126,10 +132,99 @@ defmodule SiteWeb.ScheduleController.LineApi do
...
         |> json_safe_alerts(date),
       route_stop: RouteStop.to_json_safe(map),
-      stop_data: Enum.map(data, fn {key, value} -> %{branch: key, type: value} end)
+      stop_data:
+        Enum.map(data, fn {key, value} -> %{branch: key, type: value, has_disruption?: false} end)
     }
   end
 
+  @spec maybe_add_stop_disruptions([line_diagram_stop]) :: [line_diagram_stop]
+  defp maybe_add_stop_disruptions(stops_list) do
```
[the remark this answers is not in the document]
thecristen: Have a few tests in the latest commit!

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -126,10 +132,99 @@ defmodule SiteWeb.ScheduleController.LineApi do
...
+
+      stop_data
+      |> List.update_at(stop_or_terminus, fn sd ->
+        Map.replace!(sd, :has_disruption?, true)
+      end)
+    end)
+  end
+
+  defp do_stops_list_with_disruptions(stops_list) do
+    disrupted_stop_indices =
```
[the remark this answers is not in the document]
thecristen: Refactored into an `Enum.chunk_while` call, it's still a little complex but much more readable now.

thecristen:
Ready for re-review.

@arkadyan, the final commit of the five added commits is purely a refactor for readability / test adding commit. Whereas the two preceding it, https://github.com/mbta/dotcom/pull/618/commits/9cda4ac21337eca5b4081a6f052e0367f64383bc and https://github.com/mbta/dotcom/pull/618/commits/cfbd56c0b18ab6d6ee9dfd3847777bee3b3252b6, are small unforeseen bugfixes that change (and improve) the behavior of alert filtering and matching.

ingridpierre:
Looking great so far! Small thing, can we also have the detour text align with the presence/absence of the crowding column? See this example where the right edge of the right-aligned text doesn't match above and below:
image

Once that's addressed I think I might be at the limit of what I can observe right now, would have liked to validate that it works for CR and Subway diversion, but there are none active right now. Is there any way to mock this up in the data? I wouldn't want to wait til the October diversions to launch here.

apps/site/lib/site_web/controllers/schedule/line/diagram_format.ex
```diff
@@ -0,0 +1,121 @@
+defmodule SiteWeb.ScheduleController.Line.DiagramFormat do
```
arkadyan, line 1: I like the idea to encapsulate this.

apps/site/lib/site_web/controllers/schedule/line/diagram_format.ex
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  defp shift_indices_by_effect(indices, [effect | _])
+       when effect in [:stop_closure, :station_closure, :detour]
+       when hd(indices) > 0 do
+    [hd(indices) - 1] ++ indices
+  end
+
+  defp shift_indices_by_effect(indices, [_ | effects]),
+    do: shift_indices_by_effect(indices, effects)
```
arkadyan, line 30: Of all the lines Codecov is commenting on this one feels like possibly the one most worth considering adding a quick test for.

arkadyan (approved):
Radical! 🤙

ingridpierre:
Apart from the previous comment (ok too if you want to address in separate pr), I think the rules and behavior look good, shuttles have the right limits in the diagram styling and the rules around when to show a prediction vs a label make sense at this point in time. 👍 

For the record, we looked at:
- CR alert on a particular trip/branch
- Multiple Green line alerts
- Multiple alerts on single line (OL)
- Detours and other effects on bus routes

thecristen:
Thanks all! I did find a few more edge cases I wanted to tackle, so I ended up working towards that.

Here's one example of an edge case I found: pictured below, both are shuttles, ending at Stony Brook and Assembly respectively. However they were rendering slightly differently because it's not handling **multiple** shuttles on the **same** line correctly.

image

Other weak points involved rendering in and around merge points.

So, there are a _**lot**_ more tests added in this last set of changes, along with some light refactoring. Hoping the commits make sense.

apps/site/lib/site_web/controllers/schedule/line/diagram_format.ex
thecristen: @arkadyan Rewrote this and added a test for this and many others..

apps/site/lib/site_web/controllers/schedule/line/diagram_format.ex
```diff
@@ -66,19 +66,35 @@ defmodule SiteWeb.ScheduleController.Line.DiagramFormat do
...
-        [{_, last_effects} | _] ->
-          if stop_effects == last_effects do
-            {:cont, Enum.reverse([{index, stop_effects} | chunk]), []}
+        chunk ->
+          last_stop = List.last(chunk)
+          if stop_fx == effects_for_stop(last_stop) do
+            {:cont, chunk ++ [stop]}
           else
-            {:cont, [{index, stop_effects}]}
+            case stop_fx do
```
arkadyan, line 153: I generally think of multiple types of conditionals (e.g. a `case` within an `if`) to be a red flag, potentially a sign that the conditionals should be combined or the function is responsible for too many things and could use refactoring. This is now 4 layers of nested conditionals deep. Are there ways we could consider refactoring out sub-functions that might each handle a single thing?

arkadyan (commented):
Added a comment about potential refactoring, but it doesn't have to be blocking.

ingridpierre:
It looks like there are still some issues with coloring diversions when a branch is expanded/collapsed. First image shows diversion correctly applied, second image shows diversion styling breaking when expansion happens:

image

image

thecristen:
> It looks like there are still some issues with coloring diversions when a branch is expanded/collapsed. First image shows diversion correctly applied, second image shows diversion styling breaking when expansion happens:
> 
> image
> 
> image

Noted in https://app.asana.com/0/385363666817452/1197587990510618/f to be addressed soon - we'll ship without this for now as we don't expect to encounter this edge case in the near future.

PR #619, opened 2020-09-10
⏱️ Commuter Rail Schedules | Some "via Fairmount" trips missing from Franklin Line 

thecristen:
#### Summary of changes
**Asana Ticket:** [⏱️ Commuter Rail Schedules | Some "via Fairmount" trips missing from Franklin Line ](https://app.asana.com/0/385363666817452/1188018759297667/f)

Added a few missing trips so the timetable display matches the PDF.

---


arkadyan (approved):
I re-ran the tests. This looks good to me once those pass.

PR #621, opened 2020-09-11
☎️ Add support for First and Last names (to replace Full Name)

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [☎️ Add support for First and Last names (to replace Full Name)](https://app.asana.com/0/555089885850811/1188547161220115)

Update message and mailer logic (and adjust tests): account for first & last name instead of just one field for name.


arkadyan (approved):
👍 Looks great!

PR #622, opened 2020-09-14
Update test to handle both sides of the Nubian stop ID change

arkadyan:
Update test to handle both sides of the Nubian stop ID change

This will pass against either the old or the new value. Once the
`place-nubn` data change is live in production, we can revert to the
simple equality check.

No ticket, just noticed the failure when investigating the new data on dev-green.

---


thecristen (approved):
👍

PR #623, opened 2020-09-15
CMS Script Embed ¶ Type

amaisano:
#### Summary of changes
**Asana Ticket:** [Whitelist script tag domains (Formstack, Tableau)](https://app.asana.com/0/399597520748778/1190828056070456)

Allows pure, unsanitized HTML from the CMS to be rendered 1:1 as it was input. See mbta/cms#347 for back-end regulations on this.

Using this approach since:
- Only system admins will use this feature, and not very frequently.
- Some sites (like Tableau) have embed code that is a mix of script, object, and div tags, which are nearly impossible to regulate as one acceptable rule in the HTML Scrubber as each of these elements are co-dependent.
- Paragraph fields can be steered away from the near-universal rewriter approach most HTML content from the CMS goes through, which means we can avoid the Scrubber.
- We can still white/black list script domains if needed, but doing so will result in very broken output depending on how complicated the embed code is (see the Tableau example).

Front-end:
embeds-in-body-and-right-rail

Back-end:
image

---

### DIFF
--- a/apps/cms/lib/partial/paragraph/code_embed.ex
+++ b/apps/cms/lib/partial/paragraph/code_embed.ex
@@ -1,0 +1,30 @@
+defmodule CMS.Partial.Paragraph.CodeEmbed do
+  @moduledoc """
+  Represents a CodeEmbed Paragraph in the CMS.
+  Body is raw HTML from CMS, and is not scrubbed or rewritten.
+  """
+
+  alias CMS.Partial.Paragraph.ColumnMultiHeader
+  alias Phoenix.HTML
+
+  import CMS.Helpers, only: [field_value: 2, parse_paragraphs: 3]
+
+  defstruct header: nil,
+            right_rail: false,
+            body: HTML.raw("")
+
+  @type t :: %__MODULE__{
+          header: ColumnMultiHeader.t() | nil,
+          right_rail: boolean,
+          body: HTML.safe()
+        }
+
+  @spec from_api(map, Keyword.t()) :: t
+  def from_api(data, preview_opts) do
+    %__MODULE__{
+      header: data |> parse_paragraphs(preview_opts, "field_multi_column_header") |> List.first(),
+      right_rail: field_value(data, "field_right_rail"),
+      body: field_value(data, "field_code")
+    }
+  end
+end
--- a/apps/cms/lib/partial/paragraph/code_embed.ex
+++ b/apps/cms/lib/partial/paragraph/code_embed.ex
@@ -1,7 +1,7 @@
 defmodule CMS.Partial.Paragraph.CodeEmbed do
   @moduledoc """
   Represents a CodeEmbed Paragraph in the CMS.
-  Body is raw HTML from CMS, and is not scrubbed or rewritten.
+  Body is raw HTML from CMS, and should not be scrubbed or rewritten.
   """
 
   alias CMS.Partial.Paragraph.ColumnMultiHeader
@@ -24,7 +24,7 @@ defmodule CMS.Partial.Paragraph.CodeEmbed do
     %__MODULE__{
       header: data |> parse_paragraphs(preview_opts, "field_multi_column_header") |> List.first(),
       right_rail: field_value(data, "field_right_rail"),
-      body: field_value(data, "field_code")
+      body: data |> field_value("field_code") |> HTML.raw()
     }
   end
 end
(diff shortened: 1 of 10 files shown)


apps/cms/lib/partial/paragraph/code_embed.ex
```diff
@@ -0,0 +1,30 @@
...
+  alias Phoenix.HTML
+
+  import CMS.Helpers, only: [field_value: 2, parse_paragraphs: 3]
+
+  defstruct header: nil,
+            right_rail: false,
+            body: HTML.raw("")
+
+  @type t :: %__MODULE__{
+          header: ColumnMultiHeader.t() | nil,
```
amaisano, line 17: `header` is so this embedded code can be "introduced" with custom HTML text, so that this block can be easily moved around the page w/o losing context. This is the same behavior as content lists and multi column paragraphs. These "blocks" basically have an intro/context/title chunk, and the actual content chunk. Together it makes a modular block.

apps/cms/lib/partial/paragraph/code_embed.ex
```diff
@@ -0,0 +1,30 @@
...
+          right_rail: boolean,
```
amaisano, line 18: Pretty much all of our paragraphs can be set to show in the right sidebar of the page on the front-end. This allows authors to put a small code snippet/embed in the right rail of the template, similar to other paragraphs.

apps/cms/lib/partial/paragraph/code_embed.ex
```diff
@@ -0,0 +1,30 @@
...
+          body: HTML.safe()
```
amaisano, line 19: @arkadyan should this be set to `HTML.raw()`?

arkadyan: I'm not sure I follow the question, could you clarify it, please? It should be whatever the type of the data is, but I'm not sure that's what you are asking. I'm also not finding what `HTML.raw()` is.

apps/site/lib/site/content_rewriter.ex
```diff
@@ -33,6 +33,17 @@ defmodule Site.ContentRewriter do
     rewrite(Enum.join(content), conn)
   end
 
+  @doc """
+  Certain CMS content should have its HTML content passed through (scripts, iframes, etc.).
+  This content is not scrubbed, santitized, or rewritten due to its variability and complexity.
+  """
+  @spec rewrite_raw(Phoenix.HTML.safe() | String.t(), Plug.Conn.t()) :: Phoenix.HTML.safe()
+  def rewrite_raw(content, _conn) do
```
arkadyan, line 41: Could you please add some tests for this function?

apps/cms/lib/partial/paragraph/code_embed.ex
amaisano: Nevermind -- this is a typespec, and there is only a HTML.safe() type.

apps/cms/lib/partial/paragraph/code_embed.ex
arkadyan: Is this needed for the 2 embeds we have currently? Given that the whole point of this paragraph type is explicitly to not work like other paragraphs and be "outside the flow", I can see an argument for keeping it as simple as possible.

apps/cms/test/paragraph_test.exs
```diff
@@ -273,6 +274,20 @@ defmodule CMS.ParagraphTest do
...
 
+    test "parses a code embed paragraph" do
+      code_embed_data = api_paragraph("code_embed")
+
+      assert %CodeEmbed{
+               header: header,
+               body: code
+             } = from_api(code_embed_data)
+
+      assert %ColumnMultiHeader{} = header
```
arkadyan, line 285: If you want to keep the header, could you please test it's parsing behavior? Given that it works differently than the body element I can see that as a potential source of confusion, so calling out the difference seems important.

arkadyan (changes requested):
Comments about a couple of tests and the header, but otherwise looks good.

apps/cms/test/paragraph_test.exs
amaisano: Done. The header component is still scrubbed just like all other paragraph type header components -- this is not the EmbedCode property that deals with raw code embedding.

amaisano:
> Is this needed for the 2 embeds we have currently?

Yes. It is a content requirement and applies to all paragraph types in order to maintain portability/modularity for authors.

apps/site/lib/site/content_rewriter.ex
amaisano: I found this function to be unnecessary, so removed it. See recent code changes.

arkadyan (approved):
👍

PR #625, opened 2020-09-17
Do not discard predicted schedules if they are less than 24 hours away

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Stations/Stops/TNM | False "No departures within 24 hours" language](https://app.asana.com/0/555089885850811/1188548382586302)

Added a fix to compare the dates of the predicted schedules so there's no "no departures within 24 hours" message.

Before:
image

After:
image

### DIFF
--- a/apps/site/test/predicted_schedule_test.exs
+++ b/apps/site/test/predicted_schedule_test.exs
@@ -161,6 +161,11 @@ defmodule PredictedScheduleTest do
                %{schedule: %{trip: %{id: "Trip 3"}}, prediction: %{trip: %{id: "Trip 3"}}}
              ] = predicted_schedules
     end
+
+    test "returns a list of predicted schedules, including ones that are within the next 24 hours" do
+      predicted_schedules = get("15", 1, direction_id: 1, now: Util.now())
+      refute Enum.empty?(predicted_schedules)
+    end
   end
 
   describe "group/2" do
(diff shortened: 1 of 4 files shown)


apps/site/test/predicted_schedule_test.exs
```diff
@@ -161,6 +161,11 @@ defmodule PredictedScheduleTest do
                %{schedule: %{trip: %{id: "Trip 3"}}, prediction: %{trip: %{id: "Trip 3"}}}
              ] = predicted_schedules
     end
+
+    test "returns a list of predicted schedules, including ones that are within the next 24 hours" do
```
arkadyan, line 165: I'm struggling to follow how the fix works. I think this test is relying on live data from the API? Maybe would it be possible to use `schedules_fn` and `predictions_fn` to more explicitly set up mock data that describes the problem case? I'm wondering if that might help with understanding the problem and how this solution addresses it.

alicia-mbta: I can probably write a better test (and will do so) but how do the changes to `predicted_schedule.ex` look to you?
Before, the list of predicted schedules sometimes was empty because of the `DateTime` comparison.
That has been changed to only discard dates that are not in the [now, 24 hours from now] interval. Let me know if that still doesn’t look like a valid solution.

PR #627, opened 2020-09-22
Get schedules for tomorrow if there are no schedules left for today

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Stations/Stops/TNM | False "No departures within 24 hours" language](https://app.asana.com/0/555089885850811/1188548382586302)

Added a fix to compare the dates of the predicted schedules so there's no "no departures within 24 hours" message.

Before:
image

After:
image


arkadyan (approved):
Ah, so the problem was that we thought we had some times because we were checking before the filter stripped them out? Interesting.

Nice fix!

PR #628, opened 2020-09-24
Reverse the "I would like a response" default

thecristen:
#### Summary of changes
**Asana Ticket:** [☎️ Reverse the "I would like a response" default](https://app.asana.com/0/385363666817452/1188535657153270/f)

Changes the copy on the frontend, shows the response form on the front end by default, and aligns the logic on the `CUSTREQUIRERESP` field.

Also decided to change the _name_ of the relevant variable from request_response to no_request_response such that a "positive" response (checked) means "No" (it's less confusing for my brain, at least).

---

### DIFF
--- a/apps/feedback/lib/mailer.ex
+++ b/apps/feedback/lib/mailer.ex
@@ -8,7 +8,7 @@ defmodule Feedback.Mailer do
     no_request_response = if message.no_request_response, do: "No", else: "Yes"
 
     _ =
-      if message.no_request_response do
+      unless message.no_request_response do
         Logger.info("HEAT Ticket submitted by #{format_email(message.email)}")
       end
 
--- a/apps/feedback/lib/mailer.ex
+++ b/apps/feedback/lib/mailer.ex
@@ -5,10 +5,10 @@ defmodule Feedback.Mailer do
 
   @spec send_heat_ticket(Feedback.Message.t(), [map()]) :: {:ok, any} | {:error, any}
   def send_heat_ticket(message, photo_info) do
-    request_response = if message.request_response, do: "Yes", else: "No"
+    no_request_response = if message.no_request_response, do: "No", else: "Yes"
 
     _ =
-      if message.request_response do
+      if message.no_request_response do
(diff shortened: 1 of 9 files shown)


ingridpierre:
lgtm! 👍

apps/feedback/lib/mailer.ex
```diff
@@ -5,10 +5,10 @@ defmodule Feedback.Mailer do
 
   @spec send_heat_ticket(Feedback.Message.t(), [map()]) :: {:ok, any} | {:error, any}
   def send_heat_ticket(message, photo_info) do
-    request_response = if message.request_response, do: "Yes", else: "No"
+    no_request_response = if message.no_request_response, do: "No", else: "Yes"
 
     _ =
-      if message.request_response do
+      if message.no_request_response do
```
arkadyan, line 11: Does this condition need to be reversed? Or maybe even better just check for `email` instead of `no_request_response`?

arkadyan (commented):
One logic question, though maybe I'm turned around… Otherwise looks good. I think it needs a rebase though.

apps/feedback/lib/mailer.ex
thecristen: Why change it so much? Technically on the front end it's possible to have email be populated either way (e.g. if you fill it, then change your mind and hide the contact info, it doesn't automatically clear). So some additional logic would need to handle this check properly...

thecristen: Hm it does look it needs to be reversed though!

arkadyan: Yeah, sorry that was confusing. I was wondering if it should be conditional on email just because as you say email could be set regardless of this flag. But whichever seems best to you is fine by me.

arkadyan (approved):
👍

apps/feedback/lib/mailer.ex
thecristen: I've reversed it

arkadyan:
I haven't had a chance to look at the details yet but looks like there are some failing tests.

PR #630, opened 2020-09-25
Skip any gaps that are missing critical data

arkadyan:
#### Summary of changes
**Asana Ticket:** [❌ Kingston Line page broken](https://app.asana.com/0/385363666817452/1195805969093134)

Skip any gaps that are missing critical data

Stop-gap measure to get the page to render while we figure out an actual
complete solution.

---

### DIFF
--- a/apps/site/assets/ts/schedule/components/line-diagram/graphics/Merge.tsx
+++ b/apps/site/assets/ts/schedule/components/line-diagram/graphics/Merge.tsx
@@ -100,13 +100,17 @@ const Merges = ({ lineDiagram }: MergeGraphicsProps): ReactElement | null => {
 
   return (
     <g className="line-diagram-svg__merge">
-      {stopGaps.map(([from, to]) => (
-        <Line
-          key={`${from.route_stop.id}-${to.route_stop.id}-line`}
-          from={from}
-          to={to}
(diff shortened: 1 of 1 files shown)


apps/site/assets/ts/schedule/components/line-diagram/graphics/Merge.tsx
```diff
@@ -100,13 +100,17 @@ const Merges = ({ lineDiagram }: MergeGraphicsProps): ReactElement | null => {
...
-          from={from}
-          to={to}
-        />
-      ))}
+      {stopGaps.map(
+        ([from, to]) =>
+          from &&
+          to && (
+            <Line
+              key={`${from.route_stop.id}-${to.route_stop.id}-line`}
```
thecristen, line 108: Do you know if it loads if you put the check here (the string being used as a key) instead?

thecristen (approved):

apps/site/assets/ts/schedule/components/line-diagram/graphics/Merge.tsx
arkadyan: Unfortunately, no, I tried that first.

PR #631, opened 2020-09-25
Consider "terminus" as "stop" in isStopOnMainLine()

thecristen:
No ticket, builds off #630. Existing code broke for branches that comprise of only a single terminus stop. :)


arkadyan (approved):
👍 

Feel free to link it to https://app.asana.com/0/385363666817452/1195805969093134

PR #632, opened 2020-09-25
☎️ Replace Full Name field with First Name and Last Name fields

thecristen:
#### Summary of changes
**Asana Ticket:** [☎️ Replace Full Name field with First Name and Last Name fields](https://app.asana.com/0/385363666817452/1188547161220117/f)

The page seems to break on form submit so there's an extra commit meant to address that as well...

---

### DIFF
--- a/apps/site/assets/js/support-form.js
+++ b/apps/site/assets/js/support-form.js
@@ -221,9 +221,15 @@ const validators = {
   service: function($) {
     return !!$("[name='support[service]']:checked").val();
   },
-  name: function($) {
+  first_name: function($) {
     if (responseRequested($)) {
-      return $("#name").val().length !== 0;
+      return $("#first_name").val().length !== 0;
+    }
+    return true;
+  },
+  last_name: function($) {
+    if (responseRequested($)) {
+      return $("#last_name").val().length !== 0;
     }
     return true;
   },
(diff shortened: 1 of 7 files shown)


alicia-mbta (approved):
Left a small comment but nothing major.
The other question I have is whether this branch should be merged into the feature branch instead of `master`?

thecristen:
> Left a small comment but nothing major.
> The other question I have is whether this branch should be merged into the feature branch instead of `master`?

You're totally right; I updated the target branch to be the feature branch, thanks for catching!

ingridpierre:
This appears to function fine when I'm in the form, but I'm yet to see any confirmation emails populate from test submissions. Am I missing something needed here?

thecristen:
@ingridpierre with regards to no email response, [buried in the docs](https://github.com/mbta/dotcom/tree/master/apps/feedback#development-logging):

> In development, the application is configured to print the contents of the email to the log instead of sending an actual email. This is logged at the `info` log level, but the default log level is set to warn. Add `LOGGER_LEVEL=info` to your .env file to see the feedback emails in your logs.

So when developing locally you wouldn't get an email, but you can see the recorded information logged in your terminal window if you have the logger level variable set as specified above. Sorry I hadn't mentioned this when posting this PR :/

PR #636, opened 2020-10-05
Use configuration settings for test expectations

arkadyan:
Use configuration settings for test expectations

If you've changed these values in your `.env` file the tests would fail
for you.

No ticket, just fixes a problem I've been experiencing.

---

### DIFF
--- a/apps/feedback/test/mailer_test.exs
+++ b/apps/feedback/test/mailer_test.exs
@@ -17,7 +17,9 @@ defmodule Feedback.MailerTest do
         nil
       )
 
-      assert Test.latest_message()["to"] == ["test@test.com"]
+      assert Test.latest_message()["to"] == [
+               Application.get_env(:feedback, :support_ticket_to_email)
+             ]
     end
 
     test "has the body format that heat 2 expects" do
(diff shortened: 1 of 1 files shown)


thecristen:
Was there an expectation that developers would be setting these particular environment variables, and where did this expectation come from? I just wanted to ask because this is the first I'd seen of it.

Alternatively if this is a new expectation of developers, I'd like to hear more about the potential use case for it, and I'd expect to see examples noted in the example .envrc file.

arkadyan:
@thecristen They are currently in the [`.envrc.template`](https://github.com/mbta/dotcom/blob/master/.envrc.template#L14) and they are used on the servers. I don't know if there's an expectation to set them per se—by default in dev we no longer try to send actual emails—but you could set them if you are testing sending actual emails to Heat. Moreso it just seemed like tests shouldn't be relying on values that they expect to be configured in environment variables this way. That is to say, these specific values are not special and don't need to be tested.

thecristen (approved):

paulswartz:
An alternative here is to set these variables to static values in `config/test.exs`, so that local environment variables don't override them during unit testing.

PR #620, opened 2020-09-10
Authenticate our domain for Facebook

arkadyan:
#### Summary of changes
**Asana Ticket:** [Associate MBTA.com domain with Facebook account](https://app.asana.com/0/555089885850811/1192988614433558)

Authenticate our domain for Facebook

[This can be removed as soon as we complete the verification process.](https://app.asana.com/0/385363666817452/1193278706153525)

---

Before getting review, please check the following:

* [x] Does frontend functionality render and work correctly in IE?
* [x] Have we load-tested any new pages or internal API endpoints that will receive significant traffic?
* [x] Are interactive elements accessible to screen readers?
* [x] Have you checked for tech debt you can address in the area you're working in?
* [x] If this change involves routes, does it work correctly with pertinent "unusual" routes such as the combined Green Line, Silver Line, Foxboro commuter rail, and single-direction bus routes like the 170?
* [x] Are the changes organized into self-contained commits with descriptive and well-formatted commit messages?

PR #639, opened 2020-10-06
Remove Facebook domain verification meta tag

arkadyan:
Asana ticket: [Remove Facebook verification meta-tag](https://app.asana.com/0/385363666817452/1193278706153525)

Reverts mbta/dotcom#620

Our domain has been associated with our Facebook account, so this is no longer needed.


alicia-mbta (approved):
👍

PR #644, opened 2020-10-14
Add Subject field to the customer support form

thecristen:
#### Summary of changes
**Asana Ticket:** [☎️ Add Subject field to the customer support form](https://app.asana.com/0/385363666817452/1188535657153281/f)

Front end support for the (not yet required) subject field and its interaction with the value selected from the service category radio inputs.

---

### DIFF
--- a/apps/site/lib/site_web/views/customer_support_view.ex
+++ b/apps/site/lib/site_web/views/customer_support_view.ex
@@ -6,6 +6,8 @@ defmodule SiteWeb.CustomerSupportView do
 
   import Phoenix.HTML.Tag, only: [content_tag: 2, content_tag: 3]
 
+  alias Feedback.Message
+
   def photo_info(%{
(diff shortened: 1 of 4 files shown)


apps/site/lib/site_web/views/customer_support_view.ex
```diff
@@ -45,6 +47,13 @@ defmodule SiteWeb.CustomerSupportView do
     )
   end
 
+  @spec subject_map([Message.service_option_with_subjects()]) ::
```
arkadyan, line 50: Dialyzer seems unhappy with this spec. Should the output be a map instead of a tuple maybe?

arkadyan (approved):
Looks good! (Once dialyzer is happy.)

This might be a question for Ingrid or Yinka, and might be for a future ticket, but I wonder if we might want to disable the subject dropdown until you've selected a category.

PR #645, opened 2020-10-15
"CharlieCard or Ticket Number" frontend field

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [☎️ Add a new "CharlieCard or Ticket Number" field](https://app.asana.com/0/555089885850811/1188547161220084)

Adding field to the frontend. It's only visible when the subject is "CharlieCards & Tickets".

image

### DIFF
--- a/apps/site/assets/js/test/support-form_test.js
+++ b/apps/site/assets/js/test/support-form_test.js
@@ -10,7 +10,8 @@ import {
   setupSubject,
   handleSubmitClick,
   rescale,
-  handleModeChangeSelection
+  handleModeChangeSelection,
+  handleSubjectChange
 } from "../support-form";
 import testConfig from "../../ts/jest.config";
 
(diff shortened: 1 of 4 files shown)


apps/site/assets/js/test/support-form_test.js
```diff
@@ -571,7 +572,56 @@ describe("support form", () => {
...
+       </div>
+     `);
+      handleSubjectChange($);
+    });
+
+    afterEach(() => {
+      $("#test").remove();
+    });
+
+    it("shows the 'CharlieCard or Ticket number' field", () => {
```
arkadyan, line 600: For the sake of someone scanning through the tests, it might be handy to include something in these test descriptions about in what circumstances it does or does not show the field.

arkadyan (approved):
Included a tiny suggestion, but looks good to me.

PR #646, opened 2020-10-16
Run message body through profanity sanitizer

thecristen:
#### Summary of changes
**Asana Ticket:** [☎️ Profanity sanitizer](https://app.asana.com/0/385363666817452/1188547161220113/f)

Uses the `bad-words` package to filter the text in the comments field before sending to the backend.
I didn't have the heart to put curse words in our codebase, so the test that I wrote for it modifies the library's behavior a bit to add a fake curse word... hopefully you get the idea 😅

### DIFF
--- a/apps/site/assets/js/test/support-form_test.js
+++ b/apps/site/assets/js/test/support-form_test.js
@@ -3,6 +3,7 @@ import jsdom from "mocha-jsdom";
 import { File } from "file-api";
 import "custom-event-autopolyfill";
 import sinon from "sinon";
+import { cloneDeep } from "lodash";
 import {
   clearFallbacks,
   handleUploadedPhoto,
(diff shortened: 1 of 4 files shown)


apps/site/assets/js/test/support-form_test.js
```diff
@@ -521,6 +522,33 @@ describe("support form", () => {
       // not sure how to recover actual values
       assert.equal(photos.length, toUpload.length);
     });
+
+    it("passes comment thru profanity filter", () => {
+      const Filter = require("bad-words");
+      const f = new Filter();
+      f.addWords("redsox");
```
alicia-mbta, line 529: I take it you're a Yankees fan? :laughing:

alicia-mbta (approved):
Looks like a pretty easy to use package! Thanks for looking into this 👍

PR #648, opened 2020-10-20
Move confirmation message inside the hidden div

thecristen:
No ticket, just a regression I'd noticed. Right now the customer support form always says at top "Your comments have been sent to our Customer Support team." :)


arkadyan (approved):
👍 Good catch

PR #651, opened 2020-10-27
Stitch together linked lists of stops on the back-end

arkadyan:
#### Summary of changes
**Asana Ticket:** [🚆 Line Diagram | Incorrectly identified shuttle stops, Rockport Line](https://app.asana.com/0/555089885850811/1183912994482917)

Stitch together linked lists of stops on the back-end

The first commit adds some tests that pass with the current data, and the second updates the tests to mirror the future data on dev-green and stitches the shuttle stops into the branch to correct the issues in the line diagram.

Screen Shot 2020-10-27 at 11 30 31

---

### DIFF
--- a/apps/stops/lib/route_stop.ex
+++ b/apps/stops/lib/route_stop.ex
@@ -107,6 +107,7 @@ defmodule Stops.RouteStop do
     |> Enum.map(
       &do_list_from_route_pattern(&1, route, direction_id, use_route_id_for_branch_name?)
     )
+    |> maybe_stitch_chunks()
     |> merge_branch_list(direction_id)
     |> maybe_correct_for_lechmere_shuttle()
   end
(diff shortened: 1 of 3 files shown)


apps/stops/lib/route_stop.ex
```diff
@@ -383,6 +384,74 @@ defmodule Stops.RouteStop do
...
+
+  defp do_linked_patterns({_, last_stop_a}, {first_stop_b, _})
+       when last_stop_a == first_stop_b,
+       do: -1
+
+  defp do_linked_patterns(_, _), do: 0
+
+  @spec stitch([RouteStop.t()], [RouteStop.t()]) :: [RouteStop.t()]
+  defp stitch(a, b) do
+    if linked_patterns(a, b) == -1, do: do_stitch(a, b), else: do_stitch(b, a)
```
thecristen, line 440: Do we really want to `do_stitch(b, a)` even when `linked_patterns(a, b)` is 0 (e.g. last stop of b is _not_ equal to first stop of a). Wouldn't this mean we end up omitting the first stop in a when it's not needed?

thecristen (approved):
Thanks Matt. Just one question to consider, otherwise looks good.

apps/stops/lib/route_stop.ex
arkadyan: You are correct that we don't want to do that. Part of line 396 is about filtering out those 0 cases—in those cases, we don't call `stitch`.

thecristen: Ah sorry I missed that -- great!

PR #654, opened 2020-10-28
fix: unclosed  h5 tag in template

fjlanasa:
just fixing an unclosed `h5` tag that was causing slight styling issues.

## Before
Screen Shot 2020-10-27 at 8 15 56 PM

## After
Screen Shot 2020-10-27 at 8 15 19 PM


arkadyan (approved):
👍

PR #656, opened 2020-10-28
Fix for error when trying to convert "1A-2" to integer

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Elixir.ArgumentError: argument error](https://app.asana.com/0/385363666817452/1191566494821186)

The trip details could not be loaded because the logic was trying to convert "1A-2" to an integer.

Before:
image

After:
image


arkadyan (approved):
👍 This is great! As another option, I could also see moving the `Zone.general_zone` calls into `calculate_commuter_rail` before each time we call `String.to_integer`, but it would be the same effect.

PR #657, opened 2020-10-28
Use of plugin @babel/plugin-syntax-dynamic-import

alicia-mbta:
Attempt to fix the date selector issues in prod...still not sure why it works locally without adding anything.

I'm open to any other suggestions. If not, I'll create an alternative way of having a date selector.

### DIFF
--- a/apps/site/assets/babel.config.js
+++ b/apps/site/assets/babel.config.js
@@ -7,5 +7,6 @@ module.exports = {
         corejs: 2
       }
     ]
-  ]
+  ],
+  plugins: ["@babel/plugin-syntax-dynamic-import"]
 };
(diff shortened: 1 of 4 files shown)


arkadyan:
@alicia-mbta I haven't looked at this yet, but if you need to test this on the server you can probably feel free to deploy this branch to dev-green. Since other than your change it's still the feature branch we can continue testing with it.

alicia-mbta:
> @alicia-mbta I haven't looked at this yet, but if you need to test this on the server you can probably feel free to deploy this branch to dev-green. Since other than your change it's still the feature branch we can continue testing with it.

True, thanks! on it

PR #659, opened 2020-10-29
fix: mobile layout

fjlanasa:
Couple quick fixes:
* mobile layout
* zoom in map a bit

## Before
Screen Shot 2020-10-29 at 9 09 50 AM

Screen Shot 2020-10-29 at 9 08 36 AM

## After
Screen Shot 2020-10-29 at 9 07 10 AM

Screen Shot 2020-10-29 at 9 07 23 AM


thecristen (approved):
Appreciate it!

PR #660, opened 2020-10-29
Account for non-existing schedule route id for the Green Line

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Elixir.UndefinedFunctionError: function nil.route/0 is undefined](https://app.asana.com/0/385363666817452/1185865721068541)

Another bug involving trip info. This fix avoids getting a `500` error

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/finder_api.ex
+++ b/apps/site/lib/site_web/controllers/schedule/finder_api.ex
@@ -100,42 +100,53 @@ defmodule SiteWeb.ScheduleController.FinderApi do
       }) do
     {service_end_date, direction_id, _} = convert_from_string(date: date, direction: direction)
 
-    route =
+    schedule_route_id =
       route_id
       |> get_route_id(trip_id)
-      |> Routes.Repo.get()
 
-    opts = Map.get(conn.assigns, :trip_info_functions, [])
-    params = %{"origin" => origin, "trip" => trip_id}
+    if schedule_route_id do
+      route =
+        schedule_route_id
+        |> Routes.Repo.get()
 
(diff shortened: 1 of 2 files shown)


thecristen (changes requested):
Looks good on first glance! Let's not commit the formatting improvement to `VoteMap` though, since we'll be reverting that code soon.

alicia-mbta:
That was just the linter 😄 
Relatedly, I also noticed that the build fails because there is not enough code coverage due to the new "polling location finder" work (I believe).
I think this bug fix is a "nice to have" so it probably can wait to be merged after the revert.
Let me know what you think.

thecristen:
Gotcha. You're correct about the coverage failure reason. I agree with you on the wait - and I'll re-review then too :)

apps/site/lib/site_web/controllers/schedule/finder_api.ex
```diff
@@ -450,14 +461,21 @@ defmodule SiteWeb.ScheduleController.FinderApi do
...
 
-    schedule.route.id
+    if Enum.empty?(schedule_for_trip) do
+      nil
+    else
+      schedule =
+        schedule_for_trip
+        |> List.first()
+
+      schedule.route.id
```
thecristen, line 477: I think [CodeCov wants a new test to cover this piece](https://codecov.io/gh/mbta/dotcom/src/19b90a0c42932ac61a2fcb52d53156098e008cb6/apps/site/lib/site_web/controllers/schedule/finder_api.ex), maybe that'll make it pass?

thecristen (approved):

PR #661, opened 2020-10-29
Add a script to assist with creating GitHub releases

arkadyan:
#### Summary of changes
**Asana Ticket:** [Determine the info from git to be able to print out everything you need to make a release manually](https://app.asana.com/0/385363666817452/1198965985597277/f)
which is the first step for [🔧 Use a script to automate GitHub releases](https://app.asana.com/0/385363666817452/1185680283158788/f)

Add a script to assist with creating GitHub releases

Currently, it asks you to confirm some defaults and generates everything
you'll need to create a release manually in your browser. For a next
phase, it would be nice if it used the GitHub API to create the release
for you. I want to get that going next, but this already provides substantial value during the deploy process so it seems like it's worth getting in right away.

Screen Shot 2020-10-28 at 18 44 22

---

### DIFF
--- a/bin/.gitignore
+++ b/bin/.gitignore
@@ -1,2 +1,3 @@
 # Ignore wiremock jar files
 *.jar
+__pycache__/
--- a/bin/lib/git.py
+++ b/bin/lib/git.py
@@ -1,0 +1,23 @@
+import subprocess
+
+
+def current_branch():
+    return __call(["git", "branch", "--show-current"])
+
+def clean_status():
+    return __call(["git", "status", "--porcelain"]) == ""
+
+def current_commit():
+    return __call(["git", "rev-parse", "--short", "HEAD"])
+
+def latest_tag():
+    return __call(["git", "describe", "--tags", "--abbrev=0"])
+
+def sha_for_tag(tag):
+    return __call(["git", "show-ref", "-s", "--abbrev", tag])
+
+def log_between(previous, current):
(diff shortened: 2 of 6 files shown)


arkadyan:
@ianwestcott Feel free to chime in if you have any feedback on the utility or the python in general.

ianwestcott:
@arkadyan was it a deliberate choice to avoid using external libraries like [GitPython](https://gitpython.readthedocs.io/en/stable/)? (Apologies if we discussed this already and I forgot)

arkadyan:
@ianwestcott Nope! I didn't realize that existed (and I guess scripting this way for so long means I didn't think to look). Since this is working maybe I'll add what I have and then add another phase to replace it with that library.

arkadyan:
Ticket for GitPython: https://app.asana.com/0/385363666817452/1198968561086741/f

ianwestcott:
FWIW there's also a GitHub library you could use when the project progresses to that point: https://github.com/PyGithub/PyGithub

thecristen (commented):
Should we have some config or something to specify which version(s?) of Git these scripts expect?

thecristen (approved):
This is really nifty!

arkadyan:
@thecristen That's an interesting question about the git version. I think this is only using pretty basic features, so I wouldn't expect it to be very picky about the version. Let me know what you think, but my instinct would be to not worry about it until we come across a problem, and then [replace the implementation](https://app.asana.com/0/385363666817452/1198968561086741/f) soon anyway.

thecristen:
@arkadyan Sounds good to me!

PR #663, opened 2020-10-30
Date selector in IE

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Customer Support form - Date selector not working in IE](https://app.asana.com/0/385363666817452/1198980025678691)

Fix for the date selector so it works in IE.
This undoes the addition of the profanity sanitizer (in https://github.com/mbta/dotcom/pull/646).


apps/site/lib/site_web/controllers/trip_plan_controller.ex
```diff
@@ -20,14 +20,42 @@ defmodule SiteWeb.TripPlanController do
...
+      select: "plan-time-select",
+      label: "plan-time-label"
+    },
+    title: "trip-plan-departure-title"
+  }
+
   def index(conn, %{"plan" => %{"to" => _to, "from" => _fr} = plan}) do
     conn
     |> assign(:expanded, conn.query_params["expanded"])
+    |> assign(:plan_datetime_selector_fields, @plan_datetime_selector_fields)
```
thecristen, line 51: These `assign()` calls can probably be replaced with a plug...see the `optimize_for` function (which is a plug) in this module for an example. Then you could add `plug(:your_plug_function)` at top of this file and it'll run for each of these `index()` calls to assign the variable.

thecristen (approved):
Just one nudge to try using a Plug. Not blocking. Works great!

apps/site/lib/site_web/controllers/trip_plan_controller.ex
alicia-mbta: Should look better now!

thecristen: Looks great! 👍

PR #665, opened 2020-11-04
Define a RoutesApi behaviour

arkadyan:
#### Summary of changes
**Asana Ticket:** [Define a RoutesApi behaviour for V3Api.Routes](https://app.asana.com/0/1189396929591657/1189396929591660)

Define a RoutesApi behaviour

Mark V3Api.Routes as implementing this behaviour.

---

### DIFF
--- a/apps/v3_api/lib/routes.ex
+++ b/apps/v3_api/lib/routes.ex
@@ -2,36 +2,33 @@ defmodule V3Api.Routes do
   @moduledoc """
   Responsible for fetching Route data from the V3 API.
   """
+  @behaviour V3Api.RoutesApi
+
   import V3Api
 
-  alias Routes.Route
-  alias Stops.Stop
-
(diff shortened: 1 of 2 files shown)


alicia-mbta:
I think the `behaviour` approach is a good one. Any concerns, @cristen, @paulswartz?<br/>
If not, the work for this is already completed in this PR by @arkadyan 🙏

paulswartz:
I think my comment the last time was whether we wanted the behavior to be about the `V3API.Routes` interface, or the `Routes.Repo` interface?

thecristen:
This is a great question. I like the idea of doing this at the API level. But I'm trying to think about what it might look like if we extrapolate this pattern to all tests ... and wouldn't we need a similar behavior for the Routes.Repo interface anyway? 🤔

thecristen:
Closing since we have https://github.com/mbta/dotcom/pull/761

PR #666, opened 2020-11-04
Create mock routes api file

arkadyan:
#### Summary of changes
**Asana Ticket:** [Create a RoutesApiMock](https://app.asana.com/0/1189396929591657/1189396929591665)

Create mock routes api file

Stick to the bare implementation for the moment. We can flesh out more
based on what we need for specific tests.

Depends on https://github.com/mbta/dotcom/pull/665

---

Before getting review, please check the following:

* [ ] Does frontend functionality render and work correctly in IE?
* [ ] Have we load-tested any new pages or internal API endpoints that will receive significant traffic?
* [ ] Are interactive elements accessible to screen readers?
* [ ] Have you checked for tech debt you can address in the area you're working in?
* [ ] If this change involves routes, does it work correctly with pertinent "unusual" routes such as the combined Green Line, Silver Line, Foxboro commuter rail, and single-direction bus routes like the 170?
* [ ] Are the changes organized into self-contained commits with descriptive and well-formatted commit messages?

PR #667, opened 2020-11-04
Use MockRoutes when testing

arkadyan:
#### Summary of changes
**Asana Ticket:** [Configure test env to use RoutesApiMock anywhere we currently call V3Api.Routes](https://app.asana.com/0/1189396929591657/1189396929591667)

Use MockRoutes when testing

Define realistic test data to support all current tests.

WORK IN PROGRESS: Not all tests pass yet.

Depends on https://github.com/mbta/dotcom/pull/666

---

### DIFF
--- a/apps/v3_api/lib/mock_routes.ex
+++ b/apps/v3_api/lib/mock_routes.ex
@@ -5,13 +5,1652 @@ defmodule V3Api.MockRoutes do
 
   @behaviour V3Api.RoutesApi
 
+  alias JsonApi.Item
+
+  @red %Item{
+    attributes: %{
+      "color" => "DA291C",
+      "description" => "Rapid Transit",
+      "direction_destinations" => ["Ashmont/Braintree", "Alewife"],
+      "direction_names" => ["South", "North"],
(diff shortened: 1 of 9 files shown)


apps/v3_api/lib/mock_routes.ex
```diff
@@ -0,0 +1,1762 @@
+defmodule V3Api.MockRoutes do
+  @moduledoc """
+  A mock Routes API client for testing purposes.
+  """
+
+  @behaviour V3Api.RoutesApi
+
+  alias JsonApi.Item
+
+  @red %Item{
```
thecristen, line 10: Is there a better way of doing this than making huge modules with module attributes like this?

paulswartz: That's partly why I suggested doing it at the `Routes.Repo` level. There, you'd only need to provide the data that's a part of the `%Route{}` struct instead of a full JSON:API response.

thecristen:
Closing since we have https://github.com/mbta/dotcom/pull/761

PR #669, opened 2020-11-09
Show no upcoming departures when last stop

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Schedule Finder | Error Loading Trip Details](https://app.asana.com/0/555089885850811/1191999214380060)

Slight improvements related to Upcoming Departures:
* Show filtered-by-variant stops (if variant exists) in the origin modal
* Do not fetch trips if it's the last stop (and show a message instead)

### DIFF
--- a/apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
+++ b/apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
@@ -30,6 +30,7 @@ export interface Props {
   today: string;
   scheduleNote: ScheduleNoteType | null;
   initialSelectedRoutePatternId: string | null;
+  variantStops: string[];
 }
 
 export const fetchMapData = (
@@ -97,7 +98,8 @@ const ScheduleDirection = ({
   stops,
   today,
   scheduleNote,
-  initialSelectedRoutePatternId
+  initialSelectedRoutePatternId,
+  variantStops
 }: Props): ReactElement<HTMLElement> => {
   const routePatternsInCurrentDirection = routePatternsByDirection[directionId];
   const defaultRoutePattern =
@@ -223,6 +225,7 @@ const ScheduleDirection = ({
           stops={stops}
           today={today}
           scheduleNote={scheduleNote}
+          variantStops={variantStops}
         />
       )}
     </>
--- a/apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
+++ b/apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
@@ -28,6 +28,7 @@ interface Props {
   closeModal: () => void;
   modalMode: ModalMode;
   modalOpen: boolean;
+  variantStops: string[];
 }
 
 const ScheduleFinder = ({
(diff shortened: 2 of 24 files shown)


thecristen (commented):
I don't see any change for the situation cited in the original bug report - is that expected? - /schedules/15/line?schedule_direction%5Bdirection_id%5D=1&schedule_direction%5Borigin%5D=place-rugg&schedule_direction%5Bvariant%5D=15-1-1

thecristen:
BTW I deployed this branch on dev green since what was currently on there is already merged. :)

alicia-mbta:
I had forgotten to include a small addition. You should be able to see the error message now.

Changes in this branch:
- For the scenario in which we want to display upcoming departures but the origin is the last stop, there is a call to get the upcoming departures that results in error. With these changes, the call is prevented and this error message is shown instead. This is the only thing the end user would see.

- When there is a variant selected, the list of stops that the origin modal displays is all the stops of the line, regardless of the variant; so in the example above you're able to select Ruggles even though it's the last stop for variant 15-1-1 (and you shouldn't). With the changes in this branch, we're filtering the list of stops taking into account the variant and so the final list only contains the pertinent stops.


I guess the edge case is having this page bookmarked already, in which case Ruggles is grayed out (correct) but selected (wrong). This branch doesn't consider this edge case. But let me know if it should.

thecristen (commented):
Finding a case where the "last stop" isn't so straightforward. Looking at the [24](https://www.mbta.com/schedules/24/line?schedule_direction%5Bdirection_id%5D=1&schedule_direction%5Borigin%5D=place-matt&schedule_direction%5Bvariant%5D=2427-5-1)...

image

But later in the evening these buses keep going through Mattapan, to Ashmont.

image

(the headsign should probably read `INBOUND Mattapan or Ashmont`... but anyways...).

Sadly this change means we'd be hiding an upcoming departure (below is what we [currently show](https://www.mbta.com/schedules/24/line?schedule_direction%5Bdirection_id%5D=1&schedule_direction%5Borigin%5D=place-matt&schedule_direction%5Bvariant%5D=2427-5-1))

image

A similar thing happens on the 27, when going outbound on the "Wakefield Ave" variant from Mattapan https://www.mbta.com/schedules/27/line?schedule_direction[direction_id]=0&schedule_direction[origin]=place-matt&schedule_direction[variant]=2427-2-0

I suppose the problem might be that the schedules (and departures!) that are displayed in the modal might be for multiple variants (so it seems)? :(

alicia-mbta:
So...reading all your comments and looking at the changes in #599 (which I still need to fully understand), you have a point. So I don't think my approach is good, or good enough anyway.<br/>
I'm going to close this and try a different approach from scratch, and perhaps we can discuss further if that still doesn't look like a valid solution.

aararomi:
@alicia-mbta might be worth spinning up a new design ticket? Here's how to get it into my backlog: https://form.asana.com?k=pgrwg-Drarrtzzfda4WktA&d=15492006741476

thecristen:
@alicia-mbta That's fair. Do document your learnings so far in the ticket as part of the investigation!

PR #670, opened 2020-11-09
docs: add new doc for temporary features

thecristen:
I wanted to document how to bring the polling location map back and figured I can start a page to store notes on similar features, welcome any thoughts!


alicia-mbta (approved):
I like this format. The only small suggestion I have is to add a reference to this document in the `README.md`.

PR #672, opened 2020-11-12
Customer support form: subway should not include Silver Line

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Silver Line should not show under Subway mode](https://app.asana.com/0/555089885850811/1198980474686141)

Removing Silver Line from the list of options for Subway.


thecristen (approved):
👍

PR #676, opened 2020-11-16
Filter trips without tripInfo

alicia-mbta:
#### Summary of changes
**Asana Ticket:** Partial solution to [🐞 Schedule Finder | Error Loading Trip Details](https://app.asana.com/0/555089885850811/1191999214380060)

Filter trips with a `tripInfo` value of `null` in Upcoming Departures.


thecristen (approved):
It seems to work well!

PR #678, opened 2020-11-20
Adjust date and convert to UTC

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Times (not) coming through in UTC time](https://app.asana.com/0/555089885850811/1198956720242891)

Ensure the date and time of incident is properly converted to UTC time in the message sent to HEAT.

### DIFF
--- a/apps/feedback/lib/mailer.ex
+++ b/apps/feedback/lib/mailer.ex
@@ -109,8 +109,10 @@ defmodule Feedback.Mailer do
     end
   end
 
-  defp formatted_utc_timestamp(date) do
+  @spec formatted_utc_timestamp(map) :: String.t()
+  def formatted_utc_timestamp(date) do
     date
+    |> Timex.Timezone.convert("Etc/UTC")
     |> Timex.format!("{0M}/{0D}/{YYYY} {h24}:{m}")
   end
 
--- a/apps/feedback/test/mailer_test.exs
+++ b/apps/feedback/test/mailer_test.exs
@@ -251,5 +251,25 @@ defmodule Feedback.MailerTest do
       refute Test.latest_message()["text"]
              |> String.contains?("123abc")
     end
+
+    test "converts local (America/New_York) date to UTC" do
+      assert Mailer.formatted_utc_timestamp(%{
+               "year" => 2020,
+               "month" => 11,
+               "day" => 20,
+               "hour" => 22,
+               "minute" => 15,
+               "time_zone" => "America/New_York"
+             }) == "11/21/2020 03:15"
+
+      assert Mailer.formatted_utc_timestamp(%{
+               "year" => 2020,
+               "month" => 11,
+               "day" => 20,
+               "hour" => 10,
+               "minute" => 15,
+               "time_zone" => "America/New_York"
+             }) == "11/20/2020 15:15"
+    end
   end
 end
--- a/apps/site/lib/site_web/controllers/customer_support_controller.ex
+++ b/apps/site/lib/site_web/controllers/customer_support_controller.ex
@@ -288,28 +288,49 @@ defmodule SiteWeb.CustomerSupportController do
     end
   end
 
-  @spec validate_incident_date_time(map) :: DateTime.t()
-  defp validate_incident_date_time(incident_date_time) do
-    now = Util.now() |> Util.to_local_time()
+  @spec validate_incident_date_time(map) :: map
+  def validate_incident_date_time(incident_date_time) do
+    local_now = Util.now() |> Util.to_local_time()
 
-    parsed_date_time =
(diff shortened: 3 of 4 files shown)


apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -288,28 +288,49 @@ defmodule SiteWeb.CustomerSupportController do
...
+      |> Map.put("second", 0)
+      |> Map.put("microsecond", {0, 0})
+      |> Map.put("time_zone", "America/New_York")
+
+    case Util.parse(incident_date_time) do
+      {:error, :invalid_date} ->
+        destructured_local_now
+
+      parsed_dt ->
+        parsed_date_time = parsed_dt
```
paulswartz, line 317: You can match to a different name: `parsed_date_time ->`

apps/feedback/test/mailer_test.exs
```diff
@@ -251,5 +251,25 @@ defmodule Feedback.MailerTest do
...
+
+      assert Mailer.formatted_utc_timestamp(%{
+               "year" => 2020,
+               "month" => 11,
+               "day" => 20,
+               "hour" => 10,
+               "minute" => 15,
+               "time_zone" => "America/New_York"
+             }) == "11/20/2020 15:15"
+    end
```
paulswartz, line 273: Can we add some tests with dates/time around both DST conversions?

apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -288,28 +288,49 @@ defmodule SiteWeb.CustomerSupportController do
...
-          parsed_dt
-      end
-
-    local_parsed_date_time = Util.to_local_time(parsed_date_time)
-
-    # if date and time is in the future, set it to now
-    # otherwise leave as it is
+  @spec validate_incident_date_time(map) :: map
+  def validate_incident_date_time(incident_date_time) do
+    local_now = Util.now() |> Util.to_local_time()
```
paulswartz, line 293: It might make it easier to test if the `now` can also be an argument to the function: `def validate_incident_date_time(incident_date_time, now \\ Util.now()) do`

apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -288,28 +288,49 @@ defmodule SiteWeb.CustomerSupportController do
...
+
+    destructured_local_now = %{
+      "day" => local_now.day,
+      "hour" => local_now.hour,
+      "microsecond" => {0, 0},
+      "minute" => local_now.minute,
+      "month" => local_now.month,
+      "second" => 0,
+      "time_zone" => "America/New_York",
```
paulswartz, line 302: This doesn't look like an argument that `Util.parse` uses?

apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -288,28 +288,49 @@ defmodule SiteWeb.CustomerSupportController do
...
+
+        local_parsed_date_time = Util.to_local_time(parsed_date_time)
+
+        # if date and time is in the future, set it to now
+        # otherwise leave it as it is
+        if DateTime.compare(local_parsed_date_time, local_now) in [:lt, :eq] do
+          if local_incident_date_time["am_pm"] == "PM" and
+               local_incident_date_time["hour"] != "12" do
+            {hour_int, _} = Integer.parse(local_incident_date_time["hour"])
+            local_incident_date_time |> Map.replace!("hour", Integer.to_string(hour_int + 12))
```
paulswartz, line 327: This seems like it should the responsibility of `Util.parse`?

apps/feedback/lib/mailer.ex
```diff
@@ -109,8 +109,10 @@ defmodule Feedback.Mailer do
     end
   end
 
-  defp formatted_utc_timestamp(date) do
+  @spec formatted_utc_timestamp(map) :: String.t()
+  def formatted_utc_timestamp(date) do
```
paulswartz, line 113: This should be `DateTime.t()`, since the type of `Feedback.Message.incident_date_time` is also `DateTime.t()`.
You could then also use `DateTime.shift_zone!("Etc/UTC")` instead of going through the Timex overhead.

paulswartz (commented):
I'm a little confused about what `validate_incident_date_time` is doing. It looks like it's parsing the input (a little, to ensure that the date time is in the future), but then passing the map back. The type that `Feedback.Message` expects here is a `DateTime`, so I think that's the type it should be returning. This also lets you clean up the function a bit: you can get the current time (zone doesn't matter), parse the input time, and take the smaller of the two, returning an already parsed DateTime.

PR #680, opened 2020-11-25
Restore profanity sanitizer to an IE-compatible version

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Restore profanity sanitizer in Customer Support form](https://app.asana.com/0/555089885850811/1199207412558094)

It's essentially the same changes from PR #646 but with the downgraded version so the new customer support form works in IE without breaking.


thecristen (approved):
👍 👍

PR #681, opened 2020-11-25
fix: always assign `:all_options_per_mode`

paulswartz:
Previously, we only assigned this data in the `index/2` function. However,
this is also required if the submission fails, when we re-render the form. We
can see this by removing all of extra `assign/3` calls in the unit tests.

To fix, we pull this behavior into a plug, so that we will always assign this
data regardless of function is originally called.

Asana: https://app.asana.com/0/555089885850811/1198960312594553
 
---


thecristen (approved):
👍 👍

PR #682, opened 2020-11-25
Adjust the javascript code so it works in IE

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [New Customer Support form not submitting in IE](https://app.asana.com/0/555089885850811/1199361102542406)

Some properties and methods are not supported in IE so some workarounds are needed.

### DIFF
--- a/apps/site/assets/js/support-form.js
+++ b/apps/site/assets/js/support-form.js
@@ -1,5 +1,10 @@
 /* eslint-disable */
 
+// Some FormData methods like .set or .get are not supported in IE:
+// (https://developer.mozilla.org/en-US/docs/Web/API/FormData#Browser_compatibility)
+// so we need to make use of a polyfill:
+import FormData from "formdata-polyfill";
+
 export default function($ = window.jQuery) {
   document.addEventListener(
     "turbolinks:load",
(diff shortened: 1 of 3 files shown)


alicia-mbta:
I'm not that familiar with Can I Use, I can take a look at possible options.

paulswartz:
https://caniuse.com/?search=textLength

alicia-mbta:
Oh I was sort of familiar with caniuse's lookup table but I thought there was some way to "scan" the file to look for unsupported features. But it doesn't seem like caniuse has plans to do that for now.
We do have the script `npm run eslint` (but it doesn't look like it caught any issues with `support-form.js` in the past) and I run into [this checker](https://seedmanc.github.io/jscc/) too, do you think any of these might be good options moving forward?

paulswartz:
What about something like https://github.com/amilajack/eslint-plugin-compat?

thecristen:
The `npm run eslint` script is set up a bit unusally - from the root directory it runs `git diff --name-only --diff-filter=dx origin/master... | grep apps/site/assets/js/.*\.js | xargs apps/site/assets/node_modules/.bin/eslint -c apps/site/assets/.eslintrc.js` which I think will only run on files with changes from master.

If you run it from `/apps/site/assets` it runs `cd js && npx eslint --ext .js .` catches some 1202 problems (1102 errors, 100 warnings) 🤡 Our JavaScript is old.

Anyway!! That aside...

The eslint-plugin is a good idea 👍 though since JavaScript won't throw a fatal error when a property like this fails... I don't think any tool to catch these things would be 100% foolproof. Let's consider it in our upcoming tech debt epic!

paulswartz (approved):
🍰

PR #683, opened 2020-11-27
correct CR trip messages for Fall 2020

paulswartz:
#### Summary of changes
**Asana Ticket:** [Adjust "via Fairmount" type trips](https://app.asana.com/0/555089885850811/1198686557681794)

This updates the `trip_messages/2` function to have the correct messages for
the Fall Rating. I checked it against the relevant PDF schedules. The one
liberty I took was converting the "VIA HAV" in the Lowell schedule to a full
"Via Have-rhill" text.

---


apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
```diff
@@ -75,61 +75,47 @@ defmodule SiteWeb.ScheduleController.TimetableController do
...
-      {"221", "place-WR-0075"} => "Lowell",
-      {"221", "place-WR-0085"} => "Line"
-    }
-  end
+  @doc """
+  Additional text to be included in the timetable.
 
+  We use this for Commuter Rail trips which travel via atypical routes, in
+  order to match the PDF schedules. Each rating, this should be checked
+  against the new PDFs to ensure it's kept up to date.
```
alicia-mbta, line 83: Should we create an Asana ticket or some sort of reminder every time the rating changes so we don't forget about this? cc @thecristen

apps/site/lib/site_web/views/schedule_view.ex
```diff
@@ -465,34 +465,12 @@ defmodule SiteWeb.ScheduleView do
...
         content_tag(:p, [
           content_tag(:strong, "Note:"),
           " Trains depart from Foxboro 30 minutes after conclusion of events"
-        ]),
-        content_tag(:p, [
-          content_tag(:strong, "Note:"),
-          " For regular weekday service to and from Foxboro please visit ",
-          link(to: timetable_path(SiteWeb.Endpoint, :show, "CR-Franklin")) do
-            "Franklin Line"
-          end
```
alicia-mbta, line 495: Are we removing this piece because the Foxboro service is suspended and doesn't even go via Franklin? (sorry if this is a very obvious question and I'm just not seeing it)

paulswartz: Yes, there's no weekday service on the Franklin Line in the Fall schedule. Compare the [spring PDF](https://cdn.mbta.com/sites/default/files/route_pdfs/2020-summer/2020-06-22-cr-franklin-accessible.pdf) to the current [Fall PDF](https://cdn.mbta.com/sites/default/files/route_pdfs/2020-fall/2020-11-02-cr-franklin-accessible-v2.pdf). You can see that Foxboro has dropped out of the list of stops to the left.

alicia-mbta: Got it, thanks for clarifying!

alicia-mbta (approved):
Thank you for simplifying this logic  🍰

PR #684, opened 2020-11-27
Customer Support Form improvements

thecristen:
This PR serves as a final check, there should be no surprises here. ;)

#### Summary of changes
**Asana Ticket:** [New CS Form Feature Complete + Live](https://app.asana.com/0/1139110203450767/1194612062768617/f)

All the changes to the customer support form! Including but not limited to:

* Reworked layout
* Backend support for new fields
* New fields added on front end, including a few interdependent fields 
* Reversed logic for requiring contact info on the form

---


alicia-mbta:
I think that the test that's breaking (added in #660) is because the parameters that I specified might be only valid for the day of. With the shuttles and other changes that the Green Line is going through, I wonder if it makes more sense to remove that test and try another approach?

thecristen:
I can confirm it works* in IE! 😄 

*Well, the photo _preview_ does not work in IE. But it turns out it didn't work in IE to begin with. 🥴 
Anyway, I'm not sure if/how this affects the uploaded photo being submitted.

alicia-mbta:
> I can confirm it works* in IE! 😄
> 
> *Well, the photo _preview_ does not work in IE. But it turns out it didn't work in IE to begin with. 🥴
> Anyway, I'm not sure if/how this affects the uploaded photo being submitted.

Would it be worth to submit a couple more tickets using IE (to see what's received by HEAT) and create a new ticket to investigate the photo preview in IE? not sure how important this is since you mentioned it doesn't even work in the current customer support form.

thecristen:
@alicia-mbta  definitely let's create a ticket for this existing IE bug.

aararomi:
lgtm

alicia-mbta:
> definitely let's create a ticket for this existing IE bug.

For reference, here is the [Asana ticket](https://app.asana.com/0/385363666817452/1199514633650926) for this.

paulswartz (approved):
:cake:

PR #685, opened 2020-11-29
Adjust _alert_announcement.html.eex to handle nil alert

thecristen:
Fix for the `Elixir.BadMapError: expected a map, got: nil` bug with the following trace

```shell
Elixir.BadMapError: expected a map, got: nil
  File "lib/site_web/templates/layout/_alert_announcement.html.eex", line 2, in SiteWeb.LayoutView."_alert_announcement.html"/1
  File "lib/site_web/templates/layout/app.html.eex", line 53, in SiteWeb.LayoutView."app.html"/1
  File "lib/phoenix/view.ex", line 399, in Phoenix.View.render_to_iodata/3
  File "lib/phoenix/controller.ex", line 729, in Phoenix.Controller.__put_render__/5
  File "lib/phoenix/controller.ex", line 746, in Phoenix.Controller.instrument_render_and_send/4
...
(3 additional frame(s) were not displayed)

(BadMapError) expected a map, got: nil
```

, which appears in a variety of tickets.

Since `Alerts.Repo.by_id/1` can potentially return `nil`, we adjust the`_alert_announcement.html.eex` template to account for this.

**Asana Tickets:** 
https://app.asana.com/0/0/1181081199001391/f 
https://app.asana.com/0/0/1185069682012440/f 
https://app.asana.com/0/0/1188955543885440/f 
https://app.asana.com/0/0/1196810601137545/f 
https://app.asana.com/0/0/1198058255091329/f 
https://app.asana.com/0/0/1198680950728266/f 
https://app.asana.com/0/0/1198902731845612/f 
https://app.asana.com/0/0/1185867812729748/f

### DIFF
--- a/apps/site/lib/site_web/banneralert.ex
+++ b/apps/site/lib/site_web/banneralert.ex
@@ -39,7 +39,7 @@ defimpl BannerAlert, for: Alerts.Banner do
 
   defp alert_from_banner(banner) do
     Alerts.Alert
-    |> struct(banner)
+    |> struct(Map.from_struct(banner))
     |> Map.put(:priority, :system)
   end
 end
--- a/apps/site/lib/site_web/banneralert.ex
+++ b/apps/site/lib/site_web/banneralert.ex
@@ -38,7 +38,8 @@ defimpl BannerAlert, for: Alerts.Banner do
   end
 
   defp alert_from_banner(banner) do
-    Kernel.struct(Alerts.Alert, banner)
-    |> Map.update!(:priority, :system)
+    Alerts.Alert
+    |> struct(banner)
+    |> Map.put(:priority, :system)
   end
 end
--- a/apps/site/lib/site_web/banneralert.ex
+++ b/apps/site/lib/site_web/banneralert.ex
@@ -3,7 +3,7 @@ defprotocol BannerAlert do
   def human_effect(obj)
   def human_label(obj)
   def icon(obj)
-  def alert_for_label_class(obj)
+  def label_class(obj)
 end
 
 defimpl BannerAlert, for: Alerts.Alert do
@@ -11,7 +11,7 @@ defimpl BannerAlert, for: Alerts.Alert do
   defdelegate human_effect(alert), to: Alerts.Alert
   defdelegate human_label(alert), to: Alerts.Alert
   defdelegate icon(alert), to: Alerts.Alert
-  def alert_for_label_class(alert), do: alert
+  def label_class(alert), do: SiteWeb.AlertView.alert_label_class(alert)
 end
 
 defimpl BannerAlert, for: Alerts.Banner do
@@ -32,8 +32,9 @@ defimpl BannerAlert, for: Alerts.Banner do
     |> Alerts.Alert.icon()
   end
 
-  def alert_for_label_class(banner) do
+  def label_class(banner) do
     alert_from_banner(banner)
+    |> SiteWeb.AlertView.alert_label_class()
   end
 
   defp alert_from_banner(banner) do
--- a/apps/site/lib/site_web/banneralert.ex
+++ b/apps/site/lib/site_web/banneralert.ex
@@ -1,0 +1,43 @@
+defprotocol BannerAlert do
+  def header(obj)
+  def human_effect(obj)
+  def human_label(obj)
+  def icon(obj)
+  def alert_for_label_class(obj)
+end
+
+defimpl BannerAlert, for: Alerts.Alert do
+  def header(alert), do: alert.header
+  defdelegate human_effect(alert), to: Alerts.Alert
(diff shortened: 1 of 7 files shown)


apps/site/lib/site_web/templates/layout/_alert_announcement.html.eex
```diff
@@ -1,3 +1,6 @@
+<% alert = Alerts.Repo.by_id(@alert_banner.id) %>
+<%= if alert do %>
 <div class="alert-announcement__header" onclick="location.href='<%= @alert_banner.url || alert_path(@conn, :index) %>';">
-  <%= SiteWeb.AlertView.render "_banner.html", alert: %{Alerts.Repo.by_id(@alert_banner.id) | header: @alert_banner.title, priority: :system} %>
+  <%= SiteWeb.AlertView.render "_banner.html", alert: %{alert | header: @alert_banner.title, priority: :system} %>
```
paulswartz, line 4: Isn't `@alert_banner` already an alert? Why the extra fetch?

thecristen: `Alerts.Alert.t()` and `Alerts.Banner.t()` are _slightly_ different for reasons I don't completely understand yet, but a few notes in [`Alerts.Cache.Store`](https://github.com/mbta/dotcom/commit/eaac9fdb9b32f0c55f64689ec23de01de283a513) are here.. `alert_banner` doesn't seem used in many places beyond this though, so maybe I should try refactoring it out?

paulswartz: The main difference is that what we want to show for the `header` in a banner alert is separate from the actual `header` that's returned for the alert, and that we're overriding the priority. We should be able to address that by either 1) (easer) giving `Banner.t()` a `header` instead of a `title` 2) creating a Protocol that both `Alert.t` and `Banner.t` can implement which provides the relevant data for both types.

thecristen: Looking at it more closely, the `_banner.html` template relies on certain fields that are **not** present in `Banner.t()` - `priority` to render the icon, and `lifecycle` to generate a CSS class name. I guess this is why the extra fetch. I'm intrigued by the Protocol idea but am not sure what it's solving there?

paulswartz: It would allow you to pass a `Banner.t()` to that template, by providing an interface to get the appropriate priority/lifecycle. Adding those fields to the Banner struct works out about the same.

thecristen:
@paulswartz Wondering if this update is along the lines of what you were suggesting.

paulswartz (commented):
It's the right direction yes! But what I was thinking was that instead of having the protocol return an Alert, `_banner.html` should be using the protocol to get the relevant data from either an `Alert` or a `Banner`.  From a quick glance, it looks like you'd need functions for

- human label
- human effect
- icon (or the content that goes into the icon)
- the data that `alert_label_class` needs
- header
- URL

Also, the protocol should live in `site_web` somewhere, since the `alert` app doesn't really care about the logic the frontend needs to display an Alert or a Banner.

thecristen:
Got it. I think something like this should be close...

apps/site/lib/site_web/templates/alert/_banner.html.eex
```diff
@@ -1,18 +1,18 @@
...
+    <%= @alert |> BannerAlert.icon() |> alert_icon() %>
   </div>
   <div class="c-alert-item__top">
     <div class="c-alert-item__top-text-container">
       <div class="c-alert-item__effect">
-        <%= Alerts.Alert.human_effect(@alert) %>
+        <%= BannerAlert.human_effect(@alert) %>
         <%= unless label == "" do
-          content_tag(:span, [label], class: alert_label_class(@alert))
+          content_tag(:span, [label], class: @alert |> BannerAlert.alert_for_label_class() |> alert_label_class)
```
paulswartz, line 11: Can you make `label_class` a part of the `BannerAlert` interface? `BannerAlert.label_class(@alert)`

apps/site/lib/site_web/banneralert.ex
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def alert_for_label_class(banner) do
+    alert_from_banner(banner)
+  end
+
+  defp alert_from_banner(banner) do
+    Kernel.struct(Alerts.Alert, banner)
+    |> Map.update!(:priority, :system)
+  end
```
paulswartz, line 42: Kernel functions are always available:
```elixir
Alerts.Alert
|> struct(banner)
|> Map.put(:priority, :system)
```

paulswartz (approved):
🍰 pending CI; nice work!

PR #686, opened 2020-11-30
Banner in focused state

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Projects | SCSS: Focus state causes blue underlines below white text](https://app.asana.com/0/555089885850811/1156601525508738)

Set a color for the `Banner` text when focused depending on the mode.

### DIFF
--- a/apps/site/assets/css/_banner.scss
+++ b/apps/site/assets/css/_banner.scss
@@ -3,7 +3,8 @@
   display: block;
   margin-top: $base-spacing * 3;
 
-  &:hover {
+  &:hover,
+  &:focus {
     color: inherit;
     text-decoration: none;
 


apps/site/assets/css/_colors.scss
```diff
@@ -83,7 +83,6 @@ $hp-hero-gradient-end: rgba(28, 30, 35, .8);
 $hp-hero-gradient: linear-gradient(to bottom, $hp-hero-gradient-start, $hp-hero-gradient-end);
 
 // General
-$link-hover-color: $brand-primary-darkest;
```
thecristen, line 86: Not sure we want to remove this - it's picked up by Bootstrap's SCSS in `apps/site/assets/node_modules/bootstrap/scss/_reboot.scss` line 163. ;)

apps/site/assets/css/_banner.scss
```diff
@@ -171,3 +171,22 @@
 .c-banner__date {
   font-weight: bold;
 }
+
+.c-focus--bus {
+  &:focus {
+    color: inherit;
```
thecristen, line 177: Not 100% sure about this but it might make sense to explicitly use the `$black` color here.

thecristen (changes requested):
The way you wrote the `.c-focus--` classes strikes me as more of a utility class that could be useful across the site, and thus I think it should (1) renamed to start with u like the rest of our utility classes, e.g. `.u-focus--bus` and (2) moved from this file to `_utilities.scss`.

Otherwise looks great!! 👍

apps/site/assets/css/_colors.scss
alicia-mbta: Ouch, I didn't see that -- I will restore it.

apps/site/assets/css/_banner.scss
alicia-mbta: I think for buses it's a specific shade of dark blue, that is why Adamo suggested in the ticket to use `inherit`.
@amaisano can you please confirm? otherwise `$black` totally makes sense to me :)

amaisano: I don't recall if I meant "inherit" in the literal property sense. I do remember thinking the color combos (white on red, white on blue, etc) were already present in the CSS hierarchy, so that all we really needed to do was remove hard-coded text color properties, and/or use the `:hover, :focus { color: inherit; }` rule so that the pseudo state text color matched the static state text color, without needing to hard code every mode (which I do not like).

apps/site/assets/css/_utilities.scss
```diff
@@ -79,6 +79,25 @@
...
+}
+
+.u-focus--green-line,
+.u-focus--red-line,
+.u-focus--silver-line,
+.u-focus--blue-line,
+.u-focus--subway,
+.u-focus--mattapan,
+.u-focus--commuter-rail,
+.u-focus--unknown {
```
amaisano, line 95: This seems a little unnecessary. Would making _any_ text in here white by default (no need to call out each mode type, but fine to keep semantic class), and then just making the `--bus` the single override outlier, be more efficient?

apps/site/assets/css/_banner.scss
amaisano: PS: I'm sorry this ticket was so vague. I'm sure I knew exactly how to fix it when I wrote it 😬

amaisano:
Ah ha! @alicia-mbta @thecristen  I think all we need to do is add the `:focus` selector to this helper:

https://github.com/mbta/dotcom/blob/2e98069435d2c17da60a4ded0e7b80b789a03bbc/apps/site/assets/css/_utilities.scss#L1-L15

Like I thought, the `.u-bg--` family of rules already code the text/bg colors. There is no need to add new rules, unless there is some kind of specificity conflict.

thecristen:
> Ah ha! @alicia-mbta @thecristen I think all we need to do is add the `:focus` selector to this helper:
> 
> https://github.com/mbta/dotcom/blob/2e98069435d2c17da60a4ded0e7b80b789a03bbc/apps/site/assets/css/_utilities.scss#L1-L15
> 
> Like I thought, the `.u-bg--` family of rules already code the text/bg colors. There is no need to add new rules, unless there is some kind of specificity conflict.

Yeah! As long as these colors(focus/bg/text/hover) **all** work as-is for the banner here, then I think this approach would be 💯

alicia-mbta:
> Ah ha! @alicia-mbta @thecristen I think all we need to do is add the `:focus` selector to this helper:
> 
> https://github.com/mbta/dotcom/blob/2e98069435d2c17da60a4ded0e7b80b789a03bbc/apps/site/assets/css/_utilities.scss#L1-L15
> 
> Like I thought, the `.u-bg--` family of rules already code the text/bg colors. There is no need to add new rules, unless there is some kind of specificity conflict.

Actually I found that adding the `:focus` pseudoclass to `c-banner` causes the same effect so the solution might be even simpler (unless I'm missing something?)

apps/site/assets/css/_banner.scss
```diff
@@ -3,7 +3,7 @@
   display: block;
   margin-top: $base-spacing * 3;
 
-  &:hover {
+  &:hover, &:focus {
```
amaisano, line 6: I believe each selector should be on its own line, but that might depend on the linter we are using for SCSS.

alicia-mbta:
@amaisano feel free to re-review.

amaisano (approved):
I'm not able to check out the code myself and test it, but I think this is all we need. So, pending it working in the browser(s), LGTM!

alicia-mbta:
It turns out that adding just the `:focus` pseudoclass was all we needed to do to achieve the same result, so all the previous work was discarded 🙂

thecristen:
OMG. Amazing

thecristen (approved):

PR #690, opened 2020-12-03
Fix for Green Line test

alicia-mbta:
No ticket, test was breaking.


paulswartz (commented):
Can you also update the comment on line 376?

paulswartz (approved):
🍰

PR #691, opened 2020-12-04
Add date and time fields if not present in the form

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [☎️ Elixir.KeyError: key "date_time" not found in: %{...}](https://app.asana.com/0/555089885850811/1198712992208784)

Default to "now" if said fields are not submitted.


thecristen (approved):
Thanks for the additional test! I think this would work great. 👍 

For the sake of merging, can you update this branch so it's only this last commit, branched off of master?

PR #692, opened 2020-12-08
Make subject field required

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [☎️ Make subject field required](https://app.asana.com/0/555089885850811/1188547161220081)

Do not allow submission if field is not filled:<br/>
image

### DIFF
--- a/apps/site/assets/js/support-form.js
+++ b/apps/site/assets/js/support-form.js
@@ -314,6 +314,9 @@ const validators = {
   comments: function($) {
     return $("#comments").val().length !== 0;
   },
+  subject: function($) {
+    return $("#support_subject").val().length !== 0;
+  },
   service: function($) {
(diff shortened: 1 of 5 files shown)


thecristen:
I'm not sure this needs design review as it doesn't incorporate the design proposal (https://app.asana.com/0/385363666817452/1199530274677101/f) yet... but otherwise it does the job 👍

thecristen (approved):

paulswartz (commented):
should there also be server-side validation that subject is present?

thecristen (commented):
I think the subject might also need to be included in the setupValidation() function in support-form.js, otherwise looks like the rest of the pieces are all there.

thecristen (approved):
👍

PR #693, opened 2020-12-09
Multistage docker

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🔧 Multi-stage Docker build](https://app.asana.com/0/555089885850811/1198998494356289)


alicia-mbta:
Work in progress. Essentially we consolidated the two Dockerfiles into one but this error is showing (which looks like it's related to a wrong path):
`Error: No such container:path: 453fa31b38b593626141e118287f75e4460ded233d4042907dba5f68bb929006:/_build/prod/rel/site/releases/0.0.1/site.tar.gz`

paulswartz:
That error makes sense, since we'll no longer be copying that `.tgz` out of the container to upload to ECS. What we should be able to do afterwards is instead something like:
```
docker build -t dotcom .
docker run -it --env-file .envrc -p 4001:4001 dotcom
```
That will run the newly built dotcom container (using your local environment). Does something like that work here? 

Once that works, we can work on pushing the build to Elastic Container Registry, and configuring Elastic Beanstalk to reference the build that's there (example: https://github.com/mbta/delta/tree/main/semaphore in particular the `build_push.sh` and `deploy_beanstalk.sh` scripts).

alicia-mbta:
Do you mean adding those two lines instead of the two `docker cp` commands in file `build.sh`? I'm confused because aren't lines 14 and 15 doing a similar thing?

paulswartz:
~Line 15 in the existing `build.sh` is basically the same as the `COPY` command in your new Dockerfile.~

I didn't realize originally that there's no `COPY` command in the Dockerfile, which makes my reply confusing. I'll try to describe how the build works now, and then we can see what the future looks like.

## Current state

The `Dockerfile` in the root of the repo is used to run an Elixir release: we build an image tagged `dotcom:_build` Then, `build.sh` runs a container using that tag, and copies those files out of `/root/rel/_build/prod`, and creates a `.zip` file containing the release, with a new `Dockerfile` (from `rel/Dockerfile`). We then upload this `.zip` to ElasticBeanstalk, which runs that `.zip` as a Docker container.

If you wanted to run this container on your local machine, you can do something like:

```
sh build.sh
mkdir tmp
cd tmp
unzip ../site-build.zip
docker build -t dotcom .
docker run -it -p 4000:4000 --env-file .envrc dotcom
```

You unzip the `site-build.zip` to get at the `Dockerfile` that ElasticBeanstalk would be running.
 
## Future state

The alternative is to build a Docker image ourselves, upload it to the Elastic Container Registry (ECR), and tell ElasticBeanstalk to run that already-built image. The way that works is that instead of uploading a `.zip` with all the release files, we upload a `.zip` containing a single `Dockerrun.aws.json` file, pointing at the Docker image we've already uploaded to ECR (example from github.com/mbta/delta):
```json
{
  "AWSEBDockerrunVersion": "1",
  "Image": {
    "Name": "ECR_ROOT/delta:git-876dec2",
    "Update": "true"
  },
  "Ports": [
    {
      "ContainerPort": "4000"
    }
  ]
}
```

So, how do we create the image that we upload to ECR? That's where the two-stage Docker image comes in. In the first stage, we do the release build as normal. In the second stage, we `COPY` the release files from the first stage, like the `build.sh` script is doing in the current state. From there, the rest of the `Dockerfile` is the same as the original `rel/Dockerfile`, which is responsible for running our server. This build is tagged with the Git SHA (so for us, would be something like `dotcom:git-XXXXXXX`), which we then push to ECR for ElasticBeanstalk to use.

paulswartz:
I got an error running `docker build . -t dotcom`:
```
Reading package lists...
W: GPG error: http://security.debian.org/debian-security stretch/updates InRelease: At least one invalid signature was encountered.
W: The repository 'http://security.debian.org/debian-security stretch/updates InRelease' is not signed.
W: GPG error: http://deb.debian.org/debian stretch-updates InRelease: At least one invalid signature was encountered.
W: The repository 'http://deb.debian.org/debian stretch-updates InRelease' is not signed.
W: GPG error: http://deb.debian.org/debian stretch Release: At least one invalid signature was encountered.
W: The repository 'http://deb.debian.org/debian stretch Release' is not signed.

## Installing packages required for setup: apt-transport-https lsb-release...

+ apt-get install -y apt-transport-https lsb-release > /dev/null 2>&1
Error executing command, exiting
The command '/bin/sh -c curl -sL https://deb.nodesource.com/setup_8.x | bash - &&   apt-get install -y nodejs' returned a non-zero code: 1
```

I think this is due to something about the Debian image that `elixir:1.8.1` is using. Can you throw in an `apt-get update -y` before running the NodeJS install scripts?

Dockerfile
```diff
@@ -37,3 +47,28 @@ RUN mix phx.digest
 
 WORKDIR /root
 RUN mix release --verbose
+
+# Copy the release files from the first stage in order to proceed to the second stage
+ADD . rel/site
```
alicia-mbta, line 52: Is this instruction correct, @paulswartz ?

alicia-mbta:
> I think this is due to something about the Debian image that `elixir:1.8.1` is using. Can you throw in an `apt-get update -y` before running the NodeJS install scripts?

That seems to indeed solve the issue you were having.

paulswartz:
I'm actually still having trouble, even on `master`. I wonder if something changed with the Node 8.x repository we're using?

Maybe we can land #695 first, and rebase this on top?

Dockerfile
paulswartz: This was already added on line 24/34, so I don't think you need it here.

Dockerfile
```diff
@@ -37,3 +47,28 @@ RUN mix phx.digest
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+	libssl1.1 libsctp1 curl \
+	&& rm -rf /var/lib/apt/lists/*
+
+# Set exposed ports
+EXPOSE 4000
+ENV PORT=4000 MIX_ENV=prod TERM=xterm LANG="C.UTF-8" REPLACE_OS_VARS=true
+
+ADD rel rel/site
```
paulswartz, line 69: Instead of `ADD` you'll need `COPY --from=builder`

alicia-mbta:
After #695 was merged I rebased the changes from this branch. Apologies if I'm still missing something.

Dockerfile
```diff
@@ -41,3 +51,26 @@ RUN mix phx.digest
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+	libssl1.1 libsctp1 curl \
+	&& rm -rf /var/lib/apt/lists/*
+
+# Set exposed ports
+EXPOSE 4000
+ENV PORT=4000 MIX_ENV=prod TERM=xterm LANG="C.UTF-8" REPLACE_OS_VARS=true
+
+COPY --from=builder . rel/site
```
paulswartz, line 71: The `.` here should be a path from the container. Base on the `build.sh` script, maybe `/root/_build/prod/rel`? The goal is to have `/root/rel/site/bin/startup` as a file that `CMD` can execute once the container is built.

The way I generally try to debug things like this is adding lines like `RUN ls -l <path>` to the Dockerfile, so I can see what files are where during the build process.

alicia-mbta: I thought of running `RUN ls -l <path>` but for some reason it wasn't showing anything for the different paths I was giving it.
But the `/root/_build/prod/rel` path makes sense -- sorry I didn't catch that earlier.

paulswartz:
So I think I tracked this down to two overlapping issues:
1) We need to manually copy the `rel/bin/startup` script
2) We need that script to be available for copying, not ignored by `.dockerignore`

Unrelatedly, the path for the output is `/root/rel`, not `/root/rel/site`.

```diff
diff --git a/.dockerignore b/.dockerignore
index f3df566ce..78522862d 100644
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,9 +5,6 @@ _build
 deps
 **/node_modules
 **/test
-rel
-!rel/vm.args
-!rel/config.exs
 erl_crash.dump
 build.sh
 deploy.sh
@@ -15,3 +12,5 @@ apps/site/priv
 aws
 site-build.zip
 bin
+Dockerfile
+.dockerignore
diff --git a/Dockerfile b/Dockerfile
index 55b5ca343..aad74b6e6 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 EXPOSE 4000
 ENV PORT=4000 MIX_ENV=prod TERM=xterm LANG="C.UTF-8" REPLACE_OS_VARS=true

-COPY --from=builder /root/_build/prod/rel rel/site
+COPY --from=builder /root/_build/prod/rel /root/rel
+ADD rel/bin/startup /root/rel/site/bin/startup
+
 RUN mkdir /root/work

 WORKDIR /root/work
```

For me, this built an image I could run:
```shell
$ docker build . -t dotcom
$ docker run -it --env-file .env dotcom
```

paulswartz:
Looks good! Next step is to work on pushing the built image up to ECR: let's talk on Monday (or whenever you're back in the office)

semaphore/build_push.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# bash script should be called with AWS environment (dev / dev-green / dev-blue / prod)
+# Other required configuration:
+# * APP
+# * DOCKER_REPO
+awsenv=$1
+
+# build docker image and tag it with git hash and AWS environment
+githash=$(git rev-parse --short HEAD)
+docker build -t $APP:latest .
```
paulswartz, line 12: This also needs to include the Docker login changes: for an example, see https://github.com/mbta/tablespoon/commit/3ab2537550107eb64d563048875d2d8d960c85e4

paulswartz (commented):
Looks good! https://semaphoreci.com/mbta/dotcom/servers/docker-test/deploys/3

The next step is the deploy! https://github.com/mbta/delta/tree/main/semaphore has examples: I believe the relevant files are `Dockerrun.aws.json` and `deploy_beanstalk.sh`

alicia-mbta:
This is an initial approach to the 'conversion' because I know that some instructions are incorrect and I'm getting errors that might be path-related (looking into those now) but I wanted to push what I've been doing so far.

Dockerfile
```diff
@@ -1,33 +1,36 @@
...
-RUN git config --global url.https://github.com/.insteadOf git://github.com/
+# 2.) Build the frontend assets within a node.js container instead of installing node/npm
 
-# Install Hex+Rebar
-RUN mix local.hex --force && \
-  mix local.rebar --force
+# The Elixir image was built with debian stretch, use the node debian-stretch base
+FROM node:14.15.1-stretch as assets-builder
+
+ARG SENTRY_DSN=""
```
ianwestcott, line 27: This might be out of scope for this change, but why do we need to give this an empty value at build time? I see it's referenced on [line 37](https://github.com/mbta/dotcom/pull/693/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R37)... is this a workaround to some build-time configuration requirement for the Sentry JS library?

Dockerfile
```diff
@@ -36,8 +39,43 @@ RUN npm run webpack:build -- --env.SENTRY_DSN=$SENTRY_DSN
...
+# erlang-crypto requires system library libssl1.1
+RUN apt-get update && apt-get install -y --no-install-recommends \
+	libssl1.1 libsctp1 curl \
+	&& rm -rf /var/lib/apt/lists/*
 
 WORKDIR /root
-RUN mix distillery.release --verbose
+ADD . .
+
+COPY --from=app-builder /root/rel /root/_build/prod/rel
```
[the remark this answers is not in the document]
thecristen: oh oh I know this one!
dev is the `MIX_ENV` value. If run with `MIX_ENV=prod` then the release gets put in `_build/prod/rel` :)

Dockerfile
```diff
@@ -36,8 +39,43 @@ RUN npm run webpack:build -- --env.SENTRY_DSN=$SENTRY_DSN
...
+
+WORKDIR /root
+
+# add frontend assets compiled in node container, required by phx.digest
+COPY --from=assets-builder /root/apps/site/priv/static ./apps/site/priv/static
+
 WORKDIR /root/apps/site/
-RUN mix phx.digest
+RUN mix do deps.get, deps.compile
+RUN mix do deps.get, phx.digest
```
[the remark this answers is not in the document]
alicia-mbta: Yes, I don't know why it's not finding the dependencies, but adding `deps.get` feels _hacky_ so I'm removing that.
I think you might be right and the `--only prod` flag might be causing issues and hence should not be present (it's not in the original `Dockerfile` to begin with).

Dockerfile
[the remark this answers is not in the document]
alicia-mbta: You're right! replacing `prod` with `dev` to _debug locally_ seems to successfully build. BUT I'm getting this error again:
```
Successfully built 1fa8d4439622
Successfully tagged site:_build
Error: No such container:path: 4f0f904e661b0a9664414a553f951e3400fb076a474f412a49e2a73cf2a2cc30:/root/_build/prod/rel/site/releases/0.0.1/site.tar.gz
```
Any pointers?

Dockerfile
alicia-mbta: This was already in the original dotcom `Dockerfile`. I don't know enough to attest that this is needed to have a working integration with Sentry.
Is "I can remove it and see if we get yelled at in prod" an option?

Dockerfile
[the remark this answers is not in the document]
ianwestcott: My hunch is that, in addition to the `dev`/`prod` hard-coding, there's some cruft left over from the `rel/` path being the method for transferring the built application between the two containers in the old setup. In which case, you'll need to make sure each `COPY --from` command matches up with the working directory of the previous build stage. 

To confirm where the previous build stage leaves its files, you could build the image partially, just to the end of that stage, by commenting out everything after it. Then run an interactive shell in the container (`docker run -it [container-name] /bin/bash`) and poke around to confirm where the files are located.

Dockerfile
ianwestcott: > Is "I can remove it and see if we get yelled at in prod" an option?

Maybe, but it might be worth investigating how that env var is used first. In which case I would definitely call it out of scope for this change, but perhaps worth tasking out separately if the Sentry integration for JavaScript is something y'all care about. (Or ripping out entirely if you don't.) FWIW, Skate has a working Sentry integration for its frontend, so it might be worth looking at their implementation for reference.

Dockerfile
```diff
@@ -1,33 +1,36 @@
-FROM hexpm/elixir:1.10.3-erlang-22.3.3-debian-stretch-20200224
+###
+# Three-stage Dockerfile
+###
 
-WORKDIR /root
+# 1.) Get the Elixir dependencies within an Elixir container
+FROM hexpm/elixir:1.10.3-erlang-22.3.3-debian-stretch-20200224 as elixir-builder
```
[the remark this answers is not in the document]
paulswartz: @ianwestcott we moved away from using that, because it doesn't pin the Erlang version behind the scenes. This has caused issues for us in the past where a point release of Erlang has broken some of our dependencies.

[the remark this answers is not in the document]
alicia-mbta: Should we revert back to the 2-stage version of the `Dockerfile` then?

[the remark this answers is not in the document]
paulswartz: No, just the switch back to the `hexpm/elixir` image. If you want to use Debian Buster, it looks like the latest is `1.10.3-erlang-22.3.3-debian-buster-20200224`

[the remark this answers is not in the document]
ianwestcott: @paulswartz oh yikes, good to know. There are a couple places where we're still using it, I'll follow up separately.

alicia-mbta:
Running in a container the latest version that I just pushed, I get the following:
`/root/rel/site/bin/site/startup: 4: /root/rel/site/bin/site/startup: /root/rel/site/bin/site: Permission denied`

I tried running `chmod` right before the `CMD` instruction but no luck. What am I missing?

Dockerfile
```diff
@@ -36,8 +41,45 @@ RUN npm run webpack:build -- --env.SENTRY_DSN=$SENTRY_DSN
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+	libssl1.1 libsctp1 curl \
+	&& rm -rf /var/lib/apt/lists/*
+
+WORKDIR /root
+ADD . .
+
+COPY --from=app-builder /root/_build/prod/rel ./_build/prod/rel
+
+ADD rel/bin/startup /root/rel/site/bin/site/startup
```
paulswartz, line 78: I think these two lines are where the permission error is coming from. `/root/rel/site/bin/startup` should be the path to the `startup` script, and it wants to find the `site` script in `/root/rel/site/bin/site`.

paulswartz:
```diff
diff --git a/Dockerfile b/Dockerfile
index aa85737d7..f27b6abdb 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,13 +73,13 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /root
 ADD . .

-COPY --from=app-builder /root/_build/prod/rel ./_build/prod/rel
+COPY --from=app-builder /root/_build/prod/rel /root/rel

-ADD rel/bin/startup /root/rel/site/bin/site/startup
+ADD rel/bin/startup /root/rel/site/bin/startup

 RUN mkdir /root/work

 WORKDIR /root/work

 # run the application
-CMD ["/root/rel/site/bin/site/startup", "foreground"]
+CMD ["/root/rel/site/bin/startup", "foreground"]
```

alicia-mbta: Did you try running it locally? With these latest corrections I'm still getting the ol' `Error: No such container` error. Looking into it.

paulswartz: It did, although I ran into a Node error:
```
2021-02-11T20:37:35.542 [error] node=site-unknown@2a955ad56cb3 Process #PID<0.2607.0> terminating
** (exit) exited in: Site.Application.start(:normal, [])
    ** (EXIT) shutdown: failed to start child: Site.React
        ** (EXIT) shutdown: failed to start child: :react_render
            ** (EXIT) an exception was raised:
                ** (MatchError) no match of right hand side value: {:error, {:enoent, [{:erlang, :open_port, [{:spawn_executable, nil}, [args: ["/root/rel/site/app.js"]]], []}, {Site.React.Worker, :init, 1, [file: 'lib/site/react/worker.ex', line: 15]}, {:gen_server, :init_it, 2, [file: 'gen_server.erl', line: 374]}, {:gen_server, :init_it, 6, [file: 'gen_server.erl', line: 342]}, {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 249]}]}}
                    (poolboy 1.5.1) src/poolboy.erl:275: :poolboy.new_worker/1
                    (poolboy 1.5.1) src/poolboy.erl:296: :poolboy.prepopulate/3
                    (poolboy 1.5.1) src/poolboy.erl:145: :poolboy.init/3
                    (stdlib 3.12.1) gen_server.erl:374: :gen_server.init_it/2
                    (stdlib 3.12.1) gen_server.erl:342: :gen_server.init_it/6
                    (stdlib 3.12.1) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
    (kernel 6.5.2) application_master.erl:138: :application_master.init/4
    (stdlib 3.12.1) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Initial Call: :application_master.init/4
Ancestors: [#PID<0.2606.0>]
```

alicia-mbta:
Unearthing this work again, apologies for the delay.

I made some progress: I was able to successfully build an image but I'm still unable to run it in a container in my Docker desktop (it seems due to `Supervisor` issues).
@paulswartz, could you please take a look at my latest commit and let me know if it looks ok, and if you're still running into that Node error? (or any other error). Thanks.

paulswartz:
It mostly looks good for me! The only issue I ran into was that the last stage is unable to run `react_renderer`. I believe that's because Node isn't available in the last container. Changing the last stage to use the same `nodejs` container as the Javascript build stage seems to have worked.

alicia-mbta:
Do you mean replacing:
```diff
-      FROM debian:buster
+      FROM node:14.15.1-buster
```
in line `61`?

Or, out of curiosity, would adding:
```dockerfile
COPY --from=assets-builder /root/apps/site/react_renderer/dist/app.js ./apps/site/react_renderer/dist/app.js
```
(in stage 2) also work?

paulswartz:
Yep, that's what I meant! You need more than only the `app.js` script. Since we're server-rendering the React templates, we need a Javascript engine running inside the container. By default that doesn't exist, so we need to use a container that does.

This isn't an issue for the Elixir code because of how the release is packaged. With `mix distillery.release` or `mix release`, all the code (including the Elixir and Erlang runtimes) are packaged along with the code. That's why we can run them without installing anything other than some library dependencies.

alicia-mbta:
Oh that makes sense. Thanks a lot for the explanation!

paulswartz (commented):
The `build_push.sh` script looks good. I think the next step is to replace the `build.sh` script with something that creates a `Dockerrun.aws.json` file. github.com/mbta/delta has an example of this.

paulswartz (commented):
Looks good! I think you're ready to try deploying it to one of the dev environments.

alicia-mbta:
Some more progress (and errors):

When trying to deploy, I got the following error message:

```bash
Command failed on 1 out of 2 instances. Incorrect application version found on 1 out of 2 instances. Expected version "2c1693b03" (deployment 1347).
```

Diving deeper into the logs, I saw this other better-explained error message:

```bash
Error response from daemon: manifest for 434035161053.dkr.ecr.us-east-1.amazonaws.com/dotcom:git-12f2b399a not found
  Failed to pull Docker image 434035161053.dkr.ecr.us-east-1.amazonaws.com/dotcom:git-12f2b399a
```

Do we need to add a manifest file somewhere?

In terms of the missing Docker image, I ran
```bash
aws ecr list-images --repository-name "dotcom"  --region us-east-1
```
and this is the output:
```json
{
    "imageIds": [
        {
            "imageDigest": "sha256:1d03b5f61265aca295496b1fd6b3675c4c53c40b4184e388d056a228708f50cb",
            "imageTag": "dev-blue"
        },
        {
            "imageDigest": "sha256:1d03b5f61265aca295496b1fd6b3675c4c53c40b4184e388d056a228708f50cb",
            "imageTag": "git-90b8854f8"
        }
    ]
}
```

Should there be more ImageIds in there? Is there anything I'm missing?

paulswartz:
I'd expect to see an image ID with the latest SHA from this branch: `3247302` as I write. Did the `build_push` script run successfully before the deploy?

alicia-mbta:
I had missed that step  🤦‍♀️   and also had some wrong commands in there.

But now it seems like it was successful (could this be??):

image

paulswartz:
That looks right! 🎉

paulswartz (approved):
🍰

PR #694, opened 2020-12-09
feat: Add validation for vehicle number field

thecristen:
#### Summary of changes
**Asana Ticket:** [Customer Support | Don't allow letters in the vehicle number field](https://app.asana.com/0/385363666817452/1199530274677105/f)

Adds front and back end validation for the vehicle number field. Blank/no entry is ok since the field is optional, but if filled it should contain numeric characters only.

---

### DIFF
--- a/apps/site/assets/js/support-form.js
+++ b/apps/site/assets/js/support-form.js
@@ -351,6 +351,13 @@ const validators = {
   },
   recaptcha: function($) {
     return !!$("#g-recaptcha-response").val();
+  },
+  vehicle: function($) {
+    const value = $("#vehicle").val();
+    if (value && value.length > 0) {
+      return /^[0-9]+$/.test(value);
+    }
+    return true;
   }
 };
 
@@ -365,7 +372,8 @@ function setupValidation($) {
     "#comments",
     "#email",
     "#first_name",
-    "#last_name"
+    "#last_name",
+    "#vehicle"
   ].forEach(selector => {
     const $selector = $(selector);
     $selector.on("keyup blur input change", () => {
@@ -398,6 +406,7 @@ function validateForm($) {
     subject = "#support_subject",
     comments = "#comments",
     service = "#service",
+    vehicle = "#vehicle",
     email = "#email",
     first_name = "#first_name",
     last_name = "#last_name",
@@ -436,6 +445,13 @@ function validateForm($) {
     errors.push(last_name);
   } else {
     displaySuccess($, last_name);
+  }
+  // Vehicle number
+  if (!validators.vehicle($)) {
+    displayError($, vehicle);
+    errors.push(vehicle);
+  } else {
+    displaySuccess($, vehicle);
   }
   // Phone and email
   if (!validators.email($)) {
--- a/apps/site/assets/js/support-form.js
+++ b/apps/site/assets/js/support-form.js
@@ -354,10 +354,7 @@ const validators = {
   },
   vehicle: function($) {
     const value = $("#vehicle").val();
-    if (value && value.length > 0) {
-      return /^[0-9]+$/.test(value);
-    }
-    return true;
+    return /^[0-9]*$/.test(value);
   }
 };
 
--- a/apps/site/lib/site_web/controllers/customer_support_controller.ex
+++ b/apps/site/lib/site_web/controllers/customer_support_controller.ex
@@ -253,13 +253,8 @@ defmodule SiteWeb.CustomerSupportController do
   defp validate_privacy(_), do: "privacy"
 
   @spec validate_vehicle(map) :: :ok | String.t()
-  defp validate_vehicle(%{"vehicle" => ""}), do: :ok
-
   defp validate_vehicle(%{"vehicle" => vehicle_number}) do
-    case Regex.run(~r/^[0-9]+$/, vehicle_number) do
-      nil -> "vehicle"
-      [_] -> :ok
-    end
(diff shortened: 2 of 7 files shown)


apps/site/assets/js/support-form.js
```diff
@@ -348,6 +348,13 @@ const validators = {
   },
   recaptcha: function($) {
     return !!$("#g-recaptcha-response").val();
+  },
+  vehicle: function($) {
+    const value = $("#vehicle").val();
+    if (value && value.length > 0) {
+      return /^[0-9]+$/.test(value);
+    }
+    return true;
```
paulswartz, line 357: Does `/^[0-9]*$/` cover the empty case as well? Would simplify the function.

apps/site/assets/js/support-form.js
```diff
@@ -356,16 +363,21 @@ function responseRequested($) {
...
-    }
-  );
+  [
+    "#privacy",
+    "#comments",
+    "#email",
+    "#first_name",
+    "#last_name",
+    "#vehicle"
+  ].forEach(selector => {
```
paulswartz, line 373: I don't remember whether we support IE < 9, but `forEach` isn't supported there: https://caniuse.com/mdn-javascript_builtins_array_foreach

apps/site/lib/site_web/templates/customer_support/_form.html.eex
```diff
@@ -64,9 +64,10 @@
         </div>
       </div>
       <div class="plan-time-select">
-        <%= label f, "Vehicle", [id: "vehicleLabel", for: "support_vehicle", class: "form-control-label"] %>
-        <div>
-          <%= text_input(f, :vehicle, placeholder: placeholder_text("vehicle"), class: "plan-date-input") %>
+        <%= support_error_tag @errors, "vehicle" %>
+        <div class="form-group <%= class_for_error("vehicle", @errors, "has-danger", "has-success") %>">
+          <%= label f, "Vehicle", [id: "vehicleLabel", for: "vehicle", class: "form-control-label"] %>
+          <%= text_input(f, :vehicle, id: "vehicle", placeholder: placeholder_text("vehicle"), class: "form-control plan-date-input", value: assigns[:vehicle] ) %>
```
paulswartz, line 70: Is it worth using the HTML5 regex validator here as well? It's not supported everywhere, but it's pretty wide: https://caniuse.com/input-pattern

paulswartz, line 70: I realize this was there before, but is `plan-date-input` the right class? Since it's not a date input.

apps/site/assets/js/support-form.js
thecristen: Thankfully we only support IE 11+!

apps/site/assets/js/support-form.js
thecristen: How so? Empty cases are valid since the field is optional, so I'm just testing non-empty cases here.

paulswartz: Right, so something like:
```
const value = $("#vehicle").val();
return /^[0-9]*$/.test(value)
```

apps/site/lib/site_web/templates/customer_support/_form.html.eex
thecristen: I thought about using this... I think its main benefit would be a more informative message could be offered by the browser, but the way our front-end validation is currently set up the user wouldn't see that message anyway. Which is not super great, but for that reason I chose to not use the input pattern attribute here (or elsewhere in the form).

thecristen: There's this and a few other classes borrowing styling from various inputs in the Trip Planner.. I'll clean these up in a followup commit

apps/site/assets/js/support-form.js
thecristen: Ahh I missed the subtle `+` to `*` change there...

apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -249,6 +251,18 @@ defmodule SiteWeb.CustomerSupportController do
   defp validate_privacy(%{"privacy" => "on"}), do: :ok
   defp validate_privacy(_), do: "privacy"
 
+  @spec validate_vehicle(map) :: :ok | String.t()
+  defp validate_vehicle(%{"vehicle" => ""}), do: :ok
+
+  defp validate_vehicle(%{"vehicle" => vehicle_number}) do
+    case Regex.run(~r/^[0-9]+$/, vehicle_number) do
```
paulswartz, line 258: - You can use the `*` regex here as well.
- You can also use `Regex.match?` since you don't care about the results.

apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -249,6 +251,15 @@ defmodule SiteWeb.CustomerSupportController do
   defp validate_privacy(%{"privacy" => "on"}), do: :ok
   defp validate_privacy(_), do: "privacy"
 
+  @spec validate_vehicle(map) :: :ok | String.t()
+  defp validate_vehicle(%{"vehicle" => ""}), do: :ok
```
paulswartz, line 255: You don't need this case anymore, since you're using `*` in the match.

paulswartz (approved):
One small thing, otherwise 🍰

PR #695, opened 2020-12-10
Update language versions

paulswartz:
#### Summary of changes
**Asana Ticket:** [Update language versions](https://app.asana.com/0/555089885850811/1163637176608470)

Based on @arkadyan 's origin branch, rebased onto the latest main. 

---


apps/alerts/mix.exs
```diff
@@ -22,7 +22,7 @@ defmodule Alerts.Mixfile do
   # Type "mix help compile.app" for more information
   def application do
     [
-      applications: [:logger, :timex, :repo_cache, :con_cache, :tzdata, :util],
```
alicia-mbta, line 25: Are the ones but `:logger` inferred and that's why they're removed from the list?

apps/site/lib/site/green_line/cache_supervisor.ex
```diff
@@ -1,14 +1,23 @@
 defmodule Site.GreenLine.CacheSupervisor do
-  use Supervisor
+  @moduledoc false
+  use DynamicSupervisor
 
   alias Site.GreenLine.DateAgent
 
-  def start_link do
-    Supervisor.start_link(__MODULE__, nil, name: :green_line_cache_supervisor)
+  @name :green_line_cache_supervisor
```
alicia-mbta, line 7: Great refactor.

apps/alerts/mix.exs
paulswartz: Yep, they're part of the `deps` list so Elixir already knows we depend on them. `extra_applications` was added back in Elixir 1.4 (https://elixir-lang.org/blog/2017/01/05/elixir-v1-4-0-released/), but we never went back to update our configuration.

semaphore/setup.sh
```diff
@@ -20,10 +20,10 @@ sudo swapon /swapfile
 
 # Ensure cache directories exist
 CACHE=$SEMAPHORE_CACHE_DIR
-mkdir -p "$CACHE/asdf/installs" "$CACHE/mix/deps" "$CACHE/npm"
+mkdir -p "$CACHE/asdf/installs" "$CACHE/mix" "$CACHE/npm"
 
 export MIX_ENV=test
-export MIX_DEPS_PATH=$CACHE/mix/deps
+export MIX_HOME=$CACHE/mix
```
alicia-mbta, line 26: Has this changed because of analog changes on the Semaphore side?

.tool-versions
```diff
@@ -1,4 +1,3 @@
-erlang 21.2
-elixir 1.8.1-otp-21
-nodejs 8.17.0
-python 3.8.5
```
thecristen, line 4: This was added in https://github.com/mbta/dotcom/pull/574 though admittedly I've never used the script and forgot about it until just now....

If we remove Python here (which I'm probably fine with?, as most devs won't use it anyway) we should probably add a note to the deploy script documentation about it.

semaphore/setup.sh
paulswartz: Without this, I ran into an issue compiling the Phoenix Node dependencies.

thecristen (commented):
image

The spacing between the icon and text in the `<h1>` is missing. Even after extensive inspection I have not yet figured out why, so that might be neither here nor there...

.tool-versions
paulswartz: Nope, good catch. I'll put it back.

paulswartz:
> image
> 
> The spacing between the icon and text in the `<h1>` is missing. Even after extensive inspection I have not yet figured out why, so that might be neither here nor there...

That's weird...I'll poke around.

paulswartz:
On dev, the tail end of the `<h1>`:
```
</span></span>          Commuter Rail
        </h1>
```

On dev-green:
```
</span></span>Commuter Rail        </h1>
```

So it looks like a change in how we're rendering the whitespace in the HTML.

paulswartz:
Maybe related to this fix from Elixir 1.9?
> [EEx] Consistently trim newlines when you have a single EEx expression per line on multiple lines

thecristen (approved):
I wish we had our visual regression testing situation working so I could be more confident this problem isn't showing elsewhere, but in the meantime this will do.

paulswartz:
I did go through and search for text next to `svg_icon`, and addressed this issue there as well.

paulswartz:
@thecristen dev green looks better now: great catch!

PR #697, opened 2020-12-14
Return the error when one is received

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Elixir.UndefinedFunctionError: function nil.id/0 is undefined](https://app.asana.com/0/555089885850811/1196094168100899)

We weren't taking into account when something other than a list is returned: https://github.com/mbta/dotcom/blob/a34a39656e3cbc86ab0f5178e8511b9281526963/apps/site/lib/site_web/controllers/schedule/trip_info.ex#L96

<br/> With this fix we make sure even the error is returned for its subsequent handling (in `handle_trip`).

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/trip_info.ex
+++ b/apps/site/lib/site_web/controllers/schedule/trip_info.ex
@@ -93,15 +93,22 @@ defmodule SiteWeb.ScheduleController.TripInfo do
     # Compute the active stop that matches the one computed by the tooltips
     active_stop = VehicleLocations.active_stop(conn.assigns.vehicle_locations, trip_id)
 
-    with trips when is_list(trips) <- opts[:trip_fn].(trip_id, date: conn.assigns.date) do
-      trips
-      |> build_trip_times(conn.assigns, trip_id, opts[:prediction_fn])
-      |> TripInfo.from_list(
-        vehicle: opts[:vehicle_fn].(trip_id),
-        vehicle_stop_name: active_stop,
-        origin_id: conn.query_params["origin"],
-        destination_id: conn.query_params["destination"]
-      )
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/controllers/schedule/trip_info.ex
```diff
@@ -102,6 +102,8 @@ defmodule SiteWeb.ScheduleController.TripInfo do
         origin_id: conn.query_params["origin"],
         destination_id: conn.query_params["destination"]
       )
+    else
+      error -> error
```
thecristen, line 106: It looks to me like you want to return `{:error, error} `instead of only `error` here so it might get picked up in case statement in `handle_trip/3`.

thecristen: Wait, unless `error` is already that tuple.. 🤯 I don't know.

thecristen (changes requested):
I think we can add a test for this in `apps/site/test/site_web/controllers/schedule/trip_info_test.exs` with a test where we pass into `init()` a custom `trip_fn` option that returns something that's not a list of trips.

apps/site/lib/site_web/controllers/schedule/trip_info.ex
paulswartz: You can handle both if needed:
```elixir
else
  {:error, _} = error -> error
  error -> error
end
```

alicia-mbta: > Wait, unless `error` is already that tuple.. 🤯 I don't know.

Yes, exactly, the error is already a tuple :) I will add a test though.

thecristen (approved):

PR #699, opened 2020-12-15
chore: configured Dependabot

paulswartz:
It looks like the default configuration is only looking at updates for `react_renderer`.

---


thecristen (approved):
Oh what a can of worms we will open up now 😃 
I'm wary that checking daily will drown us in a deluge of PRs but hey let's try it! I'm sure we can always change to weekly if daily is too much.

paulswartz:
Hopefully once we get more up-to-date the flood will slow down :)

PR #718, opened 2020-12-18
Default to today's date and time values if they are missing

alicia-mbta:
#### Summary of changes
**Asana Tickets:**
[Elixir.ArgumentError: unrecognized time %{"hour" => "6"}](https://app.asana.com/0/385363666817452/1199213920216686)
[Elixir.ArgumentError: unrecognized date %{"am_pm" => "PM", "hour" => "4", "minute" => "20"}](https://app.asana.com/0/385363666817452/1199148636489026)

Ensure all fields for the date selector are included

### DIFF
--- a/apps/site/lib/site_web/views/date_time_selector.ex
+++ b/apps/site/lib/site_web/views/date_time_selector.ex
@@ -12,6 +12,29 @@ defmodule DateTimeSelector do
                     int -> Integer.to_string(int)
                   end)
 
+  @spec verify_params(Form.t(), DateTime.t()) :: map
+  defp verify_params(form, datetime) do
+    datetime_params = form.params["date_time"] || %{}
+
+    if Util.parse(datetime_params) == {:error, :invalid_date} do
+      [hour, am_pm_value] =
+        datetime
+        |> Timex.format!("{h12} {AM}")
+        |> String.split(" ")
+
+      %{
+        "year" => datetime.year,
+        "month" => datetime.month,
+        "day" => datetime.day,
+        "hour" => hour,
+        "minute" => datetime.minute,
+        "am_pm" => am_pm_value
+      }
+    else
+      datetime_params
+    end
+  end
+
   def custom_date_time_select(form, date_ranges, %DateTime{} = datetime \\ Util.now()) do
     time_options = [
       hour: [options: 1..12, selected: Timex.format!(datetime, "{h12}")],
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/views/date_time_selector.ex
```diff
@@ -12,6 +12,26 @@ defmodule DateTimeSelector do
                     int -> Integer.to_string(int)
                   end)
 
+  @spec verify_params(map, map) :: map
```
thecristen, line 15:
```suggestion
  @spec verify_params(Form.t(), DateTime.t()) :: map
```

I think these types might make sense for this function?

apps/site/lib/site_web/views/date_time_selector.ex
```diff
@@ -12,6 +12,26 @@ defmodule DateTimeSelector do
...
+  defp verify_params(form, datetime) do
+    default_datetime_params = %{
+      "year" => datetime.year,
+      "month" => datetime.month,
+      "day" => datetime.day,
+      "hour" => datetime.hour,
```
thecristen, line 21: I'm seeing some funny defaults, like 13 (hour) AM.

thecristen: This default datetime corresponds with `Util.now()` on first load but after I click the button to plan a trip it becomes
```
%{
  "am_pm" => "AM",
  "day" => 21,
  "hour" => 13,
  "minute" => 20,
  "month" => 12,
  "year" => 2020
}
```
(when I run it now at 6:20PM in this time zone)

This probably isn't caused by your PR but something sinister is going on here...

thecristen (changes requested):
I'm running into some trip planner bugs that make it tricky for me to test _this_ change! I left some comments but don't feel like I can fairly evaluate with this other bug in place (screenshot - the 6:25PM is correct but it shows as 1:25PM in the actual selector and that's what gets passed into the backend, so the values aren't missing but they're incorrect (and the trip planner can't plan the trip because it thinks I'm looking up a date in the past))

image

(since it's a 5 hour difference I suspect it's a time zone thing)

apps/site/lib/site_web/views/date_time_selector.ex
alicia-mbta: I think the latest changes will not cause any inconsistencies (I was not converting the time properly).
Were you entering specific values in the Trip Planner so I can reproduce the issue you were having?

thecristen (approved):
Looks good!!

PR #723, opened 2020-12-21
Implement `@behaviour` for `Routes.Repo`

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Define a RoutesApi behaviour for V3Api.Routes](https://app.asana.com/0/555089885850811/1189396929591660)

### DIFF
--- a/apps/routes/lib/repo.ex
+++ b/apps/routes/lib/repo.ex
@@ -1,6 +1,8 @@
 defmodule Routes.Repo do
   @moduledoc "Repo for fetching Route resources and their associated data from the V3 API."
 
+  @behaviour Routes.RepoApi
+
   require Logger
   use RepoCache, ttl: :timer.hours(1)
 
@@ -8,19 +10,14 @@ defmodule Routes.Repo do
 
   alias JsonApi
   alias Routes.{Route, Shape}
-  alias V3Api.{Routes, Shapes}
+  alias V3Api.{Shapes}
 
   @default_opts [include: "route_patterns"]
 
-  @doc """
-
-  Returns a list of all the routes
-
-  """
-  @spec all() :: [Route.t()]
+  @impl Routes.RepoApi
   def all do
     case cache(@default_opts, fn _ ->
-           result = handle_response(Routes.all(@default_opts))
+           result = handle_response(V3Api.Routes.all(@default_opts))
 
            for {:ok, routes} <- [result],
                route <- routes do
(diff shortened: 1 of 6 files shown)


alicia-mbta:
This draft PR is an attempt to achieve what was discussed in https://github.com/mbta/dotcom/pull/665

apps/routes/lib/repo.ex
```diff
@@ -212,7 +215,7 @@ defmodule Routes.Repo do
   @doc """
   The Green Line.
   """
```
paulswartz, line 217: Does it make sense to move the documentation into `Routes.RepoApi` as well?

alicia-mbta: You're right, I'll re-organize the documentation.

apps/routes/lib/repo.ex
```diff
@@ -1,26 +1,23 @@
...
 
   @default_opts [include: "route_patterns"]
 
-  @doc """
-
-  Returns a list of all the routes
-
-  """
-  @spec all() :: [Route.t()]
+  @impl Routes.RepoApi
```
thecristen, line 17: Do you need both the `@behavior` and the `@impl` declarations?

paulswartz: `@impl` tells the compiler which interface this is a function for. It'll help you out by failing if you forget any implementations.

thecristen (approved):
👍

apps/routes/lib/route.ex
```diff
@@ -202,6 +202,29 @@ defmodule Routes.Route do
 
   def silver_line, do: @silver_line
 
+  @doc """
+  Determines if the given route data is hidden
+  """
+  @spec route_hidden?(%{id: String.t()}) :: boolean
+  def route_hidden?(%{id: "746"}), do: true
```
paulswartz, line 209: Since you're on `Route` already, this can be `hidden?()`

apps/routes/lib/route.ex
```diff
@@ -202,6 +202,29 @@ defmodule Routes.Route do
 
   def silver_line, do: @silver_line
 
+  @doc """
+  Determines if the given route data is hidden
+  """
+  @spec hidden?(%{id: String.t()}) :: boolean
```
paulswartz, line 208: `hidden?(t()) :: boolean`

apps/routes/lib/repo_api.ex
```diff
@@ -0,0 +1,57 @@
...
+
+  @doc """
+  Given a route pattern, returns the list of routes matching it.
+  """
+  @callback by_stop_with_route_pattern(String.t()) :: [Route.t()]
+
+  @doc """
+  Parses json into a list of routes, or an error if it happened.
+  """
+  @callback handle_response(JsonApi.t() | {:error, any}) :: {:ok, [Route.t()]} | {:error, any}
```
paulswartz, line 51: This looks like it's an implementation detail of the real `RouteApi`: I don't think a fake implementation would need to implement it?

paulswartz (approved):
🍰

PR #727, opened 2020-12-23
feat: Add x-robots-tag header in non-prod envs

thecristen:
Removes redundant controller helper.

#### Summary of changes
**Asana Ticket:** [Dev Environments | Block from Search Engines](https://app.asana.com/0/385363666817452/1199225274170402/f)

I'm not sure whether this is a legitimate way to check whether the requested page is in our production instance or not, but I think this Plug would be the way to do this? I await feedback!


paulswartz:
The Mix environment `Mix.env()` is separate from where the application is running.

- `Mix.env()` == `:dev` is the server running on your local machine
- `Mix.env()` == `:test` is the server running in test mode (unit/integration tests)
- `Mix.env()` == `:prod` is everything else (for us, running in AWS)

Since `Mix.env()` will always be `:prod` in AWS, we need a different way to do it. What I'd suggest is for the default to be disabling indexing, but provide an override in the form of an environment variable. Something like:

```elixir
# apps/site/config/config.exs
config :site,
  allow_indexing: false # overridden in lib/site/application.ex

# apps/site/lib/site/application.ex in start/2
if System.get_env("ALLOW_INDEXING") == "true" do
  Application.put_env(:site, :allow_indexing, true)
end

# apps/site/lib/site/router.ex)
plug(:optional_disable_indexing)

def optional_disable_indexing(conn, _) do
  if Application.get_env(:site, :allow_indexing) do
    conn
  else
    Site.Helpers.noindex(conn)
  end
end
```

We can then add the `ALLOW_INDEXING`=`true` configuration in only the `dotcom-prod` environment: any other environment will get the header on each request.

apps/site/lib/site/application.ex
```diff
@@ -8,6 +8,10 @@ defmodule Site.Application do
   def start(_type, _args) do
     import Supervisor.Spec, warn: false
 
+    if System.get_env("ALLOW_INDEXING") == "true" do
+      Application.put_env(:site, :allow_indexing, true)
+    end
```
paulswartz, line 13: For codecov purposes, you could use into a helper function which takes the value of `System.get_env("ALLOW_INDEXING")` and returns the value to put in the environment?

paulswartz (approved):
Other than codecov, looks good! :cake:

We'll need to add this environment variable to dotcom-prod before we release it to avoid all of MBTA.com dropping out of search engines.

PR #732, opened 2020-12-24
Retirement message

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [⏱️Schedule Tab | Add retirement message](https://app.asana.com/0/555089885850811/1170822992031033)

Create a plug for a retirement message (for CR and bus) in the Schedule tab.

**Pending: specific date**

image

image

### DIFF
--- a/apps/site/lib/site_web/controllers/helpers.ex
+++ b/apps/site/lib/site_web/controllers/helpers.ex
@@ -178,4 +178,8 @@ defmodule SiteWeb.ControllerHelpers do
 
     "#{Strftime.format!(one_year_after, "%d %b %Y")} 00:00:00 EST"
   end
+
+  def call_plug_with_opts(conn, module, opts) do
+    module.call(conn, module.init(opts))
+  end
 end
--- a/apps/site/lib/site_web/controllers/schedule/green.ex
+++ b/apps/site/lib/site_web/controllers/schedule/green.ex
@@ -5,6 +5,7 @@ defmodule SiteWeb.ScheduleController.Green do
   use SiteWeb, :controller
 
   import UrlHelpers, only: [update_url: 2]
+  import SiteWeb.ControllerHelpers, only: [call_plug: 2, call_plug_with_opts: 3, assign_alerts: 2]
 
   alias Schedules.Schedule
   alias SiteWeb.ScheduleController.LineController
@@ -48,6 +49,10 @@ defmodule SiteWeb.ScheduleController.Green do
   def trip_view(conn, _params) do
     conn
     |> assign(:tab, "trip-view")
+    |> call_plug_with_opts(SiteWeb.Plugs.BannerMessage,
+      message_key: :retirement_message,
+      message: SiteWeb.ScheduleView.build_retirement_message(conn)
+    )
     |> put_view(ScheduleView)
     |> render("show.html", [])
   end
@@ -108,7 +113,7 @@ defmodule SiteWeb.ScheduleController.Green do
       |> conn_with_branches
       |> Task.async_stream(
         fn conn ->
-          call_plug(conn, SiteWeb.ScheduleController.Schedules, opts).assigns.schedules
+          call_plug_with_opts(conn, SiteWeb.ScheduleController.Schedules, opts).assigns.schedules
         end,
         timeout: @task_timeout
       )
@@ -155,7 +160,7 @@ defmodule SiteWeb.ScheduleController.Green do
       |> conn_with_branches
       |> Task.async_stream(
         fn conn ->
-          call_plug(conn, SiteWeb.ScheduleController.VehicleLocations, opts).assigns.vehicle_locations
+          call_plug_with_opts(conn, SiteWeb.ScheduleController.VehicleLocations, opts).assigns.vehicle_locations
         end,
         timeout: @task_timeout
       )
@@ -215,11 +220,6 @@ defmodule SiteWeb.ScheduleController.Green do
       |> redirect(to: url)
       |> halt
     end
-  end
-
-  # takes opts at runtime: used for testing
-  defp call_plug(conn, module, opts) do
-    module.call(conn, module.init(opts))
   end
 
   defp conn_with_branches(conn) do
--- a/apps/site/lib/site_web/controllers/schedule/trip_view_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/trip_view_controller.ex
@@ -23,6 +23,11 @@ defmodule SiteWeb.ScheduleController.TripViewController do
 
   def show(conn, _) do
     conn
+    |> SiteWeb.ControllerHelpers.call_plug_with_opts(
+      SiteWeb.Plugs.BannerMessage,
+      message_key: :retirement_message,
+      message: SiteWeb.ScheduleView.build_retirement_message(conn)
+    )
     |> put_view(SiteWeb.ScheduleView)
     |> render("show.html", [])
   end
--- a/apps/site/lib/site_web/plugs/banner_message.ex
+++ b/apps/site/lib/site_web/plugs/banner_message.ex
@@ -1,0 +1,22 @@
+defmodule SiteWeb.Plugs.BannerMessage do
+  @moduledoc """
+  Assigns a message to `conn` using the following params:
+  - `message_key` is the key that will be used to retrieve the message.
+  - `message` is a map representing the actual message.
+  """
+  @behaviour Plug
+
+  import Plug.Conn, only: [assign: 3]
+
+  @impl true
+  def init(opts) do
+    opts
+  end
+
+  @impl true
+  def call(conn, message_key: message_key, message: message) do
+    assign(conn, message_key, message)
+  end
+
+  def call(conn, _), do: conn
+end
--- a/apps/site/lib/site_web/templates/schedule/_trip_view.html.eex
+++ b/apps/site/lib/site_web/templates/schedule/_trip_view.html.eex
@@ -1,4 +1,7 @@
 <%= SiteWeb.AlertView.group(alerts: @alerts, route: @route, date_time: @date_time, priority_filter: :high) %>
+<%= if Map.has_key?(@conn.assigns, :retirement_message) do %>
+  <%= SiteWeb.ViewHelpers.banner_message(@conn, :retirement_message) %>
+<% end %>
 <%= render "_trip_view_filters.html", forward_assigns(@conn) %>
 <div class="calendar-covered">
   <%= content_tag(:div, "", [class: "calendar-cover", hidden: !@show_date_select?]) %>
--- a/apps/site/lib/site_web/views/schedule_view.ex
+++ b/apps/site/lib/site_web/views/schedule_view.ex
@@ -481,4 +481,19 @@ defmodule SiteWeb.ScheduleView do
   def json_safe_route(route) do
     Route.to_json_safe(route)
   end
+
+  def build_retirement_message(%Conn{assigns: %{route: %Route{type: 2}}}), do: nil
+
+  def build_retirement_message(conn) do
+    path = line_path(conn, :show, conn.assigns[:route].id)
+    link_to_schedule_finder = content_tag(:a, "Schedule Finder", href: path)
+
+    %{
+      header: ["We'll be retiring this tool on March 1st, 2021."],
+      body: [
+        "All the information provided here, including scheduled trips and realtime departure and arrival information, can be found in our ",
+        link_to_schedule_finder
+      ]
+    }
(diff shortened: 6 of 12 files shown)


alicia-mbta:
The specific date is still pending from Katrina but other than that, this is ready to be reviewed.

apps/site/lib/site_web/templates/schedule/_retirement_message.html.eex
```diff
@@ -0,0 +1,4 @@
+<div class="stop-filters schedule-filter-row callout">
+  <p class="inner-label"><%= @retirement_message.header %></p>
```
thecristen, line 2:
```suggestion
  <p class="u-bold"><%= @retirement_message.header %></p>
```

The bold utility class is sufficient here.

apps/site/lib/site_web/templates/schedule/_retirement_message.html.eex
```diff
@@ -0,0 +1,4 @@
+<div class="stop-filters schedule-filter-row callout">
```
thecristen, line 1:
```suggestion
<div class="callout">
```

Looks like you copied this from elsewhere! Try to use the most general classes needed to achieve the desired effect, here it looks like only the `.callout` class is necessary?

apps/site/lib/site_web/plugs/retirement_message.ex
```diff
@@ -0,0 +1,31 @@
...
+    "All the information provided here, including scheduled trips and realtime departure and arrival information, can be found in our #{
+      link_to_schedule_finder
+    }."
+  end
+
+  @impl true
+  def init(_), do: %{header: "", body: ""}
+
+  @impl true
+  def call(conn, _) do
```
thecristen, line 28: I wonder if this assignment can happen dynamically by seeing whether we're on the Schedule Tab or not? 
```suggestion
  def call(%{query_params: %{"tab" => "trip-view"} = query_params} = conn, _) do
      # This is the schedule tab I think, so add the message here. Otherwise just return conn
```

apps/site/lib/site_web/plugs/retirement_message.ex
```diff
@@ -0,0 +1,31 @@
...
+
+    link_to_schedule_finder = safe_to_string(tag)
+
+    "All the information provided here, including scheduled trips and realtime departure and arrival information, can be found in our #{
+      link_to_schedule_finder
+    }."
+  end
+
+  @impl true
+  def init(_), do: %{header: "", body: ""}
```
thecristen, line 25: Just noting here that since the second argument in the call function isn't used, this map that's returned here doesn't see the light of day. 🤷‍♀️ I think it's ok to not return anything here.

apps/site/lib/site_web/plugs/retirement_message.ex
```diff
@@ -0,0 +1,31 @@
+defmodule SiteWeb.Plugs.RetirementMessage do
```
thecristen, line 1: Since the content of this plug is only applicable to retiring the Schedule Tab, could we rename this to be more specific? Something like `SiteWeb.Plugs.ScheduleTabRetirementMessage`?

(I'm also noting here that this plug doesn't seem quite suited for more generalized writing of retirement messages for other things that might come up in the future - do correct me if that was actually your intent!)

apps/site/lib/site_web/plugs/retirement_message.ex
```diff
@@ -0,0 +1,31 @@
...
+  defp message_body(conn) do
+    path = line_path(conn, :show, conn.assigns[:route].id)
+    tag = content_tag(:a, "Schedule Finder", href: path)
+
+    link_to_schedule_finder = safe_to_string(tag)
+
+    "All the information provided here, including scheduled trips and realtime departure and arrival information, can be found in our #{
+      link_to_schedule_finder
+    }."
+  end
```
paulswartz, line 22: Is there a reason to return this as a raw HTML, rather than a `safe` tag? I believe you can do:
```elixir
# ... up to `tag`...
["All the information ... can be found in our ", tag]
```

apps/site/lib/site_web/templates/schedule/_retirement_message.html.eex
```diff
@@ -0,0 +1,4 @@
...
+  <p class="inner-label"><%= @retirement_message.header %></p>
+  <span><%= @retirement_message.body |> raw() %></span>
```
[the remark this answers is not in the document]
paulswartz: Also, see above above about avoiding `raw`.

apps/site/lib/site_web/templates/schedule/_trip_view.html.eex
```diff
@@ -1,4 +1,5 @@
 <%= SiteWeb.AlertView.group(alerts: @alerts, route: @route, date_time: @date_time, priority_filter: :high) %>
+<%= render "_retirement_message.html", retirement_message: @retirement_message %>
```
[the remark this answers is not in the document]
paulswartz: It could even take the `@route` and appropriately handle CR separately.

apps/site/lib/site_web/plugs/retirement_message.ex
alicia-mbta: Thanks for pointing me to the proper way, I wasn't super familiar with the `safe` way and using `raw` was the only way to render the tag that I was able to get to work.

apps/site/lib/site_web/plugs/retirement_message.ex
alicia-mbta: The whole point of creating a plug (besides learning to build one for the first time) was to actually have a reusable message for the future by having a header and a body.

What disadvantages do you see? You mentioned a view function would suffice but I'm curious about scenarios in which a plug would be a better fit than here.

Also let me know if you prefer said view function and I can do that instead.

thecristen: Hmm. I think if you adjust the plug's call function to be able to return different messages based on what's in the `conn` (like I mentioned in a previous comment) that'll give us that reusability. With the pattern matching approach I'm envisioning it could even support multiple retirement messages simultaneously across the site 🤩

apps/site/lib/site_web/plugs/banner_message.ex
```diff
@@ -0,0 +1,29 @@
...
+
+  import Plug.Conn, only: [assign: 3]
+
+  @impl true
+  def init(opts) do
+    opts
+  end
+
+  @impl true
+  def call(%Plug.Conn{assigns: %{tab: "trip-view"}} = conn, opts) do
```
thecristen, line 17: If this is going to be for general use we shouldn't limit it to the trip-view tab here. Also we'll want to make sure the message_key and message are both present before assigning so I think this might work better here:

```suggestion
  def call(conn, [message_key: message_key, message: message]) do
```

apps/site/lib/site_web/controllers/schedule/trip_view_controller.ex
```diff
@@ -34,4 +38,8 @@ defmodule SiteWeb.ScheduleController.TripViewController do
   defp zone_map(conn, _), do: conn
 
   defp tab_name(conn, _), do: assign(conn, :tab, "trip-view")
+
+  defp call_plug_with_opts(conn, module, opts) do
```
thecristen, line 42: 👍 Seems like this could be a broadly useful helper, maybe put this in `SiteWeb.ControllerHelpers`?

apps/site/lib/site_web/controllers/schedule/green.ex
```diff
@@ -48,6 +48,10 @@ defmodule SiteWeb.ScheduleController.Green do
   def trip_view(conn, _params) do
     conn
     |> assign(:tab, "trip-view")
+    |> call_plug(SiteWeb.Plugs.BannerMessage,
```
thecristen, line 51: Oh I just noticed this one isn't using `call_plug_with_opts`... does this work as desired then?

apps/site/lib/site_web/views/schedule_view.ex
```diff
@@ -481,4 +481,24 @@ defmodule SiteWeb.ScheduleView do
...
   end
+
+  @spec retirement_message(Conn.t()) :: Safe.t()
+  def retirement_message(%Conn{assigns: %{route: %Route{type: 2}}}), do: nil
+
+  def retirement_message(conn) do
+    SiteWeb.ViewHelpers.banner_message(conn, :retirement_message)
+  end
+
+  def build_retirement_message(conn) do
```
thecristen, line 492: I think *this* is where we'll want to return nil if the route is a commuter rail. Then for that case, when the plug is called it'll see the message returned here is nil and then it won't assign the retirement_message variable.

apps/site/lib/site_web/templates/schedule/_trip_view.html.eex
```diff
@@ -1,4 +1,5 @@
 <%= SiteWeb.AlertView.group(alerts: @alerts, route: @route, date_time: @date_time, priority_filter: :high) %>
+<%= retirement_message(@conn) %>
```
thecristen, line 2: This might be simpler (you wouldn't need the `retirement_message` view function) as something like
```
<%= if @retirement_message do %>
  <%= SiteWeb.ViewHelpers.banner_message(@conn, :retirement_message) %>
<% end %>
```
(assuming the commuter rail part of the logic was already taken care of when `@retirement_message` was added to `conn.assigns`, which I think it should be)

apps/site/lib/site_web/views/schedule_view.ex
thecristen: Oh, and I think this is also where we'd want to verify if we're on the relevant tab via `%Plug.Conn{assigns: %{tab: "trip-view"}}`

thecristen (commented):
A few comments. Mostly good, I think the logic is a little all over the place and I had a few thoughts on places where you could simplify or generalize things.

apps/site/lib/site_web/controllers/schedule/green.ex
alicia-mbta: So in this particular file, there is an existing `call_plug` function that I'm calling and has the same code that the one in `SiteWeb.ControllerHelpers`. I didn't want to remove it from this file but it's true that then we have 2 twin functions. I'm rather have just one but let me know if keeping the original one makes more sense?

apps/site/lib/site_web/views/schedule_view.ex
alicia-mbta: Turns out I realized that we don't even need to create the case for the tab `"trip-view"` because `_trip_view.html.eex` is the view file corresponding to that tab, so no need to pattern-match or check for this.

apps/site/lib/site_web/controllers/schedule/green.ex
thecristen: Ohh I see. I agree with you let's have just one - so let's remove this call_plug and update to use the one in SiteWeb.ControllerHelpers.

apps/site/lib/site_web/views/schedule_view.ex
```diff
@@ -481,4 +481,26 @@ defmodule SiteWeb.ScheduleView do
...
     Route.to_json_safe(route)
   end
+
+  @spec retirement_message(Conn.t()) :: Safe.t()
+  def retirement_message(%Conn{assigns: %{route: %Route{type: 2}}}), do: nil
+
+  def retirement_message(conn) do
+    SiteWeb.ViewHelpers.banner_message(conn, :retirement_message)
+  end
+
```
thecristen, line 491: Your code isn't using this anymore so you can delete this!

apps/site/lib/site_web/controllers/helpers.ex
```diff
@@ -178,4 +178,8 @@ defmodule SiteWeb.ControllerHelpers do
 
     "#{Strftime.format!(one_year_after, "%d %b %Y")} 00:00:00 EST"
   end
+
+  def call_plug_with_opts(conn, module, opts) do
```
thecristen, line 182: I wonder if adding a test for this would help with coverage?

thecristen (approved):
👏

apps/site/lib/site_web/views/schedule_view.ex
alicia-mbta: 🤦‍♀️

apps/site/lib/site_web/controllers/helpers.ex
alicia-mbta: Still no luck 😞 
Any chance we can merge this regardless and perhaps spin up a new PR with the additional tests for Foxboro?

PR #738, opened 2021-01-05
fix: exit Wallaby tests with a non-zero exit code

paulswartz:
They'd been exiting with 0 even if the tests didn't pass.

---

### DIFF
--- a/semaphore/wallaby.sh
+++ b/semaphore/wallaby.sh
@@ -15,3 +15,5 @@ do
     fi
   fi
 done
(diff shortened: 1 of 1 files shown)


paulswartz:
@thecristen assigning this one to you: not sure whether you'll want to keep these Wallaby tests at all once you've made your other changes.

thecristen (approved):
💯

thecristen:
Addressed in prior Github actions work

PR #744, opened 2021-01-08
🔧 Use a script to automate GitHub releases

thecristen:
#### Summary of changes
**Asana Ticket:** [🔧 Use a script to automate GitHub releases](https://app.asana.com/0/385363666817452/1185680283158788/f)

I pretty much used the approach described in the ticket. I thought about using GitHub actions, but this was easier for me to wrap my head around and quick enough to implement for now!

One change to the previous functionality is that it no longer looks at `HEAD` - it'll default to the latest commit on `origin/master`. This made it easier for me to use and I think makes more sense for this case.

It worked to create https://github.com/mbta/dotcom/releases/tag/2021.01.08.1 for me.

### DIFF
--- a/bin/lib/git.py
+++ b/bin/lib/git.py
@@ -7,6 +7,12 @@ def current_branch():
 def clean_status():
     return __call(["git", "status", "--porcelain"]) == ""
 
+def has_gh_cli():
+    try:
+        return __call(["gh", "auth", "status"]) != "You are not logged into any GitHub hosts. Run gh auth login to authenticate."
+    except:
+        return False
+
 def current_commit():
     return __call(["git", "rev-parse", "--short", "HEAD"])
 
@@ -18,6 +24,9 @@ def sha_for_tag(tag):
 
 def log_between(previous, current):
     return __call(["git", "log", "--abbrev-commit", "--no-decorate", "--pretty=oneline", "%s..%s" % (previous, current)])
+
+def release(tag, notes):
+    return __call(["gh", "release", "create", "%s" % tag, "--title", "%s" % tag, "--notes", "%s" % notes])
 
 def __call(cmd):
     return subprocess.check_output(cmd).decode('UTF-8').rstrip()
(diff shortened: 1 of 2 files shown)


bin/lib/git.py
```diff
@@ -1,14 +1,14 @@
...
+    except:
+        return False
 
-def clean_status():
-    return __call(["git", "status", "--porcelain"]) == ""
-
-def current_commit():
-    return __call(["git", "rev-parse", "--short", "HEAD"])
+def current_master_commit():
+    return __call(["git", "rev-parse", "--short", "origin/master"])
```
arkadyan, line 11: Using the origin instead of trying to enforce being up-to-date locally is a good idea.

bin/lib/git.py
```diff
@@ -1,14 +1,14 @@
 import subprocess
 
 
-def current_branch():
-    return __call(["git", "branch", "--show-current"])
+def has_gh_cli():
+    try:
+        return __call(["gh", "auth", "status"]) != "You are not logged into any GitHub hosts. Run gh auth login to authenticate."
```
arkadyan, line 6: `gh` is a new tool to me, personally. It looks great, but it might be worth documenting that it's required to run this script. And maybe even having a prompt to install it if this command fails because it's not found? (For my usage of git I guess I was relying on the assumption that the user wouldn't have this app installed unless they also had git install…)

But also bigger picture, in the long run it would probably better to replace these command line calls with the GitHub API, as mentioned in https://app.asana.com/0/1198965985597278/1198965985597278.

arkadyan (approved):
One small note about `gh`, but this looks good to me.

I don't know if you've had a chance to look at the subtasks https://app.asana.com/0/1198965985597278/1198965985597278 and https://app.asana.com/0/1198968561086741/1198968561086741. It might be worth breaking them out into future tickets of their own or something (and maybe you are already planning on this). They don't need to hold up this PR—this is certainly already a valuable improvement—but I think they would still make this better in the long-term (if you agree).

PR #746, opened 2021-01-11
fix: Daily Schedule table is aligned, even when a route is a School Trip

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Daily Schedule table misaligned when including non-school vacation trips](https://app.asana.com/0/1199335143661136/1199180904718622)

The expandable AccordionRow component accepts a colSpan, which depends on whether the row is for a Commuter Rail or School Trip route, both of which require an additional column.

---

* [x] Does frontend functionality render and work correctly in IE?  -- **Hmmm... best way of getting IE for mac at this point?**
* [x] Are the changes organized into self-contained commits with descriptive and well-formatted commit messages?  -- **I had a commit flub with my first round of commits.  So now there are 3 redundant commits when there only needed to be 2 total.  Should I clean this up?**


alicia-mbta (changes requested):
I'm still seeing the 'gap':<br/>
image

alicia-mbta:
Also since this seems to be small I think you can have just one commit for this fix instead of 5 :)

hannahpurcell:
> I'm still seeing the 'gap':

Oh I see!  I didn't realize there could be rows with and without the S column in one table.  (Makes sense now that I think about it.)  I updated to read from `anySchoolTrips`, which catches all the accordion rows instead of only the rows with an S.

hannahpurcell:
> Also since this seems to be a small change, I think you can have just one commit for this fix instead of 5 :)

Ok great, I was hoping that squashing commits would properly overwrite my previous chaotic commit statements :joy:

alicia-mbta (approved):
Looks good!
As I mentioned in standup, for the error you're seeing on Semaphore, I have the fix in PR #732 so if you rebase after that one you shouldn't see the error anymore (but let me know if otherwise!).

PR #755, opened 2021-01-13
A few fixes to avoid crashes

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [SiteWeb.ScheduleController.FinderApi optimization](https://app.asana.com/0/555089885850811/1199617436773005)

Draft pull request for now in case this file can be further optimized. It contains fixes reported in the subtasks of the ticket.


apps/site/lib/site_web/controllers/schedule/finder_api.ex
```diff
@@ -250,6 +257,10 @@ defmodule SiteWeb.ScheduleController.FinderApi do
   defp prepare_journeys_for_json(journeys) do
     journeys
     |> Enum.map(&destruct_journey/1)
+    |> Enum.filter(fn journey ->
+      get_in(journey, [:departure, :prediction, :time]) != nil ||
+        get_in(journey, [:departure, :schedule, :time]) != nil
```
thecristen, line 262: I think we're ok with journeys where the schedule is nil and the prediction is not (as seen in set_departure_time function)... but those would get filtered out here, right? I'm not so sure filtering here is the best idea.

thecristen (commented):
I'm a bit concerned about removing journeys that we might want to still display... otherwise I think it looks good.

thecristen (approved):
🎉

PR #761, opened 2021-01-21
Use Routes.MockRepoApi when testing

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Configure test env to use Routes.MockRepoApi anywhere we currently call Routes.Repo](https://app.asana.com/0/555089885850811/1189396929591667)

This way we have a mock repo for routes and we don't need to make unnecessary calls.


alicia-mbta:
I think the codecov report might be another "false positive"  😞

thecristen (approved):
I like it! I can manually merge it in to get around Codecov...

PR #764, opened 2021-01-22
fix: Don't focus on the datepicker input on load

thecristen:
#### Summary of changes
**Asana Ticket:** [☎️ Form focus should start at the first form entry](https://app.asana.com/0/1149164109017504/1199512946850968)

The date picker input (which I learned is distinct from the date picker! haha!) had no tests so I added some.

I interpreted the spirit of the ticket to be more "don't focus on the date picker on page load" rather than "must focus on the first input on page load", so that's what I did.

---


alicia-mbta (approved):
So nice that we have nifty `gainFocusOnConstruction` handy!
Thanks for the additional test file  💯

hannahpurcell (approved):
Took it out for a spin, too, and it looks great!

PR #767, opened 2021-01-26
Comments to explain why we need the fixedsticky files

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Tech Debt | Look into removing fixedsticky from front-end](https://app.asana.com/0/555089885850811/1173098961983546)

The investigation task was to decide whether the fixedsticky files were dead code.  They're not dead, but are used so infrequently - mainly just the fares page in IE.  Some comments will help us when we run into this question again.


alicia-mbta (approved):
This is really helpful, thank you!

thecristen (approved):
👍 Great!!

PR #772, opened 2021-01-28
Remove journeys without a prediction nor a schedule

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Elixir.BadMapError: expected a map, got: nil](https://app.asana.com/0/385363666817452/1199209865060987)

As mentioned in [this comment from a previous PR](https://github.com/mbta/dotcom/pull/755#discussion_r562965381), we don't want to discard journeys where the schedule is nil and the prediction is not, but when _both_ are nil we might want to, since they contain no useful information.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/finder_api.ex
+++ b/apps/site/lib/site_web/controllers/schedule/finder_api.ex
@@ -257,10 +257,17 @@ defmodule SiteWeb.ScheduleController.FinderApi do
   defp prepare_journeys_for_json(journeys) do
     journeys
     |> Enum.map(&destruct_journey/1)
+    |> Enum.filter(&journey_has_valid_departure?/1)
     |> Enum.map(&lift_up_route/1)
     |> Enum.map(&set_departure_time/1)
     |> Enum.map(&json_safe_journey/1)
   end
+
+  @spec journey_has_valid_departure?(Journey.t()) :: boolean
+  defp journey_has_valid_departure?(%{departure: departure}),
+    do: !PredictedSchedule.empty?(struct(PredictedSchedule, departure))
+
+  defp journey_has_valid_departure?(_), do: true
 
   # Break down structs in order to use Access functions
   @spec destruct_journey(Journey.t()) :: map
--- a/apps/site/test/site_web/controllers/trip_plan_controller_test.exs
+++ b/apps/site/test/site_web/controllers/trip_plan_controller_test.exs
@@ -7,6 +7,8 @@ defmodule SiteWeb.TripPlanControllerTest do
   import Phoenix.HTML, only: [html_escape: 1, safe_to_string: 1]
   doctest SiteWeb.TripPlanController
 
+  import Mock
+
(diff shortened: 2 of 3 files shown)


thecristen (commented):
Thanks!!

I noticed that in the `json_safe_journey` function used in the finder_api.ex line 265, it seems to do something similar to what you've added here. Can you see whether this is redundant to `clean_schedule_or_prediction`? Or maybe is there a reason it's necessary to do this step here, before `json_safe_journey` works its magic?

apps/site/lib/site_web/controllers/schedule/finder_api.ex
```diff
@@ -257,11 +257,19 @@ defmodule SiteWeb.ScheduleController.FinderApi do
...
     |> Enum.map(&json_safe_journey/1)
   end
 
+  @spec journey_has_valid_departure?(Journey.t()) :: boolean
+  defp journey_has_valid_departure?(%{departure: departure}) do
+    departure != %{prediction: nil, schedule: nil}
+  end
+
+  defp journey_has_valid_departure?(_), do: true
+
```
thecristen, line 272: Pattern matching might work in your favor here!

```suggestion
  defp journey_has_valid_departure?(%{departure: %{prediction: nil, schedule: nil}}), do: false
  defp journey_has_valid_departure?(_), do: true

```

thecristen: Digging around some more I found `PredictedSchedule.empty?` which looks like it could be used here instead. Maybe something like

```elixir
defp journey_has_valid_departure?(%{departure: nil}), do:  true
defp journey_has_valid_departure?(%{departure: departure}), do:  !PredictedSchedule.empty?(departure)
```

apps/site/test/site_web/controllers/trip_plan_controller_test.exs
```diff
@@ -388,6 +390,28 @@ defmodule SiteWeb.TripPlanControllerTest do
              end)
     end
 
+    test "returns all nil fares when there is not enough information", %{conn: conn} do
+      # force legs to have incomplete information:
+      with_mock TripPlan.Leg, [:passthrough],
+        is_fare_complete_transit_leg?: fn _leg ->
+          false
+        end do
```
thecristen, line 398: Oh this is really clever 🤦‍♀️ Thanks so much for adding this

thecristen (approved):
A little nitpicky suggestion but it's good!

apps/site/lib/site_web/controllers/schedule/finder_api.ex
alicia-mbta: Wow, that's exactly what we need. Keeping it DRY...

thecristen:
Can you rebase this?

alicia-mbta:
Sure; I didn't initially because the test I added is unrelated to the work done in this branch, I just added it to keep Codecov happy :)

thecristen:
Not seeing any test failures unique to this branch, so I'm going to go ahead and merge.

PR #778, opened 2021-02-05
Show fare zones only for commuter rail

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Schedules | Fare Zones should not show for bus, ferry, and subway](https://app.asana.com/0/555089885850811/1199882294203912)

In the Stations page we won't be showing fare zones. Before:
image

After:
image

In the schedule finder modal, we won't be showing zones unless it's a commuter rail route. E.g. for buses:
image
For commuter rail:
image

Same for a specific stop:
Commuter rail:
image

Bus before:
image

Bus now:
image

### DIFF
--- a/apps/site/lib/site_web/templates/stop/_detailed_stop.html.eex
+++ b/apps/site/lib/site_web/templates/stop/_detailed_stop.html.eex
@@ -4,9 +4,4 @@
     <%= if assigns[:include_route_icon?], do: stop_feature_icon(@detailed_stop.route_icon, :small) %>
     <%= feature_icons(@detailed_stop)%>
   </div>
-  <div class="m-detailed-stop__zone">
-    <%= if @detailed_stop.zone do %>
-      <span class="commuter-rail-zone">Zone <%= @detailed_stop.zone %></span>
-    <% end %>
-  </div>
 </a>
--- a/apps/stops/lib/api.ex
+++ b/apps/stops/lib/api.ex
@@ -420,5 +420,11 @@ defmodule Stops.Api do
 
   @spec get_zone_number(String.t() | nil) :: String.t() | nil
   defp get_zone_number(nil), do: nil
-  defp get_zone_number(zone), do: String.trim_leading(zone, "CR-zone-")
+
+  defp get_zone_number(zone) do
+    case zone do
+      "CR-zone-" <> zone_number -> zone_number
+      _ -> nil
+    end
+  end
 end
--- a/apps/site/lib/site_web/templates/stop/_hub_stops.html.eex
+++ b/apps/site/lib/site_web/templates/stop/_hub_stops.html.eex
@@ -10,11 +10,6 @@
               <%= link detailed_stop.stop.name, to: stop_path(@conn, :show, detailed_stop.stop.id), class: "hub-stop-name" %>
             <% end %>
             <span class="stop-features-list"><%= feature_icons(detailed_stop)%></span>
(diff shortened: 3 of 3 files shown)


apps/site/lib/site_web/templates/stop/_detailed_stop.html.eex
```diff
@@ -4,7 +4,9 @@
     <%= if assigns[:include_route_icon?], do: stop_feature_icon(@detailed_stop.route_icon, :small) %>
     <%= feature_icons(@detailed_stop)%>
   </div>
-  <div class="m-detailed-stop__zone">
+  <%# For now we will not be showing fare zones since it doesn't help riders %>
+  <%# (reusing class .m-trip-plan__hidden for hiding) %>
```
thecristen, line 8: Why hide rather than delete?

apps/site/lib/site_web/templates/stop/_detailed_stop.html.eex
```diff
@@ -4,7 +4,9 @@
...
+  <div class="m-trip-plan__hidden m-detailed-stop__zone">
```
thecristen, line 9: Bootstrap comes with utility classes, I think `hidden-xs-up` will hide it here.
(Admittedly the Bootstrap documentation for our version (v4.0.0-alpha.2) is hard to find, but the source code is [here](https://github.com/twbs/bootstrap/tree/v4.0.0-alpha.2/scss))

thecristen (commented):
I feel partially responsible for this issue since I adjusted how we get the zones in the first place. :\ And I had no idea we'd get weird zone names for buses and such...

All of this front end adjustment feels a little clunky to me, and future devs will always need to know to make similar adjustments if they show the zone anywhere else later on, which seems brittle. It'd be nice to just ensure all the stop zones we have are fit for display....

QUESTION: Have you considered adjusting how we get the zone in the backend? 

[We already do a tiny bit of parsing to remove "CR-zone-" before saving the zone here](https://github.com/mbta/dotcom/blob/master/apps/stops/lib/api.ex#L414). It wouldn't be too much of a stretch to do a little more work there to parse and fix up the zone number, and only output the number if the stop is served by commuter rail.

alicia-mbta:
You're right, I think it's more straight-forward to do it on the Elixir side. I have also removed the zone real estate instead of hiding it (we can always go back in time). Now it's all just a small commit.

apps/stops/lib/api.ex
```diff
@@ -420,5 +420,14 @@ defmodule Stops.Api do
...
+
+  defp get_zone_number(zone) do
+    is_ferry_or_bus =
+      zone |> String.starts_with?("Boat-") or zone in ["ExpressBus-Downtown", "LocalBus"]
+
+    cond do
+      is_ferry_or_bus -> nil
+      true -> String.trim_leading(zone, "CR-zone-")
+    end
+  end
```
thecristen, line 432: Would it make sense to just check if it's a commuter rail since I think that's the only case we want the zones for? Then we're covered for whatever new modes or zones show up...

Maybe like (no idea if this actual works)

```elixir
case zone do
  "CR-zone-" <> zone_number -> zone_number
  _ -> nil
end
```

thecristen (changes requested):
Looking good! I'm glad we can check the strings in a useful way. Unless we think there's other zones we need to account for, let's check for commuter rail only.

thecristen (approved):

PR #780, opened 2021-02-05
Update Typescript

thecristen:
Asana ticket https://app.asana.com/0/385363666817452/1199918165033552/f

In my recent messing with CI and other tools, I ran into a mystery error around Typescript versions, and since we've been updating lots of dependencies lately... I decided to try my hand at updating Typescript and related libraries. And I did it by hand rather than letting Dependabot try it.

This required updating our ESLint config a fair bit, I suppose a bunch of the default rules had changed.  I tweaked the config to be as accommodating of our current code as possible, but I did also fix a few linting errors.


apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/ScheduleTable.tsx
```diff
@@ -74,9 +74,7 @@ const ScheduleTable = ({
       <table className="schedule-table">
         <thead className="schedule-table__header">
           <tr>
-            {anySchoolTrips && (
-              <th scope="col" className="schedule-table__cell" />
-            )}
+            {anySchoolTrips && <td className="schedule-table__cell" />}
```
paulswartz, line 77: Is this right? should we be mixing `td` and `th`?

thecristen: Actually, yes! :D That's only because it's an empty cell we use for spacing purposes -- [table headers should never be empty, but empty table cells, even in the header, are ok](https://www.w3.org/TR/WCAG20-TECHS/H51.html).

paulswartz: learn something new every day!

apps/site/assets/.eslintrc.js
```diff
@@ -18,7 +18,7 @@ const baseConfig = {
     "react/require-default-props": "off",
     "prefer-object-spread": "off",
     "react/jsx-props-no-spreading": "off",
-    "react/no-unused-prop-types": "warn",
+    "react/no-unused-prop-types": "off",
```
alicia-mbta, line 21: why the switch from `warn` to `off`? Too much noise?

thecristen: Thanks for flagging... looking at it again it's only five errors, maybe it's worth fixing manually so that I can leave this as "warn" 🤔

thecristen: @alicia-mbta I've restored it back to "warn" and updated a few components so that it would pass.

alicia-mbta (approved):
Looks good!

paulswartz (approved):
🍰

thecristen:
It turns out I created a compilation error... through which I found out that the Typescript plugin for ESLint does not check that our app actually _compiles_ correctly. I added a new npm script to let us check that easily, and another commit to correct the error I made. :)

PR #782, opened 2021-02-09
Shell script to serve Dotcom with mock data

thecristen:
Asana ticket: https://app.asana.com/0/385363666817452/1199918165033543/f

Also deprecates a few barely-used NPM scripts

Most of the new script was adapted from the `semaphore/backstop.sh` script (which is currently unused and I'm hoping to deprecate soon).

## Why?

The intent is to use this script to run Dotcom, so that _other_ scripts can run tests against the resulting running application. Two use cases approaching are visual testing and performance testing, which require a running version of the application to take screenshots on or computer metrics for via headless browser testing.

## How?
I'm guessing the implementation should fall in one of two approaches:

1. Run as its own step in CI before the relevant test script, e.g.
image

2. Run **by** the test script before running its tests, e.g.
```bash
# this doesn't work - will run subsequent commands before server is ready
bash ./semaphore/run_mock_server.sh &
npm run screenshots
```

## Feedback needed

In addition to refinement and error checking in the new script itself, I have other questions...

* Best way to run in the background so that other tasks can run?
* How to run via another script, but also make sure to clean up its running processes on close?


paulswartz:
> In addition to refinement and error checking in the new script itself, I have other questions...
> 
>     * Best way to run in the background so that other tasks can run?
> 
>     * How to run via another script, but also make sure to clean up its running processes on close?

Can you have this take a command to run as its argument?

```bash
semaphore/run_mock_server.sh npm run screenshots
```

If you wanted it to be interactive:
```bash
semaphore/run_mock_server.sh bash
```

It *might* be easier to manage this as a Python script, where you can use the [subprocess module](https://docs.python.org/3/library/subprocess.html) to ensure that the processes you start are properly torn down. At least, I'm more familiar with than than I am with the vagaries of Bash scripting.

thecristen:
Hm, I could probably have it take an argument, but I'm not sure I want to. Maybe if it were optional. And even if I did pass in a command... I'm still not sure how I'd invoke the commands to run the way I need 😛

semaphore/run_mock_server.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
+set -e
+
+#===============================================================================
+# A standalone script for running Dotcom with mock data.
+# Shold work both locally and on CI.
```
ryan-mahoney, line 6: Shold -> Should

semaphore/run_mock_server.sh
```diff
@@ -0,0 +1,113 @@
...
+  exit "${error_code}";
+}
+trap cleanup ERR EXIT SIGINT SIGTERM
+
+#===============================================================================
+# Setting up Wiremock
+#===============================================================================
+
+export WIREMOCK_PROXY=true
+export WIREMOCK_PROXY_URL="${WIREMOCK_PROXY_URL:=https://dev.api.mbtace.com}"
```
ryan-mahoney, line 34: at some point, we were trying to keep the dev URLs out of the code... it may make sense to instead of defaulting, check for the existence of the variable and exit with a message if it is not present

thecristen:
With a little more tweaking, it works! It turns out I need to `source` the script to best use it, e.g.:

```shell
#!/bin/bash
set -e

echo "Going to start the server script"

source ./semaphore/run_mock_server.sh

echo "Completed the server script and am ready to do other things"

sleep 10
# exits everything gracefully!
```

PR #784, opened 2021-02-10
chore: Update test for CR-Fitchburg shuttle stop

thecristen:
The newly added shuttle for CR-Fitchburg has Alewife as a stop, which wasn't accounted for in our test. Let's update this value to match so CI can finally pass.


alicia-mbta (approved):
🙏

PR #795, opened 2021-02-11
Render line diagram and map on HH Ferry page

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Schedules | Hingham/Hull Ferry Page is Blank](https://app.asana.com/0/555089885850811/1199891028022261)

The `by_typicality` filter used in `helpers.ex` forced routepatterns to have a typicality 1.  In the case of the HH ferry, all routepatterns were of typicality 2.    This work is two-fold:

1. The typicality filter was replaced with `filter_by_min_typicality`, which finds the lowest typicality among the routepatterns and uses that routepattern, instead of forcing a value of 1.
2. The TRC will need to change the Long-Hingham-Hull routepattern for the HH ferry to have a typicality of 1, since that is the primary route we want rendered in the line diagram

Ferry page:  https://www.mbta.com/schedules/Boat-F1/line

---


apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -80,10 +80,36 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
   end
 
+  @spec filter_by_min_typicality([RoutePattern.t()]) :: [RoutePattern.t()]
+  defp filter_by_min_typicality(route_patterns) do
+    route_patterns
+    |> Enum.reduce({nil, []}, fn
+      %RoutePattern{typicality: typicality} = r, {t, a} when typicality < t -> {typicality, [r]}
+      %RoutePattern{typicality: typicality} = r, {typicality, a} -> {typicality, [r | a]}
+      _, acc -> acc
+    end)
```
thecristen: I'll admit I'm _still_ trying to wrap my head around how this works, but I think I understand what it's supposed to do? Kind of like a mix of Enum.group_by typicality and sorting on whatever the lowest typicality is?

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -80,10 +80,36 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
+    |> temporary_hh_ferry_only_func(route.id)
     |> Enum.map(&stops_for_route_pattern/1)
   end
 
+  @spec filter_by_min_typicality([RoutePattern.t()]) :: [RoutePattern.t()]
+  defp filter_by_min_typicality(route_patterns) do
+    route_patterns
+    |> Enum.reduce({nil, []}, fn
+      %RoutePattern{typicality: typicality} = r, {t, _a} when typicality < t -> {typicality, [r]}
+      %RoutePattern{typicality: typicality} = r, {typicality, _a} -> {typicality, [r | _a]}
```
thecristen, line 95: I'll admit I'm still trying to figure out what it's doing, I think something like grouping by typicality while also sorting by the lowest? It might be worth pulling this out into a separate function for clarity, or maybe using some more Enum functions like Enum.group_by, sort_by

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -80,10 +80,36 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
   defp do_get_branch_route_stops(route, direction_id, route_pattern_id) do
     route.id
     |> get_route_patterns(direction_id, route_pattern_id)
-    |> Enum.filter(&by_typicality(&1, route_pattern_id))
+    |> filter_by_min_typicality()
```
thecristen, line 83: I'm wondering if it might make sense to account for the ferry here? Like what if we did
```
defp filter_by_min_typicality(route_patterns, route_pattern_id, route)
  if the route is a ferry do
     # invoke temporary_hh_ferry_only_func here
  else
    # proceed normally
     Enum.filter(route_patterns, &by_typicality(&1, route_pattern_id))
  end
end)
```

thecristen (commented):
This is a great start! I've got a few little suggestions here.

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
thecristen: Whoops sorry for the duplicate comment, I had thought Github disappeared this one

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -80,10 +80,36 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
+      %RoutePattern{typicality: typicality} = r, {typicality, a} -> {typicality, [r | a]}
+      _, acc -> acc
+    end)
+    |> elem(1)
+    |> Enum.reverse()
+  end
+
+  @spec filter_by_min_typicality([RoutePattern.t()]) :: [RoutePattern.t()]
+  defp temporary_hh_ferry_only_func(route_patterns, id) do
+    if id == "Boat-F1" do
```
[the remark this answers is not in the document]
hannahpurcell: Ok!  I would guess we'll remove this code shortly, and I'm not sure how other ferries would be structured, so I didn't want to assume this should apply to all ferries.  But I'd guess other ferries would behave similarly, right?

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -80,10 +80,36 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
+      route_patterns
+      |> hd()
+      |> (fn a -> [a] end).()
```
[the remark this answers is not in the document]
hannahpurcell: The output of hd() isn't an enumerable - it's just a single stop.  Is there a simple way to toss it into an array?

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -80,10 +80,36 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
+      %RoutePattern{typicality: typicality} = r, {t, a} when typicality < t -> {typicality, [r]}
+      %RoutePattern{typicality: typicality} = r, {typicality, a} -> {typicality, [r | a]}
+      _, acc -> acc
+    end)
+    |> elem(1)
+    |> Enum.reverse()
+  end
+
+  @spec filter_by_min_typicality([RoutePattern.t()]) :: [RoutePattern.t()]
+  defp temporary_hh_ferry_only_func(route_patterns, id) do
```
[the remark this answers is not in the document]
hannahpurcell: Ah, great point!  Hmm... I'm not exactly sure what should be done if the array is empty.  That would mean for a given route id, no route patterns were found.  What do we do in that case?  It doesn't look like there was previously a case for that, which I'd be able to mimic here

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
hannahpurcell: Yeah, you're right - iterating over the list, if a typicality is found that's less than the "t" typicality currently in the accumulator, then the list of "lowest typicality route patterns" currently in the accumulator will be ditched and replaced by this new route pattern that was just found.  I should expand those variables beyond just r / t / a, but the lines were getting verbose

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
hannahpurcell: Ohhh hmmm.... Yes, I should pull it out into a function.  I'm curious about using other Enum grouping / sorting options - let's see what I can stitch together

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
hannahpurcell: Ohhhh, for other routes besides ferry, we should be filtering by typicality == 1?

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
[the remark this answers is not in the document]
thecristen: Dang I was thinking the `RoutePattern%{}` struct would be an enumerable. In that case I'm not sure I can think of a way that's any simpler than what you have already, but something like this might read a bit clearer? 🤷‍♀️ 
```Elixir
[first_route_pattern | _ ] = route_patterns
[first_route_pattern]
```

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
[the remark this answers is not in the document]
thecristen: 🤦‍♀️ Ah my bad... let's not worry about the empty case for now then

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -80,10 +80,36 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
+    if id == "Boat-F1" do
```
paulswartz, line 104: Rather than an `if` statement, you can pattern match in the function head:
```elixir
defp temporary_hh_ferry_only_func([route_pattern | _], "Boat-F1") do
  [route_pattern]
end
defp temporary_hh_ferry_only_func(route_patterns, _route_id) do
  route_patterns
end
```

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
thecristen: I'm mostly hesitant to change how route patterns are selected for _every_ single route, especially if we think the problem we are fixing is pretty isolated to this ferry.. 😬

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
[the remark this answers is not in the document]
paulswartz: In the future, `List.wrap` does what you want:
```elixir
iex> List.wrap(5)
[5]
iex> List.wrap(nil)
[]
```

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
hannahpurcell: Ok!  I'll account for this `if-else` in the parent method then, so that "filter_by_min_typicality" only does one type of action

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
hannahpurcell: Perhaps I could use grouping, but it's not exactly what I'm doing here:  these lines populate an array to keep track of the lowest-typicality routePatterns.  I've expanded it into a separate function, but not sure I made it any clearer?

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
hannahpurcell: Oh so clean!  I restructured `do_get_branch_route_stops` to contain the if-else, so the matching option went away, I think.   But great to know for next time

thecristen (changes requested):
This is great!! 🥳 

I think it'd be great if you could add a test for this specific ferry case. Please add a test under [describe get_branch_route_stops/3](https://github.com/mbta/dotcom/blob/master/apps/site/test/site_web/controllers/schedule/line/helpers_test.exs#L44) for it. 
There are a few existing tests there that you can copy for inspiration - I'd say the `test "handles the E line with the Lechmere shuttle"` test is likely the most similar, except you'd be testing the `Boat-F1` id and the result will probably have a `nil` branch name. Additionally, I don't think you need to bother asserting the presence of specific stops, but we can assert that at least there ARE stops returned.

Hoping this makes sense? Happy to share testing tips offline if you'd like -- overall, nicely done on this first bit of Elixir work!

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -78,12 +78,42 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
+  @spec reduce_by_min_typicality(RoutePattern.t(), {integer, [RoutePattern.t()]}) ::
+          {integer, [RoutePattern.t()]}
+  defp reduce_by_min_typicality(route_pattern, acc) do
+    %RoutePattern{typicality: typicality} = route_pattern
+    {min_typicality, patterns_array} = acc
+
+    cond do
+      typicality < min_typicality -> {typicality, [route_pattern]}
+      typicality == min_typicality -> {min_typicality, [route_pattern | patterns_array]}
+      true -> acc
```
hannahpurcell, line 113: Hmmm, @alicia-mbta and @thecristen, I have a Codecov question.  The bot only left a single comment, and it's about this line.  Currently, this reduce function should only be used on the HH Ferry, but the HH Ferry never reaches this line... so, should I create a really granular test just so that this line gets covered?

thecristen: As a team we're flexible on this, especially since CodeCov can be kinda wonky sometimes. So thanks for asking! 😀 Adding a test for this case seems more trouble than it's worth IMO. So don't worry about it.

Let me know when you're ready to merge and I'll override the 'failing' coverage and merge it for you.

hannahpurcell: Ah, thanks Cristen!  Makes sense to me 😄 

This should be ready to merge, then!  The "changes requested" should all be addressed, and the codecov is the only failing check

thecristen (approved):

PR #796, opened 2021-02-11
Show the Silver Line icon for South Station instead of the bus one

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 Stations/Stops | South Station should have SL icon](https://app.asana.com/0/555089885850811/1163960794131366)

Show the silver line for South Station.
Before:
image

After:
image
<br/>
image

### DIFF
--- a/apps/stops/lib/repo.ex
+++ b/apps/stops/lib/repo.ex
@@ -160,13 +160,8 @@ defmodule Stops.Repo do
     |> Keyword.get(:connections)
     |> get_stop_connections(stop_id)
     |> Enum.map(icon_fn)
-    |> Enum.map(&consolidate_silver_line_route_feature/1)
     |> Enum.uniq()
   end
-
-  @spec consolidate_silver_line_route_feature(atom) :: atom
-  def consolidate_silver_line_route_feature(:silver_line), do: :bus
-  def consolidate_silver_line_route_feature(icon_atom), do: icon_atom
 
   @spec get_stop_connections([Route.t()] | {:error, :not_fetched} | nil, Stop.id_t()) ::
           [Route.t()]
--- a/apps/site/test/site_web/controllers/schedule_controller_test.exs
+++ b/apps/site/test/site_web/controllers/schedule_controller_test.exs
@@ -302,7 +302,7 @@ defmodule SiteWeb.ScheduleControllerTest do
       # includes the stop features
       assert List.last(stops).stop_features == [
                :red_line,
-               :bus,
+               :silver_line,
                :commuter_rail,
                :access,
                :parking_lot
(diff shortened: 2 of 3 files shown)


apps/stops/lib/repo.ex
```diff
@@ -160,14 +160,9 @@ defmodule Stops.Repo do
     |> Keyword.get(:connections)
     |> get_stop_connections(stop_id)
     |> Enum.map(icon_fn)
-    |> Enum.map(&consolidate_silver_line_route_feature/1)
     |> Enum.uniq()
   end
 
-  @spec consolidate_silver_line_route_feature(atom) :: atom
-  def consolidate_silver_line_route_feature(:silver_line), do: :bus
-  def consolidate_silver_line_route_feature(icon_atom), do: icon_atom
```
hannahpurcell, line 169: I'm trying to figure out, why would this function have been needed in the first place?

thecristen:
I'm deploying this to dev green so I can see if any other pages are affected unexpectedly. :)

apps/stops/lib/repo.ex
thecristen: OK I got curious and decided to look at our prior repo, dotcom-legacy (from before the project was open sourced). It turns out this function got added with [this pull request, "make silver line silver"](https://github.com/mbta/dotcom-legacy/pull/2788) :) 

Not sure if there's anything relevant to this PR but there's lotsa interesting discussion in that PR :D

alicia-mbta: Ok so 2 things:

- If I understand that PR discussion correctly, they did want to replace the SL icon with the bus to avoid inconsistencies. I feel like keeping the function I deleted should be the way to go then.
BUT I also like seeing the gray icon in the stops page for stops that are "silver line stops". What are your thoughts on creating an "exception function" for those? I believe that'd be only for South Station and Airport but I'm not sure if that's the right approach.

- I also noticed there is an inconsistency for the Airport stop (which can be addressed separately):
image
image

thecristen: @alicia-mbta It looks to me like your PR fixed that inconsistency, maybe I'm looking in the wrong place? https://green.dev.mbtace.com/stops/subway vs https://green.dev.mbtace.com/stops/place-aport?

thecristen: As to your first point, I like seeing the grey icon too... I'm not sure exactly which inconsistencies our predecessors were concerned about, but it seems like it might no longer matter. After all, we now definitely want to see the SL icon instead of the bus icon.

Honestly after poking around the site a bit I think deleting that function did the job great 👍 but we can let the product folks review and see if they find anything amiss too.

thecristen (approved):
🚌

apps/stops/lib/repo.ex
alicia-mbta: Perfect, thanks a lot for the poking around and the feedback!

PR #801, opened 2021-02-12
Put all the months on the same page (show all events for the year)

thecristen:
#### Summary of changes
**Asana Ticket:** [Put all the months on the same page (show all events for the year)](https://app.asana.com/0/385363666817452/1199915858612058/f)

Removes the month navigation header entirely.
I thought about removing/replacing the `EventDateRange` functions that work with months, but I figure we'll want them for the calendar view later on so I left them in. 

image

---

### DIFF
--- a/apps/site/lib/site_web/query_builders/event_date_range.ex
+++ b/apps/site/lib/site_web/query_builders/event_date_range.ex
@@ -8,6 +8,13 @@ defmodule SiteWeb.EventDateRange do
     end
   end
 
+  def build(%{"year" => year}, current_date) do
+    case Date.from_iso8601(year) do
+      {:ok, date} -> for_year(date)
+      {:error, _error} -> for_year(current_date)
+    end
+  end
+
   def build(_month_missing, current_date) do
     for_month(current_date)
   end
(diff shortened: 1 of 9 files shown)


apps/site/lib/site_web/query_builders/event_date_range.ex
```diff
@@ -8,8 +8,15 @@ defmodule SiteWeb.EventDateRange do
...
   end
 
+  def build(%{"year" => year}, current_date) do
+    case Date.from_iso8601(year) do
+      {:ok, date} -> for_year(date)
+      {:error, _error} -> for_year(current_date)
+    end
+  end
+
   def build(_month_missing, current_date) do
```
hannahpurcell, line 18: Now that this function has been diversified to work for both month and year, should the "_month_missing" be made generic?  (Doesn't quite matter since it isn't read)

apps/site/lib/site_web/views/event_view.ex
```diff
@@ -20,8 +20,15 @@ defmodule SiteWeb.EventView do
   def calendar_title(month), do: name_of_month(month)
 
   @spec no_results_message(String.t()) :: String.t()
-  def no_results_message(month) do
-    "There are no events in #{name_of_month(month)}."
+  def no_results_message(year) do
+    "There are no events in #{name_of_year(year)}."
+  end
```
hannahpurcell, line 25: I believe we'll need a `no_results_message` function for both month and year.  Go to http://localhost:4001/events?month=2021-11-01, and it will say "There are no events in 2021." even though there are only no events in November.  OR will we redirect when a user goes to a month without events?

thecristen: Good question... Since we don't have much detail for how to handle empty states right now, I think I'm just going to remove this function completely for now.

hannahpurcell (approved):
Looks great, Cristen!

PR #804, opened 2021-02-17
Event listing redesign

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Redesign current event listings](https://app.asana.com/0/555089885850811/1199915858612051)

Mostly style updates, except for the time-range rendering functions.  The functions used to render the event date came from `cms_view.ex`.  For the most part, the logic needed to stay the same - but the joining "at" needed to be replaced by a bullet " • ".  So I added a new parameter `format_type` to inform the string formation and also the date formatter used from `time_helpers.ex`

---


apps/site/lib/site_web/views/helpers/time_helpers.ex
```diff
@@ -1,9 +1,25 @@
...
+  @type format_type :: atom()
+
+  @doc "Returns a string with the month, day and year, according to format type"
+  @spec format_date(DateTime.t() | Date.t(), format_type() | nil) :: String.t()
   @spec format_date(DateTime.t() | Date.t()) :: String.t()
-  def format_date(date) do
+  def format_date(date, format_type \\ nil)
+  
+  def format_date(date, :event) do
+    Timex.format!(date, "{WDshort}, {Mshort} {D}, {YYYY}")
```
alicia-mbta, line 12: Can we make sure `Timex.format` returns a valid formatted date?

apps/site/lib/site_web/views/helpers/time_helpers.ex
```diff
@@ -1,9 +1,25 @@
...
   end
+
+  @doc "Returns a string with the month"
+  @spec format_month(DateTime.t() | Date.t()) :: String.t()
+  def format_month(date), do: Timex.format!(date, "{Mshort}")
+
+  @doc "Returns a string with the day"
+  @spec format_day(DateTime.t() | Date.t()) :: String.t()
+  def format_day(date), do: Timex.format!(date, "{D}")
 end
```
alicia-mbta, line 25: Can we add a test for these functions?

alicia-mbta (changes requested):
Looks good to me but I have one design question (perhaps we need Yinka's input). I saw this particular event:
image
And I was wondering if the 'circles' need to be vertically centered?

hannahpurcell:
> image
> And I was wondering if the 'circles' need to be vertically centered?

👍 Thanks for the note!  Looks like it needs to be top-aligned:  https://zeroheight.com/2fedee66c/p/9010eb-unique-components/i/29969884

apps/site/assets/css/_events.scss
```diff
@@ -40,3 +42,30 @@
...
+  justify-content: center;
+  align-items: center;
+  margin-right: 1.5rem;
+  text-transform: uppercase;
+  color: $gray;
+  @include media-breakpoint-down(xs) {
+    display: none;
+  }
+}
+.m-date-range {
```
thecristen, line 62: Should this be `.m-event__date-range`?

apps/site/assets/css/core.scss
```diff
@@ -272,3 +273,10 @@ div {
     }
   }
 }
+
+.l-flex-container {
+  display: flex;
+}
+.l-flex-one {
+  flex: 1;
+}
```
thecristen, line 282: Let's either make these into utility classes (move to `_utilities.scss` and prefix with `u-`) _or_ move these styles to the `_events.scss` file (+ renaming the classes to use the `m-event__` convention).

apps/site/assets/css/_events.scss
```diff
@@ -40,3 +42,30 @@
 .event-inline-block {
   display: inline-block;
 }
+
+.m-event__date-circle {
```
thecristen, line 46: If you added the `font-weight: 700` here then you wouldn't need to adjust the existing small CSS.

apps/site/lib/site_web/templates/event/_event_list.html.eex
```diff
@@ -7,12 +7,18 @@
         range = %{start: event_teaser.date, stop: event_teaser.date_end}
         path_fn = fn -> cms_static_page_path(@conn, event_teaser.path) end
       %>
-      <li class="list-group-item">
-        <div>
-          <%= render_duration(range.start, range.stop) %>
+      <li class="list-group-item l-flex-container">
+        <div class="m-event__date-circle">
+          <small><%= format_month(range.start) %></small>
+          <h3 class="m-event__day"><%= format_day(range.start) %></h3>
```
thecristen, line 13: Why h3? I'd rather avoid using headers for these items..

apps/site/assets/css/_events.scss
```diff
@@ -40,3 +42,30 @@
...
+  display: flex;
+  flex-direction: column;
+  height: 4rem;
```
thecristen, line 49: We have a Sass variable called `$base-spacing` which if I recall correctly is equivalent to `1rem` -- can you adjust these styles to make use of this variable?

apps/site/assets/css/_events.scss
```diff
@@ -40,3 +42,30 @@
...
+  align-items: center;
+  margin-right: 1.5rem;
+  text-transform: uppercase;
+  color: $gray;
+  @include media-breakpoint-down(xs) {
+    display: none;
+  }
+}
+.m-date-range {
+  font-size: .88rem;
```
thecristen, line 63: This font-size is suuuuuper close to `$font-size-small-caps` - would it make sense to use that one here?

apps/site/lib/site_web/views/helpers/time_helpers.ex
```diff
@@ -1,9 +1,29 @@
...
-  @doc "Returns a string with the full month, day and year."
+  @type format_type :: atom()
+
+  @doc "Returns a string with the month, day and year, according to format type"
+  @spec format_date(DateTime.t() | Date.t(), format_type() | nil) :: String.t()
   @spec format_date(DateTime.t() | Date.t()) :: String.t()
-  def format_date(date) do
+  def format_date(date, format_type \\ nil)
+
+  def format_date(date, :event) do
```
thecristen, line 14: Hmmm. So I think if this is going to include logic pertaining to events, then it should be in a file more specific to that than this one.

However, I'm also wondering about a possibly more generalized approach that I've seen used in `SiteWeb.ViewHelpers.pretty_date()`:

```elixir
def pretty_date(date, format \\ "{Mshort} {D}") do
    if date == Util.service_date() do
      "today"
    else
      Timex.format!(date, format)
    end
  end
```

 What if we had a helper function where the arguments can dictate the format? Then you wouldn't need separate format_month or format_day, it could just be `format_event_date(date, "{Mshort}")` (for example)

thecristen (changes requested):
Wonderful! I've got a few assorted comments here.

apps/site/assets/css/_events.scss
hannahpurcell: Yeahh, that's part of why I was asking about how much we're supposed to match the ZeroHeight rules.

[The event listing anatomy](https://zeroheight.com/2fedee66c/p/9010eb-unique-components/i/29969884) has the text in the circle represented by <small>.  [And the typography definition of <small>](https://zeroheight.com/2fedee66c/p/37413e-typography/b/83bfe5) has font-weight 700.  

The only other place "small" is used is on the "Contact Us" page.  This is what it looks like if <small> has font-weight 700.  Doesn't look quite right.  So should I change this to another non-small element, or should <small> be defined differently in our typography system?

apps/site/lib/site_web/templates/event/_event_list.html.eex
hannahpurcell: Was matching the ZeroHeight again.  Looks like I should start a thread in slack about our typography system

thecristen:
Forgot to mention this earlier -- we're going to develop each milestone for the event page work on a feature branch -- please adjust this PR to merge into the `feature/event-listings-redesign` branch.

apps/site/lib/site_web/views/helpers/time_helpers.ex
thecristen: I can also see the value in the less general approach, where we have functions just to format dates for events, but I'd like to see those in the event view rather than this more general file. 
So, up to you.. There are really so so many ways to implement a thing. 😩

apps/site/assets/css/core.scss
hannahpurcell: Ok great!  I think a simple, universally accessible flex classes could be useful in the future.  Is that the idea of the utility classes?

thecristen: Yep, exactly 😄

hannahpurcell:
> Forgot to mention this earlier -- we're going to develop each milestone for the event page work on a feature branch -- please adjust this PR to merge into the `feature/event-listings-redesign` branch.

Oops!!  Thanks for noticing!

apps/site/lib/site_web/views/helpers/time_helpers.ex
hannahpurcell: Yeah, so true!!  I can see why event-specific logic should hang out in the event view.  And I ❤️ the pretty_date function, now that I know it exists.

I'm trying to think of a good way to extract _or_ extrapolate the relevant functions from the `cms_view` file, too...  The event view reaches out to the cms view for the `render_duration` function, which then reaches out to the time helpers file for the actual Timex calucaltions.  So hm... thinking through the best way to organize this all 
 in the events view without duplicating code

hannahpurcell: You know what, I don't think it'd be a crazy messy to just create a `render_event_duration` func in the event view.  Imma do it 😁

hannahpurcell:
Ugh, I made a mistake.  When a random API test was failing, I rebased with `master` instead of `feature/event-listings-redesign` as I should have.  Soooo now the codecov bot is flagging me again, and the number of changed files has shot up unnecessarily

alicia-mbta:
Not sure if related, but if you're running into test issues, I fixed a couple in https://github.com/mbta/dotcom/pull/810/commits/df3936ffde2273facfa0974a42a4697ed6030dae.
That might help fix the issues you're having testing-wise.

hannahpurcell:
> Not sure if related, but if you're running into test issues, I fixed a couple in [df3936f](https://github.com/mbta/dotcom/commit/df3936ffde2273facfa0974a42a4697ed6030dae).
> That might help fix the issues you're having testing-wise.

I had to fix a generic green-line test.  Not sure why it started failing 🤔 , but all the tests are working now -- only stumped by the codecov bot :sob:

PR #806, opened 2021-02-18
Fixing various broken line diagram issues

thecristen:
#### Summary of changes
**Asana Ticket:** [Line Diagram | Shuttles should not break the line](https://app.asana.com/0/385363666817452/1199904083874239/f)

There are actually two different problems being tackled here. Going to deploy on dev green to make it easier to check many routes with the green dev API data.

### First commit

CR-Kingston was broken because the line diagram decided that _South Station - Kingston_ and _South Station - Plymouth via Kingston_ were two separate branches. The latter route pattern covers all of the same stops the former one does, so we choose to show only _South Station - Plymouth via Kingston_ on the line diagram.

The chosen fix adds a new function, `overall_branch`, that takes lists of stops and finds the list whose stops cover all of the stops on all other lists. E.g. one branch to rule them all. If an "overall branch" is found, then use that for the line diagram instead of trying to piece together multiple branches.

### Second commit

CR-Fitchburg was broken because of new rail replacement bus shuttle service for portions of the route. These additional route patterns were creating extra "branches" for the line diagram. Both the normal rail version and the shuttle bus version are marked with same route pattern typicality value of 1.

This could have been automatically fixed by the `overall_branch` logic to select the normal rail route pattern, however this bus shuttle includes an extra stop that _isn't_ present in the rail version of the route. 

Ideally there would be a way to write some logic that inspects the rail_replacement_bus route pattern stops for stops that are not normally served by the particular rail line that that bus is replacing... but as far as I can tell there isn't.

So the fix chosen for this issue is more targeted, looking at this particular shuttle (Alewife-Littleton) and excluding the one known extra stop (Alewife) from consideration as part of the `overall_branch` logic.

---

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line/helpers.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line/helpers.ex
@@ -114,12 +114,25 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
   defp overarching_branch(route_patterns_with_stops) do
     all_stop_sets =
       route_patterns_with_stops
-      |> Enum.map(&elem(&1, 1))
+      |> Enum.map(fn {route_pattern, stops} ->
+        stops
+        |> maybe_adjust_for_rail_replacement_bus(route_pattern)
+      end)
       |> Enum.map(&MapSet.new/1)
 
     route_patterns_with_stops
     |> Enum.find(&has_all_stops?(&1, all_stop_sets))
   end
+
+  # If these stops are for a rail_replacement_bus, some manual adjustments are
+  # needed for the line diagram. The line diagram should show rail stops only
+  @spec maybe_adjust_for_rail_replacement_bus([Stop.t()], RoutePattern.t()) :: [Stop.t()]
+  defp maybe_adjust_for_rail_replacement_bus(stops, %RoutePatterns.RoutePattern{
+         id: "Shuttle-AlewifeLittleton" <> _
+       }),
+       do: Enum.reject(stops, &(&1.id == "place-alfcl"))
+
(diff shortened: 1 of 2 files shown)


paulswartz:
> The chosen fix adds a new function, `overall_branch`, that takes lists of stops and finds the list whose stops cover all of the stops on all other lists. E.g. one branch to rule them all. If an "overall branch" is found, then use that for the line diagram instead of trying to piece together multiple branches.

FYI: this will break again when Kingston goes back to having trips ending at both Kingston and Plymouth. I don't know when this will be, but something to keep an eye on.

alicia-mbta (approved):
I really like your approach.<br/>
Taking to account [Paul's comment](https://github.com/mbta/dotcom/pull/806#issuecomment-781561343), my only comment/question is whether `maybe_use_overarching_branch` should have a more Kingston-related name since that seems to be the only current exception (although you built the function to be generic)?<br/>
That way perhaps it'll be easier to identify if these changes need to be readjusted when Kingston goes back to having trips ending at both Kingston and Plymouth?

hannahpurcell:
If a shuttle is involved in the Fitchburg line, shouldn't there be a barber pole in the line diagram?  Or is the barber pole part of another situation?

thecristen:
@hannahpurcell Conceptually I think it _should_ have a barberpole pattern in the line diagram, but the data for _that_ depends on alerts existing on the appropriate stops. SO that's a different thing. :/

@alicia-mbta  We discussed the Kingston/Plymouth route's trips on Slack and based on the route patterns involved I don't think it's will be an issue .... also the function is helping with the Fitchburg route too, so I'm going to keep the name as is. :)

hannahpurcell (approved):

PR #819, opened 2021-02-25
...the rest of the navigation elements redesign

thecristen:
#### Summary of changes

Ah dang. I kept iterating on different parts of the page and ended up doing too many parts at once! Some of this intersects with what others are working on so I wanted to open a PR sooner rather than later. I'm still working on adding more JS tests, and I need to check IE...

The URL can use month and year parameters to control what is shown/scrolled to, e.g. `/events?month=3&year=2021`
It behaves slightly different with the addition of a `calendar=true` parameter, this is to filter the events when looking at the monthly calendar.

I think I'll deploy it to dev blue so we can see it there!

**Asana Tickets:** 

- [Left hand navigation of month-by-month anchor links](https://app.asana.com/0/385363666817452/1199915858612055/f)
- [On mobile, when you scroll up, dropdown appears](https://app.asana.com/0/385363666817452/1199915858612061/f)
- [When you scroll down, the current month sticks](https://app.asana.com/0/385363666817452/1199915858612059/f)
- [On page load, anchor to current day/month](https://app.asana.com/0/385363666817452/1199921715942829/f)

---

### DIFF
--- a/apps/site/assets/css/_events.scss
+++ b/apps/site/assets/css/_events.scss
@@ -6,32 +6,11 @@
   padding-top: $base-spacing-lg;
 }
 
-.arrow-icon:hover {
-  color: $brand-primary-lightest;
-  text-decoration: none;
-}
-
 .event-info-label {
   font-weight: bold;
 
   @include media-breakpoint-down(sm) {
     display: block;
-  }
-}
-
-.events-paged-list-header {
-  background-color: $brand-primary-darkest;
-  color: $white;
-  line-height: 2em;
-  margin-bottom: 0;
-  text-align: center;
-  text-transform: uppercase;
-
-  a {
-    color: $white;
-    display: inline-block;
-    padding: 0 $space-12;
-    vertical-align: bottom;
   }
 }
 
--- a/apps/site/assets/css/_events.scss
+++ b/apps/site/assets/css/_events.scss
@@ -18,6 +18,100 @@
   display: inline-block;
 }
 
+.event-listing {
+  @include media-breakpoint-down(xs) {
+    /* stylelint-disable-next-line scss/operator-no-unspaced */
+    margin: 0 -$base-spacing;
+  }
+}
+
+$mobile-control-height: 73px;
+
+.m-event-list__month-header {
+  background-color: $white;
+  margin: 0;
+  padding: $base-spacing $base-spacing/2;
+
+  &.sticky-top {
+    @include media-breakpoint-down(xs) {
+      padding-left: $base-spacing;
+      transition: top .2s ease-in-out;
+      z-index: 12; // overwrite 13 that's default so it doesn't show on top of the mobile nav
+
+      .js-nav-down & {
+        top: 0;
+      }
+
+      .js-nav-up & {
+        top: $mobile-control-height;
+      }
+    }
+  }
+
+  &[stuck] {
+    /* stylelint-disable-next-line scss/operator-no-unspaced */
+    box-shadow: 0 $space-16 $space-16 -1rem $drop-shadow-color;
+  }
+}
+
+.m-event-list__nav {
+  background-color: $white;
+  padding-top: $base-spacing;
+
+  a {
+    display: block;
+    font-weight: bold;
+    margin-bottom: $base-spacing / 2;
+  }
+
+  &.sticky-top {
+    top: $base-spacing;
+  }
+
+  @include media-breakpoint-down(xs) {
+    display: none;
+  }
+}
+
+.m-event-list__nav--mobile-controls {
+  background-color: $white;
+  border-bottom: 1px solid $gray-lightest;
+  display: none;
+  height: $mobile-control-height;
+  padding: $base-spacing;
+  transition: top .2s ease-in-out;
+
+  .m-event-list__select {
+    height: 100%;
+
+    &:first-child {
+      flex: 1;
+      margin-right: $base-spacing;
+    }
+  }
+
+  // show on scrolling up, hide on down
+  .js-nav-up & {
+    @include media-breakpoint-down(xs) {
+      display: flex;
+      top: 0;
+    }
+  }
+
+  .js-nav-down & {
+    @include media-breakpoint-down(xs) {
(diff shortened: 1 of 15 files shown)


alicia-mbta (approved):
The javascript desired stickiness looks quite tricky to accomplish so that part was more difficult to understand for me than the Elixir one. But wow, this looks great... :tada:

hannahpurcell:
Woohoo!  Can we change the base branch to be the feature to reduce the number of "changed files"?  (Also, :heart: the dialyzer commit :joy: I've been doing that a lot!)

apps/site/assets/css/_events.scss
```diff
@@ -18,6 +18,100 @@
   display: inline-block;
 }
 
+.event-listing {
```
hannahpurcell, line 21: The container that holds all of the Events Hub "guts" -- the left nav panel and the events list -- should have a reduced height, so that the "Events" title, the year dropdown, the calendar / list view toggle, and the MBTA website header are all available even after scrolling.  ZH mock:  https://zeroheight.com/2fedee66c/p/10db2c-events-hub/i/29298936

hannahpurcell: And re: this line, possibly rename "m-event-listing"

apps/site/assets/js/event-page-setup.js
```diff
@@ -0,0 +1,79 @@
...
+              el.toggleAttribute("stuck", isStuck);
+            });
+
+          // 2. show or hide the month/year dropdowns on mobile depending on
+          // whether the user scrolls up or down
+          const curScroll =
+            window.scrollY || document.documentElement.scrollTop;
+          direction =
+            // eslint-disable-next-line no-nested-ternary
+            curScroll > prevScroll ? 2 : curScroll < prevScroll ? 1 : 0;
```
hannahpurcell, line 34: 😍 whaaaat, I've never seen a nested ternary.  That's wild

apps/site/assets/js/event-page-setup.js
```diff
@@ -0,0 +1,79 @@
...
+
+          const controlHeight = control.getBoundingClientRect().height;
+          if (direction !== prevDirection) {
+            // toggle the controls
+            if (direction === 2 && curScroll > controlHeight) {
+              // hide the controls
+              eventsHubPage.classList.add("js-nav-down");
+              eventsHubPage.classList.remove("js-nav-up");
+            } else if (direction === 1) {
+              // remove the controls
```
hannahpurcell, line 44: Should this comment say `// show controls?`

hannahpurcell (changes requested):
😍 What lovely scrolly months!!  One primary requested change relating to matching the implementation to the mocks

apps/site/assets/css/_events.scss
thecristen: Thanks for flagging -- it's true my implementation doesn't quite match the mockup! [I raised it with Yinka in this ticket](https://app.asana.com/0/555089885850811/1199915858612059) and he gave me the approval for this version ;)

hannahpurcell (approved):
Aha, thanks for pointing me to the conversation with Yinka.  Looks good, then!

apps/site/assets/js/test/event-page-setup__test.js
```diff
@@ -39,10 +39,38 @@ const eventsHubHTML = `
     </div>
     `;
 
+function triggerScrollEvent(clock) {
+  document.dispatchEvent(new window.Event("scroll"));
+  clock.tick(); // fast forward set timeout (gets past window.requestAnimationFrame)
+}
+
 describe("setupEventsPage", () => {
```
alicia-mbta, line 47: Looks like it's a little bit of a pain to trigger these events so glad you found a way to test them! Thanks for adding all these comments for better understanding 🙏

thecristen:
I merged in your changes from this branch -- @hannahpurcell I made a tiny adjustment to how your previous event button script worked since now we're showing all months on one page. Hoping I resolved the merge conflicts okay, and requesting re-review on these changes :pray:

apps/site/assets/js/view-previous-events.js
```diff
@@ -1,17 +1,21 @@
...
+
+  viewPreviousEventsButtons.forEach(button => {
+    button.addEventListener("click", () => {
+      // show hidden events for this button
+      // because of how the list is laid out, we can just iterate through next <li> siblings until we find one without .m-hidden-event
+      let li = button.parentElement;
+      do {
+        li.classList.remove("m-hidden-event");
+        li = li.nextElementSibling;
+      } while (li && li.classList.contains("m-hidden-event") === true);
```
hannahpurcell, line 14: Is doing it this way better because it eliminates one `document.querySelectorAll`?  

It seems more brittle than before, since this loop is so specific about where the button sits in relationship to the rest of the list.  On the other hand, this list format shouldn't change in the future, so maybe it doesn't hurt to assert this list format?  Does that question make sense?

thecristen: Not necessarily better, just different! We still need to match specific events to each button (before only the first button was revealing _all_ hidden events on the page). Your question/concern makes sense -- another idea I was considering was adding a custom attribute to the markup that we could use to match. Like a name on each button that would match an attribute we add to every event it's responsible for. I think I will update it to do it that way instead!

hannahpurcell: Oh oh oh right, gosh.  My brain was still stuck on the "one button".  I totally understand why you looped the way you did, then!  But I'll keep an eye out for the next change

thecristen: @hannahpurcell Just push an update that will probably be less brittle!

hannahpurcell (approved):
Ah, yay!  The hub is really coming togetha with your work here 😁  It looks so good.

One suggestion that is so minor I won't request the change:  on initial render of the events page, the sticky header is already defaulted to [stuck], and the shadow under the sticky header month is already visible.  It honestly looks like the scroll-to-month only scrolled a pixel too far and triggered the header to stick??

PR #825, opened 2021-03-03
Add event page flag

thecristen:
Recently, we'd been using feature branches to manage the development of medium-to-long-term work. This has proven to be somewhat tricky to manage, particularly as the master branch diverges over time.

### There's another way!

Earlier we used to use flags to enable/disable big features. I don't think you're all familiar with the `Laboratory` module -- so I will add some details over Slack on how to find it and use it! 😉

This PR implements that approach for the **event page**. (I should've done this with the current events listing work, but I think we're too far into it to set up a flag for that now. Better late than never!)

With this enabled, we can proceed with development while continuing to merge into master. When finally ready for the reveal, we can remove the flag and tweak the code to permanently show the new features.

### DIFF
--- a/apps/site/config/config.exs
+++ b/apps/site/config/config.exs
@@ -27,7 +27,10 @@ config :logster, :allowed_headers, ["referer"]
 config :site, SiteWeb.ViewHelpers, google_tag_manager_id: System.get_env("GOOGLE_TAG_MANAGER_ID")
 
 config :laboratory,
-  features: [],
+  features: [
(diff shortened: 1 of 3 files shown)


alicia-mbta (approved):
Looks great! I was somewhat familiar with the concept of feature-flagged work but not with `:laboratory`.
Thanks for enabling this! 🙏

hannahpurcell (approved):
Oooo fantastic!  I like this approach!

PR #828, opened 2021-03-04
Add calendar grid

thecristen:
#### Summary of changes
**Asana Ticket:** [Add event listings to the grid](https://app.asana.com/0/385363666817452/1199915858612066/f)

There are a few noticeable issues which I plan to address in the 'Add toggle between views to navigation' ticket, including:

* rename URL parameter from calendar to cal
* keep all months in the left nav, not just the current one
* hide the calendar on mobile (presumably showing the list in its place)
* adding the working toggle

### DIFF
--- a/apps/site/lib/site_web/templates/event/index.html.eex
+++ b/apps/site/lib/site_web/templates/event/index.html.eex
@@ -1,4 +1,4 @@
-<div class="container m-events-hub">
+<div class="container m-events-hub <%= if Map.get(@conn.query_params, "calendar"), do: 'm-events-hub--calendar-view', else: 'm-events-hub--list-view' %>">
   <header>
     <h1>Events</h1>
     <nav class="m-events-hub__nav--controls">
--- a/apps/site/lib/site_web/templates/event/index.html.eex
+++ b/apps/site/lib/site_web/templates/event/index.html.eex
@@ -12,19 +12,29 @@
   </header>
 
   <div class="row">
-    <nav class="m-event-list__nav col-sm-3 fixedsticky sticky-top">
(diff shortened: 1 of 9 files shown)


apps/site/lib/site_web/templates/event/index.html.eex
```diff
@@ -5,19 +5,29 @@
   </header>
```
hannahpurcell, line 5: The calendar work does not play behave when event-page-setup script is run on this page.  (The conditional for the script is based on the `m-events-hub` selector, so using a selector from `_event_list` instead should do the trick.)

hannahpurcell (approved):
Looks fantastic!  One note on console errors, and one note on a couple design tweaks

apps/site/lib/site_web/templates/event/index.html.eex
thecristen: Just updated this, and I'm having trouble reproducing so I'm hoping that means it's fixed now!

hannahpurcell:
Oh gosh!  Sorry, I didn't notice your request to re-review.  I only saw it was approved and just kept looking past 😓  I thought the url to access the calendar was `{host}/events?calendar`, and looking at the code, that seems right to me...  did anything change that could impact my being able to access it there?

apps/site/lib/site_web/templates/event/index.html.eex
hannahpurcell: sure is!

hannahpurcell (approved):
The two updates are great:  grid design looks clean, and console errors are gone

PR #829, opened 2021-03-04
fix: Show description for planned disruptions

thecristen:
#### Summary of changes
**Asana Ticket:** [Schedules | Services picker should show distinctive title for CR shuttles](https://app.asana.com/0/385363666817452/1199903984955228/f)

Prior logic wouldn't show the description in the services picker when it's for a planned disruption. Probably will deploy on dev green.

Thanks @jfabi for tipping us off on this!


alicia-mbta (approved):
🎉

PR #831, opened 2021-03-05
Proposed Sales Location page

alicia-mbta:
#### Summary of changes
**Asana Tickets:**
[Upload new data source](https://app.asana.com/0/555089885850811/1199957984734799)
[Update the location cards](https://app.asana.com/0/555089885850811/1199957984734798)
[Duplicate the existing tool into a new, unique instance](https://app.asana.com/0/555089885850811/1199957984734796)
[Restyle the new tool](https://app.asana.com/0/555089885850811/1199957984734797)

Gets data from ArcGIS depending on location.

### DIFF
--- a/apps/fares/lib/proposed_locations.ex
+++ b/apps/fares/lib/proposed_locations.ex
@@ -1,0 +1,70 @@
+defmodule Fares.ProposedLocations do
+  @moduledoc """
+    Gets information from the existing ArcGIS data related to proposed sales locations, parses the json and returns it.
+  """
+
+  alias Fares.ProposedLocations.Location
+  require Logger
+
+  @base_url "https://services1.arcgis.com/ceiitspzDAHrdGO1/ArcGIS/rest/services/ProposedSalesNetworkSpringOutreach/FeatureServer/0/query?f=json&outFields=*&inSR=4326&outSR=4326&returnGeometry=true"
+
+  @distance_in_miles 10
+
+  @spec by_lat_lon(%GoogleMaps.Geocode.Address{}) :: [Location.t()] | nil
+  def by_lat_lon(%{latitude: lat, longitude: lon}) do
+    url =
+      "#{@base_url}&where=1%3D1&geometry=#{lon}%2C+#{lat}&geometryType=esriGeometryPoint&distance=#{
+        @distance_in_miles
+      }&units=esriSRUnit_StatuteMile"
+
+    get_parsed_proposed_locations(url)
+  end
+
+  def by_lat_lon(_) do
+    nil
+  end
+
+  @spec get_parsed_proposed_locations(String.t()) :: [Location.t()] | nil
+  defp get_parsed_proposed_locations(url) do
+    case HTTPoison.get(url) do
+      {:ok, %{status_code: 200, body: body, headers: _headers}} ->
+        case Poison.decode(body) do
+          {:ok, json} ->
+            Enum.map(json["features"], &parse_proposed_location(&1))
+
+          {:error, _} ->
+            Logger.warn("error decoding json -> original_request=#{url}")
+            nil
+        end
+
+      {:error, _} ->
+        Logger.warn("error in http request -> original_request=#{url}")
+        nil
+    end
+  end
+
+  @spec parse_proposed_location(map) :: Location.t()
+  defp parse_proposed_location(json) do
+    attributes = json["attributes"]
+    routes = attributes["Routes"]
+
+    list_of_routes =
+      if routes == " " do
+        []
+      else
+        String.split(routes, ", ", trim: true)
+      end
+
+    %Location{
+      fid: attributes["FID"],
+      stop_id: attributes["stop_id"],
+      stop_name: attributes["stop_name"],
+      municipality: attributes["municipali"],
+      line: attributes["Line"],
+      retail_fvm: attributes["RetailFVM"],
+      routes: list_of_routes,
+      latitude: json["geometry"]["y"],
+      longitude: json["geometry"]["x"]
+    }
+  end
+end
--- a/apps/fares/lib/proposed_locations/location.ex
+++ b/apps/fares/lib/proposed_locations/location.ex
@@ -1,0 +1,28 @@
+defmodule Fares.ProposedLocations.Location do
+  @moduledoc """
+  Proposed Sales Location structure that matches the data coming from ArcGIS
+  """
+  defstruct [
+    :fid,
+    :stop_id,
+    :stop_name,
+    :municipality,
+    :line,
+    :retail_fvm,
+    :routes,
+    :latitude,
+    :longitude
+  ]
+
+  @type t :: %__MODULE__{
+          fid: non_neg_integer,
+          stop_id: String.t(),
+          stop_name: String.t(),
+          municipality: String.t(),
+          line: String.t(),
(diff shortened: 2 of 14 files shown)


hannahpurcell:
Looks so great!  At first glance, I noticed one need-to-fix and a couple things that are cosmetic and not a priority for Monday.  I'll note them both, but maybe we could make another "polish" ticket.

Important:  the search button on the input box is a broken link.  (If I click my address from the dropdown, no problem.  But if I click the magnifying glass again later, the url changes and goes to a dead page.)

To do later:
- The address probably shouldn't convert to lat/lon coords in the input box.  It should stay in the format that the user input.  (Maybe that's coordinates, but probably it's an address.)
- The grid of locations can be styled as a "masonry grid"

apps/site/lib/site_web/templates/fare_transformation/_nearby_locations.html.eex
```diff
@@ -0,0 +1,46 @@
+<div class="c-location-cards c-location-cards--background-gray small-set grouped">
+  <%= for {%Fares.ProposedLocations.Location{stop_name: name, retail_fvm: retail_fvm, latitude: latitude, longitude: longitude} = _location, distance} <- @nearby_proposed_locations do %>
+
+    <div class="c-location-card c-location-card--resp">
+
+      <div class="c-full-width">
```
thecristen, line 6: If you remove `display: flex` from the parent then this wrapper doesn't seem to be needed.

apps/site/lib/site_web/templates/fare_transformation/_nearby_locations.html.eex
```diff
@@ -0,0 +1,46 @@
...
+
+        <div class="c-expandable-block__header-container">
```
thecristen, line 8: eh, let's make another `c-location-card` class for this instead of appropriating this one.

apps/site/lib/site_web/templates/fare_transformation/_nearby_locations.html.eex
```diff
@@ -0,0 +1,46 @@
...
+            <%= name %>
+          </div>
+
+          <div class="c-location-card__distance">
+            <%= round_distance(distance) %>
+          </div>
+        </div>
+
+        <%#= if Fares.ProposedLocations.Location.is_retail_store?(location) do %>
+          <!-- <div class="c-location-card__location">
```
thecristen, line 19: Nice

thecristen (approved):
Looks good! I had a few comments but I don't know if any should be dealbreakers. I'd get approval from design and/or product before merging.

thecristen:
+1 to Hannah's observations, also the "use my current location" link doesn't work as expected.

hannahpurcell:
Another medium-priority task for the "Do Later" ticket:  locations should be sorted by distance

apps/site/lib/site_web/controllers/fare_transformation_controller.ex
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def finder(conn, %{"location" => _location} = params) do
+    {position, formatted} =
+      SiteWeb.FareController.calculate_position(params, &GoogleMaps.Geocode.geocode/1)
+
+    nearby_proposed_locations =
+      Nearby.nearby_with_varying_radius_by_mode(position)
+      |> Enum.map(fn stop -> stop.id end)
+      |> ProposedLocations.by_stop_ids()
```
hannahpurcell, line 27: ⚠️  The `Nearby.nearby_with_varying_radius_by_mode(position)` isn't catching all possible PSLs.

Search "Boston Landing", and the PSL page will only return one option on Arthur St @ Guest St, but there is a PSL pin in the ArcGIS data directly at Boston Landing.  When I hardcode the logic in `ProposedLocationsby_stop_id()` to also include the id "Boston Landing", the PSL shows up on the page.  So there's something about the `Nearby` logic that isn't finding this station that is very close by

hannahpurcell: Does the query to ArcGIS need to be based on stops, or can it be based on lat-lon?  

I believe ArcGIS can be queried to "get all PSLs within 10 miles of this location."  If we use this query format and just pass in the user input as the central location, than we can skip this `Nearby` step

apps/fares/lib/proposed_locations.ex
```diff
@@ -0,0 +1,75 @@
...
+
+  alias Stops.Stop
+  alias Fares.ProposedLocations.Location
+
+  require Logger
+
+  @base_url "https://services1.arcgis.com/ceiitspzDAHrdGO1/ArcGIS/rest/services/ProposedSalesNetworkSpringOutreach/FeatureServer/0/query?f=json&outFields=*&outSR=4326&returnGeometry=true"
+
+  @spec by_stop_ids([Stop.t()]) :: [Location.t()] | nil
+  def by_stop_ids(stop_ids) do
```
thecristen, line 14: Why not do an initial getting and parsing of the entire dataset, and then later just filter those results by the distance? Why involve stop_ids?

apps/fares/lib/proposed_locations/location.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Fares.ProposedLocations.Location do
+  @moduledoc """
+  Proposed Sales Location structure that matches the data coming from ArcGIS
```
thecristen, line 3: We might want to limit this just to fields we care about, for simplicity - we probably aren't going to use FFFVMUpdat, LFFVMUpdat, RSTUpdat, why_sn.

apps/fares/lib/proposed_locations.ex
alicia-mbta: I just thought getting _all_ the data was a lot, that's why my idea was to grab nearby stops and query the ArcGIS data with that. However, @hannahpurcell thought of an even better query which returns nearby locations by lat/lon! 🎉 so that's what I did instead.

alicia-mbta:
> +1 to Hannah's observations, also the "use my current location" link doesn't work as expected.

I had to really make sure location services was enabled in the browser; also it seems like the changes don't really occur unless you restart it 😕

apps/fares/lib/proposed_locations.ex
```diff
@@ -0,0 +1,69 @@
...
+  @moduledoc """
+    Gets information from the existing ArcGIS data related to proposed sales locations, parses the json and returns it.
+  """
+
+  alias Fares.ProposedLocations.Location
+  require Logger
+
+  @base_url "https://services1.arcgis.com/ceiitspzDAHrdGO1/ArcGIS/rest/services/ProposedSalesNetworkSpringOutreach/FeatureServer/0/query?f=json&outFields=*&inSR=4326&outSR=4326&returnGeometry=true"
+
+  @distance_in_miles 2
```
thecristen, line 11: This might be too small a limit for the more suburban areas.... would it make sense to do a very large limit and limit the results shown to the (nine? I don't know) closest ones?

alicia-mbta: Totally -- I'll change that.

hannahpurcell:
Clicking the 🔍 doesn't go to a dead link anymore, but it also never returns any results (even if there are results)

hannahpurcell (approved):

PR #836, opened 2021-03-10
Attempt to find the analogous route pattern

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞 The schedule view changes your route variant when you change direction](https://app.asana.com/0/555089885850811/1179129122227742)

When toggling `directionId` having an existing route pattern, an attempt to get the "analogous" route pattern for the toggled directionId will be made.

E.g.: if we toggle from `directionId=0` and `route pattern=111-5-0` to `directionId=1`, we'll try to see if `route pattern=111-5-1` exists. Otherwise, default to the first route pattern from the list (for the new directionId).

### DIFF
--- a/apps/site/assets/.stylelintrc
+++ b/apps/site/assets/.stylelintrc
@@ -8,7 +8,7 @@
     "stylelint-selector-no-mergeable"
   ],
   "rules": {
-    "at-rule-blacklist": [
+    "at-rule-disallowed-list": [
       "debug"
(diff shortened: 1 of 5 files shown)


hannahpurcell (approved):
This looks great!  I'm taking notes for my own shape task 😊

thecristen (approved):
Clever! Thanks for the test, too.

PR #842, opened 2021-03-16
Assorted test optimizations

thecristen:
This PR has an assortment of test and testing-related updates, created in an attempt to make our test results a bit more predictable and robust.

It seems many of the recurring errors were due to race conditions caused by our test suite. I went about tackling this by running the test suite repeatedly.. a lot. I found that taking note of the random seed number sometimes helped in reproducing some of these errors (by rerunning the test suite with the flag `--seed integer_of_bad_seed`), and also using `--max-failures 1` and `--listen-on-stdin` options helped me test just a little faster. ;) The end result here isn't perfect but I'm hopeful it can improve something.

Overarching changes:

- Ensuring each app's dependencies are started and thus ready to work when the tests start.
- Sometimes using the `max_cases` ExUnit option to run tests one by one instead of concurrently.

### DIFF
--- a/apps/fares/test/test_helper.exs
+++ b/apps/fares/test/test_helper.exs
@@ -1,4 +1,10 @@
-ExUnit.start()
+# Ensure the deps are all started
+Application.load(:fares)
+
+for app <- Application.spec(:fares, :applications) do
+  {:ok, _} = Application.ensure_all_started(app)
+end
 
 # Report warnings as errors
 Code.compiler_options(warnings_as_errors: true)
+ExUnit.start(max_cases: 1)
--- a/apps/fares/test/test_helper.exs
+++ b/apps/fares/test/test_helper.exs
@@ -1,9 +1,5 @@
 # Ensure the deps are all started
-Application.load(:fares)
-
-for app <- Application.spec(:fares, :applications) do
-  {:ok, _} = Application.ensure_all_started(app)
-end
+Application.ensure_all_started(:fares)
 
 # Report warnings as errors
 Code.compiler_options(warnings_as_errors: true)
--- a/apps/site/test/site/trip_plan/merge_test.exs
+++ b/apps/site/test/site/trip_plan/merge_test.exs
@@ -16,8 +16,8 @@ defmodule Site.TripPlan.MergeTest do
 
   describe "merge/3" do
     test "includes no duplicates" do
-      ptest accessible: list(of: positive_int(), max: 5),
-            unknown: list(of: positive_int(), max: 5) do
+      ptest accessible: list(of: positive_int(), min: 1, max: 5),
+            unknown: list(of: positive_int(), min: 1, max: 5) do
         {_, _, merged} = do_merge(accessible, unknown)
         assert Enum.uniq(merged) == merged
       end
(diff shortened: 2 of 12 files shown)


apps/fares/test/test_helper.exs
```diff
@@ -1,4 +1,10 @@
-ExUnit.start()
+# Ensure the deps are all started
+Application.load(:fares)
+
+for app <- Application.spec(:fares, :applications) do
+  {:ok, _} = Application.ensure_all_started(app)
+end
```
paulswartz, line 6: I think you can simplify this to `Application.ensure_all_started(:fares)`?

apps/site/test/site/trip_plan/merge_test.exs
```diff
@@ -16,8 +16,8 @@ defmodule Site.TripPlan.MergeTest do
 
   describe "merge/3" do
     test "includes no duplicates" do
-      ptest accessible: list(of: positive_int(), max: 5),
-            unknown: list(of: positive_int(), max: 5) do
+      ptest accessible: list(of: positive_int(), min: 1, max: 5),
+            unknown: list(of: positive_int(), min: 1, max: 5) do
```
paulswartz, line 20: Does this mean it fails if one of the two lists is empty? Is that a problem?

apps/vehicles/test/stream_test.exs
```diff
@@ -1,6 +1,6 @@
 defmodule Vehicles.StreamTest do
-  use ExUnit.Case, async: true
-  alias ExUnit.CaptureLog
+  use ExUnit.Case, async: false
```
paulswartz, line 2: `async: false` is the default, so you can use `use ExUnit.Case` if you want.

apps/vehicles/config/config.exs
```diff
@@ -28,3 +28,7 @@ use Mix.Config
 # here (which is why it is important to import them last).
 #
 #     import_config "#{Mix.env}.exs"
+
+# let logs happen, but only show warning and up in console
+config :logger, level: :info
+config :logger, :console, level: :warn
```
paulswartz, line 34: This is probably not what you want, since it'll affect all applications.

thecristen: I went ahead and removed this as it didn't help anyway. This one test pictured below often fails (seen in build 2 of this branch), despite clearly outputting the log it seeks. It seems sometimes capture_log works, and sometimes it doesn't...

image

apps/site/test/site/trip_plan/merge_test.exs
thecristen: Good question -- yes It was failing when `unknown` was an empty list, and in trying to assess whether that's a problem or not, at the very least for the intent of this test it doesn't seem to be an issue. 

But I am reverting as it's not decisively an improvement :shrug:

apps/fares/test/test_helper.exs
```diff
@@ -1,4 +1,6 @@
-ExUnit.start()
+# Ensure the deps are all started
+Application.ensure_all_started(:fares)
```
alicia-mbta, line 2: This is smart!

alicia-mbta (approved):
I like these additions 🎉 
Hopefully this will help with the overall process!

PR #845, opened 2021-03-17
Account for trips including the CapeFLYER

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Elixir.UndefinedFunctionError: function nil.mode/0 is undefined](https://app.asana.com/0/555089885850811/1200069554962417)

Fares are not specified for the CapeFLYER, so these changes make sure the trip planner logic doesn't break in this scenario and displays an additional message for more information instead.

image

### DIFF
--- a/apps/site/lib/site_web/views/trip_plan_view.ex
+++ b/apps/site/lib/site_web/views/trip_plan_view.ex
@@ -453,10 +453,17 @@ defmodule SiteWeb.TripPlanView do
 
       one_way_total_fare = get_one_way_total_by_type(i, :highest_one_way_fare)
 
+      itinerary_is_from_or_to_airport = itinerary_satisfies_property?(i, :is_from_or_to_airport)
+
+      show_fares =
+        !itinerary_satisfies_property?(i, :contains_capeflyer) || one_way_total_fare != 0
+
       fares_estimate_html =
         "_itinerary_fares.html"
         |> render_to_string(
           itinerary: i,
+          show_fares: show_fares,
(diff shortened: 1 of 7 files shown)


thecristen:
Looks super promising! Going to push it to Dev blue so we can review while being unable to geocode the destination (because of the Google API key thing)

thecristen (approved):
I'm hoping we get some design feedback to improve this further or content feedback to verify this is what we want to show... but code-wise it looks good!

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -663,4 +671,11 @@ defmodule SiteWeb.TripPlanView do
     |> Itinerary.transit_legs()
     |> Enum.any?(fn leg -> leg_is_from_or_to_airport?(leg) end)
   end
+
+  @spec itinerary_contains_capeflyer?(Itinerary.t()) :: boolean()
```
hannahpurcell, line 675: Should both `itinerary_is_from_or_to_airport` and `itinerary_contains_capeflyer` be handled the same way?  Either as a function that takes itinerary, or as an assigned property in the render function

hannahpurcell (approved):
I learned about another mode of area transit :blush:  Looks good to me

apps/site/lib/site_web/views/trip_plan_view.ex
alicia-mbta: Good point -- I can probably refactor this.

alicia-mbta:
Merging after getting confirmation from Yinka:

image

PR #847, opened 2021-03-18
Hide line diagram branches created by multi route trips

thecristen:
#### Summary of changes
**Asana Ticket:** [Schedules | Should not include route patterns for multi route trips](https://app.asana.com/0/385363666817452/1200081044003352/f)

Added in after a test failure led me to find broken line diagrams as described in the ticket.
Also includes fixes for a few failing Green-E tests.

This fixes the line diagram but does not fix the map polylines -- I have a tentative approach for those but wanted to prioritize fixing the branching on the line diagram here, and also wanted to hold off since I know the map is being worked on right now anyway.


alicia-mbta (approved):
🎉  nice catch

PR #848, opened 2021-03-18
Mapping 57a and other bus variants

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Draw correct shape for requested variant (route pattern)](https://app.asana.com/0/555089885850811/1199671224586822)
**Asana Ticket:** [Schedules | 57A has no shape on map](https://app.asana.com/0/555089885850811/1200000330269316)

- Schedule for 57A has a line on the map:  **this is obsolete with the new Spring schedule!**
- Bus 16 has map + line diagram
- Schedule for buses with multiple variants will update the map (lines and stops) when the variant changes.  See bus 57
- Green line:  all routePatterns of typicality 1 are included on the map, meaning that the tiny 3-stop pattern representing the Lechmere shuttle is also included.  Previously, those stops looked like random dots on the map with no connector line.

One thing that needs checking:
- [x] Does the static map render properly when JS is disabled?  It's gone through some changes but couldn't be tested locally.
--- 

* [x] Are interactive elements accessible to screen readers? -- No change


apps/site/lib/site_web/controllers/schedule/line.ex
```diff
@@ -111,17 +111,31 @@ defmodule SiteWeb.ScheduleController.Line do
     variant = schedule_direction["variant"]
     expanded = Map.get(conn.query_params, "expanded")
     reverse_direction_id = reverse_direction(direction_id)
-    route_shapes = LineHelpers.get_route_shapes(route.id, direction_id)
     route_stops = LineHelpers.get_route_stops(route.id, direction_id, deps.stops_by_route_fn)
+    # Both line.ex and helpers.ex have this function defined. Redundant?
```
alicia-mbta, line 115: If there's indeed a redundancy, could we refactor?

alicia-mbta (commented):
Left some initial comments although it seems you're still working on this branch.

apps/route_patterns/lib/repo.ex
```diff
@@ -41,7 +42,8 @@ defmodule RoutePatterns.Repo do
       direction_id -> [route: route_id, direction_id: direction_id]
     end
     |> Keyword.put(:sort, "typicality,sort_order")
-    |> cache(&api_all/1)
+    |> Keyword.put(:include, "representative_trip.shape")
+    |> api_all
```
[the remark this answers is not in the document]
thecristen: +1 to this!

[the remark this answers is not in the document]
hannahpurcell: Great catch, thank you.  It was challenging to follow the data with caching turned off, but then it got forgotten!

apps/site/lib/site_web/controllers/schedule/line/maps.ex
```diff
@@ -19,58 +20,56 @@ defmodule SiteWeb.ScheduleController.Line.Maps do
...
           [String.t()] | any,
           VehicleHelpers.tooltip_index() | [] | nil
         ) :: {String.t(), MapData.t()}
-  def map_data(route, map_route_stops, [], []) do
-    map_shapes = map_polylines(map_route_stops, route)
-    static_data = map_img_src(map_route_stops, [], route)
-    dynamic_data = dynamic_map_data(route.color, map_shapes, map_route_stops, {nil, nil})
-    {static_data, dynamic_data}
+  def map_data(route, [], [], route_patterns, [], []) do
+    # static_data = map_img_src(static_map_stops, [], route)
```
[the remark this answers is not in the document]
hannahpurcell: Nope!  Thanks.  

This implementation of map_data is only ever called in `map_api.ex`.  That function call used to look like this:  
```{_map_img_src, dynamic_map_data} = Maps.map_data(route, map_stops, map_polylines, route_patterns, [], [])```

Well... wait, if `_map_img_src` was never used, then we don't really need to calculate `static_data` in the `map_data` function - we can just return nil!

If that's the case, then the function call in `map_api.ex` can pass in a lot less parameters.
```{_map_img_src, dynamic_map_data} = Maps.map_data(route, [], [], route_patterns, [], [])```

making `direction_id` from the `show` function negligible, I believe.  Did I explain that right?

File: apps/site/lib/site_web/controllers/schedule/map_api.ex [hunk 4]
```diff
@@ -9,19 +9,17 @@ defmodule SiteWeb.ScheduleController.MapApi do
...
   alias SiteWeb.ScheduleController.Line.Helpers, as: LineHelpers
   alias SiteWeb.ScheduleController.Line.Maps
+  alias RoutePatterns.Repo, as: RoutePatternRepo
 
   @type direction_id :: 0 | 1
 
   @spec show(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def show(conn, %{
-        "id" => route_id,
-        "direction_id" => direction_id
```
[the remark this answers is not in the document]
hannahpurcell: I tried answering this in a long reply to the `maps.ex` comment you made above.

apps/site/lib/site_web/controllers/schedule/line.ex
hannahpurcell: Maybe could we add a tech debt ticket for it?  I left one or two other "todo" type comments, but since I didn't _need_ to refactor it to make it work, I left the redundancy as-is.  (I didn't add the redundancy in this ticket - it was there from previously)

apps/site/lib/site_web/controllers/schedule/line/maps.ex
```diff
@@ -19,58 +20,56 @@ defmodule SiteWeb.ScheduleController.Line.Maps do
...
     MapHelpers.image(:ferry)
   end
 
-  defp map_img_src({route_stops, _shapes}, polylines, %{color: route_color} = _route) do
+  defp map_img_src(route_stops, polylines, %{color: route_color} = _route) do
     markers = Enum.map(route_stops, &build_google_stop_marker/1)
     paths = Enum.map(polylines, &Path.new(&1, color: route_color))
 
-    {600, 600}
+    {700, 700}
```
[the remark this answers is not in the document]
hannahpurcell: Hm...  At first I thought it increased the viewport of the static map, but after seeing it dev green, I'm not so sure.  I'll revert it to what it was originally, then re-deploy to test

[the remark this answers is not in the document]
hannahpurcell: I'll leave it as-is, until a user tells us they don't like that size, haha

apps/site/lib/site_web/controllers/schedule/line/maps.ex
```diff
@@ -19,54 +20,51 @@ defmodule SiteWeb.ScheduleController.Line.Maps do
...
+  def map_stops(branches) do
+    branches
+    |> Enum.flat_map(& &1.stops)
+    |> Enum.uniq_by(& &1.id)
   end
 
-  @spec map_polylines({any, [Shape.t()]}, Route.t()) :: [Shape.t()]
-  defp map_polylines(_, %Route{type: 4}), do: []
-
-  defp map_polylines({_stops, shapes}, _), do: shapes
```
alicia-mbta, line 62: Cool that we don't need this anymore :grinning:

apps/site/lib/site_web/controllers/schedule/line/maps.ex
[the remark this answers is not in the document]
alicia-mbta: Yes, thank you! 🙏

alicia-mbta (approved):
I re-reviewed and it looks great! Thanks a lot for all these changes (and the clarifications). Don't forget to squash/consolidate the commits where it makes sense to do so for extra _crispness_ :ok_hand:

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -105,14 +102,20 @@ const ScheduleDirection = ({
...
     ) || routePatternsInCurrentDirection.slice(0, 1)[0];
 
+  // If the route is a rail type, any other routePatterns
+  // of the current typicality are a branch
+  const branchPatterns = !isABusRoute(route)
+    ? routePatternsInCurrentDirection.filter(
+        pattern => pattern.typicality === defaultRoutePattern.typicality
+      )
+    : [];
+
```
hannahpurcell, line 112: Aha!  The CR also needs a special case, because only the defaultRoutePattern should go on the map

thecristen (commented):
I am seeing a noticeable lag in the map rendering when I switch direction. 

For reference I'm comparing `/schedules/CR-Providence/line?schedule_direction%5Bdirection_id%5D=1&schedule_direction%5Bvariant%5D=CR-Providence-cc900769-1` on green dev (left in screen recording) vs dev (right).

https://user-images.githubusercontent.com/2136286/112502155-375ab380-8d60-11eb-9338-8cb9805e1a2e.mov

 Is there a possible performance issue to watch out for here? 😟

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -104,11 +104,12 @@ const ScheduleDirection = ({
 
   // If the route is a rail type, any other routePatterns
   // of the current typicality are a branch
-  const branchPatterns = !isABusRoute(route)
-    ? routePatternsInCurrentDirection.filter(
-        pattern => pattern.typicality === defaultRoutePattern.typicality
-      )
-    : [];
+  const branchPatterns =
+    !isABusRoute(route) && !isACommuterRailRoute(route)
```
thecristen, line 108: But Hannah! A few of the commuter rail routes _do_ have branches! 🙈 (localhost below!) 
image

thecristen (changes requested):
With respect to the comment I left about possible performance issues, I was poking around the map in React DevTools and the markers in the map seem to be re-rendering -- not sure if that's directly related but I don't think the map markers should be re-rendering many times a minute? Anyways I do find the lagginess slightly concerning.

It also looks like the commuter rail scenario might need another tweak or so (see screenshot in comment with missing branch!).

Otherwise, this looks good and is generally working quite nicely!

apps/site/lib/leaflet/map_data/polyline.ex
```diff
@@ -20,10 +20,11 @@ defmodule Leaflet.MapData.Polyline do
 
   @default_opts [color: "#000000", weight: 5, dotted?: false]
 
-  @spec new(Shape.t() | String.t(), Keyword.t()) :: t()
+  @spec new(RoutePattern.t() | String.t(), Keyword.t()) :: t()
   def new(polyline, user_opts \\ [])
```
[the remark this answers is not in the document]
hannahpurcell: Definitely!

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
hannahpurcell: Ohhh nooooooooo 😭

hannahpurcell:
> With respect to the comment I left about possible performance issues, I was poking around the map in React DevTools and the markers in the map seem to be re-rendering -- not sure if that's directly related but I don't think the map markers should be re-rendering many times a minute? Anyways I do find the lagginess slightly concerning.

Thanks for the note on lagginess!  Yeah, I don't like it.  But gosh is it hard to debug locally.  For me, spinning up master locally is just as laggy as this branch, so I guess I have to deploy to test it?!  Grr...

As for the map markers, I think those have always re-rendered many times a minute, because of the dynamic vehicle markers

apps/site/assets/ts/schedule/components/Map.tsx
```diff
@@ -105,9 +107,9 @@ const shouldVehicleDisplay = (
   marker: Marker,
   shapeIdsInDirection: string[]
 ): boolean => {
-  if (marker.shape_id) {
-    return shapeIdsInDirection.some(shapeId => shapeId === marker.shape_id);
-  }
+  // if (marker.shape_id) {
```
ryan-mahoney, line 110: @hannahpurcell do you still need this function if it always returns `true`?

apps/site/assets/ts/schedule/components/Map.tsx
```diff
@@ -194,11 +196,13 @@ export const reducer = (
 };
 
 export default ({
+  // Data has markers and stop_markers, and they are identical?
```
ryan-mahoney, line 199: @hannahpurcell is this a `TODO` to look into this? I don't think we do a lot of code TODOs, may be best to leave it as is or throw a tech-debt ticket in the backlog to investigate this as a later date (if this question is still unresolved)

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -97,6 +94,7 @@ const ScheduleDirection = ({
   stops,
   today,
   scheduleNote,
+  // This references the current bus variant, if applicable
```
ryan-mahoney, line 97: @hannahpurcell does this comment indicate that the variable name should be changed to be more descriptive?

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -147,8 +155,9 @@ const ScheduleDirection = ({
         );
       }
     },
+    // only re-run the effect if any of these things change
```
ryan-mahoney, line 158: @hannahpurcell maybe these comments are useful because this code is so distant from where the `useEffect` hook is first defined, but I think this comment is just describing the default behavior of `useEffect`

apps/site/test/site_web/controllers/schedule_controller_test.exs
```diff
@@ -477,15 +477,10 @@ defmodule SiteWeb.ScheduleControllerTest do
...
-
-      assert variant == conn.assigns.active_shape.id
     end
 
-    test "Bus line with correct default shape", %{conn: conn} do
-      conn = get(conn, line_path(conn, :show, "9", "schedule_direction[direction_id]": 1))
-      default_shape_id = List.first(@routes_repo_api.get_shapes("9", direction_id: 1)).id
-      assert conn.assigns.active_shape.id == default_shape_id
-    end
+    # This test was useful on the one hand, but not if the frontend isn't using
```
ryan-mahoney, line 482: @hannahpurcell does this comment refer to the code that was removed? If so, might be confusing if left in ;)

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
hannahpurcell: Fixed 😅

hannahpurcell:
> ... I do find the lagginess slightly concerning.

Now, I'm not certain here, but I think recent updates may have fixed this!!  I was putting too much logic in the TS side of things, but frontloading that logic on the Elixir side seems to have improved this!

apps/site/assets/ts/schedule/components/Map.tsx
hannahpurcell: Lol, I commented it out for testing, but the vehicles seem to be more accurate when commented than previously.  So I agree with you - get rid of the function.

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
hannahpurcell: You know what?  YES!  `busVariantId` it is

thecristen (approved):

thecristen (approved):
💯 let's squash this thing!

PR #851, opened 2021-03-22
Button is keyboard navigable

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [View Previous Month button - fix accessibility](https://app.asana.com/0/555089885850811/1200032439670461)

Added button to tab order, and added an eventListener that translates the enter-key to button click.

I tried to test it with screen readers, and I think it works better than before, but I'm not super great at using a screen reader yet.

Update:  I feel like there should be a test for this, but I tried a number of approaches and couldn't get jQuery triggers and JS events to play nicely.

---


alicia-mbta (approved):
Looks good to me 👍

PR #853, opened 2021-03-23
PSL:  Adjusted search error msg copy, color, placement

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Add error message for location search](https://app.asana.com/0/555089885850811/1200068680318939)

---

### DIFF
--- a/apps/site/lib/site_web/templates/fare_transformation/proposed_sales_locations.html.eex
+++ b/apps/site/lib/site_web/templates/fare_transformation/proposed_sales_locations.html.eex
@@ -28,7 +28,7 @@
       <%= form_for @conn, @conn.request_path<>"#address-search-input", [as: :location, method: :get, class: "form"], fn _f -> %>
         <%= input_location(%InputLocation{id: "address-search-input", name: :location, name_index: :address, address: @address, address_error: get_flash(@conn, :info), submit: true, required: true, placeholder: "Enter a location", button_class: ""}) %>
       <% end %>
-      <%= if Enum.empty?(@nearby_proposed_locations) do %>
+      <%= if !is_nil(@nearby_proposed_locations) and Enum.empty?(@nearby_proposed_locations) do %>
         <p class="m-error-message">Please enter an address that is within 100 miles of an MBTA service area.</p>
       <% end %>
     </div>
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/templates/fare_transformation/proposed_sales_locations.html.eex
```diff
@@ -28,6 +28,11 @@
       <%= form_for @conn, @conn.request_path<>"#address-search-input", [as: :location, method: :get, class: "form"], fn _f -> %>
         <%= input_location(%InputLocation{id: "address-search-input", name: :location, name_index: :address, address: @address, address_error: get_flash(@conn, :info), submit: true, required: true, placeholder: "Enter a location", button_class: ""}) %>
       <% end %>
+      <%= case @nearby_proposed_locations do %>
```
alicia-mbta, line 31: If there is only one `case`, you could probably change it to an `if`.

alicia-mbta (approved):
Looks good!

apps/site/lib/site_web/templates/fare_transformation/proposed_sales_locations.html.eex
hannahpurcell: I also have to match on empty array vs anything else, and I couldn't figure out a more elegant way to do that 😅

alicia-mbta: Do you mean the "anything else" in line 34? the `end` in line 35 makes it a 'do nothing'. Let me know if I'm not following!

hannahpurcell: Lines 31-35 should only be rendered when `@nearby_proposed_locations` exists as an empty array.  I can't seem to include a matching statement in an .eex if-statement.  For example, `<%= if @nearby_proposed_locations = [] do %>` is not valid.  But I can use "case" to do the match.

Lines 48-58 do a similar thing.  I need to use "case" to weed out @nearby_proposed_locations = [], and then in all other cases the section is rendered.

Does that make sense?

alicia-mbta: Can you use `if Enum.empty?(@nearby_proposed_locations)` maybe?

hannahpurcell: Ohoo yep!!  That works!

PR #854, opened 2021-03-23
Events page renders poorly in IE

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Events listing page renders poorly in IE](https://app.asana.com/0/555089885850811/1200032439670460)

I had to use polyfills to avoid errors.
I also removed the class `fixedsticky` as it was causing issues on IE (and the "stickiness" seems to work even after said removal).

Also, the remaining errors in IE get fixed upgrading `phoenix_live_reload` to at least 1.1.4 (I checked): https://github.com/phoenixframework/phoenix_live_reload/blob/master/CHANGELOG.md

### DIFF
--- a/apps/site/assets/package.json
+++ b/apps/site/assets/package.json
@@ -31,6 +31,7 @@
     "react-redux": "^7.2.0",
     "redux": "^4.0.5",
     "sifter": "^0.6.0",
+    "stickyfilljs": "^2.1.0",
     "swr": "^0.1.18",
     "tether": "^1.4.7",
     "turbolinks": "git+https://github.com/paulswartz/turbolinks.git#82b188e",
--- a/apps/site/assets/js/event-page-setup.js
+++ b/apps/site/assets/js/event-page-setup.js
@@ -1,3 +1,34 @@
+function toggleAttributePolyfill() {
+  if (!Element.prototype.toggleAttribute) {
+    Element.prototype.toggleAttribute = function(name, force) {
+      if (force !== void 0) force = !!force;
+
+      if (this.hasAttribute(name)) {
+        if (force) return true;
+
+        this.removeAttribute(name);
+        return false;
+      }
+      if (force === false) return false;
+
+      this.setAttribute(name, "");
+      return true;
+    };
+  }
+}
+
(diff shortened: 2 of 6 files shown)


hannahpurcell (approved):
Ah, fantastic!  Works great

apps/site/assets/package.json
```diff
@@ -31,6 +31,7 @@
     "react-redux": "^7.2.0",
     "redux": "^4.0.5",
     "sifter": "^0.6.0",
+    "stickyfilljs": "^2.1.0",
```
thecristen, line 34: Is this why it works without the `fixedsticky` class? That's quite great if so!

thecristen (changes requested):
The month links on the left are no longer sticky in IE! But that's a layout thing that I remember fighting with on Friday SO I'd like to add a commit to your branch to change that. So I'm requesting a change for me to make.

apps/site/assets/package.json
alicia-mbta: Yes, a few notes on that:
- `fixedsticky` was in general breaking the whole layout of the page, regardless of the "stickiness" not working for IE. I think we should create a sort-of-follow-up ticket to examine whether `fixedsticky` works _at all_ and/or is breaking other pages in IE (I remember @hannahpurcell mentioned she saw that for accordions? I'm not sure if I remember that correctly). If that's the case, we could get rid of it completely and use this `stickyfill` package instead.

- I just noticed that `stickyfill` has been recently stopped being maintained (for a reason that I don't think is true?). But the other alternative I found ([stickybits](https://github.com/yowainwright/stickybits)) I couldn't get to work 😓

alicia-mbta:
> The month links on the left are no longer sticky in IE! But that's a layout thing that I remember fighting with on Friday SO I'd like to add a commit to your branch to change that. So I'm requesting a change for me to make.

Oh I didn't notice that! Sure, feel free to add changes to this PR.

thecristen (approved):
Let's merge it for now -- I added a merge commit where I resolved one tiny conflict I'd caused so we should be good to go

PR #859, opened 2021-03-26
Add event listings overlay

thecristen:
#### Summary of changes
**Asana Ticket:** [Add event listings overlay](https://app.asana.com/0/385363666817452/1199915858612070/f)

Added a new dependency, [a11y-dialog](https://a11y-dialog.netlify.app), to support this new popup component.
It's dependent on the adding events to calendar ticket, so in the meantime I've added a button to show an example of activating it.
A large outstanding part of this task will involve positioning the popup so it's in a reasonable place next to the event link. I suggest using the [Tether library](http://tether.io) for that if needed.


apps/site/lib/site_web/templates/event/_popup.html.eex
```diff
@@ -0,0 +1,35 @@
...
+
+    <div class="u-flex-container">
+      <%= render "_add_button.html",
+          range: %{start: @event.date, stop: @event.date_end},
+          conn: @conn,
+          class: "popup-version",
+          path: @event.path %>
+      <div class="u-flex-one">
+        <strong>Tuesday, January 5, 2021</strong>
+        <div class="m-event__date-range">6 PM - 8 PM</div>
```
alicia-mbta, line 31: These are hard-coded 😱  I assume you meant to read from the actual event?

alicia-mbta (changes requested):
A few comments:

- Looks good :ok_hand: But in IE I noticed that having one overlay open and then clicking on another event doesn’t close the previous one (something that happens in Chrome for example).
This is yet another IE issue to address and polyfill in perhaps a separate task; but something to keep in mind.

- Perhaps a design question but shouldn't links open in a new window/tab?

- Could you please add a couple of tests for this new feature?

apps/site/lib/site_web/templates/event/_popup.html.eex
hannahpurcell: OH HA!  That was from the previous round of work, so I totally looked over it

apps/site/assets/js/event-page-setup.js
```diff
@@ -29,9 +31,56 @@ function addInternetExplorerPolyfills() {
...
+  const calendarView = document.querySelector(".m-events-hub--list-view");
+
+  // find all the event popup HTML elements
+  const eventPopups = calendarView.querySelectorAll(".m-event-overlay");
+  for (let i = 0; i < eventPopups.length; i++) {
+    const el = eventPopups[i];
+    const dialog = new A11yDialog(el);
+
+    // Hide popup when user clicks outside popup, or on a different event
+    document.addEventListener("click", e => {
```
paulswartz, line 54: do these get removed at some point? You can listen for the turbolinks events to remove the listeners when the page is torn down.

apps/site/lib/site_web/views/event_view.ex
```diff
@@ -59,42 +61,70 @@ defmodule SiteWeb.EventView do
...
   defp do_render_event_duration(
          %{year: year, month: month, day: day} = start_time,
-         %{year: year, month: month, day: day} = end_time
+         %{year: year, month: month, day: day} = end_time,
+         style
        ) do
-    "#{pretty_date(start_time, "{WDshort}, {Mshort} {D}, {YYYY}")} \u2022 #{
-      format_time(start_time)
-    } - #{format_time(end_time)}"
+    if style === "list" do
```
alicia-mbta, line 104: I wonder if we should do this logic on the template side and have two 'render event' functions; as in, "if we're in the calendar view, format duration this way; otherwise, format it this other way". Given the urgency to merge this, maybe this is not that important.

apps/site/assets/js/event-page-setup.js
```diff
@@ -29,9 +31,56 @@ function addInternetExplorerPolyfills() {
...
+      const path = e.composedPath();
+      const eventClicked = path.find(
+        x => x.className === "m-event-calendar__event"
```
paulswartz, line 57: I don't remember: do these files get compiled down so they use an earlier version of JS? Otherwise I'd suspect that older browsers don't support this syntax.

alicia-mbta (approved):
Overall it looks good. This could use a test for the overlay itself but it seems like that will be addressed in a future ticket.

apps/site/lib/site_web/views/event_view.ex
```diff
@@ -59,42 +61,70 @@ defmodule SiteWeb.EventView do
...
-    "#{pretty_date(start_time, "{WDshort}, {Mshort} {D}, {YYYY}")} \u2022 #{
-      format_time(start_time)
-    }"
+  @spec do_render_event_duration(
+          NaiveDateTime.t() | DateTime.t(),
+          NaiveDateTime.t() | DateTime.t() | nil,
+          String.t()
+        ) :: String.t() | %{date: String.t(), time: String.t()}
+  defp do_render_event_duration(start_time, nil, style) do
+    if style === "list" do
```
paulswartz, line 87: Probably should be an atom `:list` since you're using it as a way of distinguishing these two cases. However, maybe it's worth making two separate functions instead? `render_event_duration_calendar` and `render_event_duration_list`?
```elixir
def render_event_duration_list(start_time, end_time) do
  calendar = render_event_duration_calendar(start_time, end_time)
  "#{calendar.date} \u2022 #{calendar.time}"
end
...
```

apps/site/lib/site_web/views/event_view.ex
alicia-mbta: Also please remember to squash and organize all your commits.

hannahpurcell: That's exactly what hung me up for so long yesterday -- I wanted to avoid this if-else situation, but also I couldn't figure out the elegant way to do it.  Both the list and calendar view need to go through the timezone shift in `render_event_duration`.  If I made a separate 'render event' function, then I would have had to duplicate all 5 functions being used by the list view, and I figured this was cleaner?

apps/site/assets/js/event-page-setup.js
hannahpurcell: huh.  A lot of our JS scripts add event listeners, but I don't see them getting removed.  And I don't see a "turbolinks:destroy" or anything like that getting listened for anywhere else.  I assume we must be handling it differently, but I don't know how

apps/site/assets/js/event-page-setup.js
hannahpurcell: Erm... I'd have to ask my teammates, but I see there is precedent for using the updated syntax in other JS files

apps/site/lib/site_web/views/event_view.ex
hannahpurcell: I'm trying to wrap my head around whether this would do it... dates are formatted differently in the calendar and list views

Screen Shot 2021-04-15 at 11 32 48 AM
Screen Shot 2021-04-15 at 11 32 29 AM

hannahpurcell: I mean the actual text itself is truncated differently, but I can pass the timex formatter as a string.  So maybe that'll work

apps/site/assets/js/event-page-setup.js
```diff
@@ -29,9 +31,64 @@ function addInternetExplorerPolyfills() {
...
+  const eventPopups = calendarView.querySelectorAll(".m-event-overlay");
+  for (let i = 0; i < eventPopups.length; i++) {
+    const el = eventPopups[i];
+    const dialog = new A11yDialog(el);
+
+    // Hide popup when user clicks outside popup, or on a different event
+    document.addEventListener("click", e => hideDialogLogic(e, dialog));
+    document.addEventListener(
+      "turbolinks:before-render",
+      document.removeEventListener("click", e => hideDialogLogic(e, dialog)),
```
paulswartz, line 57: This doesn't quite work, since the two functions aren't the same by identity. You need something like:
```
const listener = e => hideDialogLogic(e, dialog);
document.addEventListener("click", listener);
document.addEventListener("turbolinks:before-render", () => document.removeEventListener("click", listener));
```

apps/site/lib/site_web/views/event_view.ex
```diff
@@ -56,45 +58,77 @@ defmodule SiteWeb.EventView do
...
+    calendar = get_formatted_date_time_map(start_time, nil, "{WDshort}, {Mshort} {D}, {YYYY}")
+    "#{calendar.date} \u2022 #{calendar.time}"
+  end
+
+  def render_event_duration_list_view(
+        %{year: year, month: month, day: day} = start_time,
+        %{year: year, month: month, day: day} = end_time
+      ) do
+    calendar =
+      get_formatted_date_time_map(start_time, end_time, "{WDshort}, {Mshort} {D}, {YYYY}")
```
paulswartz, line 123: Is `formatter` ever a different value?

apps/site/assets/js/event-page-setup.js
hannahpurcell: Oh my goodness, I remember this stuff!  Thanks for catching

apps/site/lib/site_web/views/event_view.ex
hannahpurcell: Not for the list view 👍

alicia-mbta: Since `formatter` doesn't change that much, you could have a default formatter such as:
```elixir
@default_format "{WDshort}, {Mshort} {D}, {YYYY}"
```
Then rewrite the definition as:
```elixir
def get_formatted_date_time_map(start_time, end_time, formatter \\ @default_format)
```
That way you wouldn't have to specify the formatter value every time, only in the call that's in `_popup.html.eex`.

alicia-mbta (changes requested):
My only suggestion is a small renaming to avoid confusion. Otherwise I think it looks good.

apps/site/assets/js/event-page-setup.js
```diff
@@ -29,9 +31,70 @@ function addInternetExplorerPolyfills() {
...
+
+  // find all the event popup HTML elements
+  const eventPopups = calendarView.querySelectorAll(".m-event-overlay");
+  for (let i = 0; i < eventPopups.length; i++) {
+    const el = eventPopups[i];
+    const dialog = new A11yDialog(el);
+
+    // Hide popup when user clicks outside popup, or on a different event
+    function hideDialogLogicListener(event) {
+      console.log("event is ", event);
```
alicia-mbta, line 55: Probably forgotten while working on these changes  :)

alicia-mbta (approved):
Just one line to remove, other than that looks good!

PR #861, opened 2021-03-26
Add a toggle to switch between list and calendar views

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Add toggle between views to navigation](https://app.asana.com/0/555089885850811/1199915858612067)

In desktop mode only, we can now switch between views:

image

### DIFF
--- a/apps/site/lib/site_web/templates/event/index.html.eex
+++ b/apps/site/lib/site_web/templates/event/index.html.eex
@@ -11,11 +11,11 @@
 
       <div class="m-events-hub__nav--navigation-toggle">
         <% index_params = Keyword.delete(params, :calendar) %>
-        <%= link to: event_path(@conn, :index, index_params), role: "button", tabindex: "0" do %>
+        <%= link to: event_path(@conn, :index, index_params), role: "button", tabindex: "0", aria: [label: "List view"] do %>
           <i class="fa fa-list-ul m-nav-toggle-icon <%= if !assigns[:calendar_view], do: "active-view" %>"></i>
         <% end %>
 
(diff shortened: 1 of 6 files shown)


apps/site/lib/site_web/templates/event/index.html.eex
```diff
@@ -7,25 +7,53 @@
...
         year: @year %>
 
-      <%# Add calendar toggle, setting the active option based on if "calendar" query_param is true %>
+      <% params = Enum.map(@conn.query_params, fn({key, val}) -> {String.to_existing_atom(key), val} end) %>
+
+      <div class="m-events-hub__nav--navigation-toggle">
+        <% index_params = Keyword.delete(params, :calendar) %>
+        <%= link to: event_path(@conn, :index, index_params), role: "button", tabindex: "0" do %>
+          <i class="fa fa-list-ul m-nav-toggle-icon <%= if !assigns[:calendar_view], do: "active-view" %>"></i>
+        <% end %>
```
hannahpurcell, line 16: Each button is available to screen readers, but is just called "button".  Give the buttons a descriptive label (I think using `aria-label`?)

hannahpurcell (changes requested):
Ooooo it looks and works soooo nicely!   I'm loving it.  But first, accessibility adjustments

hannahpurcell (approved):
Looks great!  One adjustment that isn't a deal-breaker.  The formatting of the page is a little different with and without `calendar=true` in the URL.

Screen Shot 2021-04-01 at 4 01 47 PM
Screen Shot 2021-04-01 at 4 01 52 PM

alicia-mbta:
Good catch! Fixing that.

hannahpurcell (approved):
Yeah, looks good!

PR #870, opened 2021-04-07
tests: Adjust for CR-Kingston name change, & more

thecristen:
Just adding this to address a few new very minor test fails that popped up.


alicia-mbta (approved):
🙏

PR #872, opened 2021-04-07
Hide years without events

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Years without events should not appear in the year dropdown](https://app.asana.com/0/555089885850811/1200111821763338)

The logic makes sure that years without events are not part of the dropdown so they cannot be selected.

image

### DIFF
--- a/apps/site/lib/site_web/controllers/event_controller.ex
+++ b/apps/site/lib/site_web/controllers/event_controller.ex
@@ -29,11 +29,22 @@ defmodule SiteWeb.EventController do
   end
 
   defp assign_events(conn, _opts) do
+    events_by_year =
+      for year <- year_options(conn), into: %{} do
+        {year, Repo.events_for_year(year)}
+      end
+
+    years_for_selection =
+      events_by_year
+      |> Enum.filter(fn {_year, events_for_that_year} -> Enum.count(events_for_that_year) > 0 end)
+      |> Enum.map(fn {year, _events_for_that_year} -> year end)
+
     conn
+    |> assign(:years_for_selection, years_for_selection)
     |> async_assign_default(
       :events,
       fn ->
-        Repo.events_for_year(conn.assigns.year)
+        Map.get(events_by_year, conn.assigns.year)
       end,
       []
     )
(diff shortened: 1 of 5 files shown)


apps/site/lib/site_web/controllers/event_controller.ex
```diff
@@ -121,4 +125,44 @@ defmodule SiteWeb.EventController do
...
+          :first => Calendar.year(),
+          :last => Calendar.year()
+        }
+  defp do_year_options(year), do: Range.new(year - 4, year + 1)
+
+  @spec get_years_for_selection(Plug.Conn.t()) :: list
+  defp get_years_for_selection(%{assigns: %{events: events}} = conn) do
+    year_range = year_options(conn)
+
+    Enum.map(year_range, fn year ->
```
thecristen, line 152: I think `Enum.filter` where the filtering function returns `Enum.count(events_for_that_year) > 0` would make more sense here

apps/site/lib/site_web/controllers/event_controller.ex
```diff
@@ -121,4 +125,44 @@ defmodule SiteWeb.EventController do
...
+      events_for_that_year =
+        if year == conn.assigns.year do
+          events
+        else
+          Repo.events_for_year(year)
```
thecristen, line 157: Neat approach. Since you end up fetching the events for each year anyway I wonder if it might make more sense to modify the `assign_events` plug to include this function's work. As in,
- do the `Repo.events_for_year` for each of the 5 years
- assign `conn.assigns.events` based on the events corresponding to `conn.assigns.year`, (_or_ maybe assign a variable that's a map of year & events? hmm)
- assign the years for selection in the manner you do here.

apps/site/lib/site_web/views/event_view.ex
```diff
@@ -20,24 +20,12 @@ defmodule SiteWeb.EventView do
...
-    do_year_options(year)
-  end
-
-  def year_options(_) do
-    %{year: year} = Util.now()
-    do_year_options(year)
-  end
+  @doc "Returns a list of years with existing events from conn"
+  @spec years_with_events(Plug.Conn.t()) :: list
+  def years_with_events(%Plug.Conn{assigns: %{years_for_selection: years_for_selection}}),
```
thecristen, line 25: I understand this means years where there aren't no events, but my mind keeps wanting to interpret `years_with_events` as some other data structure containing both years and events. I'm not sure what naming would paint a more accurate picture, but just thought I'd flag it here though 😆

thecristen (commented):
For your consideration!

hannahpurcell (approved):
Looks good to me, pending Cristen's notes!

thecristen (approved):
The updates look good to me!

PR #874, opened 2021-04-08
Expand months in past years

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Months in past years should be expanded by default](https://app.asana.com/0/555089885850811/1200111821763336)

Whenever we want to display events from years other than the current year, those events per month will be all expanded.

image

### DIFF
--- a/apps/site/lib/site_web/templates/event/_event_teaser.html.eex
+++ b/apps/site/lib/site_web/templates/event/_event_teaser.html.eex
@@ -1,0 +1,30 @@
+<% path_fn = fn -> cms_static_page_path(@conn, @event_teaser.path) end %>
+<% range = %{start: @event_teaser.date, stop: @event_teaser.date_end} %>
+<% event_is_hidden = if @check_event_ended, do: event_ended(range), else: false %>
+<!-- If date is upcoming, display on init -->
+<!-- Otherwise, add a 'hidden' class. On previous-button click, this class is removed -->
+<li class="list-group-item u-flex-container m-event <%= if event_is_hidden, do: 'hidden' %>" data-group="<%= render_event_month_slug(@month_number, @year) %>">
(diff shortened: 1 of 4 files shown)


apps/site/lib/site_web/templates/event/_list_of_events.html.eex
```diff
@@ -2,47 +2,39 @@
...
-            <% else %>
-              <div class="m-event__ended-message">Ended</div>
-            <% end %>
-          </div>
-        </li>
+
+      <%= if Util.now().year == @year do %>
+
+        <!-- Previous months button. A JS script sets up a listener for button click -->
+        <%= if Enum.any?(event_teasers, fn event -> event_ended(%{start: event.date, stop: event.date_end}) end) do %>
```
thecristen, line 9: What if this was `if Enum.any?(event_teasers, fn event -> event_ended(%{start: event.date, stop: event.date_end}) end) and Util.now().year == @year do` instead of wrapping the entire section in the `if Util.now().year == @year` logic? Might that simplify the rest of the logic here?

apps/site/lib/site_web/templates/event/_event_teaser.html.eex
```diff
@@ -0,0 +1,28 @@
+<% path_fn = fn -> cms_static_page_path(@conn, @event_teaser.path) end %>
```
thecristen, line 1: Thanks for extracting this! :)

apps/site/lib/site_web/templates/event/_list_of_events.html.eex
alicia-mbta: ooh good point! it's indeed simpler.

alicia-mbta:
This work might get overwritten by the logic in #880 so no need to review for now.

PR #880, opened 2021-04-15
Expandable months

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Past months in current year should be collapsed in accordion](https://app.asana.com/0/555089885850811/1200111821763339)

Every month is now an expandable/collapsible block containing the events:

image


hannahpurcell (approved):
I think it looks really good!  I tried a careful comb-through and didn't catch a thing.

apps/site/lib/site_web/templates/event/_expandable_month.html.eex
```diff
@@ -0,0 +1,28 @@
+<section id="<%= @month_number %>-<%= @year %>" class="m-event-list__month <%= if @month_number == @month, do: 'm-event-list__month--active' %>">
+  <% header_id = "header-#{@month_number}" %>
+  <% panel_id = "panel-#{@month_number}" %>
+
+  <button class="c-expandable-block__link sticky-top sticky-month"
```
thecristen, line 5: I think the `sticky-top sticky-month` classes need to be with the `h2`/`.m-event-list__month-header` on line 12 below. Hopefully that will fix the two regressions I noticed:

* The h2 not getting the `stuck` attribute when the heading sticks to the top of the page, thereby no drop shadow below the header
* The sticky header not hiding/revealing when scrolling up/down (this is on mobile only).

thecristen (changes requested):
Noticed a couple regressions and a potential issue 👁️

apps/site/lib/site_web/templates/event/_event_list.html.eex
```diff
@@ -1,20 +1,29 @@
...
+  <%= render "_year_select.html",
       conn: @conn,
-      events: @events,
       month: @month,
       year: @year %>
+</nav>
 
-</div>
+<%= for {month_number, event_teasers} <- @events do %>
+  <!-- Only expand month if past year or current year and current month -->
```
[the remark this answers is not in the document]
alicia-mbta: I thought the same! but I couldn't find anything on Zeroheight so I was waiting for Yinka's input. To me it makes sense so I can add that change and check with him once he's back.

apps/site/lib/site_web/templates/event/_expandable_month.html.eex
```diff
@@ -0,0 +1,28 @@
...
+    <h2 class="m-event-list__month-header">
+      <div class="c-expandable-block__header-container">
+        <%= render_event_month(@month_number, @year) %>
+        <span class="c-expandable-block-caret--black"></span>
+      </div>
+    </h2>
+  </button>
+  <div
+    class="collapse<%= if @initially_expanded do %> in<% end %> js-focus-on-expand"
+    tabIndex="0"
```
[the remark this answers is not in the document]
alicia-mbta: That was exactly the issue, thanks for noticing that 🙏

apps/site/lib/site_web/templates/event/_expandable_month.html.eex
alicia-mbta: I think the shadow should be back with the latest amendment :)

thecristen: @alicia-mbta I don't see any update?

alicia-mbta: @thecristen Moving the sticky classes to the `h2`'s didn't work (the stickiness was being ignored), so I modified the selector instead:
https://github.com/mbta/dotcom/blob/277576ab5d5119dfa73057afa4cb6f0771c8c135/apps/site/assets/js/event-page-setup.js#L130-L132
and the nesting in the CSS property:
https://github.com/mbta/dotcom/blob/277576ab5d5119dfa73057afa4cb6f0771c8c135/apps/site/assets/css/_events.scss#L55-L60

thecristen (commented):
I see the shadow is back! Thanks for the fix.

There's one more small bug that remains, and that's the behavior where scrolling up reveals the filter dropdowns and remains sticky until the page reaches the top ([see corresponding zero height doc](https://zeroheight.com/2fedee66c/p/10db2c-events-hub)). The current behavior on prod is correct but the behavior in this branch is a little off, can you look into that?

alicia-mbta:
Ok, I think I addressed that last issue in the last commit. Can you see if you still notice anything off? Thanks!

thecristen (approved):
Changes look good!

PR #881, opened 2021-04-16
CMS: posted_on is now datetime string (news and update structs)

amaisano:
#### Summary of changes
**Asana Ticket:** [🐞 RSS | Only one news entry goes to Slack per day; add timestamp](https://app.asana.com/0/553506286097126/1177077206704269)

This change is intended for non-dotcom consumers of the API, so we are simply parsing it differently and converting back into a normal date w/o time, so no downstream changes are necessary.

---

### DIFF
--- a/apps/cms/lib/helpers.ex
+++ b/apps/cms/lib/helpers.ex
@@ -132,6 +132,24 @@ defmodule CMS.Helpers do
     end
   end
 
+  @doc """
+  Temporary handler for CMS API transition. Prior to change,
+  data will hold a date-only value. Following the change,
+  data will hold a datetime string.
+  @deprecated upon next CMS API database update.
+  """
+  @spec parse_posted_on(map) :: Date.t() | nil
+  def parse_posted_on(data) do
+    data
+    |> field_value("field_posted_on")
+    |> parse_iso_datetime()
+    |> do_parse_posted_on(data)
+  end
+
+  # Return a date regardless of input format (time not required for this consumer)
+  defp do_parse_posted_on(nil, data), do: parse_date(data, "field_posted_on")
+  defp do_parse_posted_on(date, _), do: NaiveDateTime.to_date(date)
+
   @spec parse_link(map, String.t()) :: Link.t() | nil
   def parse_link(%{} = data, field) do
     case data[field] do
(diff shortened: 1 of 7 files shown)


hannahpurcell (commented):
Hey!  I'm new to the "news" portions of the site, so I haven't yet trailed the issue all the way to its source.  But I did notice that spinning up the site locally fails with an invalid date error appearing in `partial_view.ex`.   The teasers being passed to the `news_entry` function (line 113) have nil for dates.  Still looking into it, but that's how far I've gotten at the moment

amaisano:
Hi @hannahpurcell - thanks for looking at this. The trick is, this requires database migration on the CMS. So if you run the following backend against your local of this branch, it should work, as the date data has already been converted there:

http://pr-349-mbta.pantheonsite.io

You do bring up a race condition for deployment though...

hannahpurcell:
Ohhh sweet!  Thanks, @amaisano.  I use the link you shared as the DRUPAL_ROOT var in my .envrc?  That did help the site load.  But no news entries appear (line 232 of `apps/cms/lib/repo.ex` returns an error value, passing it to `page_controller.ex`, etc).  That seems unexpected, given there are news items at the pantheonsite

amaisano:
> Ohhh sweet! Thanks, @amaisano. I use the link you shared as the DRUPAL_ROOT var in my .envrc? That did help the site load. But no news entries appear (line 232 of `apps/cms/lib/repo.ex` returns an error value, passing it to `page_controller.ex`, etc). That seems unexpected, given there are news items at the pantheonsite

After chatting on Slack, I think it'll be better to support both pre and post cases for the CMS API data format. We can remove/refactor the code after the CMS API has released the changes. This should now work with a date or datetime string. Please take another look :)

hannahpurcell (commented):
I noticed the featured capital projects on the homepage are missing when using the CMS branch for PR-349.  (They are visible when using the live CMS).  I don't see an error, and I don't believe the datetime update should impact the capital projects.  Adamo suggests it's simply an expired snapshot of the CMS.  I'd assume this is fine, but I'd rather have additional eyes on the PR.  
Screen Shot 2021-04-20 at 11 03 07 AM
Screen Shot 2021-04-20 at 11 09 19 AM

amaisano:
Going to migrate this date transformation to the CMS API, since dotcom has no need for datetime (only other non-JSON consumers do). Sorry for the run-around.

amaisano:
See https://github.com/mbta/cms/pull/440

PR #887, opened 2021-04-20
Fix for Green Line (B branch) test

alicia-mbta:
No ticket, just a test fix.


amaisano:
Does this correlate to a GTFS change?

alicia-mbta:
I think it's because of the changes the Green Line is undergoing:
image

amaisano (approved):
Ah makes sense. Basically they will be skipping that stop during the work.

PR #892, opened 2021-04-21
Add polyfill for `composedPath`

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Overlays don't automatically close in IE](https://app.asana.com/0/555089885850811/1200196372897038)

Polyfill to avoid crashes in IE (although they're not visible/noticeable).

image


thecristen (approved):
That's so cool! Works for me.

PR #899, opened 2021-04-23
Update Event struct to support paragraphs

amaisano:
#### Summary of changes
**Asana Ticket:** [Events | Support paragraph in Elixir (Struct update)](https://app.asana.com/0/553506286097126/1200235953717698)

Add support for upcoming paragraphs-on-events.

---

### DIFF
--- a/apps/cms/test/helpers_test.exs
+++ b/apps/cms/test/helpers_test.exs
@@ -390,6 +390,16 @@ defmodule CMS.HelpersTest do
 
       assert parsed_map == []
     end
+
+    test "it handles absence of paragraphs json data" do
+      map_data = %{
+        "status" => [%{"value" => true}]
+      }
+
+      parsed_map = parse_paragraphs(map_data)
+
+      assert parsed_map == []
+    end
   end
 
   describe "rewrite_url/1" do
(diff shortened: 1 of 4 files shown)


amaisano:
Related to https://github.com/mbta/cms/pull/359

apps/cms/test/helpers_test.exs
```diff
@@ -390,6 +390,16 @@ defmodule CMS.HelpersTest do
...
+
+    test "it handles absence of paragraphs json data" do
+      map_data = %{
+        "status" => [%{"value" => true}]
+      }
+
+      parsed_map = parse_paragraphs(map_data)
+
+      assert parsed_map == []
+    end
```
alicia-mbta, line 402: Question: quite a few tests for `parse_paragraphs` exist already. Shouldn't this test be evaluating `from_api` instead, since it was the function that was modified in this PR?

alicia-mbta (changes requested):

apps/cms/test/helpers_test.exs
amaisano: I think this test is valuable, but I also agree we could use another for `from_api` for this particular struct :)

alicia-mbta: Sorry for the confusion, I didn't mean we should get rid of this test (but perhaps move it to `helpers_test.exs`, where the others live), just add an additional one accounting for the changes you added to `from_api`.

alicia-mbta (approved):

PR #904, opened 2021-04-28
Account for JS disabled in the "View Previous ... Events" button

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [With js disabled "View previous ..." links don't do anything (Firefox on Mac)](https://app.asana.com/0/555089885850811/1200154247705852)

Enclosed the existing logic in a div that will be shown if JS is enabled. Otherwise, this div will be hidden and the `<noscript>` block will be displayed.

### DIFF
--- a/apps/site/assets/js/event-page-setup.js
+++ b/apps/site/assets/js/event-page-setup.js
@@ -181,6 +181,10 @@ export default function() {
   document.addEventListener(
     "turbolinks:load",
     () => {
+      const viewPreviousEventsLink = document.querySelector(
+        ".m-view-previous-events"
+      );
+      viewPreviousEventsLink.classList.remove("hidden");
       if (document.querySelector(".m-events-hub")) {
         const isIE = !!document.documentMode;
         if (isIE) {
@@ -188,9 +192,6 @@ export default function() {
         }
 
         setupEventsListing();
-      }
-
-      if (document.querySelector(".m-events-hub")) {
         setupEventPopups();
       }
     },
--- a/apps/site/assets/js/event-page-setup.js
+++ b/apps/site/assets/js/event-page-setup.js
@@ -99,6 +99,19 @@ function setupEventPopups() {
   }
 }
 
+function MakeCollapsedHeadersSticky(eventsListing) {
+  const sections = eventsListing.querySelectorAll(".m-event-list__month");
+  for (let i = 0; i < sections.length; i++) {
+    const section = sections[i];
+    const header = section.querySelectorAll(".c-expandable-block__link");
+    if (header[0].getAttribute("aria-expanded") == "false") {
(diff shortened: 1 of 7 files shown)


thecristen (approved):
It breaks the month name sticky header (if you disable JS on the current events page, the months still are sticky as you scroll)... it'd be nice if you can fix it here or maybe add a ticket to address that. But it's a fairly minor thing so am approving. :)

apps/site/assets/js/event-page-setup.js
```diff
@@ -181,16 +181,17 @@ export default function() {
   document.addEventListener(
     "turbolinks:load",
     () => {
+      const viewPreviousEventsLink = document.querySelector(
+        ".m-view-previous-events"
+      );
+      viewPreviousEventsLink.classList.remove("hidden");
```
hannahpurcell, line 187: Oh!!  Your question in the dev channel about getting attributes from the body - was it related to this PR?  [Cristen just taught me in the other JS PR](https://github.com/mbta/dotcom/pull/902#discussion_r621592263) that there is a CSS-only way that we can check to see if JS is enabled.  Although this seems to be a succinct solution as well :)

apps/site/lib/site_web/templates/event/_month.html.eex
```diff
@@ -3,20 +3,34 @@
...
+    <noscript>
+      <%= for event_teaser <- @event_teasers do %>
+        <%= render "_event_teaser.html",
+            event_teaser: event_teaser,
+            check_event_ended: false,
+            conn: @conn,
+            month_number: @month_number,
+            year: @year %>
+      <% end %>
+    </noscript>
```
hannahpurcell, line 33: Is there a way to simplify this noscript block with the duplicated else block below?

hannahpurcell (approved):
Just a couple notes/questions on simplification

thecristen:
@alicia-mbta I'm not sure what happened but I'm not able to see events for **previous** months with JS disabled. Jan-April remain collapsed and those events are hidden. I tried on Safari and FF Nightly, disabling JS via the dev tools. The current month works fine though.

alicia-mbta:
@thecristen That is a separate issue not addressed here, accordions do not expand _in general, at all_ when JS is disabled 😩 
(there's [an Asana ticket](https://app.asana.com/0/385363666817452/1200191571993850) to address that).

As far as I know, the conversion of months to expandable blocks doesn't take into account the left menu at all, and actually from the designs, past months are collapsed by default, whereas current and future months for the current year are expanded.

This ticket only deals with 2 things:
- For **the current month and year** (and JS disabled), we will have a block containing the past events instead of the 'expandable' link that says "View previous [month year] events".
- Preserve 'stickiness' for the collapsed headers.

PR #908, opened 2021-04-29
RSL refresh

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [RSL | Update page content and layout](https://app.asana.com/0/555089885850811/1200158362621991)

Updated RSL to match mocks. 
Generic location card html:  refactored PSL and RSL location cards a bit to use a generic component.
Fixed trip planner link:  the PSL trip planner link didn't direct to the correct "to".  Rewrote for the RSL and PSL.
 
---

Also checking the following:


apps/fares/lib/proposed_locations.ex
```diff
@@ -58,7 +58,7 @@ defmodule Fares.ProposedLocations do
     %Location{
       fid: attributes["FID"],
       stop_id: attributes["stop_id"],
-      stop_name: attributes["stop_name"],
+      name: attributes["stop_name"],
```
alicia-mbta, line 61: Just curious, why was the name changed?

apps/site/lib/site_web/templates/fare/retail_sales_locations.html.eex
```diff
@@ -1,74 +1,41 @@
...
+      <% end %>
+    <% end %>
+  </div>
+  <div class="col-md-4">
+    <div class="m-schedule-page__schedule-notes mt-2">
+      <h3 class="mt-0">MBTA Fares</h3>
+      <p>
+        Learn more about fare prices, pass types, and how to pay your fare on the T.
+      </p>
+      <%= link "Learn more about MBTA fares", to: "#", target: "_blank", class: "c-call-to-action" %>
```
alicia-mbta, line 31: This `to` destination has a placeholder and not the real link 😅

apps/site/lib/site_web/templates/fare/retail_sales_locations.html.eex
```diff
@@ -1,74 +1,41 @@
...
+    </div>
+    <div class="m-schedule-page__schedule-notes mt-2">
+      <h3 class="mt-0">Fares Transformation</h3>
+      <p>
+        Fare Transformation will make it easy for you to tap and board at any door with a fare card, smartphone, or contactless credit card.
+      </p>
+      <%= link "Learn more about Fare Transformation", to: "#", target: "_blank", class: "c-call-to-action" %>
```
alicia-mbta, line 38: Same issue with the placeholder.

apps/site/lib/site_web/templates/fare/_nearby_locations.html.eex
```diff
@@ -1,16 +1,55 @@
...
+          <li>
+          <% destination_address = if !is_nil(address), do: address, else: name %>
+            <%= link "View on MBTA Trip Planner", to: trip_plan_path(@conn, :index, %{
+              from: @address,
+              from_latitude: Float.to_string(@search_position.latitude, decimals: 4),
+              from_longitude: Float.to_string(@search_position.longitude, decimals: 4),
+              to: destination_address,
+              to_latitude: Float.to_string(location.latitude, decimals: 4),
+              to_longitude: Float.to_string(location.longitude, decimals: 4),
+            }), target: "_blank", class: "dropdown-item" %>
```
alicia-mbta, line 46: Although clicking on this correctly populates the 'to' address, the reason why we implemented the `/to` feature in the trip planner was to not do it in the frontend (context [here](https://app.asana.com/0/555089885850811/1199926921659622)). Also this won't put the B pin in the map.
I believe it should be:
```elixir
<%= link "View on MBTA Trip Planner", to: trip_plan_path(@conn, :to, "#{destination_address}"), target: "_blank", class: "dropdown-item" %>
```

alicia-mbta (changes requested):
It just needs a couple tweaks, otherwise looks great 💯

apps/site/lib/site_web/templates/fare/retail_sales_locations.html.eex
hannahpurcell: Oh right!!  I have no idea what it's supposed to link to!  😅  I'll ask in slack

apps/site/lib/site_web/templates/fare/_nearby_locations.html.eex
hannahpurcell: The `/to` logic was actually broken for the PSLs, and I didn't notice until trying to implement for RSLs.  [Here's a link to the PSLs near me](https://www.mbta.com/fare-transformation/proposed-sales-locations?latitude=42.340575699999995&longitude=-71.1567228#address-search-input) in West Boston.  Clicking for MBTA directions directs me to https://www.mbta.com/trip-planner/to/South%20Street, which is a location out in Fitchburg

apps/fares/lib/proposed_locations.ex
hannahpurcell: Ah - it wasn't absolutely necessary, so correct me if this was an out-of-scope change.  My reasoning is that since I combined the Proposed Location Card and Retail Location Card into a single Location Card component, I wanted to make the attributes consistent.  So location "name" worked for both, while location "stop_name" only worked for proposed locations.  (But I could see that logic living in the Location Card component instead, if needed.)

apps/site/lib/site_web/templates/fare/_nearby_locations.html.eex
alicia-mbta: mmm I don't think it's necessarily broken, it just takes a guess. You might have just found an edge case in which "South Street" is just too generic by name (other PSL's near you seem to work fine?)

hannahpurcell: Oh true, I shouldn't say broken.  But there are many instances of wrong guesses.  Here is what I've found so far:

Back Bay:  [Station](https://www.mbta.com/stops/place-bbsta) vs [Trip Planner Directions](https://www.mbta.com/trip-planner/to/Back%20Bay)
Assembly:  [Station](https://www.mbta.com/stops/place-astao) vs [Trip Planner Directions](https://www.mbta.com/trip-planner/to/Assembly)
Chinatown: [Station](https://www.mbta.com/stops/place-chncl) vs [Trip Planner Directions](https://www.mbta.com/trip-planner/to/Chinatown)
State:  [Station](https://www.mbta.com/stops/place-state) vs [Trip Planner Directions](https://www.mbta.com/trip-planner/to/State)

There are so many generic station names that really throw off the map predictions of "which location are we talking about?"

alicia-mbta: Well, unfortunately it doesn't seem like we have the specific address name of a PSL available, so the trip planner does as best as it can by name 😅 

But one thing I just thought of is that we could try to make it more accurate by concatenating the municipality, as in
`.../trip-planner/to/{stop name}, {municipality}`  ?  Just a thought but it might help.

hannahpurcell: Mmmmm the trickiness of this problem is resonating with me now.  Thanks for your details, these are helping a lot.

Using municipality is a good lead.  Unfortunately, the Back Bay and Chinatown locations still pose a problem, since the wrong guess and the actual location are in the same municipality 😞 

But I tried something else with the url and might have a good lead.  I found a way to use lat and lng values in the url, and the pin shows up on the map!  One remaining hurdle is that it populates the trip planner "to" input box with a weird value, but I had a geographer friend explain what that value means, and I might have a way to convert it to the stop name!  Let's see...

apps/site/assets/css/_location-card.scss
```diff
@@ -21,65 +18,54 @@
...
+      font-size: $font-size-base-sm;
+    }
+    width: 33%;
   }
 
-  @include media-breakpoint-only(md) {
-    flex: 1 1 33%;
-    padding: .5rem 1rem;
+  &__location {
+    font-size: $font-size-base-xxl;
```
thecristen, line 44: For the font-size and line-height: I'm looking at the mockup and it seems like these are just the standard body font size and line height... do they really need customization here?

apps/site/lib/site_web/templates/fare/_nearby_locations.html.eex
```diff
@@ -1,16 +1,55 @@
...
-    <% end %>
-  </div>
+      <%= if retail_fvm do %>
+        <div class="c-location-card__type">
+          <%= retail_fvm %>
+        </div>
+      <% end %>
+
+      <div class="dropdown show c-location-card__directions">
+        <a href="#" role="button" id="get-directions-dropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
thecristen, line 32: Since one of these gets generated for every location, we'll want a unique `id` here 
(and then should adjust the corresponding CSS to not use the id as the selector: `.c-location-card__directions [data-toggle="dropdown"]` would be a-ok selector to me)

apps/site/assets/css/_location-card.scss
```diff
@@ -99,3 +85,73 @@
...
+}
+
+.m-error-message {
+  color: $error-text;
+  padding: 7px;
+}
+
+.m-top-half {
+  max-width: 600px;
+}
```
thecristen, line 157: These seem only relevant to this page.. might it be useful to start a `.m-sales-locations` class on the overall container (kind of like how we have `.m-schedule-line` covering most of the line diagram page) and scoping these under that? In any case, we use the "m" to indicate a module (which generally covers a larger area and will have lots of bits in it) and I'm not sure that makes sense for these particular bits of CSS.

apps/site/assets/css/_location-card.scss
```diff
@@ -21,65 +18,54 @@
...
-.c-location-card__description {
-  clear: both;
-  color: $gray-dark;
-
-  &:not(:last-child) {
-    margin-bottom: .25rem;
+  &__directions {
+    font-size: $font-size-base-xxl;
+    font-weight: $font-weight-medium;
+    margin-top: $font-size-base;
```
thecristen, line 61: `$base-spacing` might make more sense for the margin-top here?

apps/site/assets/css/_location-card.scss
```diff
@@ -21,65 +18,54 @@
...
-    padding: .5rem 1rem;
+  &__place {
+    font-size: 1.5rem;
+    font-weight: $headings-font-weight;
+    line-height: 1.9rem;
+    @include media-breakpoint-down(xs) {
+      font-size: 1.15rem;
+    }
+    width: 67%;
+    overflow: hidden;
```
thecristen, line 31: I'm not sure if these are still needed since it's in the `justify-content: space-between` area?

thecristen: Ohh I think I see, it's helping make enough room to keep the distance visible? It might make more sense to specify the width for the companion `.c-location-card__distance` instead (or maybe a `white-space: nowrap` there, with a `hyphens: auto` here so long place names won't push the distance out of the container? hmmm)

thecristen (changes requested):
OK, I feel super late to the party but I only just now realized these cards show on the sidebar of the stop pages... we might want some additional CSS to preserve the current appearance _there_.

current:
image

this PR:
image

Otherwise I just have mostly styling comments -- apologies that some of it is probably relevant to code you moved around but otherwise might not have touched 😅

apps/site/assets/css/_location-card.scss
```diff
@@ -1,13 +1,10 @@
...
-  padding: $base-spacing;
+@import 'variables';
 
-  &.empty {
-    display: none;
-  }
+.c-location-cards {
+  display: grid;
+  grid-gap: $grid-gutter-width / 2;
+  grid-template-columns: repeat(auto-fill, [col-start] minmax(220px, 1fr) [col-end]);
```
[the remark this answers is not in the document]
hannahpurcell: :sob: nooooooo, haha.  You're right.  Although it doesn't look too bad - just puts all the locations in one column.  I'll share a screenshot in slack, see what ya think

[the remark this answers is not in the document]
hannahpurcell: Actually it only took a couple lines, so I made it 2 columns 😁  Didn't want to make it look WORSE than it did before, even if it's just in IE, ha

apps/site/assets/css/_location-card.scss
hannahpurcell: Hmm, when I toggle these attributes in the dev tools, the style visibly changes, so it seems that the standard sizes aren't being applied.  I'm not sure what's not connecting, then...  What would you do?

apps/site/lib/site_web/templates/fare/_nearby_locations.html.eex
hannahpurcell: The only place this id is used is in the "aria-labelledby" of the dropdown menu content itself.  I did a little searching, but I'm not totally sure what the best use of aria labelling is.  Do we need this id?  How do I properly set aria labels on the dropdown?

thecristen: Yeah, I think it's good to keep the IDs for this! 
You can (and should!) still use them here for the `aria-labelledby` attribute; I'm suggesting you give the ID a unique name (instead of "get-directions-dropdown" for every single location)

apps/site/assets/css/_location-card.scss
thecristen: Oh, I do see the difference too! It just matches the mockup better to me without these additions, but maybe it looks better for you with these additions... 😅 maybe it's for Yinka to decide anyway.

apps/site/assets/css/_location-card.scss
hannahpurcell: Yessss, ham-jamming it like I did was not the best way of doing that 😅 I adjusted it to use `text-overflow: ellipsis`, which isn't as great as hyphens, but I couldn't get hyphenation to work in a reliable way

apps/site/assets/css/_location-card.scss
hannahpurcell: Ahhh excellent pointer.  For styles that exclusively belong to location cards, but are used in different parts of the app (for example, location cards appear on the stop pages in the "Fares" section") would that be a "c-" or "m-" style?

thecristen: "c-" (for the location card "component"!)

apps/site/lib/site_web/templates/fare/_nearby_locations.html.eex
hannahpurcell: What does aria-labelledby use to make the label?  (Changing the id right now!)

thecristen: You'll have to update the value of `aria-labelledby` to match the ID you're associating it with, I think that answers your question?

apps/site/assets/css/_location-card.scss
hannahpurcell: Thanks!  I tried to clean up the style names throughout this file, mind taking another close look to see if I took the right approach?

apps/site/lib/site_web/controllers/trip_plan_controller.ex
```diff
@@ -61,7 +67,9 @@ defmodule SiteWeb.TripPlanController do
   end
 
   def to(conn, %{"address" => address}) do
-    case TripPlan.geocode(address) do
+    updated_address = Geocode.check_address(address, @options)
```
alicia-mbta, line 70: Just noting here that we might be making extra geolocation calls here and it might have an impact on performance.

apps/site/lib/site_web/templates/fare/_nearby_locations.html.eex
```diff
@@ -1,16 +1,48 @@
...
-    <% end %>
-  </div>
+      <%= if retail_fvm do %>
+        <div class="c-sales-locations__card-type">
+          <%= retail_fvm %>
+        </div>
+      <% end %>
+
+      <div class="dropdown show c-sales-locations__card-directions">
+        <% url_address = if address, do: address, else: "#{location.latitude},#{location.longitude}" %>
```
alicia-mbta, line 32: Is there a scenario in which `url_address` is `address`? I tried different locations and it was always `latitude, longitude` 🤷‍♀️

alicia-mbta:
Maybe a question for Yinka, but I noticed that the styling is different from before. As in, how many cards we're showing/how wide they are.
<br/>In prod:
image

In this PR:
image

alicia-mbta (changes requested):
I have a few questions, will defer to Cristen for an extra set of eyes for these latest changes.

apps/site/lib/site_web/templates/fare/_nearby_locations.html.eex
hannahpurcell: For PSLs, I think it'll always be address

alicia-mbta: Gotcha! Thanks, I see it now.

apps/site/assets/css/_location-card.scss
hannahpurcell: Yinka had me make a few tweaks  👍

thecristen (approved):
👍

hannahpurcell:
> Maybe a question for Yinka, but I noticed that the styling is different from before. As in, how many cards we're showing/how wide they are.

Checked with Yinka - he says it's alright because the grid layout will be reworked in a polish ticket

alicia-mbta (approved):

apps/site/lib/site_web/controllers/trip_plan_controller.ex
thecristen: So it looks like `check_address` will result in another `geocode` call only if this 'address' happens to actually be a parsable lat/lon string... luckily that's not a case we expect to encounter often here so I'm hopeful it will be OK performance-wise :pray:

PR #914, opened 2021-05-04
deps: Remove unused sweet_xml

thecristen:
Was poking around Dependabot PRs and realized we don't use this dependency.


alicia-mbta (approved):
🙌

PR #916, opened 2021-05-04
Fix Wallaby test script in CI

thecristen:
#### Summary of changes
**Asana Ticket:** [Wallaby tests don't pass in CI](https://app.asana.com/0/385363666817452/1199728396564607/f)

The script we ran always passed, even when the tests inside fail.

This update fixes the script (thanks @paulswartz) and also updates a few integration tests to bring them to passing. The old tests were mainly outdated because they were using CSS selectors that became invalid after we refactored the customer support form some months ago.


alicia-mbta (approved):
Looks good to me! Thanks for the added tests 🙏

apps/site/test/site_web/integration/support_form_test.exs
```diff
@@ -3,117 +3,114 @@ defmodule CustomerSupportTest do
...
   describe "customer support form" do
     @tag :wallaby
     test "Shows an error for missing type and comments", %{session: session} do
       session
-      |> visit("/customer-support")
-      |> assert_has(@submit_button)
-      |> click(@submit_button)
-      |> assert_has(css(".form-group.has-danger #comments"))
-      |> assert_has(css(".form-group.has-danger #service"))
+      |> Browser.visit("/customer-support")
```
alicia-mbta, line 28: I'm not very familiar with Wallaby, what is the difference between calling `visit` and `Browser.visit`?

alicia-mbta (approved):
Looks good. Also thanks for the additional tests.
I will try to incorporate Wallaby from now on because I don't think I was really using it.

thecristen:
Addressed in #1021

PR #917, opened 2021-05-05
Meeting info in sidebar

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Add Meeting info section to right-hand rail](https://app.asana.com/0/555089885850811/1199945602717945)

Meeting info details moved to sidebar.  The location doesn't match the design perfectly, because the "Location" field on events includes the floor number.  If that was a separate field, then I could tweak the design a little more precisely.

---

### DIFF
--- a/apps/site/assets/css/_event-details.scss
+++ b/apps/site/assets/css/_event-details.scss
@@ -13,4 +13,7 @@
     font-weight: bold;
   }
 
+  .c-paragraph {
+    padding-bottom: $base-spacing * 2;
+  }
 }
--- a/apps/site/assets/css/_utilities.scss
+++ b/apps/site/assets/css/_utilities.scss
@@ -196,6 +196,10 @@
   margin-bottom: 0 !important;
 }
 
+.mb-05 {
+  margin-bottom: $base-spacing / 2 !important;
+}
+
 .mb-1 {
   margin-bottom: $base-spacing !important;
 }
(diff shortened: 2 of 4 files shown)


alicia-mbta (changes requested):
The redesign looks good to me. However, on mobile the blue block disappears? 😨 
Let me know if it's a glitch on my end.

hannahpurcell:
> The redesign looks good to me. However, on mobile the blue block disappears? 😨
> Let me know if it's a glitch on my end.

Ope!  Nope, not a glitch.  Big oversight, lol 😓

thecristen:
What you're seeing is probably the funkiness of how CMS paragraphs work here: IIRC, custom paragraphs get rendered into the template twice - once in the main area (often in a user-specified area of the page) and once in the sidebar - and each is hidden or shown based on media query - this is how sometimes we can have a different order of content on mobile, and how the right side content can appear interwoven into the main content.

thecristen (approved):
This LGTM, with the same note that @alicia-mbta noted. 

TBH I wonder if that noted issue might resolve itself once the overall page layout is fixed to have the same width as the navbar (screenshot from Safari below with issue that shows on prod right now where the content + sidebar seems dramatically narrower than the page)

image

alicia-mbta:
Oh ok good point. I guess we can keep an eye on it.

alicia-mbta (approved):

hannahpurcell:
Ohhoo weird!  Thanks for noticing and discussing!  Do we have a ticket for fixing that bigger layout issue?

PR #919, opened 2021-05-06
Sentry bug | Handle nil stops when building stop markers

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Schedules Map API | Elixir.FunctionClauseError: no function clause matching in SiteWeb.ScheduleController.Line.Maps.do_build_stop_marker/1](https://app.asana.com/0/555089885850811/1200280599638079)

Sometimes, when build_stop_marker tries to find a stop by id from the api, the stop can't be found (even though it was in the stop_id list).  In these rare cases, build_stop_marker will just return nil for that stop, and the calling function (`dynamic_map_data`) will filter out the nils from the stop_marker list.

I also removed the function `dynamic_markers` because there was duplicate code, and removing the extra function layer made the logic clearer and more readable to me.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line/maps.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line/maps.ex
@@ -93,15 +93,17 @@ defmodule SiteWeb.ScheduleController.Line.Maps do
       Enum.flat_map(route_patterns, fn %{stop_ids: stop_ids} -> stop_ids end)
       |> Enum.uniq()
 
-    stop_markers = stop_ids
-    |> Enum.map(&build_stop_marker/1)
-    |> Enum.reject(&is_nil/1)
+    stop_markers =
+      stop_ids
+      |> Enum.map(&build_stop_marker/1)
+      |> Enum.reject(&is_nil/1)
 
-    all_markers = if vehicle_tooltips do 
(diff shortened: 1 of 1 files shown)


thecristen (approved):
This looks _so_ much clearer, thank you!

apps/site/lib/site_web/controllers/schedule/line/maps.ex
```diff
@@ -104,24 +115,11 @@ defmodule SiteWeb.ScheduleController.Line.Maps do
...
-  end
-
   @spec build_stop_marker(Stop.id_t()) :: Marker.t()
   defp build_stop_marker(id) do
-    id
-    |> Repo.get()
-    |> do_build_stop_marker()
+    stop = Repo.get(id)
+    # Only build the stop marker if the api returned a value
+    if stop, do: do_build_stop_marker(stop)
```
alicia-mbta, line 122: what is returned here if `stop` is `nil`?

hannahpurcell: If a return value isn't specified, the return value is nil by default (or at least that's what happens here in this case!)

alicia-mbta (approved):

PR #927, opened 2021-05-11
ical generator makes an ICS file that passes validation

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Events hub | ICS file validation errors](https://app.asana.com/0/555089885850811/1200114012363890)

Most adjustments were slight and formatting-related.  One additional change:  ICS files can't be validated without an end-time, so if there's no end-time, I just added an hour to the start-time.  Of course, this might not always be accurate, like in the case of FMCB meetings.  But I think Google calendar defaults to an hour for new meetings anyway, so does that make sense in our case?

### DIFF
--- a/apps/site/lib/site/icalendar_generator.ex
+++ b/apps/site/lib/site/icalendar_generator.ex
@@ -1,4 +1,9 @@
 defmodule Site.IcalendarGenerator do
+  @moduledoc """
+  Takes event information and generates an ICS file.
+  File validation was checked with https://icalendar.org/validator.html.
+  """
+
   import SiteWeb.CmsRouterHelpers, only: [event_path: 3]
 
   alias CMS.Page.Event
--- a/apps/site/lib/site/icalendar_generator.ex
+++ b/apps/site/lib/site/icalendar_generator.ex
@@ -10,38 +10,19 @@ defmodule Site.IcalendarGenerator do
       "VERSION:2.0\r\n",
       "PRODID:-//www.mbta.com//Events//EN\r\n",
       "BEGIN:VEVENT\r\n",
-      "UID:",
-      "event",
-      "#{event.id}",
-      "@mbta.com",
-      "\r\n",
-      "SEQUENCE:",
-      unix_time(),
-      "\r\n",
-      "DTSTAMP:",
-      timestamp(),
-      "\r\n",
-      "DTSTART:",
-      start_time(event),
-      "\r\n",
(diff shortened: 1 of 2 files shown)


thecristen (commented):
:tada: So cool!

I noticed with longer event descriptions it generates some validation warnings about [lines of text being too long](https://icalendar.org/iCalendar-RFC-5545/3-1-content-lines.html) — it'd be really nice to add some formatting/line wrapping to the descriptions to remove the warnings altogether! 😇

apps/site/lib/site/icalendar_generator.ex
```diff
@@ -6,38 +6,41 @@ defmodule Site.IcalendarGenerator do
...
+      "\r\n",
       "URL:",
       full_url(conn, event),
-      "\n",
-      "END:VEVENT\n",
-      "END:VCALENDAR\n"
+      "\r\n",
+      "END:VEVENT\r\n",
+      "END:VCALENDAR\r\n"
     ]
```
alicia-mbta, line 44: You could also use your new function `replace_newlines` here instead of adding the `\r`:

```elixir
[...]|>Enum.map(fn str -> replace_newlines(str) end)
```

alicia-mbta (approved):
🎉

hannahpurcell:
> I noticed with longer event descriptions it generates some validation warnings about [lines of text being too long](https://icalendar.org/iCalendar-RFC-5545/3-1-content-lines.html)

Added line wrapping!

I wanted to test importing this ICS file in different calendars, but I had trouble with importing into Outlook - but I have equal trouble with importing the old format, too.  Happy to report that the new format can be imported to Google cal, and the old one couldn't!  iCalendar accepted both formats (still does).

One weird thing:  the location data doesn't get read from the ICS file.  It's there, but the calendars don't import it.  Problem?  Future investigation?

apps/site/lib/site/icalendar_generator.ex
```diff
@@ -1,44 +1,33 @@
 defmodule Site.IcalendarGenerator do
+  @moduledoc """
+  Takes event information and generates an ICS file.
+  File validation was checked with https://icalendar.org/validator.html.
```
thecristen, line 4: Thanks for adding this link, super helpful!

thecristen (approved):
Validation looks good! And I think the location issue can be followed up in a separate ticket 👍

PR #930, opened 2021-05-12
Remove 553/554/556/558 from inner express routes

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Trip Planner + Line Pages | update route 553/554/556/558 fares to the local fare $1.70](https://app.asana.com/0/555089885850811/1200267766698596)

Those 4 routes have now a fare of $1.70

image


thecristen (approved):
👍 nice, planned a trip and saw the updated fare.

PR #931, opened 2021-05-12
Account for JS disabled in overlays

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Overlays don't appear when JS disabled](https://app.asana.com/0/555089885850811/1200196372897040)

When JS is disabled, instead of overlays, we'll have a system tooltip and a link to the corresponding event page:

image

### DIFF
--- a/apps/site/assets/css/_events.scss
+++ b/apps/site/assets/css/_events.scss
@@ -271,7 +271,7 @@ $mobile-control-height: 73px;
         background-color: $gray-lightest;
       }
     }
-    
+
     i {
       margin-left: 1px;
     }
@@ -323,13 +323,14 @@ $mobile-control-height: 73px;
     align-items: center;
     color: $gray-dark;
     background-color: $white;
-    text-decoration: none;
     border: 0;
     padding: 0;
-    &:hover, &:focus {
-      color: $brand-primary;
-      cursor: pointer;
+
+    // hide when JS is disabled
+    .no-js &[data-a11y-dialog-show] {
+        display: none;
     }
+
     .fa-circle {
       font-size: xx-small;
       padding-right: 8px;
@@ -347,6 +348,11 @@ $mobile-control-height: 73px;
       white-space: nowrap;
       overflow: hidden;
       text-overflow: ellipsis;
+
+      &:hover, &:focus {
+        color: $brand-primary;
+        text-decoration: none;
+      }
     }
   }
 }
--- a/apps/site/lib/site_web/templates/event/_calendar.html.eex
+++ b/apps/site/lib/site_web/templates/event/_calendar.html.eex
@@ -25,7 +25,7 @@
             <%= if todays_events != nil do %>
               <%= for event <- todays_events do %>
                 <% event_ended = event_ended(%{start: event.date, stop: event.date_end}) %>
-                <button class="m-event-calendar__event" data-a11y-dialog-show="<%= event.id  %>">
+                <button class="m-event-calendar__event" data-a11y-dialog-show="<%= event.id %>">
                   <i class="fa fa-circle <%= if event_ended, do: 'ended' %>" aria-hidden="true"></i>
                   <%= if !event_ended do%>
                     <div class='m-event-calendar__event-time'><%= format_time(event.date) %></div>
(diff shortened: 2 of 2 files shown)


apps/site/lib/site_web/templates/event/_calendar.html.eex
```diff
@@ -38,6 +38,17 @@
...
                   event: event,
                   class: if index >= 4, do: "open-left" %>
+
+                  <noscript>
+                    <%= link to: cms_static_page_path(@conn, event.path), class: "m-event-calendar__event m-event-calendar__event-title", title: event.title, target: "_blank" do %>
+                      <i class="fa fa-circle <%= if event_ended, do: 'ended' %>" aria-hidden="true"></i>
+                      <%= if !event_ended do%>
+                        <div class='m-event-calendar__event-time'><%= format_time(event.date) %></div>
+                      <% end %>
+                      <%= event.title %>
```
thecristen, line 48: Was it intentional to not use `<div class="m-event-calendar__event-title"><%= event.title %></div>` like in the regular display?

apps/site/lib/site_web/templates/event/_calendar.html.eex
```diff
@@ -38,6 +38,17 @@
                   conn: @conn,
                   event: event,
                   class: if index >= 4, do: "open-left" %>
+
+                  <noscript>
+                    <%= link to: cms_static_page_path(@conn, event.path), class: "m-event-calendar__event m-event-calendar__event-title", title: event.title, target: "_blank" do %>
```
thecristen, line 43: Can we either add a warning that this will open in a new window ([see rationale and examples in this linked accessibility guidance](https://www.w3.org/TR/WCAG20-TECHS/G201.html)) or remove the `target: "_blank"` from these links? 

I have no preference for which option to take, but removing the attribute might make more sense since the event links in the JS-enabled popup don't open in a new window anyway.

apps/site/assets/css/_events.scss
```diff
@@ -324,10 +320,21 @@ $mobile-control-height: 73px;
...
+
+      &:hover, &:focus {
+        color: $brand-primary;
+        text-decoration: none;
+      }
     }
   }
 }
 
+button[data-js='enabled'] {
```
thecristen, line 332: Would it work to target the `data-a11y-dialog-show` attribute (e.g. `.m-event-calendar__event[data-a11y-dialog-show]` to select it regardless of value) rather than adding a new attribute for this?

apps/site/assets/css/_events.scss
```diff
@@ -324,10 +320,21 @@ $mobile-control-height: 73px;
...
+  body.no-js & {
+    display: none;
+  }
+}
+
```
thecristen, line 337: This can be more simply written as

```scss
.no-js button[data-js='enabled'] {
  display: none;
}
```

I also think it should be a bit more specific, maybe to the calendar page:

```scss
.no-js .m-event-calendar button[data-js='enabled'] {
  display: none;
}
```

apps/site/lib/site_web/templates/event/_popup.html.eex
```diff
@@ -8,6 +8,7 @@
   id="<%= @id %>"
   aria-label="Event summary for <%= @event.title %>"
   aria-hidden="true"
+  data-js="enabled"
```
thecristen, line 11: I'm not sure this is needed since it won't be seen in the `.no-js` situation anyway?

thecristen (commented):
A few small questions. It works great as far as I can tell!

apps/site/lib/site_web/templates/event/_calendar.html.eex
alicia-mbta: Yes, because `m-event-calendar__event-title` is now being applied to the (parent) `<%= link ...` and so that div was no longer necessary.

apps/site/lib/site_web/templates/event/_calendar.html.eex
alicia-mbta: Sure thing, I'll remove it.

apps/site/assets/css/_events.scss
alicia-mbta: Yes, that looks a lot better 🙏

apps/site/lib/site_web/templates/event/_popup.html.eex
alicia-mbta: Correct; removing it as well 😅

apps/site/lib/site_web/templates/event/_calendar.html.eex
thecristen: Ah, makes sense!

thecristen (approved):
👍

PR #934, opened 2021-05-13
Registration button shows/hides with event status

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Show/hide registration link based on event time](https://app.asana.com/0/555089885850811/1199951623935129)

If a registration link exists on an event, then the event follows the "New Format" and should display the new event design.  Otherwise, show the old event design.  

At the moment, we don't have any events published in the new format.  So for testing, you can edit the definition of `show_event` on line 80 of `event_controller.ex` with something like the following (which just sets a nonfunctional registration link)

```
def show_event(conn, %Event{start_time: start_time} = event) do
    temp_event = %{event | registration_link: "mbta.com"}    
    conn
    |> ControllerHelpers.unavailable_after_one_year(start_time)
    |> assign_breadcrumbs(temp_event)
    |> put_view(EventView)
    |> render("show.html", event: temp_event)
  end
```

Also, `event_ended` was diversified to cover all 3 event statuses:  not_started, started, ended.


apps/site/assets/css/_event-details.scss
```diff
@@ -16,4 +16,46 @@
...
+        width: 100%;
+      }
+      .m-event__registration-header {
+        align-self: unset;
+        margin-bottom: $base-spacing / 2;
+      }
+      .m-event__registration-container {
+        align-items: flex-start;
+      }
+    }
```
thecristen, line 59: It looks like each element in mobile could be displayed with display: block; to get the desired result, would that be simpler than reconfiguring flex box things?

thecristen (changes requested):
Some scattered questions to mull over!

apps/site/lib/site_web/templates/event/_add_button.html.eex
```diff
@@ -1,5 +1,5 @@
 <div class="m-event__status-message <%= @class %>">
-  <%= if !event_ended(@range) do %>
+  <%= if event_status(@range) != :ended do %>
```
alicia-mbta, line 2: Do we want to show the link when the event is `:started` as well?

apps/site/lib/site_web/templates/event/_calendar.html.eex
```diff
@@ -24,8 +24,8 @@
             <% todays_events = Map.get(events_by_day, day.day) %>
             <%= if todays_events != nil do %>
               <%= for event <- todays_events do %>
-                <% event_ended = event_ended(%{start: event.date, stop: event.date_end}) %>
-                <button class="m-event-calendar__event" data-a11y-dialog-show="<%= event.id %>">
+                <% event_ended = event_status(%{start: event.date, stop: event.date_end}) === :ended %>
```
alicia-mbta, line 27: `===` is used more for comparing numbers, so `==` would also work here (no need to change if you don't want to, just something to keep in mind).

apps/site/lib/site_web/templates/event/_event_teaser.html.eex
```diff
@@ -1,7 +1,7 @@
 <%
   range = %{start: @event_teaser.date, stop: @event_teaser.date_end}
   path_fn = fn -> cms_static_page_path(@conn, @event_teaser.path) end
-  event_is_hidden = if @check_event_ended, do: event_ended(range), else: false
+  event_is_hidden = if @check_event_ended, do: event_status(range) === :ended, else: false
```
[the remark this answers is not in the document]
alicia-mbta: Agree -- there are a lot of repeated `event_status(range) === :ended` statements.

apps/site/lib/site_web/views/event_view.ex
```diff
@@ -126,23 +127,36 @@ defmodule SiteWeb.EventView do
     "#{Timex.month_name(month)} #{year}"
   end
 
-  @spec event_ended(%{
+  @spec event_status(%{
```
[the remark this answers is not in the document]
hannahpurcell: I like it!  Halfway through implementation, I realized this also applies to Event Teasers, but Teasers are a generic component not just used for events.  I think a "status" attribute on Teasers maybe wouldn't hurt, but Teasers of non-Events wouldn't use it.  Where's a good place to put the `event_status` function for mutual use in both components?

apps/site/assets/css/_event-details.scss
```diff
@@ -16,4 +16,46 @@
...
+      flex: 0 1 40%;
+      margin: 0;
+      align-self: center;
+    }
+    &-container {
+      flex: 1 0 auto;
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
+      align-items: center;
```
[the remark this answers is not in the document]
hannahpurcell: Ohhoooo SNAP.  So concise, so beautiful :sob:  Uh yeah, that's way better 😅

apps/site/assets/css/_event-details.scss
```diff
@@ -16,4 +16,46 @@
   .c-paragraph {
     padding-bottom: $base-spacing * 2;
   }
+
+  .m-event__section-padding {
```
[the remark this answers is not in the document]
hannahpurcell: Yeah, that'd be a clearer name.  (This class is already sorta the section wrapper, so it should be better named)

apps/site/lib/site_web/templates/event/_add_button.html.eex
hannahpurcell: Ah, hm... The mock didn't include an "in progress" state.  Just button or "ended".  I'll ask Yinka

apps/site/lib/site_web/templates/event/_calendar.html.eex
hannahpurcell: I can be sloppy with my equal signs!  😅  To learn more about the difference, did a lil search and found (this helpful reference)[https://inquisitivedeveloper.com/lwm-elixir-7/#:~:text=Like%20Javascript%2C%20Elixir%20has%20two,and%20triple%20equals%20(%20%3D%3D%3D%20).&text=The%20triple%20equals%20operator%20(%20%3D%3D%3D,are%20the%20exact%20same%20value.].  It seems they differ in strictness level.  So since I'm working with atoms and not floats, I think `===` might be the better fit.  But that was _definitely not_ an educated choice when I made it! :joy:

apps/site/lib/site_web/views/event_view.ex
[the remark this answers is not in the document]
hannahpurcell: For now, I just added it to `event.ex` and import it in `teaser.ex`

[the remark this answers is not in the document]
thecristen: @hannahpurcell That's pretty much what I was going to suggest! :)

apps/cms/lib/page/event.ex
```diff
@@ -96,9 +106,38 @@ defmodule CMS.Page.Event do
...
+  def started_status(%NaiveDateTime{} = start, stop) do
+    started_status(
+      convert_using_timezone(start, ""),
+      if(is_nil(stop), do: nil, else: convert_using_timezone(stop, ""))
+    )
+  end
+
+  def started_status(start, nil) do
+    cond do
+      time_is_greater_or_equal?(now(), start) and Date.compare(now(), start) === :gt -> :ended
```
thecristen, line 124: Since `time_is_greater_or_equal` internally does a `DateTime.compare`, would it make more sense to only look for `DateTime.compare(now(), start) === :gt` in this particular case?

apps/cms/lib/page/event.ex
```diff
@@ -96,9 +106,38 @@ defmodule CMS.Page.Event do
...
+  end
+
+  def started_status(start, stop) do
+    cond do
+      time_is_greater_or_equal?(now(), start) and !time_is_greater_or_equal?(now(), stop) ->
+        :started
+
+      time_is_greater_or_equal?(now(), stop) ->
+        :ended
+
```
thecristen, line 137: Since the conditions are evaluated in order, I wonder if it makes sense to first look for ended, and then started, e.g.

```elixir
cond do
  time_is_greater_or_equal?(now(), stop) -> :ended
  time_is_greater_or_equal?(now(), start) -> :started
  true -> :not_started
end
```

apps/site/assets/css/_event-details.scss
```diff
@@ -16,4 +16,30 @@
...
+    
+    &-header {
+      flex: 0 1 40%;
+      margin: 0;
+      align-self: center;
+    }
+
+    .btn {
+      width: 100%;
+    }
```
thecristen, line 36: Bootstrap has a helper class for this! If you add the `btn-block` class to the button it'll do exactly this for you! 🥳

thecristen (approved):
Great updates! Just have a few tiny suggestions/comments but otherwise 👍

apps/cms/lib/page/event.ex
hannahpurcell: Love all these tips!  Thanks!

alicia-mbta (approved):
Looks good to me 👍

PR #935, opened 2021-05-14
Sentry Bugfix:  Function Clause Error in CMS.Page.Basic.from_api

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Elixir.FunctionClauseError: no function clause matching in CMS.Partial.MenuLinks.from_api/1](https://app.asana.com/0/555089885850811/1199555004763320)

If a CMS page has the attribute field_sidebar_menu, the value should either be populated or an empty array.  But there was a continual issue with the page http://mbta.com/sustainability/community-programs/, in which the field_sidebar_menu value was getting parsed as [nil] (an array with just the value nil.)

In the solution, I was able to adjust the function match for this situation.  Still, I don't know how to determine why that value was getting parsed as [nil] instead of nil or [], since the above link redirects to http://mbta.com/sustainability/, a different page.

### DIFF
--- a/apps/cms/lib/page/basic.ex
+++ b/apps/cms/lib/page/basic.ex
@@ -45,7 +45,7 @@ defmodule CMS.Page.Basic do
   end
 
   @spec parse_menu_links(map) :: MenuLinks.t() | nil
-  defp parse_menu_links(%{"field_sidebar_menu" => [menu_links_data]}) do
+  defp parse_menu_links(%{"field_sidebar_menu" => [menu_links_data]}) when not is_nil(menu_links_data) do
     MenuLinks.from_api(menu_links_data)
   end
 
--- a/apps/cms/lib/page/basic.ex
+++ b/apps/cms/lib/page/basic.ex
@@ -45,7 +45,8 @@ defmodule CMS.Page.Basic do
   end
 
   @spec parse_menu_links(map) :: MenuLinks.t() | nil
-  defp parse_menu_links(%{"field_sidebar_menu" => [menu_links_data]}) when not is_nil(menu_links_data) do
+  defp parse_menu_links(%{"field_sidebar_menu" => [menu_links_data]})
+       when not is_nil(menu_links_data) do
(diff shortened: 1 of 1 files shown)


thecristen:
@amaisano I'm curious if you have any thoughts about this bug!

amaisano:
Let me check the source data. I think that menu field is totally deprecated at this point anyway.

amaisano:
FYI you can access the original page that caused the error by using this link: https://www.mbta.com/sustainability/community-programs-OLD?preview=&vid=19288&nid=3186

amaisano:
The JSON response to elixir for that was https://live-mbta.pantheonsite.io/cms/revisions/3186?vid=19288&_format=json, which indeed shows a null inside an array. This is an error state in the CMS for the field value. The problematic page has been archived, and the [current revision](https://live-mbta.pantheonsite.io/cms/revisions/3186?vid=51654&_format=json) of it is returning a `[]` value.

There is a good chance that the menu object this field was referencing was deleted some time ago. When this happens, the CMS puts a NULL value in but keeps the field placeholder. This is what I think happened.

amaisano:
The JSON response to elixir at the time the page was live indeed showed a null inside an array. This is an error state in the CMS for the field value. The problematic page has been archived, and the current revision of it is returning a [] value.

There is a good chance that the menu object this field was referencing was deleted some time ago. When this happens, the CMS puts a NULL value in but keeps the field placeholder. This is what I think happened.

amaisano:
This is a risk for any entity reference field I think, so we may want to create an abstracted CMS field struct to handle this more generally...

alicia-mbta:
If that field is deprecated, it would be good to remove it entirely.
And create a CMS-related ticket to create that field @amaisano suggested?
Hannah's fix looks good to me. For now, anyway 🙂

alicia-mbta (approved):

thecristen (approved):
I'm going to agree and say this is good to go! Good talk.

PR #937, opened 2021-05-17
Corrected the function match for when old_value is nil

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Map | Elixir.BadMapError: expected a map, got: nil](https://app.asana.com/0/555089885850811/1200334128089894)

When updating the query, we match the old parameters array with the new.  If the new params have a map value, they are merged with old params - BUT, we don't check to make sure that old params _is also_ a map before trying it.  That's fixed.


alicia-mbta (approved):
This looks good as a fix, but we should investigate why we're getting a null value for `schedule_direction`... 🤔 
Also, I had never seen `GooglePlusBot ` 🤷‍♀️

hannahpurcell:
> This looks good as a fix, but we should investigate why we're getting a null value for `schedule_direction`... 🤔
> Also, I had never seen `GooglePlusBot ` 🤷‍♀️

Yes and yes!  What in tarnation is GooglePlusBot?!  Also agree on investigating the null for `schedule_direction`, but I have no idea how to track that down.  We seem to be enabling users to set it to null in the UI (given that so many customers are generating this sentry issue), but gosh, I feel like this could be happening in a number of places  😳 On the other hand... the sentry stack trace did say the issue is happening in a particular template 🤔  maybe I should give that another look!

thecristen:
[exploring the issue a bit](https://sentry.io/organizations/mbtace/discover/results/?field=title&field=release&field=environment&field=user.display&field=timestamp&name=Elixir.BadMapError%3A+expected+a+map%2C+got%3A+nil&project=194344&query=issue.id%3A1654524133&sort=-timestamp&statsPeriod=90d&widths=-1&widths=-1&widths=-1&widths=-1&widths=-1) to see the tags - it's overwhelmingly from Firefox users on Ubuntu, seems kinda suspect to me 🤷‍♀️ though whether it's automated crawlers or not, it is a good fix!

I will chime in and say `schedule_direction` probably merits a little investigation here. IIRC it should be set to a default value when any schedule page loads (if it's not specified via the URL params), so I wonder if something is going on there.

hannahpurcell:
Oh my gosh!  Finally recreated this error in the real world!

This URL is misshapen, but it caused a different 500 error - not the one I was looking for:
`http://localhost:4001/schedules/CR-Worcester/timetable?date=2021-05-19&schedule_direction=`

But THIS URL recreates the exact BadMapError I was looking for.
`http://localhost:4001/schedules/CR-Worcester/timetable?date=2021-05-19&schedule_direction`

Just leave off the equals sign!  The `schedule_direction` parameter should hold a map, which we use to populate a default value for the `direction_id` assigned to conn.  But it doesn't look like we catch the possibility of null `schedule_direction`.  

So, **advice needed:**  I have an idea about handling the nulls -- but would it be better to strip it out as a faulty parameter?  Right now, we don't strip out bogus parameters from schedules urls; the page just simply ignores them.  (I don't know how to do that, so I'm just handling the nulls and will push that fix shortly)

hannahpurcell:
Yay!!  Fixed it, but unsure it was the best path, so marking for re-review 🙇

hannahpurcell:
I found another Sentry issue related to schedule_direction, so I'm going to alter my previous solution to capture a wider issue.  Might even make a new branch

hannahpurcell:
There!  Found a better solution that does NOT alter the URL, but just ignores the bad values.  Added tests, too

PR #938, opened 2021-05-17
Update command to use AWS CLI v2

alicia-mbta:
This is causing authentication errors in Semaphore:
https://semaphoreci.com/mbta/dotcom/servers/docker-test/deploys/4
I had included this fix in PR #693 but it got overwritten somehow 😞

### DIFF
--- a/semaphore/build_push.sh
+++ b/semaphore/build_push.sh
@@ -20,8 +20,8 @@ docker tag $APP:latest $DOCKER_REPO:$awsenv
 docker tag $APP:latest $DOCKER_REPO:git-$githash
 
 # retrieve the`docker login` command from AWS ECR and execute it
-logincmd=$(aws ecr get-login --no-include-email --region us-east-1)
-eval $logincmd
+# AWS CLI v2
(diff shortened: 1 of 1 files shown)


semaphore/build_push.sh
```diff
@@ -20,8 +20,8 @@ docker tag $APP:latest $DOCKER_REPO:$awsenv
 docker tag $APP:latest $DOCKER_REPO:git-$githash
 
 # retrieve the`docker login` command from AWS ECR and execute it
-logincmd=$(aws ecr get-login --no-include-email --region us-east-1)
-eval $logincmd
+# AWS CLI v2
+aws ecr get-login-password | docker login --username AWS --password-stdin ${DOCKER_REPO}
```
thecristen, line 24: Assuming `--password-stdin` means the script stops to let the user type the password... which only makes sense if this script is run manually (as opposed to programmatically / in CI)?  Will that always be the case?

paulswartz: This looks like the password is being piped in as the output of the `aws ecr get-login-password` command.

PR #940, opened 2021-05-18
Expand/collapse months by clicking on the left menu

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Clicking on the nav link on the left should expand the corresponding month](https://app.asana.com/0/555089885850811/1200312808499748)

Clicking on a month on the left menu will expand/collapse the corresponding block.

### DIFF
--- a/apps/site/assets/css/_events.scss
+++ b/apps/site/assets/css/_events.scss
@@ -72,12 +72,6 @@ $mobile-control-height: 73px;
   background-color: $white;
   padding-top: $base-spacing;
 
-  a {
-    display: block;
-    font-size: 18px;
-    margin-bottom: $base-spacing / 2;
-  }
-
   &.sticky-top {
     top: $base-spacing;
   }
@@ -372,4 +366,15 @@ $mobile-control-height: 73px;
 
 .c-expandable-block-caret--black {
   @include expandable-block-caret($color: $black);
+}
+
+.m-month-control {
+  color: $brand-primary;
+  display: block;
+  font-size: 18px;
+  margin-bottom: $base-spacing / 2;
+
+  &:hover {
+    text-decoration: underline;
+  }
 }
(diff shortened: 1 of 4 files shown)


apps/site/assets/css/_events.scss
```diff
@@ -373,3 +367,14 @@ $mobile-control-height: 73px;
...
+.m-month-control {
+  color: $brand-primary;
+  display: block;
+  font-size: 18px;
+  margin-bottom: $base-spacing / 2;
+
+  &:hover {
+    text-decoration: underline;
+  }
+}
```
hannahpurcell, line 380: This wasn't affected by your changes at all, but I just realized that the font-weight for the month links is different than the mocks.  Not on you to fix, but if you happen to be making other changes - mind adding `font-weight: 500` here?

hannahpurcell, line 380: This wasn't affected by your changes at all, but I just realized that the font-weight for the month links is different than the mocks. Not on you to fix, but if you happen to be making other changes - mind adding font-weight: 500 here?

hannahpurcell (commented):
Final polish task, woohoooo!!  I think before I go deep into review, I think this would be good to deploy to blue/green and get Yinka's review.  The Asana ticket doesn't explain what the expected behavior should be.  I interpreted that the nav links should only expand the clicked month and not collapse it.  I could also imagine that clicking on a past month might cause _other_ past months to collapse while the clicked month expands.

thecristen (commented):
This seems really complicated - to be honest I was expecting something more like adding to the existing
```
if (activeMonthElement) activeMonthElement.scrollIntoView();
```
to additionally expand the active month's events if needed (since they're collapsed by default if the active month is in the past).

(The active month was set via URL params, e.g. `/events?month=3&year=2021#3-2021` as navigated to by the month/year dropdowns and the old left menu links - IMO this is what should be controlling/triggering this expansion!)

alicia-mbta:
Giving this another try because calling `scrollIntoView` on the selected month doesn't seem to work as I expected so it might be somewhat more complicated.

alicia-mbta:
Closing this for now, might revisit depending on the decision on #971.

PR #946, opened 2021-05-25
Added livestream to top of event page

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Event | Show live stream video at top of page when event starts](https://app.asana.com/0/555089885850811/1199951623935130)

To test, the event must be started / ended, and must have a registration link (indicating it's an event with the "New Format").  Add the following after line 80 of `event_controller.ex`:
```
event = %{ event | registration_link: "mbta.com", livestream_link: "https://livestream.com/accounts/20617794/events/9667130/player" }
```


apps/site/assets/css/_event-details.scss
```diff
@@ -2,6 +2,9 @@
 
   .c-cms__title {
     margin-top: 2.5rem;
+    @media (min-width: 800px) {
```
alicia-mbta, line 5: [Comment/suggestion for the future] We have media breakpoints defined in `_variables.scss` that you could use here. Looks like
```css
@include media-breakpoint-up(md) { ... }
```
might be what you need here?

apps/site/test/site_web/views/event_view_test.exs
```diff
@@ -218,4 +218,12 @@ defmodule SiteWeb.EventViewTest do
...
     end
   end
+
+  describe "is_started?/2" do
+    test ":not_started value read by function",
+      do: assert(is_started?(%Event{started_status: :not_started}) == false)
+
+    test ":ended value verified",
+      do: assert(is_started?(%Event{started_status: :started}) == true)
+  end
```
alicia-mbta, line 228: 🙏  thanks for adding tests here.

apps/site/lib/site_web/views/event_view.ex
```diff
@@ -130,6 +130,11 @@ defmodule SiteWeb.EventView do
     event.started_status === :ended
   end
 
+  @spec is_started?(Event.t() | Teaser.t()) :: boolean
```
alicia-mbta, line 133: Perhaps a bit nit-picky (feel free to ignore) but would it make sense to name this **has**_started?

alicia-mbta (approved):
Looks good to me 🎉

PR #952, opened 2021-05-27
Revert work supporting multistage Dockerfile... for now

thecristen:
## Why?

The included scripts and consolidated `Dockerfile` mostly work as intended. The added scripts offer a way for us to deploy without relying on Semaphore, and this setup lets us upload our Docker images to Amazon ECR for the first time!

BUT we haven't quite gotten this setup to work in concert with our deployment to dev process. If you're curious you can see it in action on our dev-blue server today -- deploys are successful but the resulting web application only sort of works. :)

THEREFORE I'm opening this PR mostly so we can resume our normal deployment workflow. 

Once merged, auto-deployments from the `master` branch to our dev server should start working again (and I can continue relying on those deployments to deploy to our production server). 🙏 

This itself is temporary, since we are migrating away from Semaphore to Github Actions (in other words, our deployment to dev process is ✨ changing ✨ ).

## In the meantime
 
We'll continue refining our deployment, building upon this work (which we can reinstate by reverting this squashed commit)! I figure can manually deploy to either of our staging servers as we iterate.

Please let me know if you have any questions or concerns.


alicia-mbta (approved):
😞

PR #953, opened 2021-05-27
Sentry:  URL with a faulty direction_id caused breakage

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Line Diagram | Elixir.ArgumentError: argument error](https://app.asana.com/0/555089885850811/1200334263746937)

I think all possible ill-formed direction_ids from the URL should be filtered out now!  Added tests to catch all those cases.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line.ex
@@ -81,12 +81,11 @@ defmodule SiteWeb.ScheduleController.Line do
       schedule_direction = Map.get(conn.query_params, "schedule_direction", %{})
 
       direction_id_value =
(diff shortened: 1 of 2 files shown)


thecristen (changes requested):
Can you add a test to make sure it handles negative numbers too? e.g. if if the param happened to include a hyphen in front of the number 😨

hannahpurcell:
> Can you add a test to make sure it handles negative numbers too? e.g. if if the param happened to include a hyphen in front of the number 😨

Ahaaaaa the missed case!  I chose to set it to ignore negative direction_ids in the url and just go with the default

thecristen (approved):
Works for me!

PR #955, opened 2021-05-28
Events:  Updated calendar link

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Event | Update "Add to calendar" link logic](https://app.asana.com/0/555089885850811/1199945602717946)

Review the livestream PR first -- this is dependent on that one.

On Event Pages, the Add to Calendar link is only visible before the event starts (not during or after).  
In the Events Hub, the Add to Calendar link is visible before the event starts, Ended is displayed after it ends, and no button displays while the meeting is in progress.  (An "In progress" state will be designed in the future.)

To test, events need a registration link.  Add this after line 80 of `event_controller.ex`.

```
    event = %{ event | registration_link: "mbta.com" }
```


apps/site/assets/css/_event-details.scss
```diff
@@ -39,3 +42,9 @@
     }
   }
 }
+
+.m-event__livestream {
+  width: 100%;
+  height: 56vw;
+  max-height: 560px;
```
alicia-mbta, line 49: Does `max-height` have the same value for mobile?

hannahpurcell: Oh good question - I had to check, and yep.  It works totally fine because the video scales to the viewport size.

alicia-mbta (approved):

PR #962, opened 2021-06-02
Multistage docker reiteration

alicia-mbta:
Copy of #693 to build on top of.


ryan-mahoney:
Actually, I think I referenced the source file and not the build, testing that now.

ryan-mahoney:
Had some errors with CI running last night, but it seems to be working now. I am getting this error:

```Error: No such container:path: 44ef5b773941b86bc875d85f278c052320d699b850dc58ea25905329e9175ae7:/root/apps/site/react_renderer/dist/app.js
```

Which is think could be because this file is not being built by the build process, I just had it locally because I ran the site locally. Will investigate.

alicia-mbta:
The work was in the end done here: https://github.com/mbta/dotcom/pull/975

PR #963, opened 2021-06-02
Internet Explorer warning

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Implement IE browser warning](https://app.asana.com/0/555089885850811/1200138787176597)

Show a warning suggesting to switch to a different browser other than Internet Explorer:

ie


hannahpurcell (changes requested):
Ooooo, it's been a bit since I've reviewed a React file!  

It looks like the modal doesn't persist across the site (at least not when I test it with BrowserStack, so maybe it's a BS problem [hehe]).  Whatever page I land on initially will show the modal, but if I navigate to another page, the modal disappears.

alicia-mbta:
Did you click on the 'Ok, got it!' button? Once you click on it, it should not appear again for the rest of your session.
From the specs, this was so the overlay would remain unobtrusive when collapsed (so you can still use the page).

hannahpurcell:
No, didn't click the "got it" button.  I just clicked a link to another page (schedules), and the warning goes away.  I've tested with a few other links, and the warning doesn't persist for me.  Does it stay for you?

thecristen:
Hm, hadn't thought about how to persist the unanswered warning when clicking around the website. Maybe we can track whether the IE message was dismissed by using a cookie - see `SiteWeb.Plugs.Cookies` for how we show recently visited routes on the home page.

hannahpurcell (approved):
Looks and works great for me!

thecristen:
Going to merge this for you 🥰

PR #964, opened 2021-06-03
Handle an empty list of RouteStop structs

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Elixir.FunctionClauseError: no function clause matching in Stops.RouteStops.branch/1](https://app.asana.com/0/385363666817452/1200033257113092)

I was unable to reproduce the error but I did force an 'empty list scenario' and added this fix for it.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line/diagram_helpers.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line/diagram_helpers.ex
@@ -140,6 +140,11 @@ defmodule SiteWeb.ScheduleController.Line.DiagramHelpers do
 
   # Splits green branch into a tuple of shared stops and stops that are unique to that branch.
   @spec split_green_branch(RouteStops.t(), direction_id) :: {[RouteStop.t()], [RouteStop.t()]}
+
+  defp split_green_branch(%RouteStops{stops: []}, _direction_id) do
+    {[], []}
+  end
+
   defp split_green_branch(%RouteStops{branch: "Green-E", stops: stops}, _direction_id),
     do: {[], stops}
 
--- a/apps/stops/lib/route_stops.ex
+++ b/apps/stops/lib/route_stops.ex
@@ -46,7 +46,7 @@ defmodule Stops.RouteStops do
   end
 
   @spec from_list([RouteStop.t()]) :: t()
-  def from_list(stops) do
+  def from_list(stops) when stops != [] do
     %__MODULE__{
       branch: branch(stops),
       stops: stops
@@ -59,6 +59,8 @@ defmodule Stops.RouteStops do
     |> Enum.find(first, &green_branch?/1)
     |> Map.get(:branch)
   end
+
+  defp branch(_), do: nil
 
   @spec green_branch?(RouteStop.t()) :: boolean()
   defp green_branch?(%RouteStop{branch: branch}) when is_binary(branch),
--- a/apps/stops/lib/route_stops.ex
+++ b/apps/stops/lib/route_stops.ex
@@ -46,7 +46,7 @@ defmodule Stops.RouteStops do
   end
 
   @spec from_list([RouteStop.t()]) :: t()
-  def from_list(stops) when stops != [] do
+  def from_list(stops) do
     %__MODULE__{
       branch: branch(stops),
       stops: stops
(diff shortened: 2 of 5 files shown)


apps/stops/lib/route_stops.ex
```diff
@@ -46,7 +46,7 @@ defmodule Stops.RouteStops do
   end
 
   @spec from_list([RouteStop.t()]) :: t()
-  def from_list(stops) do
+  def from_list(stops) when stops != [] do
```
hannahpurcell, line 49: I think this will cause the same FunctionClauseError when stops _do_ equal [], just further up the call stack.  If stops is [], then there will be no function match for `from_list/1`.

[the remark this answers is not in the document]
alicia-mbta: I could not reproduce the Green Line issue anymore so I'm a bit in the dark there in terms of troubleshooting that...

I did take a different approach plus a for the frontend too so let me know if this looks a bit more like what you had in mind?

apps/site/lib/site_web/controllers/schedule/line/diagram_helpers.ex
```diff
@@ -140,6 +140,11 @@ defmodule SiteWeb.ScheduleController.Line.DiagramHelpers do
 
   # Splits green branch into a tuple of shared stops and stops that are unique to that branch.
   @spec split_green_branch(RouteStops.t(), direction_id) :: {[RouteStop.t()], [RouteStop.t()]}
+
+  defp split_green_branch(%RouteStops{stops: []}, _direction_id) do
+    {[], []}
+  end
+
```
hannahpurcell, line 147: Not sure if I'm getting my wires tangled, but I think `split_green_branch` gets called by `DiagramHelpers.build_stop_list()` on line 118 of line_api.ex.  But I think the Sentry error is getting triggered the line before, in `LineHelpers.get_branch_route_stops`.

The stack trace:
`branch(stops)` route_stops line 51
`Enum.map(route_stop_groups, &from_list/1)` route_stops line 21
`RouteStops.from_route_stop_groups()`, line/helpers line 65
`LineHelpers.get_branch_route_stops`, line_api line 117

So -- not _totally_ sure in elixir -- but I think the error will happen before `DiagramHelpers.build_stop_list()` is even executed.

I messaged back and forth with Gabe about this weird issue, and he thinks it might be a caching issue with dotcom.  (oh boy...)  The error always happens a few hours after a GTFS deployment.  Gabe hypothesized that "once we deploy a new version [of GTFS], the `If-Modified-Since` for a lot of the [GTFS] resources will be changing. maybe some of dotcom's requests time out, but over time the successful ones get cached, and then eventually it's steady state again".  And he pointed out this has mostly been an issue with the D line (because of all the shuttle weirdness, I think).

Ok so, wow.  Caching is a little out of my league at the moment.  My first idea was to just filter out empty arrays on line 63 of `LineHelpers.get_branch_route_stops`, because I think that's the root problem.  Somewhere in there, a request fails and an empty array is returned instead of the D line stops array.  But if we did that, then would we lose the D branch from the line diagram?!  Ha, that's no good either.  Sorry, comment is too long.  But I'm stumped for the day!

hannahpurcell:
OH!  Just realized -- while the caching situation might be a separate thing worth making a ticket about, we can still resolve the cause of the Sentry issue.  I think your original solution is halfway there.  Add a `when stops != []` guard on `Stops.RouteStops.from_list`, but also make a `def from_list(_)` clause that matches everything else.

thecristen (commented):
Adding my 2 cents to say I think Hannah's right. But I think it'd also work to add that fallback in the `branch` function, e.g.

```
# Stops.RouteStops.branch
defp branch(_), do: nil
```

alicia-mbta:
Yes, that's more or less what I had initially. Let me know if this looks better?

hannahpurcell:
Ahhh!  I'm understanding so much more now, about the issue and about your solution.  Sorry, I wasn't getting it earlier.

To test this fix, I replaced `branch: branch(stops)` on line 51 of route_stops.ex with an empty array [].  There's no longer an elixir error, which is great.  But the Green line diagram goes funky.  Does that need to be resolved, also?  Hypothetically, I can see that issue showing up user-side after a GTFS update.

thecristen:
> But the Green line diagram goes funky.

Hm, but the alternative is a 500 error, right? 🤔 So maybe funky page might be better than broken page.

Out of curiosity I took a similar gander. Assuming `stops` and `branch(stops)` are empty lists, the line diagram page will now just show a map (no line diagram below it). Very interesting. 😌 

I think this might suffice until [the caching fix](https://github.com/mbta/dotcom/pull/982) or maybe some other fix kicks in. 😓

thecristen (approved):
👌🏼

PR #965, opened 2021-06-04
Ensure alerts are for the proper route using its id

alicia-mbta:
#### Summary of changes:
**Asana Ticket:** [⏱️ Schedules | Clicking an alert filter should not change which line page you are on](https://app.asana.com/0/385363666817452/1199671198523838)

E.g. for the case of the Green Line branches, it shouldn't show the generic Green Line alerts but the ones for the particular branch (see link at bottom of screenshot):

image


hannahpurcell (approved):
:chefs-kiss:  Looks good!  Fabulously simple one-liner, and thanks for the test

PR #971, opened 2021-06-09
Expand and scroll to selected month

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Clicking on the nav link on the left should expand the corresponding month](https://app.asana.com/0/555089885850811/1200312808499748)

A different approach in an attempt to achieve the desired behavior. Related to #940.

### DIFF
--- a/apps/site/lib/site_web/templates/event/_event_list.html.eex
+++ b/apps/site/lib/site_web/templates/event/_event_list.html.eex
@@ -13,7 +13,7 @@
 
   <%= for {month_number, event_teasers} <- @events do %>
     <!-- Only expand month if past year OR current year and current-or-future month -->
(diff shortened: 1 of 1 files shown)


hannahpurcell:
The first 5 or so events of the month are visible, but you can't scroll to the end of the event list for a month.  Here's a gif of trying to scroll to the end of April, but the events are covered by the May header

Jun-10-2021 08-30-53

alicia-mbta:
Ouch that's not good...reverting.

hannahpurcell (approved):
Oh nice!  Simple and sweet.  I think the note on the asana ticket said this task would go back into the backlog, but this looks 100% finished to me, right?

thecristen (approved):
LGTM!

PR #982, opened 2021-06-18
Sentry:  Created a "no-cache" option when a dead trip id is found

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Stops | Elixir.KeyError: key :id not found in: %Schedules.Schedule{early_departure?: false, flag?: false, last_stop?: fals...](https://app.asana.com/0/555089885850811/1200280640565186)

Okay, is this janky?  Yeah, I think so!  But I'm glad to have given it my best shot, I think it's one possible solution, and I'm more than happy to go a different direction if you have pointers.

The issue:  some PredictionSchedules were making it to `PredictedSchedule.group_transform/1` with `trip: nil` and no id value, which means it wasn't a proper Prediction OR Schedule.  Something was wrong, and there wasn't a function to catch it.  Also, these nil trip errors only show up on days after GTFS deployments.

The origin was in the request for Schedules that happens in `PredictedSchedule.get/3` (line 28).  This calls `Schedules.Repo.by_route_ids`, which is cached.  Gabe Durazo pointed out to me that Trip IDs are not maintained between GTFS deployments.  Bingo.  We're still trying to use the cached trip IDs but the trip IDs have changed from a deploy.  So line 255 of Schedules.Repo (`trip: trip(trip_id)`) returns a nil trip and breaks stuff down the line.

So... how do I force a new request and ignore the cached stuff?  That's what I tried to add here.


alicia-mbta:
Just adding some comments here.<br/>
I like the changes in `apps/schedules/lib/repo.ex` but I'm not sure about the other two files. I _think_ I remember there were approaches in the past trying to address `nil` trips but they were not successful. Perhaps @thecristen remembers better but if this PR looks like a reasonable approach, I have no issues approving as well 👍

apps/schedules/lib/repo.ex
```diff
@@ -20,18 +20,23 @@ defmodule Schedules.Repo do
...
-    |> cache(&all_from_params/1)
+    |> cache_condition(no_cache)
     |> filter_by_min_time(Keyword.get(opts, :min_time))
     |> load_from_other_repos
   end
 
+  # Will almost always cache, unless the calling function explicitly passes "no_cache"
+  defp cache_condition(params, nil), do: cache(params, &all_from_params/1)
+  defp cache_condition(params, _no_cache), do: all_from_params(params)
+
```
thecristen, line 39: 🤪 the pessimist in me is thinking in the worse cast scenario, someone somewhere might pass in a `no_cache: false` here, which would produce the undesired result.
Might want to tweak these function headers a bit, OR adjust line 23 to define no_cache as a boolean, e.g. something like

```elixir
no_cache = Keyword.get(opts, :no_cache) == true
```

apps/schedules/lib/repo.ex
```diff
@@ -20,18 +20,23 @@ defmodule Schedules.Repo do
...
+    no_cache = Keyword.get(opts, :no_cache)
 
     @default_params
     |> Keyword.put(:route, Enum.join(route_ids, ","))
     |> Keyword.put(:date, Keyword.fetch!(opts, :date) |> to_string())
     |> add_optional_param(opts, :direction_id)
     |> add_optional_param(opts, :stop_sequences, :stop_sequence)
     |> add_optional_param(opts, :stop_ids, :stop)
-    |> cache(&all_from_params/1)
+    |> cache_condition(no_cache)
```
thecristen, line 31: I wonder if there's a way to get at this caching issue earlier, through the initial fetching of schedules. Like if we could get the  [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) header somehow before actually fetching the schedules and trips via the `V3Api.Schedules.all(params)` call, and use that to fetch from cache or not. But then why isn't the `cache` function already doing this? Hm 🤔 🤔

thecristen (approved):
This seems like a pretty reasonable approach to me too -- it might be interesting to deploy onto dev-green for a few days to see if we encounter any surprises.

apps/schedules/lib/repo.ex
hannahpurcell: Right??  I'm a little baffled by that bit.  Maybe we could do a partner programming sesh to look at how caching works?

thecristen: Yeah. Hmm. Can you start a spinoff investigation ticket for this? I'd like to merge in this PR in the meantime (who knows, maybe it's already the whole fix we need!)

hannahpurcell: Great!  Ticket made:  https://app.asana.com/0/385363666817452/1200531793045978

apps/site/lib/site/transit_near_me.ex
```diff
@@ -125,6 +125,15 @@ defmodule Site.TransitNearMe do
       |> expand_route_id()
       |> schedules_fn.(direction_id: direction_id, date: date)
 
+    # if there are any schedules without a trip, we need to ignore the cache
+    # because the trip ids in the cache has probably been changed during a deploy
+    if Enum.any?(schedule_data, fn sched -> Map.get(sched, :trip) === nil end) do
```
paulswartz, line 130: @thecristen this is the issue with the deploy. If the provided date is outside the current GTFS file, `schedule_data` is an `{:error, _}` tuple, which is not an Enumerable.

thecristen: 💡 💡 💡 💡 💡 💡 💡

PR #983, opened 2021-06-18
Uncaught error was preventing some JS from loading

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Events | When you're in a previous year, you can't click to a different year in the drop down](https://app.asana.com/0/555089885850811/1200448490117396)

Turns out, the JS setup function wasn't running all the way because we didn't check to make sure the the "previous events" button existed before trying to alter its classlist.  Simple conditional fixed the whole shebang.


alicia-mbta (approved):
I noticed that and had the same fix in one of my branches/PR's. Thanks for the separate fix, it's probably better this way.

PR #984, opened 2021-06-21
Refactor and simplify line page

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [⏱ Schedule Finder | Modal appears twice](https://app.asana.com/0/555089885850811/1200159550591672)

The page should look unchanged (including mobile), in terms of look and feel, and behavior.
<br/>
But...underneath this there were quite a few changes:

Previously, we had two `ScheduleFinder`s that were being hidden/shown depending on whether the page was being displayed on desktop or mobile. This was causing issues in which sometimes a 'double modal' was appearing. This refactor should get rid of that.

- This page has now 3 'blocks':
    - the map and line diagram.
    - the schedule finder (or schedule note, depending on mode).
    - additional information: news, pdf's, fares, etc...
    
- `SchedulePage` has been then renamed to `AdditionalLineInfo` to match more closely what it contains. Props have been adjusted to only pass what's necessary.


hannahpurcell (changes requested):
Ah, what a great refactor!  At a quick look, this looks clean 🧼 
Just doing a cosmetic review as a first pass, and I see that the Schedule Note isn't appearing for the subways.

alicia-mbta:
Arg, good catch. I think it should be fixed now.

hannahpurcell (approved):
One CSS question, but otherwise - looks great

apps/site/assets/css/_schedule-page.scss
```diff
@@ -1,3 +1,19 @@
...
+  }
+  .m-schedule-page__schedule-finder-or-note {
+    order: 1;
+  }
+  .m-schedule-page__main-content {
+    order: 2;
+  }
+  .m-schedule-page__side-content {
+    order: 3;
+  }
```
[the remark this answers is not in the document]
alicia-mbta: In this case, it's only meant to have an effect on mobile; I used `order` to preserve the layout. So in smaller screens, the order of the 'widgets' should be: schedule finder (or note), then the line diagram, then the additional information (pdf's, fares, etc.). Let me know if you see something different!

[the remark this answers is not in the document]
hannahpurcell: Oh yep!  Perfect, it looks perfect.  Thanks!

thecristen (changes requested):
This is really wonderful!

One thing. I notice when I change direction, both the main section and the additional info on the right side of the page refresh (I can see them flicker). Do you think there might be a way to avoid having the right side news flicker?

https://user-images.githubusercontent.com/2136286/123720624-867a1e00-d852-11eb-95a6-2012a204904d.mov

A second, more minor thing - if you can update this branch, we can take advantage of the new and improved PR checks that were just merged. I think that'd be nice to see 😅

apps/site/assets/css/_schedule-page.scss
```diff
@@ -1,3 +1,28 @@
+@include media-breakpoint-down(sm) {
+  .m-schedule-page {
+    display: flex;
+    flex-direction: column;
+  }
```
thecristen, line 5: This is a really cool use of the `order` attribute to get the result we want on mobile.
Since you're using flexbox I'm wondering if we can simplify the template a bit by removing some of the bootstrap wrappers. 

Currently the page is like this:

```html
<div class="page-section m-schedule-line">
    <div class="row">
        <div class="col-md-12">
           <ul class="c-alert-group">alerts</ul>
        </div>
        <div class="m-schedule-page">
            <div class="col-xs-12 col-md-7 m-schedule-page__main-content"></div>
            <div class="col-xs-12 col-md-5 col-lg-4 col-lg-offset-1 m-schedule-page__schedule-finder-or-note"></div>
            <div class="col-xs-12 col-md-5 col-lg-4 col-lg-offset-1 m-schedule-page__side-content"></div>
        </div>
    </div>
</div>
```

This should look the same and be a lot cleaner:

```html
<div class="page-section m-schedule-line">
    <ul class="c-alert-group">alerts</ul>
    <div class="m-schedule-page row">
        <div class="col-md-7 m-schedule-page__main-content"></div>
        <div class="col-md-5 col-lg-4 col-lg-offset-1 m-schedule-page__schedule-finder-or-note"></div>
        <div class="col-md-5 col-lg-4 col-lg-offset-1 m-schedule-page__side-content"></div>
    </div>
</div>
```

* Since the alerts are always full width we don't really need to manage its layout with any special classes
* We can move the `.row` class to directly wrap the rest of the `.col-` elements
* We can remove `.col-xs-12` generally... this is the default behavior! :)

apps/site/assets/ts/schedule/components/schedule-finder/upcoming-departures/__tests__/TripDetailsTest.tsx
```diff
@@ -69,6 +69,20 @@ describe("TripDetails", () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it("it renders the loading state", () => {
```
thecristen, line 72: Thanks for the additional tests :pray:

thecristen (changes requested):
This is great! I don't know how but I don't see the flickering I mentioned earlier anymore. Cool.

I do have two mild suggestions, the second of which should fix the  failing test:

1) The layout markup suggested in a comment I left
2) A change you recently added in `LineDiagram.tsx` is creating a few linting errors that we should resolve. Some options:
* remove `if (!lineDiagram[0]) return null;` (which I'm pretty sure will remove the error) and hope for the best
* add an eslint comment to disable the `react-hooks/rules-of-hooks` check and hope for the best
* (probably the best option) remove `if (!lineDiagram[0]) return null;`  **and** add a check to `ScheduleDirection.tsx`, so changing that from this
* 
```tsx
{lineState.data && (
     <LineDiagramAndStopListPage
          lineDiagram={lineState.data}
          ...
      />
)}
```

to that

```tsx
{lineState.data && lineState.data[0] && (
     <LineDiagramAndStopListPage
          lineDiagram={lineState.data}
          ...
      />
)}
```

apps/site/assets/css/_schedule-page.scss
alicia-mbta: Wow that is a lot simpler indeed...the latest changes should contain your suggested changes!

thecristen (approved):
LGTM pending test coverage!

alicia-mbta:
To merge after #1009.

PR #990, opened 2021-06-23
PSL type filters

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Implement filtering retail/vending machine](https://app.asana.com/0/555089885850811/1200068680318929)

---


alicia-mbta (changes requested):
Codewise looks good for what it’s trying to do but I wonder if we could avoid refreshing the page every time. Some notes on this:<br/>
Not consistently, but the page sometimes takes me to the very top of the page whenever I click on a ‘pill’ to filter. I assume this is because of the refreshing with every click.
We have all the PSL’s at this point, do we really need to refresh? Could we just filter the _existing_ data?
(from my work on the events hub page, this turned out to be a little more complicated than I thought because of the ‘month menu’ on the left side 😅 so this might be a similar scenario, but I thought I’d mention it in case we might be able to prevent so many refreshes).

Also, hovering over the pills does not change the background color to a light blue but underlines the button text instead. That's from what I saw in Zeroheight but I don't know if Yinka has updated anything.

hannahpurcell:
Ooooo, that's a thought.  I guess I assumed that this was a limitation of Elixir.  It doesn't have state, so I didn't think a tabbed scenario was possible without refresh.  All the toggles I found around the site used the URL to manage the tab state.  Do we have an example of "tabs without refresh"?  Or I guess I could build it in TS if it's not possible, but that's a complexity layer I haven't tried played with yet.  

Also, let's say a user really wants to find the nearest Charlie Retailer PSL to them, but the 10 closest results are all Fare Vending Machines.  The page doesn't have all the PSLs it needs in that case.  So when the user toggles on the Charlie Retailer filter, we'll need to re-request PSLs.

alicia-mbta:
I'm not seeing the 'pills' properly...
image
Do you see the same?

hannahpurcell:
> Do you see the same?

I saw the same at first, until the CSS updates caught up in my browser.  Try using an incognito window?

alicia-mbta:
I had to reinstall everything but I can see the new pills now 👍 <br/>
I checked on mobile and I think we should make sure that they are not "broken" into multiple lines. See here:<br/>
image

thecristen (commented):
Not much to add to Alicia's review, looks good!

apps/site/assets/css/_location-card.scss
```diff
@@ -15,6 +15,20 @@
     margin-top: $base-spacing * 5;
     margin-bottom: $base-spacing * 2;
   }
+  .psl-type-buttons {
+    border-color: $brand-primary;
+    border-style: solid;
+    border-width: 1px;
+    border-radius: 4px;
```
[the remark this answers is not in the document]
hannahpurcell: Oh nifty!!  Fixed it :)

hannahpurcell:
> I checked on mobile and I think we should make sure that they are not "broken" into multiple lines.

Oh goodness, thanks for catching.  Fixed

alicia-mbta:
> 
> Oh goodness, thanks for catching. Fixed

Looks good now! 🎉

alicia-mbta (approved):

apps/site/lib/site_web/controllers/fare_transformation_controller.ex
```diff
@@ -37,7 +37,9 @@ defmodule SiteWeb.FareTransformationController do
         # Return the 10 closest locations, sorted by distance
         nearby_proposed_locations
         |> Util.Distance.sort(position)
-        |> Enum.slice(0, 10)
+        # There are 1178 PSLs. Since we are rendering them as hidden and showing/hiding them based on the filter,
+        # we should really narrow down how many we are passing 
+        |> Enum.slice(0, 50)
```
alicia-mbta, line 42: Why are you selecting 50 if we're only showing 10?

hannahpurcell: Let's say there's only 1 "Charlie Retailer" in the top 10.  Then when the user filters by "Charlie Retailer", only 1 PSL will appear (even though there are plenty other Charlie Retailers just a little further away).  

I use a filter in the JS file to narrow down the 50 to the top 12

thecristen: It's a little arbitrary and not especially sophisticated, but absolutely helpful and in my opinion adequate (at least for now)! 😄

hannahpurcell: Haha oh yeah, not at all sophisticated.  But when JS is disabled and can't filter down the list, it does prevent us from rendering 1100+ cards  😵

alicia-mbta: In that case, I think we should filter in the backend and not in JS...

PR #991, opened 2021-06-24
GitHub Actions | Auto-deploy workflow

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Automate deploys from master every week Thursday at 2 PM](https://app.asana.com/0/385363666817452/1200386237750202)

Wow, thanks Cristen for setting up these workflows  🙌 
This small addition to automate the prod deploy was really simple... too simple?


thecristen (approved):
Hm, the ticket says 2PM but this appears to be for 6AM? But now that I think about it I have no idea when the scheduler thinks time zones are, or if that's even a thing... well anyway I'm happy to have us merge this in and see what happens Thursday! 🤠

.github/workflows/deploy-prod.yml
```diff
@@ -2,6 +2,8 @@ name: Deploy to Production
 
 on:
   workflow_dispatch:
+  schedule:
+  - cron: '0 6 * * 4'
```
hannahpurcell, line 6:
```suggestion
  - cron: '0 18 * * 4'
```

hannahpurcell:
Aha!  I set it for 2am instead of 2pm, lol.  It's the UTC time  😁

PR #992, opened 2021-06-25
Hide fares from CR shuttle trips in Schedule Finder

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [⏱️ Schedule Finder | Don't show normal fare next to shuttle](https://app.asana.com/0/385363666817452/1199903976071291)

showFares was already set up for another component in the same file, so only needed to copy it so it shows up in the ScheduleFinder

### DIFF
--- a/apps/site/assets/js/event-page-setup.js
+++ b/apps/site/assets/js/event-page-setup.js
@@ -199,7 +199,7 @@ export default function() {
       const viewPreviousEventsLink = document.querySelector(
         ".m-view-previous-events"
       );
-      if (viewPreviousEventsLink) { 
+      if (viewPreviousEventsLink) {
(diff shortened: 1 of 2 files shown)


hannahpurcell:
Huh, eslint is catching a lot more issues than it used to, but many are for files we've never updated since the first-ever commit (like address-search.js).  How do I quickly address those linting problems?

thecristen:
@hannahpurcell 😅 Yeah, sorry about that! The short answer is you don't (and I can merge for you). This should be fixed when #993 is merged (which goes back to the old way of linting JavaScript files, which only looked at the files that changed, because as you noticed there were a LOT of existing errors). Hope this helps!

thecristen (approved):
Already implemented feature? Nice.

PR #994, opened 2021-06-28
hotfix: Handle service date correctly

thecristen:
#### Summary of changes
**Asana Ticket:** [Schedule Finder | Says no service on date with service](https://app.asana.com/0/385363666817452/1200532032193761/f)

Even though the `today` parameter was correct (2021-06-28), inputting it into `new Date()` brought it back to yesterday (a Sunday with no service on this and many lines), thanks UTC. 🤦🏻‍♀️ So when we checked if there was service today, today meant yesterday instead of today.

This corrects for that so that `CR-Franklin` can now show upcoming departures instead of the "there is no schedule service today" message.

I also decided to apply the same correction in a couple other locations where I'd similarly used `new Date()`.

### DIFF
--- a/apps/site/assets/js/event-page-setup.js
+++ b/apps/site/assets/js/event-page-setup.js
@@ -199,7 +199,7 @@ export default function() {
       const viewPreviousEventsLink = document.querySelector(
         ".m-view-previous-events"
       );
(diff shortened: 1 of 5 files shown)


alicia-mbta (approved):
LGTM!

PR #998, opened 2021-06-29
Upgrade react_renderer dependencies

thecristen:
No ticket, just part of the ongoing chore of updating dependencies.

I eventually realized that this whole module was kind of simple, just a single `app.js` that webpack builds. And dependabot keeps trying to update each of its dependencies. So I thought... how hard could this be? To just update _everything_?

It turns out it was quite easy, so I'm opening a PR here.

I'm just aiming for not breaking anything! So I've deployed it onto dev blue so I could poke around it and check if anything broke. I invite reviewers to help me with this check. :)

Not sure what to look for? 👇🏼 Read on.

---

### React Renderer in a nutshell

As far as I can tell, the node script takes each of these React components:

https://github.com/mbta/dotcom/blob/d6f1b7c41be84768697108b7b3750e2b1ce3b7ef/apps/site/react_renderer/app.js#L56-L65

and takes in some initial props, and spits out a string containing the generated HTML for each component. This HTML string is eventually served in an Elixir template. 

For users without JS, these components generated by React Renderer are all they will see. For everyone else,  these components are eventually updated with whatever dynamic data and shiny UI interactivity the components call for. 

The UX for these simplified server-rendered components varies widely between components, and probably leaves a lot to be desired in some spots. But for the purpose of this PR, the experience on dev blue should be identical to that seen on production :)

---

### DIFF
--- a/apps/site/react_renderer/package-lock.json
+++ b/apps/site/react_renderer/package-lock.json
@@ -1,216 +1,2813 @@
 {
   "name": "react-renderer",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
-  "dependencies": {
-    "@babel/code-frame": {
-      "version": "7.12.13",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.13.tgz",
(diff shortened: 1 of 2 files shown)


alicia-mbta (approved):
Thanks for this upgrade.
I agree with Hannah, I took a rough look at dev-blue and didn't spot anything out of the ordinary.

PR #1008, opened 2021-07-07
Fix for IE warning logic

alicia-mbta:
A different way of loading the files needed to display the IE warning without making use of Turbolinks. Should go together with/fixes faulty behavior from #963.

### DIFF
--- a/apps/site/assets/ts/ie-warning/__tests__/IEWarningTest.tsx
+++ b/apps/site/assets/ts/ie-warning/__tests__/IEWarningTest.tsx
@@ -1,0 +1,92 @@
+import React from "react";
+import { mount } from "enzyme";
+import * as cookiesModule from "../../../js/cookies";
+import * as IEWarningModule from "../IEWarning";
+import * as IEWarningBannerModule from "../components/IEWarningBanner";
+
+const body = '<div id="ie-warning"></div>';
+
+it("it renders a warning", () => {
+  document.body.innerHTML = body;
+
+  const browserIsIEFn = jest
+    .spyOn(IEWarningModule, "browserIsIE")
+    .mockImplementation(() => true);
+
+  IEWarningModule.default();
+
+  expect(document.body.innerHTML).toEqual(expect.stringContaining("<aside"));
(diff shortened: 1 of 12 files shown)


apps/site/assets/ts/ie-warning/__tests__/IEWarningTest.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const browserIsIEFn = jest
+    .spyOn(IEWarningModule, "browserIsIE")
+    .mockImplementation(() => true);
+
+  IEWarningModule.default();
+
+  expect(document.body.innerHTML).toEqual(expect.stringContaining("<aside"));
+
+  browserIsIEFn.mockRestore();
```
thecristen, line 20: TIL about `.mockRestore()`

apps/site/lib/site_web/templates/layout/app.html.eex
```diff
@@ -70,6 +70,14 @@
         <%= content_tag :main, render(@view_module, @view_template, assigns), id: "main", tabindex: -1 %>
       </div> <%# /container %>
       <%= render __MODULE__, "_footer.html", conn: @conn %>
+
+      <%# Show a warning when browser is Internet Explorer %>
+      <%= if Application.get_env(:site, :dev_server?) do %>
+        <script defer src="<%= "#{Application.get_env(:site, :webpack_path)}/iewarning.js" %>"></script>
+      <% else %>
+        <script defer src="<%= static_url(@conn, "/js/react.js") %>"></script>
```
thecristen, line 78: Is this script tag needed if React is already imported in the IEWarning?

thecristen (approved):
Just a question about a possibly extraneous tag... but it LGTM!

apps/site/lib/site_web/templates/layout/app.html.eex
thecristen: Now that I think about it it'd be nice if we didn't have to load React for everyone who visits 😕

PR #1013, opened 2021-07-09
Item can be a mode or a Route in Alerts pages

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Alerts | Alerts should be for proper route_id when route_id is relevant](https://app.asana.com/0/555089885850811/1200500758083708)

Refining the work done in #965 because it was breaking the alert page for modes. With these changes, all alert pages should not give any errors.


thecristen (approved):
👍🏼

PR #1016, opened 2021-07-15
Expandable blocks in line page for suspended routes

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Schedules | 🐞 Right side accordions don't open on suspended routes](https://app.asana.com/0/555089885850811/1200532824750507)

`renderAdditionalLineInformation` should also be called for suspended routes, so that React _kicks in_ and enables the expand/collapse functionality of the blocks on the right rail.

Also, for these same suspended routes, we only want to show teasers and upcoming holidays, no other information like fares or connections, etc.


thecristen (approved):
Looks solid to me!

hannahpurcell (approved):
Looks great!

PR #1020, opened 2021-07-22
Github Actions: build cache, run all tests on master branch

thecristen:
#### Summary of changes
**Asana Ticket:** [Tech Debt | All CI steps should run on main branch](https://app.asana.com/0/385363666817452/1200549651204429/f)

The first commit is unrelated to the linked Asana ticket. It restructures the `tests.yml` workflow jobs, mainly adding a `build` job that basically does the same as `npm run build` but more methodically and with caching. The whole workflow ends up a bit faster since it's not building the same build several times.
You can see the effect of this change: https://github.com/thecristen/dotcom/actions/runs/1054828882

The second and third commit: lets the `tests.yml` workflow run on pushes to the master branch. Tweaks the job logic so that every test will run on master (whereas in PRs, some tests might get skipped depending on the types of files that have changed). The last commit tweaks this further so that PRs opened by Dependabot should trigger all tests to run.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,15 +85,13 @@ jobs:
     needs: setup
     outputs:
       build-key: ${{ steps.cache-keys.outputs.build }}
-      digest-key: ${{ steps.cache-keys.outputs.digest }}
       assets-key: ${{ steps.cache-keys.outputs.assets }}
     steps:
       - uses: actions/checkout@v2
       - name: Set cache keys
         id: cache-keys
         run: |
-          echo "::set-output name=digest::${{ runner.os }}-digest-cache-${{ hashFiles('apps/site/priv/static/**') }}"
-          echo "::set-output name=build::${{ runner.os }}-build-cache-${{ hashFiles('**/lib/**.ex', 'apps/site/priv/static/**', '**/mix.lock') }}"
+          echo "::set-output name=build::${{ runner.os }}-build-cache-${{ hashFiles('**/lib/**.ex', '**/mix.lock') }}"
           echo "::set-output name=assets::${{ runner.os }}-assets-cache-${{ hashFiles('apps/site/assets/**', 'apps/site/react_renderer/**.js') }}"
       - uses: actions/cache@v2
         with:
@@ -115,7 +113,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: |
-            apps/site/priv/static/js
+            apps/site/priv/static
             apps/site/react_renderer/dist/app.js
           key: ${{ steps.cache-keys.outputs.assets }}
         id: assets-cache
@@ -125,14 +123,8 @@ jobs:
       - run: npm run build
         if: ${{ steps.assets-cache.outputs.cache-hit != 'true' }}
         working-directory: apps/site/react_renderer
-
-      - name: Digest assets # always, as the cache key hash needs these files
-        run: mix phx.digest
-      - name: Cache asset digest
-        uses: actions/cache@v2
-        with:
-          path: apps/site/priv/static
-          key: ${{ steps.cache-keys.outputs.digest }}
+      - run: mix phx.digest
+        if: ${{ steps.assets-cache.outputs.cache-hit != 'true' }}
       
       - name: Cache _build
         uses: actions/cache@v2
@@ -281,15 +273,10 @@ jobs:
             apps/site/assets/node_modules
             apps/site/react_renderer/node_modules
           key: ${{ needs.setup.outputs.nodejs-key }}
-
-      - uses: actions/cache@v2
-        with:
-          path: apps/site/priv/static
-          key: ${{ needs.build_app.outputs.digest-key }}
       - uses: actions/cache@v2
         with:
           path: |
-            apps/site/priv/static/js
+            apps/site/priv/static
             apps/site/react_renderer/dist/app.js
           key: ${{ needs.build_app.outputs.assets-key }}
       - uses: actions/cache@v2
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -61,9 +63,66 @@ jobs:
...
+          key: ${{ runner.os }}-build-${{ hashFiles('**/mix.lock') }}-${{ hashFiles('apps/site/*/package-lock.json') }}
+        id: build-cache
+      - run: mix compile --warnings-as-errors
+        if: steps.build-cache.outputs.cache-hit != 'true'
+
   file_changes:
     name: Before all / Note files changed
     runs-on: ubuntu-latest
+    env:
+      isDevPR: ${{ github.event_name == 'pull_request' && github.actor != "dependabot[bot]" }}
```
alicia-mbta, line 125: Neat!

alicia-mbta (approved):
As far as my knowledge of GitHub Actions goes (which is not a lot) this looks good to me!<br/>
I also liked the tweak to distinguish between a PR from a human and dependabot :D

.github/workflows/tests.yml
```diff
@@ -61,9 +63,66 @@ jobs:
...
+          key: ${{ runner.os }}-digest-${{ hashFiles('apps/site/assets/package-lock.json') }}
+        id: digest-cache
+      - run: mix phx.digest
+        if: steps.digest-cache.outputs.cache-hit != 'true'
+      
+      - name: Cache _build
+        uses: actions/cache@v2
+        with:
+          path: _build
+          key: ${{ runner.os }}-build-${{ hashFiles('**/mix.lock') }}-${{ hashFiles('apps/site/*/package-lock.json') }}
```
paulswartz, line 116: I think you mentioned this in Slack or Asana, but this also needs to hash the files themselves, since they're getting compiled as a part of the next step.

thecristen: d'oh, that's completely right and it slipped my mind! thank you

thecristen: the workflow's been reworked a bit, but I'm hoping `${{ runner.os }}-build-cache-${{ hashFiles('**/lib/**.ex', 'apps/site/priv/static/**', '**/mix.lock') }}` might suffice as a useful cache key for the `_build` folder?

thecristen:
Some more changes added - mainly to rebase and accommodate the new tests added in #1021 . I also extracted out the build step to its own job that includes caching of certain artifacts there. Then I realized running these tests on different ports might be useful...

.github/workflows/tests.yml
```diff
@@ -62,11 +74,75 @@ jobs:
...
+        id: assets-cache
+      - run: npm run webpack:build
+        if: ${{ steps.assets-cache.outputs.cache-hit != 'true' }}
+        working-directory: apps/site/assets
+      - run: npm run build
+        if: ${{ steps.assets-cache.outputs.cache-hit != 'true' }}
+        working-directory: apps/site/react_renderer
+
+      - name: Digest assets # always, as the cache key hash needs these files
+        run: mix phx.digest
```
paulswartz, line 130: I'm not sure this works: the hash of the files is happening in an earlier step, but the relevant files aren't created until this step. Can you see what happens if you change one of the CSS files which gets compiled into `apps/site/priv/static`?

.github/workflows/tests.yml
```diff
@@ -80,14 +156,26 @@ jobs:
...
+          if [[ "${{ github.event_name }}" == "pull_request" ]] && [[ "${{ github.actor }}" != "dependabot[bot]" ]]; then
+              changes () { git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} ; }
+              echo "changed files? $(changes)"
+              echo "::set-output name=js::$(changes | grep -E 'apps/site/assets/js/.*\.jsx?' | xargs)"
+              echo "::set-output name=ts::$(changes | grep -E 'apps/site/assets/ts/.*\.tsx?' | xargs)"
+              echo "::set-output name=scss::$(changes | grep -E 'apps/site/assets/css/.*\.scss' | xargs)"
+              echo "::set-output name=ex::$(changes | grep -E '.*(\.exs?|\.eex)' | xargs)"
+          else
+              echo "Not a PR by a dev. Use placeholder value so all tests run"
+              echo "::set-output name=js::skip"
```
paulswartz, line 171: Can you change the value to "all" instead of "skip"? Since I believe the intent is to run all the tests, not skip all the tests.

apps/site/config/test.exs
```diff
@@ -3,7 +3,7 @@ use Mix.Config
 # We don't run a server during test. If one is required,
 # you can enable the server option below.
 config :site, SiteWeb.Endpoint,
-  http: [port: 4002],
+  http: [port: System.get_env("PORT") || 4002],
```
paulswartz, line 6: Does this work? `System.get_env("PORT")` is going to be a string, not an integer.

thecristen: It does work for the sake of running the app, despite being a string!

.github/workflows/tests.yml
thecristen: Ahh good point - I hadn't thought about the timing, but that's very true that since I'm running `hashFiles` in `step.cache-keys`, the `digest-key` has no files to hash yet. 🤔 And that probably explains why I saw my digest cache failing my expectations on earlier runs. 

I had decided to _always_ run `mix phx.digest` to get around that, but that's a very blunt approach and I can probably adjust the timing/sequence of operations here to get what I need. Time to revise....

thecristen: Simplified this slightly -- I now cache the `apps/site/priv/static` directory and use the cache key I was using for the assets (involving `hashFiles('apps/site/assets/**', 'apps/site/react_renderer/**.js')`). I figure that whenever I need to run the webpack build, I'll need to subsequently re-run `mix phx.digest` as well.

paulswartz (approved):
:cake: This looks good to me! I'd definitely keep an eye out for any weird CI issues in the next few weeks though, in case there are any kinks which need to be ironed out.

PR #1022, opened 2021-07-22
Fix for year not updating properly in Trip Planner selection (left panel)

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [🐞Trip Planner | Fields should clear when new form submitted](https://app.asana.com/0/555089885850811/1150378678618193)

The affected piece of code was mixing JS and jQuery. I updated it so the dreaded `[object Object]` doesn't get added to the selector.

### DIFF
--- a/apps/site/assets/js/datepicker-input.js
+++ b/apps/site/assets/js/datepicker-input.js
@@ -131,17 +131,17 @@ export default class DatePickerInput {
   updateSelect(select) {
     const type = Object.keys(select)[0];
     const options = document.getElementById(this.selectors[type]);
-    const currentOpt = options.querySelector("option[selected='selected']");
+    const currentOpt =
+      options.querySelector("option[selected='selected']") ||
+      options.querySelector("option[selected]");
     const newOpt = options.querySelector(`option[value='${select[type]}']`);
     if (currentOpt) currentOpt.removeAttribute("selected");
     if (!newOpt) {
-      options.append(
-        $(
-          `<option value="${select[type]}" selected="selected">${
-            select[type]
-          }</option>`
-        )
-      );
+      const opt = document.createElement("option");
+      opt.value = select[type];
+      opt.text = select[type];
+      opt.setAttribute("selected", "selected");
+      options.append(opt);
     } else {
       newOpt.setAttribute("selected", "selected");
     }


thecristen:
Can you give me a test case for this? I read the ticket and tinkered with it a bit but I'm still having trouble understanding where the problem appears and what part is supposed to change with this.

alicia-mbta:
> Can you give me a test case for this? I read the ticket and tinkered with it a bit but I'm still having trouble understanding where the problem appears and what part is supposed to change with this.

The way I understood the issue was as follows:
This ticket is from 2019 so if you click on [the link that Ingrid provided](https://www.mbta.com/trip-planner?_utf8=%E2%9C%93&plan%5Bfrom%5D=North+Station&plan%5Bfrom_latitude%5D=42.365577&plan%5Bfrom_longitude%5D=-71.06129&plan%5Bto%5D=South+Station&plan%5Bto_latitude%5D=42.35302&plan%5Bto_longitude%5D=-71.06459&plan%5Btime%5D=arrive&plan%5Bdate_time%5D%5Bhour%5D=9&plan%5Bdate_time%5D%5Bminute%5D=00&plan%5Bdate_time%5D%5Bam_pm%5D=PM&plan%5Bdate_time%5D%5Bmonth%5D=11&plan%5Bdate_time%5D%5Bday%5D=20&plan%5Bdate_time%5D%5Byear%5D=2019&plan%5Bmodes%5D%5Bsubway%5D=false&plan%5Bmodes%5D%5Bsubway%5D=true&plan%5Bmodes%5D%5Bcommuter_rail%5D=false&plan%5Bmodes%5D%5Bcommuter_rail%5D=true&plan%5Bmodes%5D%5Bbus%5D=false&plan%5Bmodes%5D%5Bbus%5D=true&plan%5Bmodes%5D%5Bferry%5D=false&plan%5Bmodes%5D%5Bferry%5D=true&plan%5Boptimize_for%5D=best_route#plan_result_focus), you'll get the error stating that parameters are in the past. However, if you select a valid date (i.e. from today on), you'll still get the error because the 'year' field does not update properly (day and month seemed to update correctly).
Turns out that there was some mix up with jQuery and pure JavaScript that I believe gets fixed with this PR.<br/>

I _now see_ there's an additional trip planner-related issue in the ticket but I think that's related to the actual trip planner logic's way to calculate itineraries. But this PR does not address that one.

thecristen (approved):
Thanks for the clarification - looks good to me!

PR #1024, opened 2021-07-23
Account for `end_of_rating` returning nil

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Elixir.UndefinedFunctionError: function nil.calendar/0 is undefined](https://app.asana.com/0/555089885850811/1199903866102716)

This error happened back in late January and the details are not available in Sentry anymore.
I tried my best to investigate what happened and it looks like `Schedules.Repo.end_of_rating` might return nil in some scenarios. So I added a small tweak to avoid `Date.compare` to break.


thecristen (approved):
👍🏼 nice

PR #1028, opened 2021-07-27
logs: Add log for when schedules don't have trip

thecristen:
#### Summary of changes
**Asana Ticket:** [🐞 Elixir.UndefinedFunctionError: function nil.id/0 is undefined](https://app.asana.com/0/385363666817452/1195676305098336/f)

Also returns a sensible value instead of erroring in Map.new so that the entry can still be added to the map that's being created. 
As far as I can tell it shouldn't cause issues downstream (where a template fetches information from that map).

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
@@ -5,6 +5,8 @@ defmodule SiteWeb.ScheduleController.TimetableController do
   alias Routes.Route
   alias SiteWeb.ScheduleView
 
+  require Logger
+
   plug(SiteWeb.Plugs.Route)
   plug(SiteWeb.Plugs.DateInRating)
   plug(:tab_name)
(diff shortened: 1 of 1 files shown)


apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
```diff
@@ -124,7 +126,7 @@ defmodule SiteWeb.ScheduleController.TimetableController do
           required(:all_stops) => [Stops.Stop.t()]
         }
   def build_timetable(all_stops, schedules) do
-    trip_schedules = Map.new(schedules, &{{&1.trip.id, &1.stop.id}, &1})
+    trip_schedules = Map.new(schedules, &trip_schedule(&1))
```
alicia-mbta, line 129: So much cleaner 👌

alicia-mbta (approved):
I really like this, I think it'll help us troubleshoot better in the future with the addition of the logging.

PR #1029, opened 2021-07-27
Add a plug so that if URI is malformed, the 404 page is rendered

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Elixir.ArgumentError: malformed URI "Potential%20MBTA%"](https://app.asana.com/0/555089885850811/1198971266994640)

I created a simple plug that attempts to check whether `conn.path_info` contains malformed information before calling the router. If so, it redirects to the 404 page.<br/>
Refactor: since I was going to use `render_not_found` in this new plug, I moved it to `SiteWeb.ControllerHelpers`.


thecristen (approved):
This is so cool!

PR #1033, opened 2021-07-29
Account for empty route id

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Schedules Map API | Elixir.CaseClauseError: no case clause matching: %JsonApi{data: %JsonApi.Item{attributes: %{"color" => "DA291C", "descri...](https://app.asana.com/0/555089885850811/1200280548002897)

A small check to avoid a crash when route_id is empty.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line/helpers.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line/helpers.ex
@@ -17,6 +17,8 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
   @typep stops_by_route :: %{String.t() => [Stop.t()]}
 
   @spec get_route(String.t()) :: {:ok, Route.t()} | :not_found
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -18,12 +18,18 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
 
   @spec get_route(String.t()) :: {:ok, Route.t()} | :not_found
   def get_route(route_id) do
```
thecristen, line 20: I feel like pattern matching might be a better fit here - e.g. adding the function `def get_route(""), do: :not_found`

thecristen (changes requested):

thecristen (approved):

PR #1035, opened 2021-08-02
Updated Event Agenda Display

thecristen:
#### Summary of changes
**Asana Ticket:** [[Elixir] Updated agenda display (CSS/HTML)](https://app.asana.com/0/385363666817452/1200184323508002/f)

Building upon @hannahpurcell's initial work, makes the agenda all pretty and stuff.

image

---

### DIFF
--- a/apps/site/assets/.stylelintrc
+++ b/apps/site/assets/.stylelintrc
@@ -39,6 +39,7 @@
       2,
       {
         "ignore": [
+          "param",
           "value"
         ]
       }
--- a/apps/site/assets/css/_event-details.scss
+++ b/apps/site/assets/css/_event-details.scss
@@ -41,10 +41,51 @@
       }
     }
   }
-}
 
-.m-event__livestream {
-  width: 100%;
-  height: 56vw;
-  max-height: 560px;
+  .m-event__livestream {
+    width: 100%;
+    height: 56vw;
+    max-height: 560px;
+  }
+
+  .agenda-topics {
+    counter-reset: topicCounter 0 subCounter 0;
+    padding: 0;
+    list-style: none;
+
+    // All bullets should look like:
+    li:before {
+      display: inline-block;
+      height: 32px;
+      width: 32px;
+      border-radius: 16px;
+      margin-right: 12px;
+      margin-top: 8px;
+      padding-top: 3px;
+      text-align: center;
+      font-weight: 700;
+    }
+    // main bullets
+    > li:before {
+      content: counter(topicCounter, upper-alpha);
+      background-color: $gray;
+      color: $white;
+      font-size: 18px;
+      counter-increment: topicCounter;
+    }
+    // subtopic bullets
+    ol > li:before {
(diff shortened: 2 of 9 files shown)


apps/site/assets/css/_event-details.scss
```diff
@@ -41,10 +43,129 @@
...
+    }
+
+    counter-increment: $counter;
+    display: inline-block;
+    font-weight: 700;
+    height: $bullet-size;
+    line-height: $bullet-size;
+    text-align: center;
+    width: $bullet-size;
+  }
```
alicia-mbta, line 79: I love this and its reusability 💯

apps/site/assets/.stylelintrc
```diff
@@ -39,6 +39,7 @@
       2,
       {
         "ignore": [
+          "param",
```
alicia-mbta, line 42: What does this do?

apps/site/lib/site_web/controllers/customer_support_controller.ex
```diff
@@ -140,8 +140,7 @@ defmodule SiteWeb.CustomerSupportController do
 
   @spec do_submit(Plug.Conn.t(), map) :: Plug.Conn.t()
   defp do_submit(%Plug.Conn{assigns: %{ip_address: {:ok, ip}}} = conn, data) do
-    rate_limit_interval =
-      if Application.get_env(:feedback, :rate_limit, true), do: 60_000, else: 1_000
+    rate_limit_interval = Application.get_env(:feedback, :rate_limit, 60_000)
```
alicia-mbta, line 143: Is this to avoid timeouts when submitting the form (in real life and in the tests)?

alicia-mbta (approved):
Just a few questions (for me to understand) but otherwise it looks so good!

apps/site/lib/site_web/controllers/customer_support_controller.ex
thecristen: It leaves the default rate limit at 60s (so if someone tries to submit the form more than once a minute, they get stopped). Using the env variable lets me override that, which I do for tests (where many tests running very quickly sometimes in parallel will very easily hit the rate limit! so in the config file I set this value to 1s instead)

apps/site/assets/.stylelintrc
thecristen: https://stylelint.io/user-guide/rules/list/indentation/#ignore-inside-parens-param-value

Allows arbitrary indentation of function params. 
TLDR Lets me write this without complaining haha

```scss
@mixin agenda-bullet($bg-color: $gray, $color: $white,
                     $counter, $counter-type: false)
```

apps/site/lib/site_web/controllers/customer_support_controller.ex
thecristen: @amaisano Hm yeah, that was pretty out of scope of me 🤔

PR #1037, opened 2021-08-03
Stops | Elixir.UndefinedFunctionError: function nil.id/0 is undefined

thecristen:
#### Summary of changes
**Asana Ticket:** [Stops | Elixir.UndefinedFunctionError: function nil.id/0 is undefined](https://app.asana.com/0/385363666817452/1200280592452254/f)

This error prominently came up while creating a new map via the `PredictedSchedule.group_transform` function, so I added some more logic to mitigate against those errors.


alicia-mbta (changes requested):
Love this granularity!
Would you mind adding a couple tests for it?

alicia-mbta (approved):
🥇 so good

PR #1043, opened 2021-08-06
Display track number (if available) for CR in map in line pages

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Add arriving track number to vehicle tooltip on the timetable, line diagram, and map when we have it](https://app.asana.com/0/1200650448662357/1199653833933398)

- Mode is now part of the marker info for later checking if it's CR (optional because some markers only contain stop information).
- Moved the polling to get live data from `LineDiagram` to `scheduleDirection` so liveData can be passed to both the line diagram and the map.
- The 'core' of this PR is some frontend work to have liveData available when needed and contrast it with the vehicle data in order to obtain the proper track number (if it's not null) when Commuter Rail and if so, update the tooltip.

Screen Shot 2021-08-06 at 17 00 38


alicia-mbta:
I added a new commit to periodically request the live data so the tooltip can be updated. That way, we can then update the tooltip with the track number in the timetable as well:<br/>
Screen Shot 2021-08-09 at 17 26 27

thecristen (changes requested):
Two things to note, a bug and a question:

### Different vehicles

Not sure what's going on, but I'm having trouble seeing the right vehicles for the _other_ direction. If I switch directions, live vehicles don't show on the timetable at all, and the vehicles for the _original_ direction show on the line diagram.

See example below. Looks like predictions are affected too? Notice how the line diagram originates from Worcester, but the prediction and the vehicle tooltip indicate going _to_ Worcester.

image

### Live data

It's an interesting approach to use the live data that we're already getting through polling! I was wondering if you had considered editing the `marker.tooltip_text` that's already on the vehicle instead (e.g. getting the prediction and adding the track info in Elixir).

I'd poked around `SiteWeb.VehicleChannel` a bit and it looks like if we add the prediction in here (where it's `nil`) it'll insert the track number on the map for us.

https://github.com/mbta/dotcom/blob/085582d82dd267e88c42a36a8eb6b064a8c4b50f/apps/site/lib/site_web/channels/vehicle_channel.ex#L60-L68

And I _think_ we can get the appropriate prediction with the vehicle:

```elixir
prediction =
      Predictions.Repo.all(
        route: vehicle.route_id,
        stop: vehicle.stop_id,
        trip: vehicle.trip_id,
        direction_id: vehicle.direction_id
      ) |> List.first()
```

(now I wonder if that prediction was added as `nil` for a good reason... just hoping to provide some interesting idea here!)

PR #1046, opened 2021-08-10
Add load test for line diagrams

thecristen:
#### Summary of changes
**Asana Ticket:** [Add docs/process around load testing](https://app.asana.com/0/385363666817452/1200761946703394/f)

Enhances our load testing script by

1. Linking to its docs in a relevant spot (in this case the PR template)
2. Updating the Locust dependency (it was very outdated!)
3. Adding additional tests on the line diagram page. 

Much of that script is adapted from the `locustfile.py` in the [API repo](https://github.com/mbta/api/blob/master/load_tests/locustfile.py). Additional logic is added to simulate the multiple JS-driven API calls that happen simultaneously on the page. There is also logic inspecting the request responses for certain values, as we expect to see not only a valid non-error HTTP response, also data! It's a little convoluted but it does okay ✨

### DIFF
--- a/apps/site/load_tests/README.md
+++ b/apps/site/load_tests/README.md
@@ -2,6 +2,8 @@
 
 This directory contains our load testing scripts which use [Locust](https://locust.io).
 
+Load testing can be a helpful way to verify that a server will stay responsive under heavy use.
+
 ## Installing
 
 Follow the [Locust instructions](https://docs.locust.io/en/2.1.0/installation.html) for installing system dependencies (notably `libev` on the macOS).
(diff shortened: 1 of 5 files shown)


apps/site/load_tests/README.md
```diff
@@ -4,13 +4,14 @@ This directory contains our load testing scripts which use [Locust](https://locu
...
 
-Follow the [Locust instructions](https://docs.locust.io/en/stable/installation.html) for installing system dependencies (notably `libev` on the macOS).
+Follow the [Locust instructions](https://docs.locust.io/en/2.1.0/installation.html) for installing system dependencies (notably `libev` on the macOS).
 
 Install required libraries:
 
     cd apps/site/load_tests/
     pip3 install -r requirements.txt
 
+If using `asdf`, you may have to run `asdf reshim python 3.8.5 locust` before using Locust in your shell.
```
arkadyan, line 16: Nit: You might want a blank line between this line and the following header.

apps/site/load_tests/locustfile.py
```diff
@@ -1,6 +1,5 @@
-from locust import HttpLocust, TaskSet, task
-from page import Page
```
arkadyan, line 2: It looks like the `page.py` file is no longer used anywhere and can probably just be deleted.

apps/site/load_tests/locustfile.py
```diff
@@ -46,8 +45,19 @@ def paris(self):
...
+    tasks = {
+        TransitNearMe: 3,
+        LineDiagramCR: 5,
+        LineDiagramBus: 5,
+        LineDiagramSubway: 3,
+        LineDiagramFerry: 1
+    }
+
+    min_wait = 1000  # 1 seconds
+    max_wait = 15000  # 15 seconds
```
arkadyan, line 63: Nit: Missing newline at EOF.

apps/site/load_tests/locustfile.py
```diff
@@ -1,6 +1,5 @@
-from locust import HttpLocust, TaskSet, task
-from page import Page
-
+from locust import HttpUser, TaskSet, task
+from linediagram import LineDiagramTaskSet
 
 class TransitNearMe(TaskSet):
```
arkadyan, line 4: Not related to this commit, so feel free to ignore, but I could see the `TransitNearMe` class moving to its own file.

arkadyan (approved):
This is fun! I added a few comments, but they're only nitpicks and tangents, nothing blocking.

alicia-mbta (approved):
As far as I can see, it looks good to me too!
It was cool to see the charts updating in real time 📈

PR #1049, opened 2021-08-12
Show scheduled time (not predicted) for CR when status is 'on time'

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [⏱ Schedules | "Arrival" time is confusing to customers](https://app.asana.com/0/555089885850811/1200119447648715)

Putting this together in case it's a valid solution.<br/>
I just added a slight modification whenever the commuter rail's status is 'On time'. If so, the time displayed is the one from the scheduled time and not the prediction.

### DIFF
--- a/apps/site/assets/ts/helpers/prediction-helpers.tsx
+++ b/apps/site/assets/ts/helpers/prediction-helpers.tsx
@@ -2,6 +2,7 @@ import React, { ReactElement } from "react";
 import { PredictedOrScheduledTime } from "../__v3api";
 import { isSkippedOrCancelled } from "../models/prediction";
 import { TripPrediction } from "../schedule/components/__trips";
+import { compareStringTimes } from "./date";
 
 const delayForCommuterRail = (
   data: PredictedOrScheduledTime,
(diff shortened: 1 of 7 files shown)


apps/site/assets/ts/helpers/prediction-helpers.tsx
```diff
@@ -27,7 +27,11 @@ export const timeForCommuterRail = (
     return delayForCommuterRail(data, className);
   }
 
-  const time = prediction ? prediction.time : scheduledTime;
+  let time = prediction ? prediction.time : scheduledTime;
+  if (scheduledTime && prediction) {
+    // "On time"
```
thecristen, line 32: I feel like there should be a more granular way to conclude whether the train is "on time" or not? 🤔 Also this would show the `scheduledTime` even if there's a delay of 4 or fewer minutes, which doesn't seem quite desirable. 

I think what might make sense is to show the prediction time only if it happens to be later than the scheduled time.

thecristen (changes requested):

PR #1054, opened 2021-08-17
Avoid Docker error

alicia-mbta:
No ticket. Trying to solve an error I'm seeing when running our Dockerfile:

image

I'm not sure if this is a valid solution so please advice.


thecristen (approved):
Should be good. Thank you so much!

PR #1057, opened 2021-08-19
Refactor: pass channel data to tooltips

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Add arriving track number to vehicle tooltip on the timetable, line diagram, and map when we have it](https://app.asana.com/0/1200650448662357/1199653833933398)

A big-ish refactor to have a non-nil prediction so the vehicle tooltip gets properly built and is the same in all three places:
- The `Timetable` tab.
- The map and the line diagram in the `Schedule & Maps` tab.

Screen Shot 2021-08-19 at 09 45 37

Screen Shot 2021-08-19 at 09 45 58


thecristen:
Once we get #1054 merged in I'd like to deploy this to Dev blue!

thecristen (commented):
image

I was comparing this side by side with what's on prod, they're mostly the same. But interestingly, now the prediction inside the tooltip seems to update more quickly than the vehicle icon location. (in this example, the icon is one stop behind)

(I'll have more comments but they'll have to wait until next day! :))

apps/site/assets/ts/schedule/components/line-diagram/LineDiagramWithStops.tsx
```diff
@@ -2,20 +2,27 @@ import React, { ReactElement } from "react";
...
 import useStopPositions, { RefList } from "./graphics/useStopPositions";
 import StopCard from "./StopCard";
 import { hasPredictionTime } from "../../../models/prediction";
+import { MapMarker } from "../../../leaflet/components/__mapdata";
+
+interface Props {
+  stops: LineDiagramStop[];
+  handleStopClick: (stop: RouteStop) => void;
+  liveData: LiveDataByStop;
+  vehicleMarkers: MapMarker[];
```
thecristen, line 16: The way this component is set up right now, the entire line diagram component tree will re-render every time the channel updates any of the vehicle markers. That's a lot! :sweater

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -202,6 +216,44 @@ const ScheduleDirection = ({
...
+            currentShapes.some(shape => shape === p.id)
+          )
+        : [],
+    markers: channelState.markers.concat(stopMarkers)
+  };
+  const bounds = useRef(getBounds(stopMarkers));
+
+  const vehicleMarkers = mapDataFromChannel.markers
+    ? mapDataFromChannel.markers.filter(
+        (marker: MapMarker) => marker.icon === "vehicle-bordered-expanded"
```
thecristen, line 253: You had added an `isVehicleMarker` function elsewhere that you could totally use here 😉

thecristen: But why not use `channelState.markers` for the vehicle markers?

apps/site/assets/ts/schedule/components/reducer.ts
```diff
@@ -82,37 +51,17 @@ export const iconOpts = (
...
-
 const shouldRemoveMarker = (id: string | null, idHash: IdHash): boolean =>
   id !== null && idHash[id] === true;
 
-interface State {
-  channel: string;
-  markers: Marker[];
-}
+const isVehicleMarker = (marker: Marker): boolean =>
+  marker.icon ? marker.icon.includes("vehicle") : false;
```
thecristen, line 64: These map-related functions feel kind of out of place here, could we move these to a helper file or something?

apps/site/assets/ts/schedule/components/line-diagram/VehicleIcons.tsx
```diff
@@ -5,26 +5,25 @@ import { RouteStop, LineDiagramVehicle } from "../__schedule";
...
-  headsigns: HeadsignWithCrowding[];
+  vehicleMarkers: MapMarker[];
 }
 
-const tooltipText = (
-  routeType: RouteType | null,
-  stopName: string | null,
-  vehicle: LineDiagramVehicle,
-  prediction: Prediction | null
+// fallback to get the tooltip if vehicle channel data is not available
```
thecristen, line 20: This is a nice thought but I think we should proceed assuming the channel data is available! Please use the info from `vehicleMarkers` in place of the prior `vehicles` :)

apps/site/assets/ts/schedule/components/line-diagram/LineDiagramWithStops.tsx
```diff
@@ -39,7 +46,11 @@ const LineDiagramWithStops = (
           !anyCrowding ? "u-no-crowding-data" : ""
         }`}
       >
-        <Diagram lineDiagram={stops} liveData={liveData} />
+        <Diagram
+          lineDiagram={stops}
+          liveData={liveData}
```
thecristen, line 51: we can probably stop passing `liveData` here and just use the data from `vehicleMarkers`

thecristen (changes requested):
Here's a wider assortment of comments. I think this is definitely a good way to go!
I think additionally we should reconsider our `liveData` prop. Currently it is an object with both headsigns and vehicles, but these are used in different components so IMO there's no longer a good reason to lump them together in the same variable. And it seems to me like those vehicles can be replaced by the vehicles you're getting from the channel anyway. So much refactoring potential (on top of a good amount of refactoring already!) 🤩 

I think the most crucial comment I have here is about performance, on how everything will end up rerendering when the `vehicleMarkers` prop changes. Happy to collaborate on ideas for how to tackle that if you need some direction on that.

apps/site/assets/ts/schedule/components/reducer.ts
thecristen: Oops, this didn't highlight the functions I wanted -- I was talking about `iconOpts`, `zIndex`, `updateMarker`, `shouldRemoveMarker`, `isVehicleMarker`

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -257,6 +306,7 @@ const ScheduleDirection = ({
           stops={stops}
           today={today}
           scheduleNote={scheduleNote}
+          vehicleMarkers={vehicleMarkers}
```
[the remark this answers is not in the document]
hannahpurcell: Sanity check:  multiple components can connect to a channel, right @thecristen?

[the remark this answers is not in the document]
thecristen: @hannahpurcell You probably already ran into the answer by now but yes!

PR #1064, opened 2021-08-30
Schedule Finder for ferry

alicia-mbta:
#### Summary of changes
**Asana Ticket:** [Ferry redesign | Schedules & Maps tab](https://app.asana.com/0/1200690266799825/1200759598958472)

Ready for test deploy and Yinka's review


apps/site/assets/ts/schedule/components/ScheduleFinder.tsx
```diff
@@ -81,19 +83,37 @@ const ScheduleFinder = ({
...
+        <ScheduleFinderFormVertical
+          onDirectionChange={changeDirection}
+          onOriginChange={changeOrigin}
+          onOriginSelectClick={openOriginModal}
+          onSubmit={openScheduleModal}
+          route={route}
+          selectedDirection={directionId}
+          selectedOrigin={selectedOrigin}
+          stopsByDirection={stops}
+        />
```
alicia-mbta, line 104: It might not be necessary to create a separate component/file but instead apply specific styles to the existing one

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleFinderForm.tsx
```diff
@@ -57,73 +58,85 @@ export default ({
...
-    </form>
+        <div className="schedule-finder__submit text-right">
+          <input
+            className="btn btn-primary"
+            type="submit"
+            value="Get schedules"
+          />
+        </div>
+      </form>
+      {isFerryRoute && <hr />}
```
thecristen, line 139: Looks like this `<hr />` appears in the schedule finder modal too and it looks super out of place! How about using CSS to draw this line instead when inside a `.schedule-finder-vertical` class?

apps/site/assets/ts/schedule/components/ScheduleLoader.tsx
```diff
@@ -220,7 +223,9 @@ export const ScheduleLoader = ({
     }
 
     if (component === "SCHEDULE_FINDER") {
-      return (
+      return isFerryRoute ? (
```
thecristen, line 226: Why not put this in with the if clause above (e.g `if (component === "SCHEDULE_FINDER" && !isFerryRoute)`?

apps/site/lib/site_web/templates/schedule/_line.html.eex
```diff
@@ -15,11 +15,29 @@
...
+              4 -> nil
+              _ -> "Schedules & Maps"
+            end %>
+          <%= if title do %>
+            <h2><%= title %></h2>
+          <% end %>
 
           <div id="react-schedule-direction-root"></div>
 
+          <div id="react-schedule-finder-root">
```
thecristen, line 29: suggestion: Combine this logic into the #react-schedule-finder-root element at line 69. And we don't want to end up with two `<div id="react-schedule-finder-root">` elements.

thecristen (changes requested):
A few small comments.
Also! I'm not sure if it's just me but on other routes I'm finding myself unable to interact with the sidebar Schedule Finder anymore. Can you confirm/deny? 

image

apps/site/lib/site_web/templates/schedule/_line.html.eex
hannahpurcell: Ok, I couldn't figure out how to combine them... because the placement of each chunk of code has layout implications (but if there's a way to put component-alize the chunk, I'm all-ears!)  But I DID fix the duplicate react-schedule-finder-root elements (which fixed the fact that the schedule-finder was unusuable on some pages 😅)

thecristen: That works, yay!

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleModalContent.tsx
```diff
@@ -125,6 +125,7 @@ const ScheduleModalContent = ({
           selectedDirection={selectedDirection}
           selectedOrigin={selectedOrigin}
           stopsByDirection={stops}
+          typeModal
```
thecristen, line 128: What does this mean? :eyes:

hannahpurcell: Boolean to render type "modal" of ScheduleFinderForm.  All modal versions of ScheduleFinderForm should look the same (even for ferry) but the ferry version of ScheduleFinderForm that lives right on the page is the outlier to be styled differently.  I could make this value into a "className" but figured the boolean was... simpler?

hannahpurcell:
I do not know why DST tests are failing 😓

apps/site/assets/ts/schedule/components/schedule-finder/ScheduleModalContent.tsx
thecristen: Is this different than what the `schedule-finder-vertical` class is already doing?

thecristen: Maybe a different name would make me less confused... like `insideModal` or something?

hannahpurcell: Oh hey!  True -- I can do this all with CSS without too much headache.  One less inflexible prop on a component 👍

thecristen (approved):

PR #1066, opened 2021-09-01
feat: Add class name when map tiles loaded

thecristen:
Asana ticket TBA

This change listens to Leaflet's `onload` event to add a class to the component after all visible tiles have loaded.

I plan to use that class in the future to get visual tests to wait until all the tiles are loaded before taking a screenshot!

### DIFF
--- a/apps/site/assets/ts/leaflet/components/Map.tsx
+++ b/apps/site/assets/ts/leaflet/components/Map.tsx
@@ -74,8 +74,15 @@ const Component = ({
     const boundsOrByMarkers = bounds || (boundsByMarkers && getBounds(markers));
     const position = mapCenter(markers, defaultCenter);
     const nonNullZoom = zoom === null ? undefined : zoom;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const mapRef = React.createRef<any>();
+    const onTilesLoaded = (): void => {
+      mapRef.current.container.parentElement.classList += " map--loaded";
+    };
+
     return (
       <Map
+        ref={mapRef}
         bounds={boundsOrByMarkers}
         center={position}
         zoom={nonNullZoom}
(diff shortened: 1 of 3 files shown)


apps/site/assets/ts/leaflet/components/Map.tsx
```diff
@@ -74,8 +74,15 @@ const Component = ({
     const boundsOrByMarkers = bounds || (boundsByMarkers && getBounds(markers));
     const position = mapCenter(markers, defaultCenter);
     const nonNullZoom = zoom === null ? undefined : zoom;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const mapRef = React.createRef<any>();
+    const onTilesLoaded = (): void => {
+      mapRef.current.container.parentElement.classList += " map--loaded";
+    };
+
```
hannahpurcell, line 82: I added a breakpoint right here (with a console.log statement, just so there was something to ground the breakpoint) and I seemed to see the "map-loaded" class getting added juuuuuust ahead of leaflet's actual load of the tiles, I think?  But the className doesn't get added when the polyline appears, or the markers, or the vehicle tooltips -- it consistently happens just *as* the tiles are being loaded (I think...).  So if that's enough for the test, then 👍 👍 

Screen Shot 2021-09-07 at 2 22 00 PM

hannahpurcell (approved):
I think it looks great, though it's tough to test out.  I tried one kinda silly method of testing and left the comment below.  But if that outcome is expected, then I approve!

apps/site/assets/ts/leaflet/components/Map.tsx
thecristen: you'd think a breakpoint there would mean that the whole map isn't even instantiated yet, right? 🤔 oh my gosh, that's super interesting!

thecristen: that said

> But the className doesn't get added when the polyline appears, or the markers, or the vehicle tooltips

that is expected, as the function is only on the `<TileLayer />`, which is where the map tiles (and nothing else) is 😅

PR #1070, opened 2021-09-04
refactor(react_renderer): Bring into main config

thecristen:
We don't need a separate node_modules setup for this one script, especially with all the overlapping dependencies! So we simplify the setup a bit.

We leave the output file in the same place (react_renderer/dist/app.js) because why not.

Asana issue: [ 🔧 Tech Debt | Server-side rendering - consolidate react renderer script into rest of assets](https://app.asana.com/0/385363666817452/1200953801815490/f)


thecristen:
Note to self, I forgot to update the Github Actions workflow 🤦🏼‍♀️ !

thecristen:
Going to deploy this to dev-green for good measure!

thecristen:
Yay, finally fixed the tests! Deployed on dev green and ready for review.

hannahpurcell (approved):
A lot of this is deep magic to me, but I see what you're doing and think it's ready to give it a whirl

PR #1071, opened 2021-09-04
fix(Application): Make SiteWeb.Endpoint last child

thecristen:
Now the endpoint will wait for the other processes to initialize before itself starting.

Asana ticket: [Tech Debt | Call supervisor children in correct order](https://app.asana.com/0/385363666817452/1200943226791246/f)


hannahpurcell (approved):
👍

PR #1082, opened 2021-09-10
Fix CI tests for Dependabot PRs

thecristen:
#### Summary of changes
**Asana Ticket:** [Fix CI tests for Dependabot PRs](https://app.asana.com/0/385363666817452/1200701591068667/f)

Turns out it seems to make more sense to create two similar-but-different CI workflows, one which gets triggered by Dependabot PRs, and other by everything else.

The Dependabot PR CI workflow uses the `pull_request_target` event to get around the issues around accessing secrets (which are needed to run a few CI tests) cited here: https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544

I mainly identify Dependabot PRs by checking `github.head_ref` and `github.actor`. 
* The head ref value will be the branch name in the case of PRs (e.g. `refs/heads/dependabot/npm_and_yarn/apps/site/react_renderer/babel/core-7.15.5`). I don't believe this has a value on pushes to master branch.
* The actor value is who initiated the event, which varies. It's `dependabot[bot]` who opens a PR, but if I trigger a rebase on that branch then the new actor for the next workflow run is me! 

I tested these workflows out a bit [on my fork](https://github.com/thecristen/dotcom/actions). 

The screenshot below shows a PR I made with a small JS change running the normal CI workflow, and a dependabot PR triggering the CI (Dependabot) workflow.

image

(The first happens to fail because I neglected to update snapshots, and the second happens to fail because the branch doesn't have the recent Elixir test fixes. But this is just to demonstrate that the right workflows are running for each PR!)

### DIFF
--- a/.github/workflows/tests-dependabot.yml
+++ b/.github/workflows/tests-dependabot.yml
@@ -55,7 +55,6 @@ jobs:
     # The Tests
     - name: Linting / TypeScript
       run: npm run ci:lint:ts
-      if: always()
     - name: Linting / JavaScript
       run: npm run ci:lint:js
       if: always()
--- a/.github/workflows/tests-dependabot.yml
+++ b/.github/workflows/tests-dependabot.yml
@@ -1,0 +1,94 @@
+name: CI (Dependabot branches)
+on:
+  pull_request_target:
+    branches: [master]
+
+env:
+  MIX_ENV: test
+  V3_URL: ${{ secrets.V3_URL }}
+  V3_API_KEY: ${{ secrets.V3_API_KEY }}
+  RECAPTCHA_PUBLIC_KEY: 6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
+  RECAPTCHA_PRIVATE_KEY: 6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
+  OPEN_TRIP_PLANNER_URL: ${{ secrets.OPEN_TRIP_PLANNER_URL }}
+
+# Run everything, no skipping, minimal caching.
+jobs:
+  ci:
+    name: Run entire CI suite
+    if: ${{ contains(github.head_ref, 'dependabot') }}
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    
+    # ASDF. Used the cached.
+    - uses: asdf-vm/actions/setup@v1
+    - uses: actions/cache@v2
+      with:
+        path: ~/.asdf
+        key: ${{ runner.os }}-asdf-cache-${{ hashFiles('.tool-versions') }}
+      id: asdf-cache
+    - uses: asdf-vm/actions/install@v1
+      if: ${{ steps.asdf-cache.outputs.cache-hit != 'true' }}
+
+    - name: Install Elixir dependencies
+      run: |
+        mix local.hex --force
+        mix local.rebar --force
+        mix deps.get
+
+    - name: Install Node dependencies
+      run: |
+        git config --global url."https://github.com/".insteadOf ssh://git@github.com/
+        npm run install:ci
+
+    # Build assets
+    - run: npm run webpack:build
+      working-directory: apps/site/assets
+    - run: npm run build
+      working-directory: apps/site/react_renderer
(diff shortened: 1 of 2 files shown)


thecristen:
@paulswartz Tagging you since you have much experience with our GitHub Actions and thought you might have some insight or can help gut check this, feel free to take a look if you have time!

.github/workflows/tests-dependabot.yml
```diff
@@ -0,0 +1,94 @@
...
+      working-directory: apps/site/react_renderer
+    - run: mix phx.digest
+
+    - name: Build application
+      run: mix compile --warnings-as-errors
+
+    # The Tests
+    - name: Linting / TypeScript
+      run: npm run ci:lint:ts
+      if: always()
```
paulswartz, line 58: Do you need this? This will run the tests even if it wasn't possible to build the application.

.github/workflows/tests.yml
```diff
@@ -435,7 +430,7 @@ jobs:
     name: Integration tests / Elixir
     runs-on: ubuntu-latest
     needs: [setup, file_changes, build_app]
-    if: ${{ needs.file_changes.outputs.ex }}
+    if: ${{ !(contains(github.head_ref, 'dependabot') || github.actor == 'dependabot[bot]') && needs.file_changes.outputs.ex }}
```
paulswartz, line 433: Do you need the new case? It looks like `needs.file_changes.outputs` will all be false because that step also has the "not a dependabot PR" check.

.github/workflows/tests-dependabot.yml
thecristen: oooh good point. I'll remove it from this first test.

paulswartz: Do any of them need it?

thecristen: When this workflow finishes I'd be able to see ALL tests that fail. Without `always()` it'll abort the workflow and not run any jobs after a failing test.

thecristen: Unrelated, upon reflection it seems silly for me to run linters on Dependabot PRs though... hmmmmm 😑

.github/workflows/tests.yml
thecristen: removing the case seems to work great! thanks

.github/workflows/tests-dependabot.yml
```diff
@@ -0,0 +1,69 @@
...
+  V3_API_KEY: ${{ secrets.V3_API_KEY }}
+  RECAPTCHA_PUBLIC_KEY: 6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
+  RECAPTCHA_PRIVATE_KEY: 6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
+  OPEN_TRIP_PLANNER_URL: ${{ secrets.OPEN_TRIP_PLANNER_URL }}
+
+# Run unit and integration tests, no skipping, minimal caching.
+jobs:
+  ci:
+    name: Run entire CI suite
+    if: ${{ contains(github.head_ref, 'dependabot') }}
```
hannahpurcell, line 18: Is there ever the case that a branch could NOT have a head_ref that contains 'dependabot' but DOES have `github.actor == 'dependabot[bot]'`?  The ifs in `tests.yml` reference github.actor, but not the ifs here in `tests-dependabot.yml`.

thecristen: Yeah! I think when I wrote this I wasn't sure of the answer to that question and wanted to be extra safe just in case! 
But for `tests-dependabot.yml` I want the tests to run even if I'm the `github.actor` -- hence it's not included in the ifs here.

Now that I've experimented a bit more on my fork, I probably _could_ leave the `github.actor == 'dependabot[bot]'` check off the `tests.yml` jobs -- it looks like the answer to your question is "probably not"!

thecristen: @hannahpurcell I decided to leave out the github.actor part, but I also edited the contains() to be more specific. otherwise it'll work on _any_ PR branch with dependabot in the name (include this branch!).

hannahpurcell (approved):
I like it!  I think we should take it for a test drive

PR #1083, opened 2021-09-10
Return of the Ferry Timetable

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Ferry redesign | Timetable tab](https://app.asana.com/0/555089885850811/1172866781307225)

- Timetable is the default page when you click on more info about a ferry, just like for CR
- All references to trains was replaced with boats in the details of this page
- Adjusted how bike info was handled on the ferry timetable
- Parking icon was added next to station names (this impacts the CR, too!)

---


apps/site/lib/site_web/templates/schedule/_timetable.html.eex
```diff
@@ -13,9 +13,9 @@
       <div class="m-timetable__cell m-timetable__cell--gray m-timetable__cell--first-column m-timetable__cell--first-column-header m-timetable__row-header--empty"></div>
       <div class="m-timetable__col-headers" aria-hidden="true">
         <%= if should_show_scroll_controls? do %>
-          <span class="m-timetable__trains-label">Trains</span>
-          <button class="m-timetable__scroll-btn m-timetable__scroll-btn--left btn btn-outline-primary btn-sm" data-scroll="earlier"><%= fa "angle-left m-timetable__scroll-btn-arrow" %> Earlier Trains</button>
-          <button class="m-timetable__scroll-btn m-timetable__scroll-btn--right btn btn-outline-primary btn-sm" data-scroll="later">Later Trains <%= fa "angle-right m-timetable__scroll-btn-arrow" %></button>
+          <span class="m-timetable__trains-label"><%= trains_or_boats(@route.type) <> "s" %></span>
```
thecristen, line 16: I think you can replace `trains_or_boats(@route.type)` with `Routes.Route.vehicle_name(@route)`! 😃

apps/site/lib/site_web/views/schedule/timetable.ex
```diff
@@ -101,4 +120,12 @@ defmodule SiteWeb.ScheduleView.Timetable do
...
   def cell_via_class(nil), do: ""
   def cell_via_class(<<_::binary>>), do: " m-timetable__cell--via"
+
+  @spec trains_or_boats(integer) :: String.t()
+  def trains_or_boats(2), do: "train"
+  def trains_or_boats(4), do: "boat"
+
+  @spec is_ferry(integer) :: boolean
+  def is_ferry(4), do: true
+  def is_ferry(_), do: false
```
thecristen, line 130: We often identify route modes by their `type_atom`, so I think this would be better defined as `def is_ferry(route), do: Routes.Route.type_atom(route) == :ferry`

apps/site/lib/site_web/templates/schedule/_timetable.html.eex
thecristen: OOH I just realized `Routes.Route.vehicle_name` is only used in two places right now and they're both to show realtime data, so they don't even show ferries yet. @hannahpurcell I give you permission to edit `Routes.Route.vehicle_name` so that it outputs "Boat" so that you can use it here.

thecristen (changes requested):
There's a couple unit tests failing because `trains_or_boats/1` is being given a route with a type of 0 (that's a trolley!!), so I am extra pushing to use a more general function such as the `Routes.Route.vehicle_name/1` function there instead! 😉

thecristen (approved):
👏🏼

PR #1088, opened 2021-09-21
fix(ControllerHelpers): no overwrite x-robots-tag

thecristen:
#### Summary of changes
**Asana Ticket:** [Really block dev environments from search engines](https://app.asana.com/0/385363666817452/1201031900628965/f)

It turns out a function was adding an `unavailable_after` value for `x-robots-tag` indiscriminately for events and news entries, which was butting against how we add the `noindex` value for `x-robots-tag` for all pages in our non-prod environments. This change aims to fix that.

---


hannahpurcell (approved):
Wow, nifty.  Great catch and succinct solution!  Thanks for the test, too

PR #1090, opened 2021-09-23
List.first Sentry bug

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Schedules | Elixir.FunctionClauseError: no function clause matching in List.first/1](https://app.asana.com/0/555089885850811/1201042269068385)

We assign CMS teasers with an async function.  The default value for "featured_content" teasers is set to nil until the content is populated by the async function.  We didn't have a render_teaser function that could accept nil, so any time the assign was going on too long, an error is generated.


thecristen (approved):
👍🏼

PR #1091, opened 2021-09-23
Sentry issue: nil lat / lng should be discarded in Transit Near Me

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Transit Near Me | Elixir.FunctionClauseError: no function clause matching in SiteWeb.TransitNearMeController.Location.parse_float/1](https://app.asana.com/0/555089885850811/1200280531064861)

Not quite sure how "nil" lat or lng is getting input in the url string -- is there any way for us to verify whether we generate a url like that anywhere?  That seems like an impossibly hard to answer question.  But anywho -- now the nil value is getting handled properly.

### DIFF
--- a/apps/site/lib/site_web/controllers/transit_near_me_controller.ex
+++ b/apps/site/lib/site_web/controllers/transit_near_me_controller.ex
@@ -29,7 +29,6 @@ defmodule SiteWeb.TransitNearMeController do
   end
 
   defp assign_location(conn) do
-    # Why do we check for location_fn?  It's never assigned
     location_fn = Map.get(conn.assigns, :location_fn, &Location.get/2)
 
     location = location_fn.(conn.params, [])
--- a/apps/site/lib/site_web/controllers/transit_near_me_controller.ex
+++ b/apps/site/lib/site_web/controllers/transit_near_me_controller.ex
@@ -29,6 +29,7 @@ defmodule SiteWeb.TransitNearMeController do
   end
 
   defp assign_location(conn) do
+    # Why do we check for location_fn?  It's never assigned
     location_fn = Map.get(conn.assigns, :location_fn, &Location.get/2)
 
     location = location_fn.(conn.params, [])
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/controllers/transit_near_me_controller.ex
```diff
@@ -29,6 +29,7 @@ defmodule SiteWeb.TransitNearMeController do
   end
 
   defp assign_location(conn) do
+    # Why do we check for location_fn?  It's never assigned
```
thecristen, line 32: Hmm! I think it’s a convention that arose via certain developers… there are similar examples of passing in functions with defaults like this throughout the codebase. Sometimes it seems like a helpful way to easily enable alternate implementations i.e. from one page we want function A but from a different page we want to use function B and from the test case we will pass in mocked function C etc.

thecristen (approved):
I’m a bit curious why we were running into so many nil lat/lon to begin with, but this update definitely helps!

thecristen:
Are we so sure the nil is coming from the URL as opposed to some of the other logic going on here?

hannahpurcell:
> Are we so sure the nil is coming from the URL as opposed to some of the other logic going on here?

I spent today trying to digest what's going on with `algolia-autocomplete-with-geo.js`, and I've made progress but still cannot follow everything.  Can't figure out exactly how the longitude / latitude is getting set after running a search, and how the parameters are added to the url.

hannahpurcell:
On another pass, it *seems like everything is properly hooked up, passing around lats / lngs as desired.

Also noticed that all the issues are happening on Chrome Mobile, particularly on Samsung phones... Why?

apps/site/lib/site_web/controllers/transit_near_me_controller.ex
hannahpurcell, line 32:
```suggestion
```

PR #1092, opened 2021-09-28
Fix tests affected by how we set x-robots-tag

thecristen:
No ticket - quick fix for some failing Elixir tests as a result of merging in #1088. 

These values change based on the application configuration, which varies between prod/dev/test environments. So these particular tests need to make sure to be set a particular way to enable these tests to test the things they are trying to test. 🙂


hannahpurcell (approved):
Looks good, and it's passing!

PR #1093, opened 2021-09-30
Added a `get_shapes_by_direction` test 

hannahpurcell:
... to catch cases where there aren't scheduled trips.  Prompted during the development of this branch, waaaay back when:  https://github.com/mbta/dotcom/pull/848 

#### Summary of changes
**Asana Ticket:** [Testing | Cover route without scheduled trips](https://app.asana.com/0/555089885850811/1200172434828846)

### DIFF
--- a/apps/routes/lib/mock_repo_api.ex
+++ b/apps/routes/lib/mock_repo_api.ex
@@ -893,6 +893,9 @@ defmodule Routes.MockRepoApi do
   end
 
   @impl Routes.RepoApi
+  def get_shapes("27", direction_id: 0), do: []
+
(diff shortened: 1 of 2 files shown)


thecristen (changes requested):
There's a compilation error coming up, I think it doesn't like mixing around `get_shapes/3` and `get_shapes/2` functions in `Routes.MockRepoApi`. Other than that looks good!

```
warning: clauses with the same name should be grouped together, "def get_shapes/2" was previously defined (lib/mock_repo_api.ex:219)
```

thecristen (approved):

PR #1094, opened 2021-10-01
fix(TripPlanController): set Massport route type

thecristen:
Our last workaround, accommodating "Massport-1" as a route type, stopped working because now the Massport buses are returning as route type "1". So for the sake of the Trip Planner, let's stop relying on the given type and stick to "Massport-1". This ensures that the correct icon (bus) and naming ("Massport shuttle") shows in the itinerary information.

#### Summary of changes
**Asana Ticket:** [Trip Planner | Crashes when using Massport routes](https://app.asana.com/0/385363666817452/1201105996476530/f)

### DIFF
--- a/apps/site/lib/site_web/controllers/trip_plan_controller.ex
+++ b/apps/site/lib/site_web/controllers/trip_plan_controller.ex
@@ -383,10 +383,11 @@ defmodule SiteWeb.TripPlanController do
       mode: mode,
       long_name: long_name,
       name: name,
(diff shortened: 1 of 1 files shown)


losvedir (approved):
Thanks for jumping on this! Code looks good to me, as far as I know without being a `dotcom` expert. :) Looks like the `url` comes from the feed's `agency.txt`, so I'll try to remember to keep an eye on that in the future.

PR #1097, opened 2021-10-05
fix(Router): ensure redirect to https:// dotcom

thecristen:
fixes security issue where mbtace.com first redirected to http://www.mbta.com before that redirected to https://www.mbta.com

...admittedly not sure how to actually test this outside prod. But I think this shortcut should work.

#### Summary of changes
**Asana Ticket:** [Insecure redirect in dotcom (HTTP vs. HTTPS)](https://app.asana.com/0/385363666817452/1201080431534538/f)

### DIFF
--- a/apps/site/lib/site_web/plugs/canonical_hostname.ex
+++ b/apps/site/lib/site_web/plugs/canonical_hostname.ex
@@ -27,8 +27,12 @@ defmodule SiteWeb.Plugs.CanonicalHostname do
       conn
     else
       rewritten_url =
-        Plug.Conn.request_url(conn)
-        |> String.replace(requested_hostname, canonical_hostname, global: false)
+        if requested_hostname == "www.mbtace.com" do
+          "https://www.mbta.com" <> conn.request_path
+        else
+          Plug.Conn.request_url(conn)
+          |> String.replace(requested_hostname, canonical_hostname, global: false)
+        end
 
       conn
       |> put_status(:moved_permanently)
--- a/apps/site/test/site_web/router_test.exs
+++ b/apps/site/test/site_web/router_test.exs
@@ -113,6 +113,18 @@ defmodule Phoenix.Router.RoutingTest do
       conn = get(conn, "/trip-planner/to/")
       assert redirected_to(conn, 301) == "/trip-planner"
     end
+
+    test "www.mbtace.com to www.mbta.com", %{conn: conn} do
+      System.put_env("HOST", "www.mbta.com")
+
+      on_exit(fn ->
+        System.delete_env("HOST")
+      end)
+
+      conn = %Plug.Conn{conn | host: "www.mbtace.com"}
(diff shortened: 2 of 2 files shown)


apps/site/lib/site_web/plugs/canonical_hostname.ex
```diff
@@ -27,8 +27,12 @@ defmodule SiteWeb.Plugs.CanonicalHostname do
...
     else
       rewritten_url =
-        Plug.Conn.request_url(conn)
-        |> String.replace(requested_hostname, canonical_hostname, global: false)
+        if requested_hostname == "www.mbtace.com" do
+          "https://www.mbta.com" <> conn.request_path
+        else
+          Plug.Conn.request_url(conn)
+          |> String.replace(requested_hostname, canonical_hostname, global: false)
+        end
```
paulswartz, line 35: The [code for request_url/1](https://github.com/elixir-plug/plug/blob/v1.12.1/lib/plug/conn.ex#L1676-L1685) looks like it shouldb e taking the scheme of the request into account. If it thinks that the scheme is `http`, perhaps the `Plug.SSL` plug is mis-configured?

paulswartz: Oh, I see: this plug is configured in `SiteWeb.Endpoint` before `SiteWeb.Router`, so it runs before`Plug.SSL` has a chance to correct the scheme.

thecristen: Oh my gosh thank you!!! I was trying to figure out for the longest time why `Plug.SSL` wasn't helping us out here, hmmmm!

paulswartz: Can you push up a branch with what you have? I'll take a look tomorrow.

thecristen: @paulswartz I eventually realized I can add some stuff to the test config to make the test behave more in line with the prod environment, so I think I've got it now. Pushed up the changes.

thecristen:
Am deploying this to dev-blue to verify it produces the desired result.

thecristen:
Going to give that suggestion a try. If that doesn't work, might make sense to try `url: [scheme: "https", host: {:system, "HOST"}, port: 443],` here?

thecristen:
> Going to give that suggestion a try. If that doesn't work, might make sense to try `url: [scheme: "https", host: {:system, "HOST"}, port: 443],` here?

Ah wait, just realized there's already such a declaration in https://github.com/mbta/dotcom/blob/master/apps/site/config/prod.exs#L84

thecristen:
Didn't notice any change. Going to try updating `Plug` next because of this relevant-sounding bug fix noted in the 1.9 release notes:
```
[Plug.SSL] Rewrite port when rewriting a request coming to a standard port
```

PR #1099, opened 2021-10-06
Listing 62/76 twice in the bus lists

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Schedules | Blended routes are confusing (62/76)](https://app.asana.com/0/555089885850811/1200937969737836)

Added a helper function so we can specify "blended routes" that we'd like to list twice on the bus routes page.  Right now, it's only the 62/76, but it's pooooossible there could be other routes in the future?  If not, then this logic could probably be simplified.


apps/routes/lib/repo.ex
```diff
@@ -180,4 +183,33 @@ defmodule Routes.Repo do
       color: "00843D"
     }
   end
+
+  # Currently, we list the 62/76 in both places of the bus route list
+  # This function accepts the name of the blended route and inserts it into the route list
+  @spec duplicate_blended_route([Route.t()], [String.t()]) :: [Route.t()]
+  defp duplicate_blended_route(routes, blended_route_names) do
+    blended_route_names
+    |> Enum.reduce(routes, fn current_blended_name, acc ->
```
thecristen, line 192: Curious about this reduce flow as opposed to something like

```elixir
# passing in a route id via duplicate_blended_route(routes, blended_route_id)

blended_route = Routes.Repo.get(blended_route_id) #627
for route_name1, route_name2 <- String.split(blended_route.name, "/") do
   # logic here to find the right index to eventually
   # routes = List.insert_at(routes, index, blended_route)
end

routes
```

hannahpurcell: I believe I've seen three-blended routes, but I'm not certain.  Is that ever possible?

Also, chose to make `duplicate_blended_route` accept an array of names because we tend to discuss blended routes by name, so if we ever get asked to handle a new blended route in the future, we can easily add the new route to the array.  From `|> duplicate_blended_route(["62/76"])` to `|> duplicate_blended_route(["62/76", "18/19", ...])`.  Thoughts?  

Also perhaps we should store the blended routes in a config spot?

hannahpurcell: Ok finally tested this out.  The syntax of changing the value of `routes` from inside the for loop doesn't work, so the only way I can think of doing this is a reduce.  What do you think?  Specifically, this doesn't update `routes`:

```
  defp duplicate_blended_route(routes) do
    for route <- Enum.filter(routes, &Route.combined_route?/1) do

      for name <- String.split(route.name, "/") do
        index = Enum.find_index(routes, fn x -> 
          !is_nil(int_or_string_to_int(x.name)) and int_or_string_to_int(x.name) > int_or_string_to_int(name)
        end)

        routes = List.insert_at(routes, index, route)
      end
    end

    routes
  end
```

hannahpurcell: Ok, thanks for the advice and review!  I never love using reduce, ugh.  But I guess I'm stuck with it here.  I tried cleaning up the reduces a bit, though.  Let me know if this new iteration is cleaner?

thecristen (approved):

PR #1105, opened 2021-10-15
Trip Planner | 💡 Auto-disambiguate Massport GTFS IDs

thecristen:
#### Summary of changes
**Asana Ticket:** [Trip Planner | 💡 Auto-disambiguate Massport GTFS IDs](https://app.asana.com/0/385363666817452/1200911564672890/f)

When parsing results from Open Trip Planner, we distinguish between MBTA and Massport data based on the agency ID, and append `"Massport-"` to the identifiers in Massport data.

The other changes ensure that this renders the correct Massport-related information in the trip planner results.

---


apps/routes/lib/route.ex
```diff
@@ -65,7 +65,7 @@ defmodule Routes.Route do
   def type_atom("ferry"), do: :ferry
   def type_atom("909"), do: :logan_express
   def type_atom("983"), do: :massport_shuttle
-  def type_atom("Massport-1"), do: :massport_shuttle
```
hannahpurcell, line 68: There's another reference to "Massport-1" in `site_web/views/helpers.ex`.  I'm not sure if that needs to be replaced in the same way, or if `"Massport-" <> _` is even needed because it also checks for :massport_shuttle.

apps/site/lib/site/trip_plan/related_link.ex
```diff
@@ -88,20 +88,25 @@ defmodule Site.TripPlan.RelatedLink do
...
-        else
-          route.name
-        end
-
-      date = Date.to_iso8601(itinerary.start)
-      url = schedule_path(SiteWeb.Endpoint, :show, route, date: date, trip: trip_id)
-      new([base_text, " schedules"], url, icon_name)
+    cond do
+      String.starts_with?(route.id, "Massport") ->
+        new("Massport schedules", "https://massport.com/", icon_name)
```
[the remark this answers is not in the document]
thecristen: This is a great idea – but I can't figure out how to map the shuttle numbers/names to more specific URLs! Was also concerned about links potentially breaking over time if they were changed.

apps/routes/lib/route.ex
thecristen: Done

PR #1107, opened 2021-10-21
Error page layout

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Update error page layout and content](https://app.asana.com/0/555089885850811/1201073069115358)

Set up layout for error pages.  Added new `_typography` file to explicitly set the fonts from our style guide.  

Currently, only 404 and 500 errors are set up in the app.  All pages have been built (403, 503, 404 file not found) but we'll need a future ticket to do the new error handling.

---


apps/site/assets/css/_typography.scss
```diff
@@ -0,0 +1,125 @@
...
+  @include media-breakpoint-down(sm) {
+    font-size: 14px;
+    line-height: 17.85px;
+  }
+}
+
+.font-body {
+  font-family: $font-family-base;
+  font-weight: 400;
+  font-size: 18px;
```
thecristen, line 76: The body, link, and white fonts share enough (same sizes, probably same line heights etc) that it might be useful to extract into some sort of base-font-size mixin that includes the breakpoints too. But maybe the existing `@mixin base-font-size` does that fine? 🤔

apps/site/lib/site_web/templates/error/error_layout.html.eex
```diff
@@ -0,0 +1,42 @@
+
+<!-- Default -->
+<div class="container error-page average-sizes">
```
thecristen, line 3: Unless I'm mistaken it looks like the only difference between Default and XL is the moving of `<div class="heading-1 error-title"> <%= @error_title %></div>`. Instead of using a class to toggle one entire `.error-page` vs another (loading the SVG and everything else twice and whatnot), can we limit that strategy to the .error-title?

I'm thinking, you can have two instances of `<div class="heading-1 error-title"> <%= @error_title %></div>`, the one at this level that only shows on average sizes, and then the other one inside the `.error-details-container` that shows only at the largest breakpoint.

thecristen: Oh, and you wouldn't even need an .xl-only class! You could select it via `.error-details-container .error-title {}` vs `.error-pre-title + .error-title {}` (just as an example)

thecristen (approved):
It all looks good so let's go with it!

As for the base-font styles, they're looking the same despite you removing that bit from core.scss, which is super curious but let's think about this more elsewhere! I'm thinking we should add a ticket for overhauling the core typography CSS, which the understanding that you've done most of the legwork here already so we can focus on applying this "new" styles everywhere and cleaning up and/or replacing redundant old CSS if needed.

PR #1113, opened 2021-10-27
Update Agenda Topic HTML+CSS

thecristen:
#### Summary of changes
**Asana Ticket:** [Update Agenda Topic HTML+CSS](https://app.asana.com/0/385363666817452/1201218504169421/f)

Updating our implementation of the Agenda topics and subtopics to account for https://github.com/mbta/cms/pull/376, with some test data... 
image

### DIFF
--- a/apps/site/assets/css/_event-details.scss
+++ b/apps/site/assets/css/_event-details.scss
@@ -53,7 +53,6 @@
 
 $bullet-size: 32px / $font-size-base-xxl * $base-spacing; // 32px on large viewport
 $separator-border: solid 1px $gray-lightest;
-$agenda-row-gap: $base-spacing * .25;
 $agenda-column-gap: $base-spacing * .75;
 
 %small-grey-font {
@@ -107,20 +106,32 @@ $agenda-column-gap: $base-spacing * .75;
     .agenda-topic,
     .agenda-subtopic {
       border-top: $separator-border;
+      column-gap: $agenda-column-gap;
       display: grid;
-      gap: $agenda-row-gap $agenda-column-gap;
       grid-template: auto auto / $bullet-size 1fr;
       padding: $base-spacing * .75 0 $base-spacing;
+
+      &:last-child {
+        padding-bottom: 0;
+      }
     }
   }
 
   .agenda-topic__title {
     grid-area: 1 / 2 / 2 / 3;
-    margin-top: 0;
+    margin: 0;
+
+    + .agenda-topic__content {
+      margin-top: $base-spacing * .5;
+    }
   }
 
   .agenda-topic__content {
     grid-area: 2 / 2 / 3 / 3;
+
+    :last-child {
+      margin-bottom: 0;
+    }
   }
 
   // top-level list
@@ -137,7 +148,7 @@ $agenda-column-gap: $base-spacing * .75;
 
   // second-level list
   .agenda-subtopics {
-    margin-top: $base-spacing * .75;
+    margin-top: $base-spacing * .125;
 
     // further nested content gets styled as small muted text
     p {
@@ -153,6 +164,8 @@ $agenda-column-gap: $base-spacing * .75;
   .agenda-subtopic {
     @include agenda-bullet($bg-color: $gray-bordered-background,
     $color: $gray, $counter: subCounter);
+
+    align-items: baseline;
   }
 
   .agenda-topic__timestamp {
--- a/apps/site/assets/css/_event-details.scss
+++ b/apps/site/assets/css/_event-details.scss
@@ -90,7 +90,6 @@ $agenda-column-gap: $base-spacing * .75;
 .m-event__agenda {
   p {
     line-height: $headings-line-height;
-    margin-bottom: 0;
   }
 
   .agenda-topics,
@@ -159,10 +158,23 @@ $agenda-column-gap: $base-spacing * .75;
   .agenda-topic__timestamp {
     @extend %small-grey-font;
     font-weight: bold;
-    margin-right: $base-spacing * .5;
 
-    + p {
-      display: inline;
+    &:not(:last-child) {
+      // hacky way of showing it "in" the first paragraph
+      margin-right: $base-spacing * .25;
+
+      // first paragraph
+      + p {
+        // display paragraph inline so text is adjacent to timestamp
+        display: inline;
+
+        // display:inline removes margin from the paragraph.
+        // to make up for that, add a top margin to the next element
+        /* stylelint-disable-next-line max-nesting-depth */
+        + * {
+          margin-top: 1em;
+        }
+      }
     }
   }
 }
--- a/apps/site/assets/css/_event-details.scss
+++ b/apps/site/assets/css/_event-details.scss
@@ -52,8 +52,9 @@
 }
 
 $bullet-size: 32px / $font-size-base-xxl * $base-spacing; // 32px on large viewport
-$bullet-space: $base-spacing * .75;
 $separator-border: solid 1px $gray-lightest;
+$agenda-row-gap: $base-spacing * .25;
+$agenda-column-gap: $base-spacing * .75;
 
 // adds a large round colored circle with number from counter
 @mixin agenda-bullet($bg-color: $gray, $color: $white,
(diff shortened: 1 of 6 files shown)


hannahpurcell:
Hmm, is this the proper link to view the event?
http://localhost:4001/events/2021-02-22/test-joint-meeting-the-massdot-board-and-the-fiscal-and-management-control-board?preview=&vid=52097&nid=5353

I viewed the event last week and it was beautiful (didn't do a full review because of the ongoing chats in Asana).  But now things are looking funky for me, and it doesn't look like anything is changed code-side, so has something been altered in the CMS?

Screen Shot 2021-11-02 at 4 19 32 PM

lund5000:
@hannahpurcell How does the localhost version of https://www.mbta.com/events/2021-02-22/test-joint-meeting-the-massdot-board-and-the-fiscal-and-management-control-board?preview=&vid=55501&nid=5353 work for you? It should have the updated agenda attached to it.

thecristen:
@hannahpurcell Does **vid=55501** work any better? I think that's the one with the updated data.

hannahpurcell (approved):
Awesome!  I think it looks great.  I couldn't tell if Yinka already 👍 the design, and that's the only thing I didn't review with a fine-toothed comb.  Everything else looks great!  🎉

apps/site/assets/css/_event-details.scss
```diff
@@ -52,8 +52,16 @@
 }
 
 $bullet-size: 32px / $font-size-base-xxl * $base-spacing; // 32px on large viewport
-$bullet-space: $base-spacing * .75;
 $separator-border: solid 1px $gray-lightest;
+$agenda-row-gap: $base-spacing * .25;
+$agenda-column-gap: $base-spacing * .75;
```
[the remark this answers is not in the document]
thecristen: I don't assume so, but who knows what the future may bring.. ⭐

apps/site/assets/css/_event-details.scss
```diff
@@ -80,92 +88,79 @@ $separator-border: solid 1px $gray-lightest;
...
-    border-top: $separator-border;
-    padding: $base-spacing * .75 0 $base-spacing;
+  p {
+    line-height: $headings-line-height;
+    margin-bottom: 0;
   }
 
-  // top-level topic
-  .agenda-topic {
-    @include agenda-bullet($counter: topicCounter, $counter-type: upper-alpha);
```
amaisano, line 92: I think we've lost the sub-topic counter reset. The numbered subtopics should reset to `1` after each topic. This preview (may be outdated) shows this is in error:

image

hannahpurcell: This is intentional.  It matches the mocks:  https://zeroheight.com/2fedee66c/p/42cbbf-event-pages/i/31027605

amaisano: This does not seem correct, even in the mocks. Can you please confirm with design? It makes no sense to me for sub-topics to have a cross-topic number progression... it might be a design bug.

hannahpurcell: Yinka's not available for a few days, but in past conversations, we've discussed the numbering.  That day in the summer when we were meeting as a full group to talk through the tech challenges of this design, I said I was worried we wouldn't be able to implement that style of numbering -- turned out to be an easy solution, so no worries there.  But Yinka didn't say it was a mistake.

Viewed as a full agenda, it's really easy to tell someone to look at agenda item 5.  Little faster than specifying C3.

amaisano: OK. Very awkward from a semantic POV IMO, but fine. You can go further and make CSS rules for applying "C1, C2" as part of the numbering scheme, I believe, just as an FYI.

https://clickhelp.com/clickhelp-technical-writing-blog/advanced-autonumbering-techniques-with-css/

```
1 item
   1.a
   1.b
2 item
   2.a
```

lund5000:
_[copied from Asana for ease of reference in the next set of reviews]_

I spoke with Katrina this morning about our prospective users for the CMS side and Cristen about effort/priorities of the things that aren't quite right yet. Here's what we resolved to try and get done now vs. defer, all on the current PR:

# Fixing on this PR if we can
- Video bookmarks
  - When a video bookmark value is added
    - always display it as a label instead of a link
    - only show that label after the event date and time have passed
    - allow for values in a "00:00" format for meetings that end in less than an hour
    - more gracefully handle values that don't match a "00:00" or "00:00:00" format (i.e., don't 500)
  - Investigate and hopefully fix the 500 error caused by entering a link in the livestream URL field of the event (example: https://dev-blue.mbtace.com/events/2021-02-22/test-joint-meeting-the-massdot-board-and-the-fiscal-and-management-control-board?preview=&vid=55800&nid=5353#)
- Files related to topics
  - When a related file has been added to a topic, convert the topic text into a link to the first file (but ignore any others)
- External resources / "View documents"
  - See if there's an easy way to restore the FontAwesome substitution in topic and subtopic descriptions (e.g. `{{fa "external-link"}}`)
- Description formatting
  - Add 0.5em margin between paragraphs in topic and sub-topic descriptions
  - Nice-to-have: Handle ordered and unordered lists in descriptions

# Deferring for now
I'll add these to the backlog and work with Avery to document limitations and workarounds for CMS users.

- Related files
  - Adding related files to subtopics
  - Handling additional related files added to topics (besides the first)
- External resources
  - Keep in the CMS, but the website won't do anything with these URL/text combinations for now

thecristen:
@lund5000 All the suggested fixes have been made/attempted, and re-deployed, so please feel free to try to break it again 😈 

The one other thing I did was hide the livestream video for now -- from what I can figure out, a URL alone isn't sufficient to embed a video (tried to show that YouTube link in an iframe, got a nice CORS error)

amaisano:
> from what I can figure out, a URL alone isn't sufficient to embed a video

Oh no! Bad assumption on my part I think. Is there no way to reverse engineer the URL arguments to form an embed template (this is what I assumed could happen - to keep things simple for the content editors, and consistent for the devs/template).

If this doesn't pan out and you need a full embed code/script from the CMS, I can tweak the input field configuration for that, but then we're transferring raw HTML to the front-end and will by default be limited by the HTML scrubber (which no-likey `embed` and `script` tags).

thecristen:
@amaisano So I can put a video _file_ in a `<video>`, or use whatever embed code whatever livestream site recommends, not sure what the options look like beyond this -- happy to brainstorm strategies for this (to be addressed in a later ticket)!

amaisano:
🤔  def. need a brainstorm session. We're talking about the _post_ event video, right? In that case perhaps it could be downloaded from the source, then uploaded to the CMS and returned as a file URI, but there could be better ways. Please keep me in the loop :)

lund5000:
@thecristen It's looking and working so much better! Just a few more things to check out from me before we approve and release these changes: 

1. Blocking: Find and fix whatever is causing https://dev-blue.mbtace.com/events/2021-02-22/test-joint-meeting-the-massdot-board-and-the-fiscal-and-management-control-board?preview=&vid=55800&nid=5353# to generate 500 errors
2. Should have: The video timestamp field is much more permissive, but I was still able to generate a 500 with it by saving a YouTube URL as the value. We don't need to support URLs in that field and links on the event pages, but could do something like swallow the error and display no timestamp label on the page if it happens. (I think this is unrelated to the 500 condition above, but maybe not.)
3. Nice to have, but non-blocking: The `<ol>` and `<ul>` formatting is much nicer, but the `font-size` resets to the default body value of 16px rather than matching the rest of the topic or sub-topic description sizes.

apps/site/lib/site_web/views/event_view.ex
```diff
@@ -173,4 +173,41 @@ defmodule SiteWeb.EventView do
...
+
+  @spec agenda_title(String.t(), :h3 | :h4) :: Phoenix.HTML.Safe.t()
+  def agenda_title(title, tag_type \\ :h3)
+
+  def agenda_title(title, tag_type)
+      when not is_nil(title) do
+    content_tag(tag_type, [title], class: "agenda-topic__title")
+  end
+
+  def agenda_title(_, _), do: ""
```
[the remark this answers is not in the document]
thecristen: Added

thecristen:
@lund5000 kicking off a deploy to dev-blue with all three fixes!

PR #1116, opened 2021-11-02
Stop Pages | Show predictions when we don't have arrival or departure times

thecristen:
#### Summary of changes
**Asana Ticket:** [🚉 Stop Pages | Show predictions when we don't have arrival or departure times](https://app.asana.com/0/385363666817452/1199651847560606/f)

The data feeding into the `<StopCard>` component which displays on Transit Near Me and on Stop pages had `status: null` hardcoded into the prediction when processing data from the backend to frontend 🤦🏻‍♀️

The other commit is because I saw a `Train null` -- now we check if there's a `train_number` first.

For example, looking at rail results for /transit-near-me?from=search-homepage&query=1%20International%20Place%2C%20Boston%2C%20MA%2C%20USA&latitude=42.35604&longitude=-71.0521404&address=1%20International%20Place%2C%20Boston%2C%20MA%2C%20USA

**Updated version on left -------- original on right**

image

image

---

### DIFF
--- a/apps/site/assets/ts/components/Headsign.tsx
+++ b/apps/site/assets/ts/components/Headsign.tsx
@@ -115,7 +115,9 @@ const HeadsignComponent = (props: Props): ReactElement<HTMLElement> => {
       <div className="m-tnm-sidebar__headsign">
         {renderHeadsignName(props)}
 
-        {routeType === 2 && renderTrainName(`Train ${headsign.train_number}`)}
+        {routeType === 2 && headsign.train_number
+          ? renderTrainName(`Train ${headsign.train_number}`)
+          : null}
       </div>
       <div className="m-tnm-sidebar__schedules">
         {headsign.times.map((time, idx) => {
--- a/apps/site/assets/ts/tnm/helpers/process-realtime-data.ts
+++ b/apps/site/assets/ts/tnm/helpers/process-realtime-data.ts
@@ -121,7 +121,7 @@ const buildHeadsign = (
           ? {
               track: prediction.track,
               time: prediction.time,
-              status: null,
+              status: prediction.status,
               // eslint-disable-next-line camelcase
               schedule_relationship: prediction.schedule_relationship
             }


hannahpurcell:
HA!  Omg, glad you found the hardcoded bit.  (I wonder why it was done in the first place?)

I'm trying to test it out, but seems like all the CR schedules are properly built atm with boarding status and time-based prediction.  To mimic the GTFS data that TRC provided in the screenshot (where `arrival_time` and `departure_time` is `null`), I tried just replacing line 30 in `predictions/lib/parser.ex` with nil.  I _think_ that effectively recreates the scenario TRC was seeing, but not positive.

That caused all upcoming trips to be marked as "Scheduled", even if the train was currently boarding (which I believe is what TRC was seeing?)  Is this a proper way to test this?

MBTA.com:
Screen Shot 2021-11-02 at 3 46 48 PM

Localhost:
Screen Shot 2021-11-02 at 3 46 57 PM

thecristen:
@hannahpurcell A LOT of the time, the `prediction.status` is _actually_ nil, so you might have to manually add one (in the same place you were editing) to see this in action.

hannahpurcell:
> @hannahpurcell A LOT of the time, the `prediction.status` is _actually_ nil, so you might have to manually add one (in the same place you were editing) to see this in action.

ah, cool!  Thanks!  I changed line 34 to be `"All aboard"`, which is the standard string on predictions, and I got the same result as the screenshot above.  Still seems like when prediction times are nil and board status exists, we're ignoring the board status?  (Or my testing is off!)

thecristen:
Hey I think you're onto something; if I edit the time to `nil` as you describe, the `status` stops showing too. Thanks!

thecristen:
@hannahpurcell It turns out trying to create predictions with nil time to render in Transit Near Me is surprisingly tricky! I took a try at it at the `cbj/prediction-without-arrival-testing` branch -- the results on TNM are super interesting/weird! Will dig into this more tomorrow. 🤔 

image

thecristen:
We were filtering out nil prediction times pretty early on, so now I'm letting those show in the frontend but have to make a lot of upstream adjustments! I have some additional filtering to figure out but making good progress on this! 

image

thecristen:
Closing as this'll all be redundant due to #1476 and the stops page redesign.

PR #1119, opened 2021-11-03
More retroactive test fixes + updates

thecristen:
No ticket, just took some time to dig into what was going on with the tests that we'd recently skipped. 
* The first two commits make additional fixes that were needed after we updated the GL eastbound termini, and updates the corresponding tests. 
* The third commit unskips a test that now works. 
* The fourth commit refactors several tests to get rid of the `Floki.parse` deprecation message. And then I decided to stop and make this PR. :)

### DIFF
--- a/apps/site/lib/green_line.ex
+++ b/apps/site/lib/green_line.ex
@@ -13,11 +13,11 @@ defmodule GreenLine do
 
   @termini %{
     {"Green-B", 0} => "place-lake",
-    {"Green-B", 1} => "place-pktrm",
+    {"Green-B", 1} => "place-gover",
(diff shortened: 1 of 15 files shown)


hannahpurcell (approved):
Wow!!  I can't figure out how all these skipped tests were ultimately sorted out??  Was it a TRC change, or did the green_line_test file somehow have something to do with it?

thecristen:
@hannahpurcell It was partly that I'd forgotten to update the termini that are hardcoded into the `GreenLine` module, but ALSO the TRC change _happened_ to break our implementation in `get_green_branch` that picks the right shapes for each GL branch. 🥲

PR #1120, opened 2021-11-03
refactor(Site.Stream.Vehicles): add rescue

thecristen:
No ticket. This mostly just affects the console on application startup.

Because `Site.Stream.Vehicles` starts before `Site.Endpoint`, but the broadcasted events work with ETS tables that are set up with the Endpoint, unhandled errors come up when broadcasting events while the application continues to start up. Here we catch those errors only, and discard the events that happen before the application finishes starting up.

Open to alternate ways of handling this, but this change can be a stopgap measure to perhaps a smarter change. One obvious alternative is starting `Site.Endpoint` _before_ `Site.Stream.Vehicles` in `Site.Application`, but I found this to degrade the user experience quite a lot (visiting a schedule page, vehicles take a noticeably long time to populate the map, since they start getting streamed/downloaded later).

### DIFF
--- a/apps/site/lib/site/stream/vehicles.ex
+++ b/apps/site/lib/site/stream/vehicles.ex
@@ -15,13 +15,7 @@ defmodule Site.Stream.Vehicles do
   end
 
   def handle_info({:remove, ids}, state) do
-    _ =
-      Endpoint.broadcast(
-        "vehicles:remove",
-        "remove",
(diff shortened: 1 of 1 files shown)


hannahpurcell (approved):
Awwwwhhh!  no more lump of huge red text on local app load!  I like it

PR #1121, opened 2021-11-04
Update package-lock.json and Dockerfile

thecristen:
From what I can tell, the version of npm used to generate the `package-lock.json` file happens to be newer than the version of npm inside the docker image, causing old lockfile warning messages and such.

> `npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!`

I upgraded `npm` both locally and in the `Dockerfile` (happens to be version 8), [the warning is gone from the corresponding docker build](https://github.com/mbta/dotcom/runs/4097756709). I am faintly hoping this might solve or at least make progress toward the issues presented in [failing docker builds on dependabot PRs](https://github.com/mbta/dotcom/runs/4077843009?check_suite_focus=true).

This also deletes the extraneous root level `package-lock.json`.


hannahpurcell (approved):
Oh my goursh, I didn't realize the root package-lock was unneeded 😅  Yay!  Thanks

PR #1123, opened 2021-11-05
Remove duplicate related links from trip planner

i0sea:
#### Summary of changes
**Asana Ticket:** [Trip Planner | Remove duplicate related link](https://app.asana.com/0/385363666817452/1200984438570060/f)

Filtered out duplicate schedule links in trip planner itinerary.

### DIFF
--- a/apps/site/lib/site_web/controllers/trip_plan_controller.ex
+++ b/apps/site/lib/site_web/controllers/trip_plan_controller.ex
@@ -114,10 +114,10 @@ defmodule SiteWeb.TripPlanController do
   end
 
   defp get_route(link) do
-    if is_list(Enum.at(link.text, 0)) do
-      Enum.at(Enum.at(link.text, 0), 1)
+    if is_bitstring(link.text) do
+      link.text
     else
-      Enum.at(link.text, 0)
+      link.text |> List.to_string()
     end
   end
 
--- a/apps/site/lib/site_web/controllers/trip_plan_controller.ex
+++ b/apps/site/lib/site_web/controllers/trip_plan_controller.ex
@@ -113,6 +113,18 @@ defmodule SiteWeb.TripPlanController do
     end
   end
 
+  defp get_route(link) do
+    if is_list(Enum.at(link.text, 0)) do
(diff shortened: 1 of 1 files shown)


apps/site/lib/site_web/controllers/trip_plan_controller.ex
```diff
@@ -113,6 +113,18 @@ defmodule SiteWeb.TripPlanController do
     end
   end
 
+  defp get_route(link) do
+    if is_list(Enum.at(link.text, 0)) do
```
thecristen, line 117: This will throw an error when `link.text` does not happen to be a list, which happens sometimes on non-MBTA routes. One example link I ran into when planning a trip from South Station to Boston Logan Airport:

```
%Site.TripPlan.RelatedLink{
  icon_name: :massport_shuttle,
  text: "Massport schedules",
  url: "https://massport.com/"
}
```

Can we add a condition to check for/handle this case where text is a single string?

thecristen: Also! I'd suggest using `List.to_string()` here to get the text, e.g.

```elixir
[["Route ", "SL1"], " schedules"] |> List.to_string()
# returns "Route SL1 schedules"
```

thecristen (changes requested):
Good start! I noted a case where the logic breaks a bit so we'll need to fix that; I think addressing that will probably fix the test failures too. 👍🏼

apps/site/lib/site_web/controllers/trip_plan_controller.ex
i0sea: @thecristen should I be using `List.to_string()` for all the cases where link.text is an array ie.
```
["Green Line E", " schedules"],
[["Route ", "SL1"], " schedules"],
["View fare information"],
```

thecristen: @i0sea Yeah I'd say so!

thecristen (approved):

thecristen:
@i0sea Looks good! 👍 Please squash and merge, rewriting the commit title to be a bit more formal when you do.

PR #1124, opened 2021-11-09
Layout for mobile menu

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Mobile menu content and layout](https://app.asana.com/0/555089885850811/1201306740497114)

Need to:
- add tests
- handle the MD breakpoint when feature flag is OFF.  what do I do when some of my changes were CSS only?

---


thecristen (commented):
Ideally, I'd like to minimize modifications to the existing header/nav code. Can we separate the new and old navs out by checking for the flag one level higher, in the `app.html.eex`, e.g.

```eex
<%# leave the original header and nav %>
<%= render "_header.html", @conn |> Plug.Conn.assign(:page, get_page_classes(@view_module, @view_template)) |> forward_assigns %>

<%= if Laboratory.enabled?(@conn, :nav_redesign) do %>
    <%= render "_new_header.html", @conn |> Plug.Conn.assign(:page, get_page_classes(@view_module, @view_template)) |> forward_assigns %>
<% end %>
```

And `_new_header.html.eex` can render a `_new_nav_mobile.html.eex` and whatever other new pieces of the new nav. 

I'm guessing we'd still want to make some modifications to the original nav, say to hide the old mobile nav if the flag is enabled so that the new mobile nav can shine. That's fine -- I still think separating out the files will make it easier to reason about the changes and eventually let us delete the old file when we're ready to change over.

(More review coming)

thecristen:
> handle the MD breakpoint when feature flag is OFF. what do I do when some of my changes were CSS only?

oof. great question. I think in that case your best bet is to use the feature flag to add a special class name in the template, and to make your CSS adjustments using that class.

apps/site/lib/site_web/templates/layout/_nav_mobile.html.eex
```diff
@@ -1,22 +1,51 @@
...
-      <%= Phoenix.HTML.Link.link to: search_path(@conn, :index), class: "mobile-nav-link" do %>
-        <span class="mobile-nav-item-header">
-          <%# we put the icon and mode name on one line to avoid the whitespace being collapsed. %>
-          <%= svg_icon(%SvgIcon{icon: :search, show_tooltip?: false}) %> <span>Search</span>
-        </span>
+<%= if Laboratory.enabled?(@conn, :nav_redesign) do %>
+  <!-- New nav -->
+  <nav class="m-menu" id="navmenu">
+    <div class="m-menu__cover"></div>
+    <!-- Do we need to use ul for access? -->
```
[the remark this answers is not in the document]
thecristen: I've been thinking about how to best structure these submenus, and there are several options -- from an accessibility standpoint it's most important that the menu structures can be understood programmatically, and nested unordered lists or a list of `<nav>` elements (one nav for each submenu, each nav having a distinct `aria-label`) both can work great for that. [Some more about that here](https://www.w3.org/WAI/WCAG21/Techniques/html/H97#examples).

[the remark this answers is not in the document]
thecristen: Ultimately we want to make sure each section is accessibly labelled, [example here](https://inclusive-components.design/menus-menu-buttons/#tablesofcontent) (the whole page is a neat read!)

apps/site/test/site_web/controllers/event_controller_test.exs
```diff
@@ -69,6 +69,7 @@ defmodule SiteWeb.EventControllerTest do
     end
   end
 
+  # setup similar test for mobile menu?
```
thecristen, line 72: Yeah, probably for testing the layout view https://github.com/mbta/dotcom/blob/eaac9fdb9b32f0c55f64689ec23de01de283a513/apps/site/test/site_web/views/layout_view_test.exs#L32

thecristen (changes requested):
I feel like I should reserve review on the CSS until we tackle the styling ticket? At the very least, I have plenty of comments already, mostly around structure and accessibility things. The interaction feels pretty tight and I think this is off to a great start.

hannahpurcell:
> And `_new_header.html.eex` can render a `_new_nav_mobile.html.eex` and whatever other new pieces of the new nav.

Aha, ok yeah!

> ```
> <%# leave the original header and nav %>
> ```

This is written to leave the header in all cases, but I'm putting it under the feature flag "else" (if Lab.enabled, show the new one, else show the old).  That's what I understand the goal to be here

apps/site/lib/site_web/templates/layout/_nav_mobile.html.eex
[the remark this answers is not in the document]
hannahpurcell: How will converting the navs to accordions impact the question of best-labeling?  Not sure how much I should do on this piece because the components will be replaced in the accordion ticket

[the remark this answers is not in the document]
thecristen: @hannahpurcell whatever labels you have here will likely have to be transferred over to work with the accordion markup.

apps/site/assets/css/_header.scss
```diff
@@ -240,3 +241,96 @@
...
+      display: none;
+    }
+  }
+  .header-search__toggle {
+    @include media-breakpoint-up(sm) {
+      display: none;
+    }
+  }
+}
+nav.m-menu {
```
thecristen, line 263: The mobile menu is technically still "visible" to assistive tech -- can we hide this in the larger breakpoints? `display: none` or `visibility: hidden` should do it.

image

(screenshot is via the accessibility inspector in Firefox, don't remember what the equivalent in Chrome's dev tools are)

apps/site/lib/site_web/templates/layout/_new_nav_mobile.html.eex
```diff
@@ -0,0 +1,29 @@
...
+        <%= for %{sub_menu_section: sub_menu_section} = sub_menu <- sub_menus do %>
+          <!-- 'accordion' class marks the intended pass-of-focus from the menu button to the navbar, though divs don't have focus... -->
+          <!-- Todo: so when accordions are updated, also update mobile-menu.js -->
+          <nav aria-labelledby="accordion-heading">
+            <div id="accordion-heading" class="heading-4 accordion" style="color:white;"><%= sub_menu_section %></div>
+            <%= if Map.has_key?(sub_menu, :links) do %>
+              <div style="background:#e2ecf9;">
+                <%= for {link_name, href, link_host} <- Map.get(sub_menu, :links) do %>
+                  <% icon = if link_host === :external_link, do: fa("external-link"), else: "" %>
+                  <%= link to: href, do: [content_tag(:div, link_name ), icon] %>
```
thecristen, line 17:
```suggestion
                  <%= link to: href, do: [link_name, icon] %>
```

Leaving note for future styles work: let's remove the `content_tag` so we're not adding `<div>` in every link (which will only work after we style the links reasonably).

thecristen (approved):
This is a great foundation, thank you! Approving because any notes I'd add could be solved in the upcoming tickets!

apps/site/assets/js/mobile-menu.js
```diff
@@ -0,0 +1,83 @@
...
+        closeMenus();
+      }
+    })
+  }
+  
+  document.addEventListener(
+    "turbolinks:load",
+    () => window.nextTick(setupMobileMenu),
+    { passive: true }
+  );
```
thecristen, line 82: With the flag _not_ enabled this file raises a few errors e.g. `Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')` because the various `document.querySelector` calls (rightfully) return no nodes -- let's only setup the mobile menu if the `button.m-menu__toggle` is present in the DOM.

thecristen (changes requested):
OK just kidding I do have one change to request! This will probably fix the broken integration tests too.

apps/site/assets/js/mobile-menu.js
hannahpurcell: OHHHHHH!  Thank you

apps/site/assets/js/mobile-menu.js
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    // Veil click closes
+    veil.addEventListener("click", () => {
+      closeMenus();
+    })
+
+    // Esc key closes
+    document.body.addEventListener("keydown", (e) => {
+      handleNativeEscapeKeyPress(e, closeMenus);
```
thecristen, line 72: Sorry I didn't explicitly include it as part of my suggestion earlier, but I'm pretty sure you'll need to also import the `handleNativeEscapeKeyPress` function 😅

hannahpurcell: Oh yeah, I haven't pushed up my latest fixes yet!  Caught that suggestion 😁

apps/site/assets/js/test/psl-page-setup_test.js
```diff
@@ -39,7 +39,7 @@ const pslHtml = `
   </div>
 `;
 
-describe.only("psl-filters", () => {
+describe("psl-filters", () => {
```
hannahpurcell, line 42:
```suggestion
describe.only("psl-filters", () => {
```
There are a couple tests to fix as a result of this, so should be a separate PR

thecristen (approved):
🚀

PR #1125, opened 2021-11-09
feat: Enable experimental features in API requests

thecristen:
Configurable via environment variable, defaults to enabled on non-production environments.

#### Summary of changes
**Asana Ticket:** [Enable MBTA V3 API experimental features](https://app.asana.com/0/555089885850811/1201306297717355/f)

### DIFF
--- a/apps/site/config/config.exs
+++ b/apps/site/config/config.exs
@@ -76,7 +76,8 @@ config :site, :react,
   build_path: Path.join(File.cwd!(), "/apps/site/react_renderer/dist/app.js")
 
 config :site,
-  allow_indexing: false
+  allow_indexing: false,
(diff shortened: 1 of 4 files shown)


i0sea (commented):
LGTM! One question I have is do the experimental features also need to be enabled in the config files for dev and test as well?

thecristen:
@i0sea Good question - not in this case. The compiled configuration for all environments (dev, test, prod) uses the configuration from `config/config.exs` as a base, with values only getting overwritten from that if explicitly reset (as it is in this PR for `config/prod.exs` for this feature)

```elixir
# In config/config.exs

# Import environment specific config. This must remain at the bottom
# of this file so it overrides the configuration defined above.
import_config "#{Mix.env()}.exs"
```

PR #1127, opened 2021-11-11
debounce the Algolia autocomplete result list

thecristen:
Instead of sending queries on every input change, limit queries to being sent every .5 second. This should dramatically reduce the number of extraneous requests sent to Google's Places API while maintaining the interactivity and accuracy of autocomplete search results.


hannahpurcell (approved):
I don't seem to have an algolia key to test it, but the one-line change looks great

PR #1128, opened 2021-11-12
Adjust navbar breakpoint between mobile / desktop views

hannahpurcell:
No ticket.  Seemed easier to just set up the new breakpoint instead of setting up a bunch of temporary classes to feature-flag this in the navbar rework.

### DIFF
--- a/apps/site/assets/css/core.scss
+++ b/apps/site/assets/css/core.scss
@@ -13,11 +13,9 @@
 @import 'layout-core-shared';
 
 html {
-  width: 100%;
   height: 100%;
   position: relative;
   scroll-behavior: smooth;
-  overflow-x: hidden;
 }
 
 body {
--- a/apps/site/assets/css/core.scss
+++ b/apps/site/assets/css/core.scss
@@ -13,9 +13,11 @@
 @import 'layout-core-shared';
 
 html {
+  width: 100%;
   height: 100%;
   position: relative;
   scroll-behavior: smooth;
+  overflow-x: hidden;
 }
 
 body {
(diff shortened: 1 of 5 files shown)


apps/site/assets/css/core.scss
```diff
@@ -13,9 +13,11 @@
 @import 'layout-core-shared';
 
 html {
+  width: 100%;
   height: 100%;
   position: relative;
   scroll-behavior: smooth;
+  overflow-x: hidden;
```
thecristen, line 20: I'm a bit concerned about this having unintended consequences on other pages, particularly with any other elements that might have or rely on horizontal overflow/scroll behavior. 

So I'm curious what this particular change is meant to do, and whether this is the least-intrusive change we can do to accomplish it?

hannahpurcell: Ah ok, I can omit this bit.  I noticed that some viewport sizes make the homepage have a horizontal scroll when it shouldn't.  I didn't think we had any pages on dotcom where there is horizontal scrolling -- but forgot to account for when the page is zoomed in.  Thanks for catching!

hannahpurcell, line 20:
```suggestion
  height: 100%;
  position: relative;
  scroll-behavior: smooth;
```

apps/site/lib/site_web/views/event_view.ex
```diff
@@ -218,7 +218,7 @@ defmodule SiteWeb.EventView do
         |> Timex.Duration.from_clock()
         |> Timex.Duration.to_string()
       rescue
-        e ->
+        _e ->
```
thecristen, line 221: 😅 heh thanks!

thecristen (approved):
👉🏼 looks good!

PR #1129, opened 2021-11-15
Use Silver Line Icon in Search

i0sea:
#### Summary of changes
**Asana Ticket:** [🐞 Inconsistency for South Station in search results](https://app.asana.com/0/385363666817452/1199926429840347/f)

### DIFF
--- a/apps/site/assets/js/algolia-result.js
+++ b/apps/site/assets/js/algolia-result.js
@@ -109,12 +109,16 @@ function _subwayRouteIcon(routeId) {
   return mapper[routeId];
 }
 
-function iconFromGTFSId(id) {
+function iconFromGTFSId(id, route_name) {
+  console.log("in the gtfs " + route_name);
   const toSubway = _subwayRouteIcon(id);
   if (toSubway) {
     return Icons.getFeatureIcon(toSubway);
   }
-  if (id === "Silver Line") return Icons.getFeatureIcon("bus");
+  if (route_name.startsWith('SL')) {
+    console.log("in the silver zone " + id);
+    return Icons.getFeatureIcon("silver_line");
+  }
   if (id in ["commuter_rail", "bus", "ferry"]) {
     return Icons.getFeatureIcon(id);
   }
@@ -124,18 +128,21 @@ function iconFromGTFSId(id) {
   return Icons.getFeatureIcon(id);
 }
 
-function iconsFromGTFSIds(id) {
+function iconsFromGTFSIds(id, route_name) {
+  // console.log(id); id for silver line is bus!
+  console.log(route_name);
   if (Array.isArray(id)) {
-    return id.map(icon => iconFromGTFSId(icon));
+    return id.map(icon => iconFromGTFSId(icon,route_name));
   }
-  return [iconFromGTFSId(id)];
+  return [iconFromGTFSId(id,route_name)];
 }
 
-export function iconFromGTFS(id, ancestry) {
+export function iconFromGTFS(id, ancestry, route_name) {
+  console.log("iconFromGTFS "+ route_name)
   if (!id) {
     return TEMPLATES.fontAwesomeIcon.render({ icon: "fa-info" });
   }
-  let icons = iconsFromGTFSIds(id);
+  let icons = iconsFromGTFSIds(id, route_name);
   if (ancestry) {
     icons = [...new Set([...icons, ...iconsFromGTFSAncestry(ancestry)])];
   }
@@ -220,6 +227,7 @@ function getPopularIcon(icon) {
 }
 
 export function getIcon(hit, type) {
+  console.log("in getIcon " + hit._highlightResult.route.name.value);
   switch (type) {
     case "locations":
       return _contentIcon({ ...hit, _content_type: "locations" });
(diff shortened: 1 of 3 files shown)


apps/site/lib/site_web/templates/partial/_hidden_icons.html.eex
```diff
@@ -22,7 +22,7 @@
     end
   %>
   <%=
-    for mode <- [:subway, :commuter_rail, :bus, :ferry] do
+    for mode <- [:subway, :commuter_rail, :bus, :ferry, :silver_line] do
       content_tag(:div, [mode_icon(mode, :small)], id: ["icon-feature-", Atom.to_string(mode)])
```
thecristen, line 26: I found the culprit that makes `mode_icon(:silver_line, :small)` always return the bus icon!! 

https://github.com/mbta/dotcom/blob/867be60394ddf736e264eac875d9eb6aeb959ea5/apps/site/lib/site_web/views/helpers.ex#L79-L81

apps/site/lib/site_web/templates/partial/_hidden_icons.html.eex
```diff
@@ -57,4 +57,7 @@
   <div id="icon-feature-map-station-marker">
     <%= svg("icon-map-station-marker.svg") %>
   </div>
+  <div id="icon-feature-silver_line">
+    <%= svg("icon-silver-line-default.svg") %>
+  </div>
```
thecristen, line 62: Right now the `for mode <- [:subway, :commuter_rail, :bus, :ferry, :silver_line] do` loop is outputting another `<div id="icon-feature-silver_line">` with a bus icon inside, and since that's earlier up the template that one's getting picked up first. So you've got two options here -- you can leave this here and remove `:silver_line` from the loop, or remove this HTML and fix the `mode_icon` function above.

apps/site/assets/js/algolia-result.js
```diff
@@ -109,12 +109,16 @@ function _subwayRouteIcon(routeId) {
...
 
-function iconFromGTFSId(id) {
+function iconFromGTFSId(id, route_name) {
+  console.log("in the gtfs " + route_name);
   const toSubway = _subwayRouteIcon(id);
   if (toSubway) {
     return Icons.getFeatureIcon(toSubway);
   }
-  if (id === "Silver Line") return Icons.getFeatureIcon("bus");
+  if (route_name.startsWith('SL')) {
```
thecristen, line 118: definitely on the right track using this function to pick out the Silver Line

apps/site/assets/js/algolia-result.js
```diff
@@ -220,6 +227,7 @@ function getPopularIcon(icon) {
 }
 
 export function getIcon(hit, type) {
+  console.log("in getIcon " + hit._highlightResult.route.name.value);
```
thecristen, line 230: `hit.route` returns the Silver Line route here for me (at least which the searches I've been trying) --  and yep` _iconFromRoute`(in line 238) is currently returning "bus" for it -- so we'll have to add an extra bit of logic so it returns something different. I'd try inspecting/adjusting `_iconFromRoute` function so that it returns `silver_line`

thecristen (commented):
Left a few comments, hoping some of it is helpful!

apps/site/assets/js/algolia-result.js
i0sea: What stops are you searching? I've been trying to use `south` and `Chelsea`

i0sea: What searches were you using? I've tried `Chelsea` and `South`

thecristen: `Silver` (to pull up the Silver line route in the dropdown)

apps/site/assets/js/algolia-result.js
```diff
@@ -259,7 +261,8 @@ function getTransitIcons(hit) {
   }
   const icons = iconFromGTFS(
     hit.related_transit_gtfs_id,
-    hit.related_transit_gtfs_ancestry
+    hit.related_transit_gtfs_ancestry,
+    hit._highlightResult.route.name.value
```
thecristen, line 265:
```suggestion
    hit.related_transit_gtfs_ancestry
```

I don't see any changes to `iconFromGTFS` to have it accept a third argument so I'm presuming this was left here mistakenly.

apps/site/assets/js/algolia-result.js
```diff
@@ -114,7 +114,6 @@ function iconFromGTFSId(id) {
   if (toSubway) {
     return Icons.getFeatureIcon(toSubway);
   }
-  if (id === "Silver Line") return Icons.getFeatureIcon("bus");
```
thecristen, line 117: 🤔 I wonder if this should return `Icons.getFeatureIcon("silver_line")`?

thecristen (changes requested):
Just a couple tiny comments!

apps/site/assets/js/algolia-result.js
i0sea: Ah the id is `silver_line` now so the id can just be passed as the parameter like the other types.

thecristen (approved):
:tada:

PR #1130, opened 2021-11-16
Link to Trip Planner From Stop Pages

i0sea:
#### Summary of changes
**Asana Ticket:** [Stop pages | Replace directions with our Trip Planner](https://app.asana.com/0/385363666817452/1200761810257407/f)
Link to Trip Planner directions instead of maps' on stop pages


hannahpurcell (changes requested):
This link works in many cases, but there are several other cases where the simple `/trip-planner/to/{stop.name}` construction isn't specific enough.  For example, Back Bay is a neighborhood _and_ a station.  So `/trip-planner/to/back-bay` directs to the center of Back Bay, but the station is in a totally different spot.  You'll may need to use the actual stop address to create the link.

i0sea:
> This link works in many cases, but there are several other cases where the simple `/trip-planner/to/{stop.name}` construction isn't specific enough. For example, Back Bay is a neighborhood _and_ a station. So `/trip-planner/to/back-bay` directs to the center of Back Bay, but the station is in a totally different spot. You'll may need to use the actual stop address to create the link.

@hannahpurcell I see there's a check if `stop.address` is null. In what case is it null and what should be used for the route in such a case?

apps/site/assets/ts/stop/components/LocationBlock.tsx
```diff
@@ -33,18 +32,14 @@ const addressOrMunicipality = (stop: Stop): ReactElement | null => {
   return null;
 };
 
+const formatMapLink = (location: string | null): string =>
+  location
+    ? `/trip-planner/to/${location.replace(/\s+/g, "-").toLowerCase()}`
+    : "";
+
```
thecristen, line 39: I think we need the encoder back! :laughing

thecristen, line 39: I think we need the encoder back! :smile: I found this works pretty well for stations so far, except we have a few stations with a `/` in the name - Charles/MGH, Kendall/MIT, JFK/Umass - for those stations the link leads to a 404.

If we use the encoder (`window.encodeURIComponent()`) function on the name/address that fixes the issue, so `/trip-planner/to/charles/mgh-station` (doesn't work) becomes `/trip-planner/to/charles%2Fmgh-station` (which works!)

apps/site/assets/ts/stop/components/LocationBlock.tsx
```diff
@@ -53,15 +48,16 @@ const streetViewLink = (stop: Stop): string =>
...
   streetViewUrl
 }: Props): ReactElement<HTMLElement> => (
   <div className="m-stop-page__location-block">
     {addressOrMunicipality(stop)}
     <div className="m-stop-page__location-links">
       <div className="m-stop-page__location-link">
         <a
-          href={directionLink(stop, encoder)}
+          href={formatMapLink(
+            stop["station?"] ? `${stop.name}-station` : stop.address
```
thecristen, line 59: When `stop.address` is `null` this ends up taking the user back to the stop page. BUT I just realized that the link seems to work fine with latitude/longitude - I just tried `/trip-planner/to/42.255009,-71.006352` and it worked great. So let's use `latLngString(stop)` here in place of null address (or maybe in place of address altogether? I haven't yet found a not-station with a valid address though...)

thecristen (changes requested):

apps/site/assets/ts/stop/components/LocationBlock.tsx
i0sea: Are there any other cases the encoder would handle? Would a replace `/` with `%2F` suffice?

thecristen: Not sure off the top of my head, but we can try to look at all the stops and see.. but there are hundreds (maybe 1000?) of those. Not to mention we don't know what future stops (rare, but does happen) might look like!

thecristen (approved):

thecristen:
Approving! I think this should work OK but I'd like to do a bit more QA before releasing it. @i0sea welcome to merge into master and we can test on the dev server, or leave this open and deploy this branch to one of our staging servers so we can test there.

i0sea:
> 

I'll try deploying to one of the staging servers

hannahpurcell (approved):
I think this looks good!  Checked out dev-green and poked around the weirdly-named stations (Harvard Station vs Harvard Avenue, JFK/UMass, etc) and so far so good.  Think it's ready, @thecristen ?

thecristen:
Okay, I found a few and these look like bonafide edge cases:
* /stops/Boat-Hingham (ferry)
* /stops/Boat-Hull (ferry)
* /stops/place-FR-0201 (Concord commuter rail station)

I was about to propose implementing workarounds for these specific edge cases, but your idea to use lat/lng is better (future proof in case place names/ids/addresses change, safer and more accurate (every single station will definitely have coordinates), go for it 🚀 )

hannahpurcell (approved):
Oh my gosh!  I totally forgot that the Trip Planner attempts to reverse-geolocate lon/lat combos to populate the input box with a human-readable name!  That's perfect!  That's super user-friendly.  I feel like I sent us on a wild goose chase earlier, but super great final solution, yay Isza!!  Still might want a test deploy before merging?  Check in with Cristen

thecristen (approved):
Good with me! Thanks for going back and forth on this so much during QA.

i0sea:
@thecristen can you squash and merge this?

PR #1131, opened 2021-11-16
Commuter rail crowding icons and tooltips

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Display crowding info in CR timetable](https://app.asana.com/0/555089885850811/1201303796115903)

---

### DIFF
--- a/apps/schedules/lib/repo.ex
+++ b/apps/schedules/lib/repo.ex
@@ -11,7 +11,7 @@ defmodule Schedules.Repo do
 
   @default_timeout 10_000
   @default_params [
-    include: "trip",
+    include: "trip,trip.occupancies",
     "fields[schedule]":
       "departure_time,arrival_time,drop_off_type,pickup_type,stop_sequence,timepoint",
     "fields[trip]": "name,headsign,direction_id,bikes_allowed"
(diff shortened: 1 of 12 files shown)


apps/site/assets/ts/schedule/components/line-diagram/LiveCrowdingIcon.tsx
```diff
@@ -1,26 +1,45 @@
...
-          placement: "left",
+          placement: tooltipPlacement || "left",
           animation: false,
           html: true
         }}
       >
-        {crowdingIcon(`c-icon__crowding--${crowding}`)}
+        {crowdingIcon(
+          `c-icon__crowding--${crowding} ${
+            isCommuterRail ? "commuter-rail" : ""
```
thecristen, line 40: We have a `.commuter-rail` class name used elsewhere, I'm just afraid of collision... can we name this something more unique?

apps/schedules/lib/repo.ex
```diff
@@ -11,7 +11,7 @@ defmodule Schedules.Repo do
 
   @default_timeout 10_000
   @default_params [
-    include: "trip",
+    include: "trip,trip.occupancies",
```
thecristen, line 14: Nice find! Does this fail gracefully if the experimental header is **not** enabled?

apps/site/lib/site_web/templates/schedule/_timetable.html.eex
```diff
@@ -63,6 +63,11 @@
                 <% else %>
                   <div class="sr-only">Bicycles not allowed</div>
                 <% end %>
+                <%= Site.React.render("LiveCrowdingIcon", %{
+                  crowding: schedule.trip.occupancy,
+                  tooltipPlacement: "top",
+                  isCommuterRail: true
+                }) %>
```
thecristen, line 70: This was such a great idea to use the existing react component in this way!

thecristen (changes requested):
Super! A couple minor notes/questions:

thecristen (approved):
💯 !

PR #1133, opened 2021-11-17
refactor(FinderApi): Get trip route ID via route pattern

thecristen:
No ticket, just noticed a `[warn] route_id_not_found route_id=Green, trip_id=ADDED-1591071232` logged while working on another task. Similar warnings are logged many times a day (see chart from Splunk below!).

image

---

Sometimes trips don't have associated schedules: for example trip id `ADDED-1591071232` was added to the rotation but wasn't originally scheduled to run. This can also happen with shuttle trips. In these cases, using `Schedules.Repo.schedule_for_trip()` will return empty.

Since `get_route_id` is just trying to fetch the route ID for the trip, we can do that by fetching the trip directly, and then fetching the trip's associated route pattern, from which we can get the ID.


hannahpurcell (approved):
Oh hey, yeah!  Opening the LazyAccordion on the Green Line Schedule Finder triggered the warning and now gone 👍

PR #1135, opened 2021-11-17
CR crowding improvements

hannahpurcell:
Updated wording of accordion icon key.
Don't render crowding icon when crowding data is nil.


thecristen (approved):
Sounds about right

PR #1136, opened 2021-11-18
Fetch occupancy every time a request includes "trip"

hannahpurcell:
Also updated wording on crowding, after Keolis request.

### DIFF
--- a/apps/schedules/lib/repo.ex
+++ b/apps/schedules/lib/repo.ex
@@ -277,6 +277,8 @@ defmodule Schedules.Repo do
     |> Enum.map(&Task.await/1)
   end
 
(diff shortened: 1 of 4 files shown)


apps/schedules/lib/repo.ex
```diff
@@ -277,6 +277,8 @@ defmodule Schedules.Repo do
     |> Enum.map(&Task.await/1)
   end
 
+  # Fetching predictions will also insert trips into cache using this function
+  # and that happens more frequently than fetching schedules due to realtime tracking
```
thecristen, line 281: 👏🏼

thecristen (approved):
Thank you!

PR #1137, opened 2021-11-22
Hide  Registration if Null

i0sea:
#### Summary of changes
**Asana Ticket:** [Hide registration section when registration URL isn't provided](https://app.asana.com/0/385363666817452/1201348288803391/f)

Don''t show registration link if it isn't available. Instead uses previous copy of `Virtual meeting registration will be available soon`


thecristen (approved):
Sorry to have missed this earlier -- looking good to me! ✨

PR #1141, opened 2021-11-30
tests: Update tests now that 2 GL-B stops are out

thecristen:
No ticket, just wanted to fix failing Green line tests now that the old stops are fully phased out of the data. Also noticed some unused CSS along the way.


hannahpurcell (approved):
Looks good!  other than the failing snapshot test 😕

thecristen:
fixed the snapshot and found that the Prettier rules all changed and an ESLint rule started flagging some of our code.... which is super strange, as I don't think any of the related configuration or dependencies had changed (maybe the `package-lock.json` update did it?). Anyhow now that everything is passing I'll merge it in and we'll see what happens!

PR #1143, opened 2021-12-02
Updated subway map

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [🔧  Maps | update map preview photo](https://app.asana.com/0/555089885850811/1201417381255112)

I wish this image could be dynamically updated by the CMS, but that would take a lot more work.  Backlog idea?


thecristen (approved):
Well okay! 💃🏼

PR #1144, opened 2021-12-03
Prevent scroll to top in ScheduleFinderModal opened from LineDiagram

hannahpurcell:
#### Summary of changes
**Asana Ticket:** [Schedule Finder resets scroll position to top ](https://app.asana.com/0/555089885850811/1201439846310850)

EDIT:  Skipped the tests that were failing.  Originally, I removed them as unhelpful, but I'm actually leaving them for investigative context.

Wow, I made a draft PR, wrote a massive description, and that gave me an idea to redo the whole solution.  

Ok so, turns out there are 2 separate instances of ScheduleFinderModal on a schedules page, which is not great.  ScheduleFinderModal was a child component of the ScheduleFinder AND it's a child component of the Line Diagram.  The latter instance re-rendered every single time the Line Diagram receiveed new realtime data. 

So I removed the Line Diagram instance and use the modal logic from ScheduleLoader to open the ScheduleFinder's instance of the modal.

### DIFF
--- a/apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
+++ b/apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
@@ -91,10 +91,6 @@ const ScheduleDirection = ({
   mapData,
   staticMapData,
   lineDiagram,
-  services,
-  stops,
-  today,
-  scheduleNote,
   busVariantId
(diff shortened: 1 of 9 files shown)


thecristen (approved):
Such a nice simplification!!

PR #1145, opened 2021-12-07
Test server improvements

thecristen:
#### Summary of changes
**Asana Ticket:** [Fix Cypress testing of mobile navigation menu](https://app.asana.com/0/385363666817452/1201439454001317/f) (final commit)

So this update changes a few things.
* new `mix compile.assets` command runs all the build steps needed to compile production front-end assets.
* The `mix phx.server` command has been aliased and will now behave differently when run in the test environment (invoked with `MIX_ENV=test`). Under test conditions it will compile assets first (since there is no dev server to handle asset compilation). 

Those should get the test server in good condition. So the rest of the commits address Cypress and using it to test the mobile navigation menu:
* Update Cypress to the latest version because why not
* Adds new Cypress command to simplify enabling flagged features
* Adds several tests to cover the mobile navigation menu.


hannahpurcell (approved):
Wow, this is so great!!  The compilation changes you make is absolute black-box magic to me.  But I understand and love the Cypress tests you added.  Thanks

PR #1148, opened 2021-12-13
mobile accordion

i0sea:
#### Summary of changes
**Asana Ticket:** [Accordion containers in mobile menu](https://app.asana.com/0/555089885850811/1201306740497115/f)

A draft pr for the mobile accordion. Currently this work wraps the mobile menu links within a toggle-able accordion. This doesn't include no javascript accordion(it is however built into the component) or accessibility checks


hannahpurcell (approved):
Cool!  

Figured out why the test is failing:  lines 46-49 of `mobile-menu.js` are no longer finding an element called "accordion", and the focus is not passing.  If you'd like, comment that piece out, and this should be ready to merge.  I'll add the focus in my CSS branch.  The accessibility work and no-JS work will be new, separate tickets (I think).

PR #1149, opened 2021-12-14
Refactor out shared data structuring code from Stops and Transit Near Me

thecristen:
#### Summary of changes
**Asana Ticket:** [🔧 Refactor out shared data structuring code from Stops and Transit Near Me](https://app.asana.com/0/385363666817452/1116698265374657/f) (partial)

This is part of refactoring the shared logic around Transit Near Me and Stops. Each commit message has additional commentary.

The output from the code changes in this PR (rendered predictions and schedules in Transit near me and in Stop pages) should be unchanged from current master.

Coming soon:
* There's even more refactoring of this shared logic left here. But I wanted to avoid making an _even bigger_ PR.
* Adding an enhancement to display prediction status even when we don't have prediction _time_
---


thecristen:
Closing as this'll change after the Stops redesign work.

PR #1154, opened 2021-12-17
Remove Feature Flag for New Event Agendas

i0sea:
#### Summary of changes
**Asana Ticket:** [Remove feature flag for new event agendas](https://app.asana.com/0/555089885850811/1201507862161340/f)

Remove feature flag from new event agendas. If new agenda is not used fallback to old style.


thecristen (approved):
Works great!

PR #1156, opened 2022-01-04
Update search result styling

i0sea:
#### Summary of changes
**Asana Ticket:** [Update search result styling](https://app.asana.com/0/555089885850811/1201244628068090/f)
Make some styling changes to search results

- Add project icon for results for projects
- Update placement of mode/line icons to be next to station/stop or project icon
- Update placement of text on to second line if more than two icons

### DIFF
--- a/apps/site/assets/js/algolia-result.js
+++ b/apps/site/assets/js/algolia-result.js
@@ -34,6 +34,7 @@ export const TEMPLATES = {
     {{#hasDate}}
     <div class="c-search-result__hit--vertical">
     {{/hasDate}}
+    <span>{{{hitIcon}}}</span>
     {{#id}}
     <a id="hit-{{id}}" class="${SELECTORS.result} c-search-result__link" href="{{hitUrl}}">
     {{/id}}
@@ -42,7 +43,6 @@ export const TEMPLATES = {
     {{/id}}
       <span class="c-search-result__hit-name">{{{hitTitle}}}</span>
     </a>
-    <span>{{{hitIcon}}}</span>
     {{#hasDate}}
     </div>
     {{/hasDate}}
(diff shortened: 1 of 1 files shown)


thecristen (changes requested):
I'm seeing all the icons on the left! 👍🏼 A few requested updates:

* The text of the result is supposed to be bumped to a new line if there are more than two icons with it, I'm not seeing it here:

image

* I'm not seeing a project icon here? It's a little hammer icon (example in mock 2.1.1.3)
* There are some cases where it looks like we might need to add a space between the icon and text, example in two spots here. Can we get the icon spacing consistent?

image

i0sea:
What are some examples of projects that should have a project icon?

i0sea:
Screen Shot 2022-01-14 at 11 52 39 AM
updated text to be on second line but now the icons are underlined on hover. Moved the icons span to be outside the anchor tag which got rid of the underlines but moved everything back to one line. Open to suggestions on a fix to resolve both issues.

apps/site/assets/js/algolia-result.js
```diff
@@ -30,17 +30,23 @@ export const TEMPLATES = {
...
     {{^id}}
     <a class="${SELECTORS.result} c-search-result__link u-no-underline" href="{{hitUrl}}" data-queryid="{{analyticsData.queryID}}" data-hit-position="{{analyticsData.position}}" data-objectid="{{analyticsData.objectID}}">
     {{/id}}
+      <span>{{{hitIcon}}}</span>
+      <span class="c-search-result__feature-icons">
+      {{#hitFeatureIcons}}
+        {{{.}}}
+      {{/hitFeatureIcons}}
+      <br/>
+    </span>
```
thecristen, line 49: Would it work to put these lines immediately before the <a> on line 41? I'm thinking that taking them outside the link might help with the underlining issue.

Otherwise I think it makes sense to file a new ticket for that formatting and let that part go.

thecristen:
> Screen Shot 2022-01-14 at 11 52 39 AM
> 
> updated text to be on second line but now the icons are underlined on hover. Moved the icons span to be outside the anchor tag which got rid of the underlines but moved everything back to one line. Open to suggestions on a fix to resolve both issues.

Looked at the code and the link has a `.u-no-underline` class on it... it's not even supposed to be underlined in the first place. 😩 So thanks for catching that! Can you file a ticket to remove this underlining?

thecristen (approved):

thecristen:
Bypassing the unrelated failing test to merge in, let's goooo

PR #1158, opened 2022-01-05
add theme-color meta tag to app template

bfauble:
#### Summary of changes
**Asana Ticket:** [Mobile menu: Dark blue background should extend to very top of screen](https://app.asana.com/0/385363666817452/1201529194279336/f)

Change via the theme-color meta tag which explicitly sets the status bar background color rather than the os/browser trying to determine what color it should be.

### DIFF
--- a/apps/site/lib/site_web/templates/layout/app.html.eex
+++ b/apps/site/lib/site_web/templates/layout/app.html.eex
@@ -10,7 +10,7 @@
     <% meta_description = assigns[:meta_description] || "Official website of the MBTA -- schedules, maps, and fare information for Greater Boston's public transportation system, including subway, commuter rail, bus routes, and boat lines." %>
     <meta name="description" content="<%= Phoenix.HTML.raw(meta_description) %>">
     <meta name="author" content="Massachusetts Bay Transportation Authority">
-    <meta name="theme-color" content="#071d2e">
+    <meta name="theme-color" content="#0b2f4c">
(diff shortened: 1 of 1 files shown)


apps/site/lib/site_web/templates/layout/app.html.eex
```diff
@@ -10,6 +10,7 @@
     <% meta_description = assigns[:meta_description] || "Official website of the MBTA -- schedules, maps, and fare information for Greater Boston's public transportation system, including subway, commuter rail, bus routes, and boat lines." %>
     <meta name="description" content="<%= Phoenix.HTML.raw(meta_description) %>">
     <meta name="author" content="Massachusetts Bay Transportation Authority">
+    <meta name="theme-color" content="#071d2e">
```
thecristen, line 13: How'd you get this particular hex value? Sorry, we should have been more explicit/linked the ticket to the mocks -- I'm pretty sure the right color is:

```suggestion
    <meta name="theme-color" content="#0b2f4c">
```

*But it works!

thecristen (changes requested):

apps/site/lib/site_web/templates/layout/app.html.eex
bfauble: I was matching off of the footer background color which is what the ss in the ticket appeared to be using, but no problem can swap over to the other

thecristen (approved):
🎨

PR #1160, opened 2022-01-10
Expand col width on larger devices

i0sea:
#### Summary of changes
**Asana Ticket:** [Mobile menu | Expand col width of fare prices on larger devices](https://app.asana.com/0/555089885850811/1201529194279333/f)

Have column width of fare price box expand based on screen size.


thecristen (approved):
It's beautiful!

PR #1162, opened 2022-01-13
tests: Restore previously failing tests

thecristen:
#### Summary of changes
**Asana Ticket:** [Restore access to forbidden static files](https://app.asana.com/0/385363666817452/1201648677863847/f)

The root problem appears to have been solved, let's restore the tests and see everything pass!


i0sea (commented):
The uploadedimages tests(line 15 and 22) seem to be returning 404 instead of the expected 200. Should some of these remained skip for now?

thecristen:
This is obsolete as of #1293

PR #1165, opened 2022-01-19
fix for Investigation: Blue bars appear above and below open sub-menu

thecristen:
#### Summary of changes
**Asana Ticket:** [Investigation: Blue bars appear above and below open sub-menu](https://app.asana.com/0/385363666817452/1201529194279341/f)

Awkwardly, this was mainly a Safari problem. 
I tried...

* unsetting the blue outline through CSS, but no `!important` or higher specificity made the Safari user agent stop rendering its blue outline.
* listening to the `show.bs.collapse` or `shown.bs.collapse` events from Bootstrap's collapse plugin (which the accordions use when opening/closing) to intercept the adding of the blue outline. But I couldn't get this working either.

I settled on listening to the `focus` event on the accordion target, and removing the outline there.

### DIFF
--- a/apps/site/assets/ts/app/mobile-menu.ts
+++ b/apps/site/assets/ts/app/mobile-menu.ts
@@ -81,6 +81,15 @@ const setupMobileMenu = (): void => {
 
   // Changing viewport size closes
   window.addEventListener("resize", closeMenu);
+
+  // removes focus outline in Safari from open accordions
+  document
+    .querySelectorAll(".m-menu__content .js-focus-on-expand")
+    .forEach(acc => {
+      acc.addEventListener("focus", function undoOutline(this: HTMLElement) {
+        this.style.outline = "none";
+      });
+    });
 };
 
 export default (): void => {


apps/site/assets/ts/app/mobile-menu.ts
```diff
@@ -81,6 +81,15 @@ const setupMobileMenu = (): void => {
 
   // Changing viewport size closes
   window.addEventListener("resize", closeMenu);
+
+  // removes focus outline in Safari from open accordions
+  document
+    .querySelectorAll(".m-menu__content .js-focus-on-expand")
+    .forEach(acc => {
+      acc.addEventListener("focus", function undoOutline(this: HTMLElement) {
+        this.style.outline = "none";
```
bfauble, line 90: I think you can add this to the focus call [here](https://github.com/mbta/dotcom/blob/master/apps/site/assets/js/collapse.js#L20) and have the same effect. I'm not sure if it's wanted to have the focus outline for any other case though

thecristen: Yeah, the particular component this is operating on (the accordion) is reused in many places across the website! But in this particular case the accordion content has focusable content inside, which has its own visible focus outline, so the default focus outline is redundant.

PR #1166, opened 2022-01-19
chore: remove unused vendor collapse plugin

thecristen:
this was used on elements with `mbta-` prefaced class names. we no longer have any of those.

(no ticket)


thecristen:
Deploying to dev-green just to really make sure nothing's broken...

i0sea (approved):

PR #1169, opened 2022-01-20
Mobile Menu Styling Fixes

i0sea:
#### Summary of changes
**Asana Ticket:** [Mobile menu | Fix multiple styling issues](https://app.asana.com/0/385363666817452/1201613685853923/f)

Make Search button a square
Stop T logo on header from shifting when opening menu or search
Make submenu heading normal blue at all times

Couldn't recreate [white bar issue](https://app.asana.com/0/1154777277498386/1201529194279324/f)

### DIFF
--- a/apps/site/assets/css/_header.scss
+++ b/apps/site/assets/css/_header.scss
@@ -478,7 +478,7 @@ nav.m-menu {
       @include heading-4();
       color: $white;
       &:hover, &:active {
-        background-color: $brand-primary-darkest;
+        background-color: $brand-primary;
       }
     }
     .c-accordion-ui__content {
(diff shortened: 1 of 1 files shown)


apps/site/assets/css/_header.scss
```diff
@@ -476,7 +478,7 @@ nav.m-menu {
       @include heading-4();
       color: $white;
       &:hover, &:active {
-        background-color: $brand-primary-darkest;
+        background-color: $brand-primary;
```
thecristen, line 481: Since we're setting it to the same color it was when not hovered, let's use `inherit` for the value.

thecristen (approved):
Just a small comment, otherwise LGTM

PR #1177, opened 2022-02-04
Set up AWS Location access in Dotcom

thecristen:
#### Summary of changes
**Asana Ticket:** [Set up AWS Location access in Dotcom](https://app.asana.com/0/385363666817452/1201335974613194/f)

Using an updated version of `ExAws` to interact with Amazon Location Service. I expect this would Just Work from an AWS authentication perspective.

The test file demonstrates a basic way we can interact with the service and the returned data.

Note: I had to update the Dockerfile to compile the application digest from the umbrella root instead of from `apps/site/`, such that it would install the required `ex_aws` as defined in the root `mix.exs` file without raising an error.


paulswartz:
I can verify that these tests pass for me once I provide valid AWS credentials.

i0sea (approved):

PR #1178, opened 2022-02-04
Abstract API calls into separate services

i0sea:
#### Summary of changes
**Asana Ticket:** [Abstract API calls into separate services](https://app.asana.com/0/385363666817452/1201652088279936/f)
Create a new folder for AWS location service specific implementation. Rename files and env variables to specify source of geocode, map data, etc.


thecristen:
I added a tiny commit so that it'll compile, and I'm going to deploy it to dev-green to confirm it works as expected in the prod Mix environment. Nice work!! I will add a proper review shortly.

apps/site/lib/site_web/templates/layout/app.html.eex
```diff
@@ -40,11 +40,11 @@
       <script defer src="<%= static_url(@conn, "/js/app.js") %>" data-turbolinks-track="reload"></script>
     <% end %>
 
-    <%= if assigns[:requires_google_maps?] do %>
+    <%= if assigns[:requires_location_service?] do %>
```
thecristen, line 43: TIL about this variable, interesting!

thecristen (approved):
Looks good to go! Thanks for all the cleanup/renaming/moving 🧹

PR #1179, opened 2022-02-08
Fix scrolling issues around mobile nav

thecristen:
#### Summary of changes
**Asana Ticket:** [🛑 Fix scrolling issues](https://app.asana.com/0/555089885850811/1201764117487437/f)

It turns out preventing scrolling a page while having an interactive sticky element is really tricky in iOS Safari only... who knew? There are a variety of CSS and JS-based approaches, some of which I tried unsuccessfully. See [this blog post](https://medium.com/jsdownunder/locking-body-scroll-for-all-devices-22def9615177) for some background on that. I don't usually add new npm libraries but [body-scroll-lock](https://www.npmjs.com/package/body-scroll-lock) did the job, fixing both the mobile-user-can-still-scroll-page-with-search-open problem _and_ the closing-search-scrolls-slightly-down problem.


zovt (approved):
Interesting issue... did some Googling and didn't find a great definitive answer to solve this, so I think it's ok to pull in the library here. Not sure if there is any more formal process around adding deps but from my perspective 🚢

PR #1180, opened 2022-02-08
Close search completely when closing keyboard

thecristen:
#### Summary of changes
**Asana Ticket:** [🛑 Close search completely when closing keyboard](https://app.asana.com/0/555089885850811/1201764117487453/f)

Added an extra event listener to close open menus (includes search bar) when the search input has a `focusout` event.


i0sea (approved):
Testing on BrowserStack it looks like the issue persists(iphone 13 w/ safari). This approach seems reasonable though.

thecristen:
After talking about this in standup, it turns out this actually might have been fixed in #1179

PR #1187, opened 2022-02-14
Show veil when focusing header search on <=tablet widths

zovt:
#### Summary of changes
**Asana Ticket:** [Add "veil"/modal cover when search is open](https://app.asana.com/0/555089885850811/1201764117511338/f)

When the viewport is "tablet-sized" (I used the `lg` breakpoint for this), focusing the nav search bar will show the modal veil and prevent scrolling.


thecristen (approved):
I forgot we even had those helpers, nice!
Thanks for the accompanying test -- it looks good to me.

PR #1190, opened 2022-02-15
Button focus state

zovt:
#### Summary of changes
**Asana Ticket:** [Update button toggle color to default after it's been pressed](https://app.asana.com/0/555089885850811/1201764117487423/f)

This PR:
- Adds `focus-visible` to our dependencies (so that we can use `:focus-visible` in Safari. WebKit has recently merged support for the selector but it's still not widely available).
- Configures webpack to copy `focus-visible.min.js` to our output directory
- Conditionally loads the dependency based on whether or not the browser supports `:focus-visible`
- Updates the styles for the header menu toggle buttons to use `:focus-visible` and use `$brand-secondary` as the border color (matching the accordion buttons)

Let me know if the webpack configuration and load strategy makes sense :)

I've tested this locally on Firefox, Safari (desktop), and Chrome and they all seem to be working (though there's an additional outline added in Chrome, I didn't locate where it was coming from but I'm fairly certain it's not caused by this PR).

Demo:
https://user-images.githubusercontent.com/5061820/154130813-b7955ac4-efce-4a66-ace8-680f6db86166.mp4

### DIFF
--- a/apps/site/assets/css/_header.scss
+++ b/apps/site/assets/css/_header.scss
@@ -560,27 +560,39 @@ nav.m-menu--mobile {
   }
 }
 
+@mixin focus-visible {
+  &.focus-visible {
+    @content;
+  }
+
+  &:focus-visible {
+    @content;
+  }
+}
+
 .m-menu__toggle, .m-menu--desktop__toggle, .header-search__toggle {
   background-color: unset;
   border: 2px solid $brand-primary-light-contrast;
   border-radius: 4px;
   margin: auto 0;
 
-  &:hover,
-  &:focus-visible,
-  &.focus-visible {
+  &:hover {
     background-color: $brand-primary-darkest;
   }
 
-  &:focus-visible,
-  &.focus-visible {
+  @include focus-visible {
+    background-color: $brand-primary-darkest;
     outline: 2px solid $brand-secondary;
   }
 
-  &[aria-expanded="true"]:hover,
-  &[aria-expanded="true"]:focus-visible,
-  &[aria-expanded="true"].focus-visible {
-    background-color: $brand-primary;
+  &[aria-expanded="true"] {
+    &:hover {
+      background-color: $brand-primary;
+    }
+
+    @include focus-visible {
+      background-color: $brand-primary;
+    }
   }
 }
 
--- a/apps/site/assets/css/_header.scss
+++ b/apps/site/assets/css/_header.scss
@@ -562,6 +562,7 @@ nav.m-menu--mobile {
 
 @mixin focus-visible {
   &.focus-visible {
+    empty-cells: inherit;
     @content;
   }
 
--- a/apps/site/assets/css/_header.scss
+++ b/apps/site/assets/css/_header.scss
@@ -577,8 +577,10 @@ nav.m-menu--mobile {
   border-radius: 4px;
   margin: auto 0;
 
-  &:hover {
-    background-color: $brand-primary-darkest;
+  @media (hover: hover) {
+    &:hover {
+      background-color: $brand-primary-darkest;
+    }
   }
 
   @include focus-visible {
@@ -587,8 +589,10 @@ nav.m-menu--mobile {
   }
 
   &[aria-expanded="true"] {
-    &:hover {
-      background-color: $brand-primary;
+    @media (hover: hover) {
+      &:hover {
+        background-color: $brand-primary;
+      }
     }
 
     @include focus-visible {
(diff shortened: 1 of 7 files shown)


apps/site/assets/css/_header.scss
```diff
@@ -567,16 +567,16 @@ nav.m-menu--mobile {
   margin: auto 0;
 
   &:hover,
-  &:focus {
+  &:focus-visible {
     background-color: $brand-primary-darkest;
   }
 
-  &:focus {
-    border-color: $brand-primary-light;
```
thecristen, line 575: Can we add these back in? This removed the focus styling applied when tabbing through the desktop navigation buttons (it used to look like below). Maybe adjusting these to `:focus:not(:focus-visible)` might get what we want for the border, or maybe we might need to style the desktop menu buttons and mobile menu buttons differently 🤔 

image

apps/site/assets/webpack.config.base.js
```diff
@@ -128,7 +128,10 @@ module.exports = {
   },
 
   plugins: [
-    new CopyWebpackPlugin([{ from: "static/**/*", to: "../../" }], {}),
+    new CopyWebpackPlugin([
+      { from: "static/**/*", to: "../../" },
+      { from: "node_modules/focus-visible/dist/focus-visible.min.js", to: "../js" },
```
thecristen, line 133: I see why you want to do this rather than importing it into our existing `app.js`, since we're only planning to load the file if needed but we can't evaluate whether it's needed until we load the page in the browser. Pretty neat approach.

thecristen (changes requested):
I'll look into the failing Cypress test that's plaguing every PR lately.

Looks good! The only change I request is to restore the prior focus style.

apps/site/assets/css/_header.scss
zovt: Just looked into this, it appears that the focus styling is applying when tabbing through these with a keyboard on my end. Unless you're referring to the fact that the outline is now `$brand-secondary` which I thought was desired according to the mocks I saw. 

I could also be totally misunderstanding what you're pointing out so let me know :)

thecristen: Hm interesting!
Is this on dev-blue? On that the button background is transparent/idle blue when I tab to it -- this is only on Safari 15.3, it looks fine in FF... 
image

zovt: This is on dev-blue yes (you should see a network request for `focus-visible.js`). 

I see what's going on, one sec

zovt: Should be good now (was missing the polyfill selector in the stylesheet)

thecristen: Okay, the script tag was being built and `focus-visible.js` was being requested but Safari wasn't loading it for me. I ran it locally and finally got it to work in my Safari by `require`ing it conditionally -- let me know what you think!

apps/site/assets/css/_header.scss
```diff
@@ -560,24 +560,44 @@ nav.m-menu--mobile {
   }
 }
 
+@mixin focus-visible {
+  &.focus-visible {
+    empty-cells: inherit;
```
zovt, line 565: This is used to "break" the CSS optimizer and force it to keep these rules separate so WebKit doesn't end up choking on comma'd rules with `:focus-visible` in them. I don't _think_ this should have any noticeable affect other than that, but let me know if there's a better property to use here.

apps/site/assets/css/_header.scss
```diff
@@ -560,24 +560,44 @@ nav.m-menu--mobile {
...
 .m-menu__toggle, .m-menu--desktop__toggle, .header-search__toggle {
   background-color: unset;
   border: 2px solid $brand-primary-light-contrast;
   border-radius: 4px;
   margin: auto 0;
 
-  &:hover,
-  &:focus {
-    background-color: $brand-primary-darkest;
+  @media (hover: hover) {
```
zovt, line 580: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover

thecristen:
Going to deploy this on dev-green to play with it some more.

apps/site/assets/css/_header.scss
thecristen: Hm, this pick of property seems good. Can you please add a comment in the CSS about this too?

apps/site/assets/css/_header.scss
```diff
@@ -560,24 +560,44 @@ nav.m-menu--mobile {
   }
 }
 
+@mixin focus-visible {
```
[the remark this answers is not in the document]
thecristen: Yeah, let's add this to our main mixins file over at `_mixins.scss`

thecristen (changes requested):
Just some additional comments requested -- it works great

thecristen (approved):
💯

PR #1192, opened 2022-02-16
Call `closeAllMenus` when clicking T logo

zovt:
#### Summary of changes
**Asana Ticket:** [Remove flashing when clicking T logo](https://app.asana.com/0/555089885850811/1201764117487461/f)

This should fix a bunch of issues on the homepage when the mobile menu
is open, including:
- menu popping in when the T logo is clicked twice
- theme flickering on mobile Chrome and Safari
- modal veil remaining open

### DIFF
--- a/apps/site/assets/ts/app/global-navigation.ts
+++ b/apps/site/assets/ts/app/global-navigation.ts
@@ -181,6 +181,9 @@ export default function setupGlobalNavigation(): void {
         menu_links[i].addEventListener("click", closeAllMenus);
       }
 
+      // T logo click closes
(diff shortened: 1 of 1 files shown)


i0sea (approved):
When opening the menu->clicking the T logo->clicking search->clicking the T logo, the nav menu still flashes. Very specific edge case so maybe let's just log it in the backlog for now. Thoughts @thecristen ?

thecristen:
I wonder if it's different on different iOS versions? This PR actually does fix the flashing for me, at least on iOS / Safari 15.3 :)

PR #1193, opened 2022-02-17
Use AWS with HERE data for reverse geocoding

i0sea:
#### Summary of changes
**Asana Ticket:** [Use AWS with HERE data for reverse geocoding](https://app.asana.com/0/555089885850811/1201614287788933/f)

Use AWS for reverse geocoding. Updated references to use HERE data instead of google maps. 
*Note: Comparing against cbj/aws-geocoding so that only reverse geo changes are shown in diff. When time to merge will be against master


apps/site/lib/site_web/controllers/fare_controller.ex
```diff
@@ -11,7 +11,7 @@ defmodule SiteWeb.FareController do
 
   @options %{
     geocode_fn: &LocationService.Geocode.geocode/1,
-    reverse_geocode_fn: &Geocode.reverse_geocode/2,
+    reverse_geocode_fn: &LocationService.Geocode.reverse_geocode/1,
```
thecristen: I ended up correcting these references to point to the module you wrote, `LocationService.ReverseGeocode.reverse_geocode/2`.

thecristen: However, totally not opposed to consolidating that function with the `LocationService.Geocode.geocode/1` function into one module that has both functions...

thecristen (approved):
Thanks for the tests! Let's go 🚀

PR #1194, opened 2022-02-18
feat: activate feature flag via URL

thecristen:
No ticket.

Adds the ability to activate a feature flag via URL. This was added with the intent of sharing flagged features with stakeholders in an easier manner. Maybe it'll help! 🙂

Deployed on dev-blue for the time being.

Some open questions...

* Technically this sets a cookie that expires after a month (which is exactly what the "enable" button does). So you can activate it by navigating to the appropriate URL with query parameter. But if you subsequently go to the URL without the query parameter, the feature is still activated! So how problematic is that?
* Should this support activating more than one feature at once?
* Maybe the function belongs in `SiteWeb.Plugs.Cookies`

### DIFF
--- a/apps/site/lib/site_web/router.ex
+++ b/apps/site/lib/site_web/router.ex
@@ -29,6 +29,7 @@ defmodule SiteWeb.Router do
     plug(SiteWeb.Plugs.ClearCookies)
     plug(SiteWeb.Plugs.Cookies)
     plug(:optional_disable_indexing)
+    plug(:activate_flag)
   end
 
   pipeline :api do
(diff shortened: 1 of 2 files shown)


i0sea (approved):
I think it's not too problematic for the flag to be enabled since being able to turn it on is only accessible on a need to know basis.
Being able to activate multiple features at once would be nice. If there isn't really an instance where the features are dependent on each other it might not be necessary. Could happen in the future though.
Moving this to `SiteWeb.Plugs.Cookies` would have what effect?

thecristen:
> Moving this to SiteWeb.Plugs.Cookies would have what effect?
@i0sea Just a bit more organized I guess, since this function sets a cookie, which is what everything in `SiteWeb.Plugs.Cookies` does (which just sets different cookies)

PR #1197, opened 2022-02-18
chore: Update API dev URL in crawler workflow

losvedir:
#### Summary of changes
**Asana Ticket:** n/a

The `api.mbtace.com` host points to the old dev EB instance. I'm not sure yet if we're going to maintain it, so this updates the crawler to use the new `api-dev.mbtace.com` host and ECS instances.

---


thecristen (approved):
yay! thank you

PR #1199, opened 2022-02-23
feat: Add RR crossing number to updated mobile and desktop menus

bfauble:
#### Summary of changes
**Asana Ticket:** [Add RR crossing number to updated mobile and desktop menus](https://app.asana.com/0/385363666817452/1201824241791486/f)

Add the railroad crossing gate issue number to the new desktop and mobile menus.

image
image
image

### DIFF
--- a/apps/site/lib/site_web/views/helpers.ex
+++ b/apps/site/lib/site_web/views/helpers.ex
@@ -264,6 +264,7 @@ defmodule SiteWeb.ViewHelpers do
 
       machine_formatted ->
         content_tag(:a, pretty_formatted,
+          class: "no-wrap",
           href: "tel:#{machine_formatted}",
           aria: [label: Site.PhoneNumber.aria_format(number)]
         )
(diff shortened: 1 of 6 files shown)


apps/site/lib/site_web/views/helpers.ex
```diff
@@ -264,6 +264,7 @@ defmodule SiteWeb.ViewHelpers do
 
       machine_formatted ->
         content_tag(:a, pretty_formatted,
+          class: "no-wrap",
```
thecristen, line 267: 👍🏼 Good addition.

thecristen (changes requested):
💯 on mobile/tablet view, thanks for the attention to white space/text wrapping.

The mock for the desktop view structures this new section a bit differently, placing it in the bottom of the center column instead of below the two sections in the right column. Can you update this to reflect that too?

bfauble:
https://www.w3.org/TR/css-grid-1/#repeat-syntax
https://www.w3.org/TR/css-grid-1/#valdef-repeat-auto-fit

couple relevant pieces of info regarding generating auto-fit columns

thecristen (approved):
Great!

PR #1202, opened 2022-02-28
Release redesigned header navigation

thecristen:
No ticket. No more flag to activate the new header.
There's a lot more cleanup I want to do, particularly around the CSS, but I wanted to keep this PR a bit small.


i0sea (approved):
Looks like the railroad emergency contact info is missing from the contact section. Besides that looks good to me 😎

thecristen:
> Looks like the railroad emergency contact info is missing from the contact section. Besides that looks good to me 😎

Ha, you totally should have "requested changes" @i0sea! Anyway thanks so much for letting me know.

PR #1206, opened 2022-03-01
Scroll menu accordion to top

zovt:
#### Summary of changes
**Asana Ticket:** [Update scroll behavior when opening accordion](https://app.asana.com/0/555089885850811/1201764117487465/f)

This PR implements a scroll-to-top behavior when the accordions in the mobile nav menu are expanded. This functionality is not enabled if the user has indicated that they prefer reduced motion.

WebKit has only very, very recently gotten support for smooth scrolling behavior in the various `scroll*` functions, and as far as I can tell there are no WebKit-based browser releases that support it yet. Thus, I've added a polyfill for the smooth scrolling functionality, and taken an approach similar to #1190 for feature detection and lazy-loading.

### DIFF
--- a/apps/site/assets/package-lock.json
+++ b/apps/site/assets/package-lock.json
@@ -38,6 +38,7 @@
         "react-redux": "^7.2.0",
         "redux": "^4.0.5",
         "sifter": "^0.6.0",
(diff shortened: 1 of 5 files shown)


i0sea (approved):
🧈

PR #1207, opened 2022-03-02
Prevent overscroll in mobile menu 

zovt:
#### Summary of changes
**Asana Ticket:** [Update sticky behavior when pulling down menu](https://app.asana.com/0/555089885850811/1201764117487427/f)

This prevents overscroll effects in the mobile menu. WebKit has only recently merged support for this property, so the behavior is still observable on macOS and iOS Safari, but once broader support for the property rolls out this should solve the issue.


thecristen:
Wow, [in Safari](https://bugs.webkit.org/show_bug.cgi?id=236060) that feature is even fixed and merged... to be released any minute now... sheeeesh

thecristen (approved):
👍🏼

PR #1209, opened 2022-03-02
fix trip planner returning 500 when suggested itinerary route is not queryable via api

bfauble:
#### Summary of changes
**Asana Ticket:** [Fix: Trip Planner returns error on Union Square - NEU plan](https://app.asana.com/0/555089885850811/1201500059383446/f)

Change allow the trip planner to suggest the returned OTP itinerary for an MBTA Bus even if a specific leg of a trip is not present via the api to fetch additional details for things such as shuttle buses. 

Details not fully present so from the detailed view it's not indicated fully that this is a shuttle bus, but the icon indicates that it is a shuttle bus. 

image
image


zovt (approved):
Seems reasonable 👍

PR #1210, opened 2022-03-03
Fix shifting logo on large tablet breakpoint

zovt:
#### Summary of changes
**Asana Ticket:** [Update T logo to stay in the same place](https://app.asana.com/0/555089885850811/1201764117487441/f)

Specifically on the large tablet breakpoint, the text changing from `Menu` to `Close` caused the menu button to expand ever so slightly, causing visual shifting of the MBTA logo. To fix this, I've set a static width and centered the text, and there appears to be no more visual shifting when expanding the menu.

### DIFF
--- a/apps/site/assets/css/_header.scss
+++ b/apps/site/assets/css/_header.scss
@@ -120,7 +120,7 @@
   }
 }
 
-nav.m-menu, .m-menu__toggle {
+nav.m-menu, .header-navbar .m-menu__toggle {
   @include media-breakpoint-up(lg) {
     display: none;
   }
@@ -356,6 +356,14 @@ nav.m-menu--mobile {
   .m-menu__toggle {
     @extend %hide-button;
   }
+}
+
+.m-menu__toggle {
+  width: 5em;
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
 
 .m-menu__toggle, .m-menu--desktop__toggle, .header-search__toggle {
--- a/apps/site/assets/css/_header.scss
+++ b/apps/site/assets/css/_header.scss
@@ -120,7 +120,7 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


apps/site/assets/css/_header.scss
```diff
@@ -358,6 +358,18 @@ nav.m-menu--mobile {
   }
 }
 
+.m-menu__toggle {
+  width: 5em;
```
bfauble, line 362: Looking into this issue a bit more it seems like the issue is tied to whether or not the scrollbar is showing or not and the container getting wider to fill the extra space when the search overlay is show and subsequently hides the scrollbar.

Because of this I think this solution may only resolve the issue for some browser types due to the differences in scroll bar widths etc across different platforms. 

A possible solution to resolve the issue across all platforms is to add a `width: 100vw` to the `<body>` element to always have the body expanded to the entire viewport and not subtract the scrollbar width. But this could potentially cause undesirable results due to seemingly mismatched margins/padding on either side depending on the browser and may be a question for design.

zovt: Ah I see what you're saying now, I was testing using responsive mode and my scrollbars were the kind that hovers over the content instead of being next to it. There's definitely a much more noticeable shift when the scrollbar pops in and out with the default scrollbars (though this is still solving the very small amount of shifting I see in responsive design mode).

I'll noodle on this a bit, thanks for the call out 👍

zovt: I've logged a separate ticket here for the desktop fixes: https://app.asana.com/0/555089885850811/1201920171846633/f

I think we can safely merge this in to fix the minor issue on mobile while thinking about how to address the bigger issue on desktop

bfauble (approved):

PR #1214, opened 2022-03-03
docs: Updated instructions for Google API key

thecristen:
This was brought to my attention after @i0sea tried following the old instructions. There's a better way!


zovt (approved):

zovt:
Could also be worth clarifying that this is for internal use only, other users should go through the typical Google API key flow

PR #1217, opened 2022-03-04
WIP: front-end logic to call geocoding & related functions

thecristen:
While investigating/refining tickets I created a few new endpoints to experiment with, just wanted to share them here for whoever will work on [Add front-end logic to use AWS for geocoding + place lookup](https://app.asana.com/0/0/1201846410385721/f) or [Add front-end logic to use AWS for reverse geocoding](https://app.asana.com/0/0/1201846410385722/f)

They will return both AWS and Google results in a JSON response, which I found useful for a quick view/comparison to debug what's returned for a given query. 🤷🏼‍♀️ 

Just to be clear, it's super hacky.


thecristen:
Superseded by #1222

PR #1218, opened 2022-03-07
Fix: Transit Near Me not handling Union Square stop

i0sea:
#### Summary of changes
**Asana Ticket:** [Fix: Transit Near Me not handling Union Square stop](https://app.asana.com/0/555089885850811/1201500059383445/f)

Make Union Square stop visible on transit near me map pages nearby routes/stops section. Removing route check populates predicted schedules which determines whether a stop is rendered.


thecristen (approved):
What a find! I've been wracking my brain trying to piece together why this condition was here in the first place, and frankly I'm not sure! Reviewing the rest of the site, nothing stood out as behaving much different, so if this change causes any unforeseen problems..well, we can cross that bridge when we get to it 😆

thecristen:
@i0sea there's a new failing test, might that need adjustment?

i0sea:
> @i0sea there's a new failing test, might that need adjustment?

Decided to comment out failing tests since they don't quite tests the new changes. Thoughts @thecristen ?

thecristen (changes requested):
I'd try a few things before resorting to commenting out tests
* sometimes rerunning the test might make it pass (the underlying API response might change, or could be a weird race condition in the tests (many of the tests run async), etc)
* Relating to the underlying API responses changing, sometimes stop IDs or route pattern IDs might change. So a test might be failing because the particular ID it's looking for is no longer in use (in which case we can replace it with a relevant working one)
* sometimes a tests is written assuming a specific implementation, and once you change the implementation you might have to update the test to match. This one seems the most likely to be happening here but I haven't looked into it enough to be sure.

(Maybe you tried some of these already, but just wanted to make note)

Also, it looks like you were commenting out some unrelated tests? From what it looks like, the failing test is in `Site.RealitimeScheduleTest` - I can reproduce it locally running `mix test apps/site/test/site/realtime_schedule_test.exs:160`

```elixir

  1) test stop_data/3 returns stop (Site.RealtimeScheduleTest)
     apps/site/test/site/realtime_schedule_test.exs:160
     ** (EXIT from #PID<0.1148.0>) an exception was raised:
         ** (KeyError) key :stop_id not found in: %Schedules.Schedule{early_departure?: true, flag?: false, last_stop?: false, pickup_type: 0, route: %Routes.Route{color: "ED8B00", custom_route?: false, description: :rapid_transit, direction_destinations: %{0 => "Forest Hills", 1 => "Oak Grove"}, direction_names: %{0 => "Southbound", 1 => "Northbound"}, id: "Orange", long_name: "Orange Line", name: "Orange Line", sort_order: 99999, type: 1}, stop: %Stops.Stop{accessibility: [], address: nil, bike_storage: [], child_ids: [], closed_stop_info: nil, description: nil, fare_facilities: [], has_charlie_card_vendor?: false, has_fare_machine?: false, id: "place-ogmnl", is_child?: false, latitude: nil, longitude: nil, municipality: nil, name: nil, note: nil, parent_id: nil, parking_lots: [], platform_code: nil, platform_name: nil, station?: false, type: nil, zone: nil}, stop_sequence: 1, time: ~U[2030-02-19 12:00:00Z], trip: %Schedules.Trip{bikes_allowed?: false, direction_id: 1, headsign: "Oak Grove", id: "40709317", name: "", occupancy: nil, route_pattern_id: "Orange-3-1", shape_id: "903_0017"}}
             (site 0.0.1) lib/site/realtime_schedule.ex:216: anonymous fn/2 in Site.RealtimeSchedule.do_get_schedules/5
             (elixir 1.10.3) lib/enum.ex:3291: Enum.filter_list/2
             (site 0.0.1) lib/site/realtime_schedule.ex:216: Site.RealtimeSchedule.do_get_schedules/5
             (elixir 1.10.3) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
             (elixir 1.10.3) lib/task/supervised.ex:35: Task.Supervised.reply/5
             (stdlib 3.12.1) proc_lib.erl:249: :proc_lib.init_p_do_apply/3

     The following output was logged:
     2022-03-08 13:41:09.790 [error] Task #PID<0.1153.0> started from #PID<0.1157.0> terminating
     ** (KeyError) key :stop_id not found in: %Schedules.Schedule{early_departure?: true, flag?: false, last_stop?: false, pickup_type: 0, route: %Routes.Route{color: "ED8B00", custom_route?: false, description: :rapid_transit, direction_destinations: %{0 => "Forest Hills", 1 => "Oak Grove"}, direction_names: %{0 => "Southbound", 1 => "Northbound"}, id: "Orange", long_name: "Orange Line", name: "Orange Line", sort_order: 99999, type: 1}, stop: %Stops.Stop{accessibility: [], address: nil, bike_storage: [], child_ids: [], closed_stop_info: nil, description: nil, fare_facilities: [], has_charlie_card_vendor?: false, has_fare_machine?: false, id: "place-ogmnl", is_child?: false, latitude: nil, longitude: nil, municipality: nil, name: nil, note: nil, parent_id: nil, parking_lots: [], platform_code: nil, platform_name: nil, station?: false, type: nil, zone: nil}, stop_sequence: 1, time: ~U[2030-02-19 12:00:00Z], trip: %Schedules.Trip{bikes_allowed?: false, direction_id: 1, headsign: "Oak Grove", id: "40709317", name: "", occupancy: nil, route_pattern_id: "Orange-3-1", shape_id: "903_0017"}}
         (site 0.0.1) lib/site/realtime_schedule.ex:216: anonymous fn/2 in Site.RealtimeSchedule.do_get_schedules/5
         (elixir 1.10.3) lib/enum.ex:3291: Enum.filter_list/2
         (site 0.0.1) lib/site/realtime_schedule.ex:216: Site.RealtimeSchedule.do_get_schedules/5
         (elixir 1.10.3) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
         (elixir 1.10.3) lib/task/supervised.ex:35: Task.Supervised.reply/5
         (stdlib 3.12.1) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
     Function: #Function<28.104581810/0 in Site.RealtimeSchedule.get_schedules/3>
         Args: []
     
```

It looks like this failing test is running into an exception at `lib/site/realtime_schedule.ex:216` - I'd look into that next and see if there's a way to handle that more gracefully or fix it.

thecristen (approved):
🎉 🟢

thecristen:
> What a find! I've been wracking my brain trying to piece together why this condition was here in the first place, and frankly I'm not sure! Reviewing the rest of the site, nothing stood out as behaving much different, so if this change causes any unforeseen problems..well, we can cross that bridge when we get to it 😆

Okay I ultimately _did_ remember why we likely rejected the 0 and 1 route types. 0 and 1 are light rail and subway, and for those we want to show predictions rather than schedules. This makes it unnecessary to get schedules for those routes anyway.

We'll see how this goes on prod!

PR #1219, opened 2022-03-08
Plug to check if within range of glx opening date

bfauble:
#### Summary of changes
**Asana Ticket:** [Store station opening dates and check age](https://app.asana.com/0/555089885850811/1201708898492590/f)

Added plug to set an assign within conn if the current date after the set opening date for the glx expansion and not more than 3 months after that date.

### DIFF
--- a/apps/site/lib/site_web/plugs/glx_now_open.ex
+++ b/apps/site/lib/site_web/plugs/glx_now_open.ex
@@ -10,7 +10,7 @@ defmodule SiteWeb.Plugs.GlxNowOpen do
       Science Park/West End (place-spmnl)
   """
 
-  @opening_date ~N[2022-03-21T03:00:00]
+  @opening_date ~N[2022-03-21T04:55:00]
 
   @behaviour Plug
   import Plug.Conn, only: [assign: 3]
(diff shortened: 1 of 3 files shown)


apps/site/lib/site_web/plugs/glx_now_open.ex
```diff
@@ -0,0 +1,39 @@
...
+  def init([]), do: [now_fn: &Util.now/0, to_local_fn: &Util.to_local_time/1]
+
+  @impl true
+  def call(conn, now_fn: now_fn, to_local_fn: to_local_fn) do
+    conn
+    |> assign(:glx_now_open?, check_current_service_date(now_fn.(), to_local_fn.(@opening_date)))
+  end
+
+  defp check_current_service_date(current_date, opening_date) do
+    after_open_date?(current_date, opening_date) && before_end_date?(current_date, opening_date)
```
thecristen, line 28: Ha, technically `after_open_date?` can also be `Util.time_is_greater_or_equal?(current_date, opening_date)`

thecristen (changes requested):
I found out this morning that this is confirmed to open at the start of service (e.g. very early morning). MBTA service dates are understood (in Dotcom's code, at least -- see `Util.service_date/1`) to run from 3am to 2:59am. Can you adjust the opening date time to 3am?

bfauble:
> I found out this morning that this is confirmed to open at the start of service (e.g. very early morning). MBTA service dates are understood (in Dotcom's code, at least -- see `Util.service_date/1`) to run from 3am to 2:59am. Can you adjust the opening date time to 3am?

updated to use 3am as the opening time, also tightened up tests to check those specific boundaries.

thecristen (approved):
⏲️

PR #1220, opened 2022-03-09
fix(PredictedSchedule): add function clause

thecristen:
#### Summary of changes
**Asana Ticket:** [Transit Near Me | Elixir.FunctionClauseError: no function clause matching in PredictedSchedule.group_transform/1](https://app.asana.com/0/385363666817452/1200951629013373/f)

(Asana ticket contains linked Sentry issue).

This particular function was erroring whenever it faced a %PredictedSchedule{} that contained a schedule where the stop was nil. This change adds the logic to handle Schedules that don't have a Stop defined. 

This seems to happen with shuttles sometimes.

### DIFF
--- a/apps/site/lib/predicted_schedule.ex
+++ b/apps/site/lib/predicted_schedule.ex
@@ -6,7 +6,8 @@ defmodule PredictedSchedule do
   * prediction: The prediction for this trip (optional)
   """
   alias Predictions.Prediction
-  alias Schedules.{Schedule, ScheduleCondensed}
+  alias Schedules.{Schedule, ScheduleCondensed, Trip}
+  alias Stops.Stop
 
   @derive Jason.Encoder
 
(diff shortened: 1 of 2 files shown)


apps/site/lib/predicted_schedule.ex
```diff
@@ -133,9 +134,11 @@ defmodule PredictedSchedule do
     {{trip_id, stop_id, stop_sequence}, ps}
   end
 
-  defp group_transform(%{trip: trip, stop: stop} = ps)
-       when not is_nil(trip) and not is_nil(stop) do
-    {{ps.trip.id, ps.stop.id, ps.stop_sequence}, ps}
+  defp group_transform(%{trip: %Trip{id: trip_id}, stop: stop} = ps) do
```
bfauble, line 137: is this a possibility for both `Schedule` and `Prediction`? or can this be narrowed down to just one type?

thecristen: I've only seen witnessed it on Schedule but theoretically it's possible on both. 🤔  Noting how we treat Predictions in the clause above it makes me wonder what should be done if a Prediction had a valid trip and nil stop. Going to look at this a bit

thecristen: Well, I was about to push an update that was more specific about handling `Schedule` vs `Prediction` but then I realized the overall logic was unchanged! 😝

In both cases, for this particular function, we want to return the same result.

bfauble (approved):

PR #1221, opened 2022-03-10
Fix new navigation bugs

thecristen:
#### Summary of changes
**Asana Ticket:** [🛑 Mobile menu: Fix search alignment](https://app.asana.com/0/385363666817452/1201937782917570/f)

After a maddening amount of debugging, found that limiting the page width so that it doesn't exceed 100% fixes it.
Also noticed the logo was slightly off-center, so modified the width of the `Menu` button to match the width+margin-left of the search button.
Also unset the padding on the search button - this seems to make it truly square (it was a few pixels too wide on iOS 15).

### DIFF
--- a/apps/site/assets/css/_header.scss
+++ b/apps/site/assets/css/_header.scss
@@ -359,7 +359,7 @@ nav.m-menu--mobile {
 }
 
 .m-menu__toggle {
-  width: 5em;
+  width: 4rem;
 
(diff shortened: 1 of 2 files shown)


zovt (approved):
This makes sense to me 👍 it could be worth doing some more significant structural changes to the header that makes the layout easier to work with but for now I think this is fine

PR #1223, opened 2022-03-10
add glx now open component and styling for stop page

bfauble:
#### Summary of changes
**Asana Ticket:** [GLX "NOW OPEN" treatment for /stops pages](https://app.asana.com/0/555089885850811/1201915536508567/f)

- render glx open stations into dom for frontend to pick up and use within react component.
- create custom component to be placed at appropriate places

image
image
image


i0sea (approved):
🟢

i0sea:
Generating some new snapshots might resolve some of the failing tests

apps/site/assets/ts/hooks/useGlxOpen.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const useGlxOpen = (stationId: string): boolean => {
+  const [isGlxOpen, setIsGlxOpen] = useState(false);
+  useEffect(() => {
+    const glxOpen = document.querySelector(".glx-is-open");
+    if (glxOpen instanceof HTMLElement) {
+      setIsGlxOpen(glxOpen.dataset.open === "true");
+    }
+  }, []);
+
```
thecristen, line 13: Is `isGlxOpen` a variable that needs to be stateful? 🤔I am wondering what React is adding here as opposed to writing this as a simpler vanilla TypeScript helper function.

bfauble: removed hook entirely and just moved into the component.

apps/site/assets/ts/hooks/useGlxOpen.ts
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from "react";
+
+const glxStations = ["place-lech", "place-unsqu", "place-spmnl"];
+
+export const getIsGlxOpen = (stationId: string): boolean => {
```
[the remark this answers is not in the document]
bfauble: updated

apps/site/test/site_web/plugs/glx_now_open_test.exs
```diff
@@ -22,24 +22,26 @@ defmodule SiteWeb.Plugs.GlxNowOpenTest do
...
+      assert updated_conn.assigns.glx_stations_open == nil
     end
 
-    test "assigns glx_now_open? with true if current date is between opening date and 3 months post opening",
+    test "assigns glx_stations_open with array of stations if current date is between opening date and 3 months post opening",
          %{conn: conn} do
       updated_conn = call(conn, now_fn: &now_between_fn/0)
 
-      assert updated_conn.assigns.glx_now_open? == true
+      assert updated_conn.assigns.glx_stations_open == "place-lech,place-unsqu,place-spmnl"
```
thecristen, line 37: We'll have to update this when the Medford branch's station open later this year. Might be sufficient to just ensure a truthy value with `assert updated_conn.assigns.glx_stations_open`

apps/site/test/site_web/plugs/glx_now_open_test.exs
```diff
@@ -22,24 +22,26 @@ defmodule SiteWeb.Plugs.GlxNowOpenTest do
...
     end
 
-    test "assigns glx_now_open? with false if current date is after 3 months post opening date",
+    test "assigns glx_stations_open with nil if current date is after 3 months post opening date",
          %{conn: conn} do
       updated_conn = call(conn, now_fn: &now_after_fn/0)
 
-      assert updated_conn.assigns.glx_now_open? == false
+      assert updated_conn.assigns.glx_stations_open == nil
```
thecristen, line 44: Can also `refute updated_conn.assigns.glx_stations_open`

thecristen (approved):
Thanks for taking my feedback into consideration -- looks great to me.

thecristen:
The failing test I think is because the call to `React.render(
      "StopPage",
      %{
        stopPageData: stop_page_data,
        mapData: map_data,
        mapId: map_id
      }
    )` is stumbling over your new addition. That function is what pre-renders the component (via a call to a Node.js script and ultimately `ReactServer.renderToString`) and writes it into the template so users with JS disabled can see a mostly-complete version of the page.

PR #1230, opened 2022-03-15
Add logging for when no-js class is added

thecristen:
Adding this so we can start to collect data on how many requests are from user agents without JavaScript enabled.

Pursued in support of [Check existing logging or add logging to trace source of Maps Static API requests](https://app.asana.com/0/385363666817452/1201938268964502/f)


thecristen:
@lund5000 Feel free to chime in, request change in format of the logged message, etc.

i0sea (approved):

PR #1231, opened 2022-03-15
Misc. bugfixes

thecristen:
A couple tweaks to avoid known sources of server errors in user flows. These two problems were found in Sentry months ago and keep coming up, usually around times when certain routes have shuttles. Investigating/debugging the underlying data is beyond the scope of fixing these errors.

#### Summary of changes
**Asana Ticket:** [no function clause matching in SiteWeb.CustomerSupportController.submit/2](https://app.asana.com/0/0/1200334203657800)

Turns out we thought we had fixed this in https://github.com/mbta/dotcom/pull/969 but had left out one case.
* Function clause added to the submit function to handle cases where the second argument does not include the "support" key.

**Asana Ticket:** [function nil.id/0 is undefined](https://app.asana.com/0/0/1200951746296209)
This seems related to https://github.com/mbta/dotcom/pull/1220 where the schedule's stop happens to be `nil`.
* Check for nil schedule before requesting its id

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
@@ -136,7 +136,7 @@ defmodule SiteWeb.ScheduleController.TimetableController do
   end
 
   @spec trip_schedule(Schedules.Schedule.t()) ::
-          {{Schedules.Trip.id_t() | nil, Stops.Stop.id_t()}, Schedules.Schedule.t()}
+          {{Schedules.Trip.id_t() | nil, Stops.Stop.id_t() | nil}, Schedules.Schedule.t()}
   defp trip_schedule(%Schedules.Schedule{trip: trip, stop: stop} = schedule)
        when not is_nil(trip) and not is_nil(stop) do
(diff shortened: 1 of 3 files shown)


apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
```diff
@@ -137,13 +137,20 @@ defmodule SiteWeb.ScheduleController.TimetableController do
 
   @spec trip_schedule(Schedules.Schedule.t()) ::
           {{Schedules.Trip.id_t() | nil, Stops.Stop.id_t()}, Schedules.Schedule.t()}
```
bfauble, line 139: I think the spec needs to be updated to include the `| nil` for the `stop.id` return as well

bfauble (changes requested):
1 quick thing then looks good

apps/site/lib/site_web/controllers/schedule/timetable_controller.ex
thecristen: Yeah, that's right. Thanks for catching - done

bfauble (approved):

PR #1233, opened 2022-03-16
Remove "Powered by Google" when using AWS

zovt:
#### Summary of changes
**Asana Ticket:** [Remove "Powered by Google" when using AWS](https://app.asana.com/0/555089885850811/1201936731451755/f)

This PR prevents the "Powered by Google" logo from rendering when we're not using Google to drive the autocomplete. The basic approach is as follows:
- Inject the current autocomplete backend into the page by querying `LocationService.active_service(:autocomplete)`
- Check that in each place where we render the "Powered by Google" logo, and prevent the rendering if we're not using Google

If there's a better way to go about this than the injection method here, let me know

### DIFF
--- a/apps/location_service/lib/location_service.ex
+++ b/apps/location_service/lib/location_service.ex
@@ -46,7 +46,7 @@ defmodule LocationService do
     end)
   end
 
-  defp active_service(key) do
+  def active_service(key) do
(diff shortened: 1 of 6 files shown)


i0sea (approved):
lgtm

PR #1239, opened 2022-03-18
WIP Update hard-coded endpoints for GL

i0sea:
#### Summary of changes
**Asana Ticket:** [Update hard-coded endpoints for GL](https://app.asana.com/0/555089885850811/1201984535304521/f)

use getIsGlxOpen to determine stops to be used in green line picker.


thecristen (commented):
I think you're getting the two exports from the GlxOpen file mixed up -- in your component it looks like you're trying to invoke the `getIsGlxOpen` function using the props for the `<GlxOpen />` component. You probably just want to call `getIsGlxOpen("place-unsqu")`.

thecristen (approved):
🟢

PR #1240, opened 2022-03-18
bugfix: correctly allow matching of 0 index in glx open stations

bfauble:
Updated glx now open station list checking to include 0 (first) index.


zovt (approved):
Could also use `includes` here I believe, but looks good to me 👍

bfauble:
> Could also use `includes` here I believe, but looks good to me 👍

probably a good call. would also keep it consistent w/ the elixir template logic. Updated

PR #1242, opened 2022-03-21
Tech debt | Schedules direction picker should get destinations from API

i0sea:
Tech debt | Schedules direction picker should get destinations from API [Asana Ticket](https://app.asana.com/0/555089885850811/1201207135598169/f)
Fetch termini of routes from v3 api and use that data to populate the green line route picker.

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/green_termini_api.ex
+++ b/apps/site/lib/site_web/controllers/schedule/green_termini_api.ex
@@ -1,4 +1,4 @@
-defmodule SiteWeb.ScheduleController.GreenTerminiApiTest do
+defmodule SiteWeb.ScheduleController.GreenTerminiApi do
   @moduledoc """
     API for retrieving the terminuses of each green line route
   """
--- a/apps/site/test/site_web/controllers/schedule/green_termini_api_test.exs
+++ b/apps/site/test/site_web/controllers/schedule/green_termini_api_test.exs
@@ -1,4 +1,4 @@
-defmodule SiteWeb.ScheduleController.GreenTerminiApi do
+defmodule SiteWeb.ScheduleController.GreenTerminiApiTest do
   use ExUnit.Case, async: true
   use SiteWeb.ConnCase, async: true
 


apps/site/assets/ts/schedule/components/direction/GreenLineMenu.tsx
```diff
@@ -43,6 +72,19 @@ interface GreenLineItem {
...
+const [state, dispatch] = useReducer(reducer, {
+  data: null,
+  isLoading: false,
+  error: false
+});
+
+const allRouteIds = ["Green-B", "Green-C", "Green-D", "Green-E"];
+let directionDestinations = new Map();
+for (var routeId in allRouteIds){
+  directionDestinations.set(routeId,fetchData(routeId, dispatch));
```
thecristen, line 84: I think it might only work if run inside a `useEffect()` callback. 😕 There's a similar example in `ScheduleModalContent`.

(I'm not sure how the for loop will play out! But the usual pattern is that the `fetchData` function will fetch the data and call the `dispatch` function, which will update the `state.data` with the fetched value. So in a loop you might end up overwriting the value with each iteration.)

apps/site/assets/ts/schedule/components/direction/GreenLineMenu.tsx
```diff
@@ -17,6 +20,32 @@ import iconGreen from "../../../../static/images/icon-green-line-small.svg";
...
+export const fetchData = (
+  routeId: string,
+  dispatch: (action: fetchAction) => void
+): Promise<void> => {
+  dispatch({ type: "FETCH_STARTED" });
+  return (
+    window.fetch &&
+    window
+      .fetch(
+        `/schedules/map_api?id=${routeId}&direction_id=0`
```
thecristen, line 37: One option would be to set up a new endpoint custom made for this admittedly peculiar scenario. Like what if `/schedules/green_termini_api` could be called here and could return the destination values? This might be more straightforward than iterating through these calls and picking the data you need from those results.

apps/site/lib/site_web/controllers/schedule/green_termini_api.ex
```diff
@@ -0,0 +1,15 @@
...
+    API for retrieving the terminuses of each green line route
+  """
+  use SiteWeb, :controller
+
+  alias SiteWeb.ScheduleController.Line.Helpers, as: LineHelpers
+
+  @spec show(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def show(_conn, %{
+  }) do
+    ["Green-B", "Green-C", "Green-D", "Green-E"]
```
thecristen, line 12: :D We have a shortcut for this one - `GreenLine.branch_ids()`

apps/site/lib/site_web/controllers/schedule/green_termini_api.ex
```diff
@@ -0,0 +1,15 @@
...
+    |> Map.new(fn [k, _v] -> {k, LineHelpers.get_route(k).direction_destinations} end)
```
thecristen, line 13: Each member of the Enumerable is just one string, so this would be `fn k ->` rather than `fn [k, v] ->`

thecristen, line 13: Careful here! `LineHelpers.get_route` doesn't directly return the route, it returns `{:ok, route}`

thecristen (commented):
✨ We're taking you through the entire stack and back - these comments focus on the backend piece. You can navigate to `http://localhost:4001/schedules/green_termini_api` directly to see what it's returning (or, what error it's giving)!

For more on this you can refer to the section in the Phoenix Framework on rendering in controllers - the JSON response will probably work best for this https://hexdocs.pm/phoenix/1.4.1/controllers.html#rendering

apps/site/lib/site_web/controllers/schedule/green_termini_api.ex
```diff
@@ -0,0 +1,19 @@
+defmodule SiteWeb.ScheduleController.GreenTerminiApi do
+  @moduledoc """
+    API for retrieving the terminuses of each green line route
+  """
+  use SiteWeb, :controller
+
+  alias SiteWeb.ScheduleController.Line.Helpers, as: LineHelpers
+
+  @spec show(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def show(conn, %{}) do
```
thecristen, line 10: Let's add a simple unit test for this?

thecristen (commented):
Shouldn't this update cover the `<ExpandedGreenMenu />` too? As far as I can tell those are still using the hard-coded values in the `greenRoutes` variable.

thecristen (changes requested):
Thanks for the test!

One thing I noticed: when I go to /schedules/Green, then click to open the menu, the component + the rest of the map/line diagram disappear thanks to a JS error.

https://user-images.githubusercontent.com/2136286/161839393-77f3b294-8db3-4f10-b276-53bee00e6458.mov

.

thecristen (approved):
✨

PR #1243, opened 2022-03-21
Remainder of GLX "NOW OPEN" treatment for /stops pages

thecristen:
#### Summary of changes
**Asana Ticket:** [Remainder of GLX "NOW OPEN" treatment for /stops pages](https://app.asana.com/0/555089885850811/1201984803295268/f)

This was fairly involved! I attempted to organize this info smaller, logically self-contained commits.

- To be able to render the desired text, I had to explicitly add the date as well as the new stop IDs to be assigned in the plug.
- Extracted a hook from the existing React component file to be able to reuse the DOM-querying logic in other places
- Added the desired coloring to the stop page's stop name and station information tab
- Added the desired text under Station Information

image

Also: Updated some failing unit tests
TODO: I still need to add a unit test to test the new hook.

### DIFF
--- a/apps/site/assets/ts/stop/components/StopPage.tsx
+++ b/apps/site/assets/ts/stop/components/StopPage.tsx
@@ -5,7 +5,7 @@ import { StopPageData, StopMapData, TypedRoutes } from "./__stop";
 import { EnhancedRoute } from "../../__v3api";
 import StopPageHeader from "./StopPageHeader";
 import { reducer, initialState, Dispatch, updateRoutesAction } from "../state";
-import { Alert } from "../../components/Alerts";
+import Alerts from "../../components/Alerts";
 import AlertsTab from "./AlertsTab";
 import Sidebar from "./Sidebar";
 import LocationBlock from "./LocationBlock";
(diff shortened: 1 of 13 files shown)


apps/site/assets/ts/stop/components/StopPage.tsx
```diff
@@ -71,13 +86,24 @@ export default ({
...
-              See upcoming departures, maps, and other features at this
-              location.
-            </p>
+          <div
+            className={`m-stop-page__info-container${
+              isGlxOpen ? " glx-open" : ""
+            }`}
+          >
+            <div className="m-stop-page__info">
+              {highPriorityAlerts.length > 0 ? (
```
bfauble, line 95: any particular reason this is being recreated here instead of just using `<Alerts/>` component as before?

thecristen: The Alerts component comes wrapped in a `<div className="container">` , which is centered and contained using  margins that work fine on a white background but not with the green. 

image

And I was hesitant to modify the `<Alerts />` component itself, since it's likely used in many places.

Going to try to import it again now that I've updated the surrounding styling anyway, will push up another commit soon

thecristen: @bfauble Updated this to just use <Alerts />.

Also modified some styling and checked it with a fake alert, it will look about this

image

bfauble (approved):

PR #1244, opened 2022-03-21
Always render acccessibility icon for bus stops

zovt:
#### Summary of changes
**Asana Ticket:** [Show accessibility icon for all bus stops](https://app.asana.com/0/385363666817452/1201894234147725/f)

This PR updates the gating logic for the accessibility icons to always render for bus routes. The following locations have been updated:
- Line Diagram
- Stop pages (both the header and the nearby transfers)
- Transit Near Me

I decided to tackle this on the front end since I figured it was more of a rendering concern, and didn't want to muck with any of the accessibility data we return from the back end.

I did some light testing, and don't believe that these changes should have any unexpected side effects, but let me know if there is a cleaner approach for any of these.

### DIFF
--- a/apps/site/assets/ts/helpers/routes.ts
+++ b/apps/site/assets/ts/helpers/routes.ts
@@ -1,0 +1,20 @@
+import { RouteStop } from "../schedule/components/__schedule";
+import { RouteWithDirection, TypedRoutes } from "../stop/components/__stop";
+import { isABusRoute } from "../models/route";
+
+export function isBusRouteStop(stop: RouteStop): boolean {
+  return !!stop.route && isABusRoute(stop.route);
+}
+
+export function routesWithDirectionsAreAllBusStops(
+  routes: RouteWithDirection[]
+): boolean {
+  return (
+    routes.length > 0 &&
+    routes.reduce((acc: boolean, r) => isABusRoute(r.route) && acc, true)
+  );
+}
+
+export function typedRoutesHasBusRoute(routes: TypedRoutes[]): boolean {
+  return !!routes.find(r => r.group_name === "bus");
+}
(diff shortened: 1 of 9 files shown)


thecristen (changes requested):
Let's add a helper function, maybe `routeHasBusStop()`, that could return the boolean for all these different scenarios (taking in `TypedRoutes[]`, `RouteWithDirection[]`, `RouteStop`, `Route`).

zovt:
@thecristen 

> Let's add a helper function, maybe `routeHasBusStop()`, that could return the boolean for all these different scenarios (taking in `TypedRoutes[]`, `RouteWithDirection[]`, `RouteStop`, `Route`).

I think the different scenarios have a different enough meaning that they warrant separately-named functions (plus the type punning necessary for the arrays is pretty ugly), but I've colocated all of them into a single file in `helpers`. There's another `route` file in `models`, but the function signatures need to pull in some types from separate locations and I didn't want to pollute that file with those. 

Let me know what you think!

apps/site/assets/ts/helpers/routes.ts
```diff
@@ -0,0 +1,20 @@
...
+): boolean {
+  return (
+    routes.length > 0 &&
+    routes.reduce((acc: boolean, r) => isABusRoute(r.route) && acc, true)
+  );
+}
+
+export function typedRoutesHasBusRoute(routes: TypedRoutes[]): boolean {
+  return !!routes.find(r => r.group_name === "bus");
+}
```
thecristen, line 20: Are these similar enough to be rewritten into one function using Typescript's function overloading? See example here https://dmitripavlutin.com/typescript-function-overloading/

thecristen (approved):

PR #1245, opened 2022-03-21
test: add unit test for useIsGlxOpen hook

thecristen:
Also adds a small refactor to the hook to avoid returning an opening date when the station will not newly open.

( Leftover task from #1243 )


zovt (approved):
🪝 🚋

PR #1248, opened 2022-03-24
no translate stop names on timetables

i0sea:
#### Summary of changes
**Asana Ticket:** [Prevent stop name translation in timetables](https://app.asana.com/0/555089885850811/1201984535304523/f)

Stop translation of stop names on timetables for commuter rail and ferry by adding `notranslate` class.


apps/site/lib/site_web/templates/schedule/_timetable.html.eex
```diff
@@ -83,7 +83,7 @@
               class: "js-tt-stop-name m-timetable__cell--first-column m-timetable__cell m-timetable__cell--" <>
                 cell_background
             ] do %>
-              <div class="m-timetable__row-header m-timetable__stop-name">
+              <div class="m-timetable__row-header m-timetable__stop-name notranslate">
```
bfauble, line 86: looks like this one works for the list on the side but the direction picker is also a part of this ticket

image

i0sea: Resolved this in the line diagram ticket since it uses the same direction picker component. I'll merge that one in first.

thecristen: @i0sea It's not the same direction picker component.

thecristen (approved):
LGTM!

PR #1249, opened 2022-03-25
Prevent stop name translation on line diagrams

i0sea:
#### Summary of changes
**Asana Ticket:** [Prevent stop name translation on line diagrams](https://app.asana.com/0/555089885850811/1201938268964507/f)

Don't translate stop names on line diagram pages, including direction selector


i0sea:
All contained elements will also not be translated

PR #1250, opened 2022-03-25
Render TikTok icon in footer

zovt:
#### Summary of changes
**Asana Ticket:** [Add MBTA's TikTok account to our footer](https://app.asana.com/0/385363666817452/1201625950270166/f)

Adds the TikTok icon in our footer, using an SVG.

Sidenote: in newer versions of fontawesome, there is a tiktok icon available. Once we upgrade fontawesome it would make sense to rework this slightly to use that.


i0sea (approved):
🚀

PR #1253, opened 2022-03-28
Prevent stop name translation in Schedule Finder

i0sea:
#### Summary of changes
**Asana Ticket:** [Prevent stop name translation in Schedule Finder](https://app.asana.com/0/555089885850811/1201984535304524/f)

Add notranslate tag to schedule finder dropdown menu, schedule modal, and stop name header.

### DIFF
--- a/apps/site/lib/site_web/templates/schedule/_line_header.html.eex
+++ b/apps/site/lib/site_web/templates/schedule/_line_header.html.eex
@@ -1,5 +1,5 @@
 <div class="schedule__header <%= header_class(@route) %>">
-  <div class="schedule__header-container notranslate">
+  <div class="schedule__header-container">
     <h1 class="schedule__route-name">
       <%= route_header_text(@route) %>
     </h1>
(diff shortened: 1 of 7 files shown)


apps/site/lib/site_web/templates/schedule/_line_header.html.eex
```diff
@@ -1,5 +1,5 @@
 <div class="schedule__header <%= header_class(@route) %>">
-  <div class="schedule__header-container">
+  <div class="schedule__header-container notranslate">
```
thecristen, line 2: Can we undo this? We haven't committed to preventing translation of route names yet. This spot also prevents translation of the tab names (schedules & info, alerts) which I don't think we want.

thecristen (changes requested):
Can you add the `notranslate` class to the `<select>` for "Choose an origin stop" too? That should fix this one spot that's still being translated (Centro de Gobierno in the screenshot):

image

thecristen (approved):
👍🏼

PR #1254, opened 2022-03-29
Broken search on projects pages

zovt:
#### Summary of changes
**Asana Ticket:** [Navigating to /search from header shows no results](https://app.asana.com/0/385363666817452/1202012916713303/f)

Previously, we waiting to initialize the global header search until after Google maps had loaded. This was causing issues on pages where Google maps did not need to get loaded (e.g. project pages).

### DIFF
--- a/apps/site/assets/js/algolia-autocomplete-with-geo.js
+++ b/apps/site/assets/js/algolia-autocomplete-with-geo.js
@@ -56,18 +56,10 @@ class AlgoliaAutocompleteWithGeo extends AlgoliaAutocomplete {
 
   bind() {
     super.bind();
-    this.onFocus = this.onFocus.bind(this);
   }
 
   _addListeners() {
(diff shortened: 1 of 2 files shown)


apps/site/assets/js/algolia-autocomplete-with-geo.js
```diff
@@ -54,26 +53,6 @@ class AlgoliaAutocompleteWithGeo extends AlgoliaAutocomplete {
     this._indices.push("popular");
   }
 
-  bind() {
-    super.bind();
-    this.onFocus = this.onFocus.bind(this);
-  }
-
-  _addListeners() {
-    super._addListeners();
```
thecristen, line 63: Are we sure removing this and the above `super.bind()` aren't going to cause an adverse side effect?

thecristen (approved):
👍🏼

PR #1259, opened 2022-03-30
Remove `resetSessionToken` call

zovt:
#### Summary of changes
**Asana Ticket:** [Selecting custom locations in trip planner should populate map](https://app.asana.com/0/385363666817452/1202050563424979/f)

This was missed when previously deleting this method and was silently breaking selecting locations in the trip planner


thecristen (approved):
It works!

PR #1263, opened 2022-04-04
Fix rendering issues with CR zones

zovt:
#### Summary of changes
**Asana Ticket:** [CR zone badge cut off in suggestions](https://app.asana.com/0/0/1201994835377135/f)

On Chrome and Firefox on Windows in particular, the commuter rail zone "pill" was getting cut off vertically in different situations. Switching to `inline-block` alleviates this without otherwise affecting layout in my testing.


zovt:
Before:


After:

bfauble (approved):

PR #1267, opened 2022-04-07
Reinstate computing and assessing Elixir test coverage

thecristen:
# Backstory

Way back in the day much of CTD used [CodeCov](https://about.codecov.io) and got coverage reports there. But they weren't very reliable, often returning wonky/incorrect results, so we weren't too enamored with it.

Then we migrated all of our projects to use Github Actions, during which we overwhelmingly adopted [lcov_ex](https://hexdocs.pm/lcov_ex/readme.html) to generate coverage information, incorporating into our workflows and [using our own fork of a Github action, github-actions-report-lcov,](https://github.com/mbta/github-actions-report-lcov) to report coverage using this data.

All well and good for most, but it never computed coverage for Dotcom, and after some frustration we disabled the coverage check and moved on with our lives.

Until now...

# The diagnosis

It turns out that the prior workflow didn't work as expected on Elixir umbrella projects. It would generate coverage for each app individually, and its output contained a mix of relative and absolute links to source files. Later on, resolving the relative links would cause the action to try to parse incorrect source files, leading to failure.

# The fix

I decided to go back to the coverage tool we were using before, ExCoveralls, and instead of reporting it to CodeCov, we'll save the output as an artifact here in GitHub.

The only caveat is there's no obvious way to generate a PR comment with more detailed coverage information -- I'm leaving that as an exercise for another day.

# What will happen in PRs

* The tests will run as normal, but will also will fail CI if the Elixir test coverage is found to be below 90%. That will look something like this ([example from my fork here](https://github.com/thecristen/dotcom/runs/5874121811?check_suite_focus=true)):

image

* A .zip file containing the output HTML coverage report is added as an artifact to the workflow run, see bottom of screenshot:

image


amaisano:
Will that `.zip` file be a human-readable HTML report like we used to get with `coveralls -html`? For less savvy users like me, I'll need a clear idea of which files->lines were missed.

thecristen:
@amaisano yep! it's a big HTML report that looks like this 

image

To your second question, yes! You'd be able to run the same command (CI uses a variation on `mix coveralls.html`) and produce the HTML output locally.

amaisano (approved):

thecristen:
Rebased / resolved some conflicts. bump @bfauble

bfauble (approved):

PR #1270, opened 2022-04-11
Sentry updates

thecristen:
#### Summary of changes
**Asana Ticket:** [Tech Debt | Add Sentry monitoring for our JS frontend](https://app.asana.com/0/385363666817452/1200329682137272/f)

Dependabot, via https://github.com/mbta/dotcom/pull/1268, prompted me to try to update our [Sentry](https://sentry.io/) settings a bit.

* Updated the Elixir Sentry library to the latest version. Followed some of the guidance for handling upgrading from v7 to v8.
* Updated the browser Sentry libraries, which we had but somehow weren't really using?
* Started using Sentry in the browser/frontend

Note: I noticed a few CTD projects set up a separate project in Sentry for the frontend and backend. I opted not to do this -- so this setup will start adding frontend errors into our existing `mbta-site` project. I might change that later. T[his StackOverflow comment](https://stackoverflow.com/a/27428603) suggests that a Sentry project's selected "Platform" setting is only used for cosmetic purposes of rendering errors/stack traces differently. We'll see!


apps/site/assets/js/app.js
```diff
@@ -46,6 +48,13 @@ import eventPageSetup from "./event-page-setup";
 import previousEventsButton from "./view-previous-events";
 import pslPageSetup from "./psl-page-setup.js";
 
+if (window.sentry) {
+  Sentry.init({
+    dsn: window.sentry.dsn,
+    integrations: [new BrowserTracing()]
+  });
```
lemald, line 55: **suggestion (non-blocking):** You may also want to [pass an environment](https://docs.sentry.io/platforms/javascript/configuration/environments/) option here to more easily differentiate between your prod and your dev errors.

skyqrose: An example of passing the environment in Glides: [frontend](https://github.com/mbta/glides/pull/431/files#diff-2692177d3a0965491f6ceb040f13619e4b9200e106bf09f3c2b7d7227f700b8dR28) [backend](https://github.com/mbta/glides/blob/babb0b117f62f6943a1b397421e66c89ba8b5812/config/prod.exs#L97)

apps/site/config/prod.exs
```diff
@@ -51,7 +51,7 @@ config :site,
 config :logger,
   level: :info,
   handle_sasl_reports: true,
-  backends: [:console]
+  backends: [:console, Sentry.LoggerBackend]
```
skyqrose, line 54: It's possible to do sourcemaps on the backend.
[docs](https://hexdocs.pm/sentry/Sentry.Sources.html).  
[config](https://github.com/mbta/glides/pull/431/files#diff-8e72c709985cb6a43cc0ab32cb487b8c17d2b2d003338ea42a4e2ad2829979dcR81).
[You also have to recompile your app an extra time](https://github.com/mbta/glides/pull/431/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R63).

apps/site/assets/js/app.js
thecristen: @skyqrose this example is 💯 
I think I'll set this up in a followup PR.

thecristen:
No idea what's going on with the dependencies here, going to try again later

PR #1274, opened 2022-04-13
Update vulnerabilities

thecristen:
Quick PR to update or remove a few npm packages in order to reduce the number of vulnerabilities (as reported via [Dependabot](https://github.com/mbta/dotcom/security/dependabot))

* Bumps `minimist` to a patched version
* Removed `backstopjs` to eliminate usage of bad versions of `jpeg-js`

We have not used Backstop in a long time, so now seemed as good a time to remove as ever.

There are lots more security vulnerabilities! But we can't fix 'em at least until we upgrade from Webpack 4 to 5 🙃


zovt (approved):
🪓 🪓 🪓

PR #1278, opened 2022-04-19
Use pre-cached object reference for checking offset

bfauble:
When navigating from a 'parent' page for a type of service:
image
These search boxes are present within the page and an embedded algolia search object is generated for each search input. When a user navigates to a more specific pages, say for the red line only:
image
The from, to, and middle search are no longer present within the dom, but the objects are still loaded in memory. So when the app fires the resize event (sometimes happens on page load, especially on mobile) it kicks off the `onOpen` call which attempts to re-query the dom for the element group which it has already been attached to, but since the 3 fields are no longer 'physically' present, the call returns null and destructuring null throws an error.  

image
image


thecristen (approved):
Was able to confirm bug with your instructions and confirm fix. Awesome sleuthing 🤩

PR #1280, opened 2022-04-22
chore: remove select-modal.js + sifter dependency

thecristen:
Looks like it was used in an [older version of the Schedule tab](https://github.com/mbta/dotcom-legacy/commit/9e9ae9d89af87836178a0beff018a1c6b509e4eb), but from what I can tell we don't need it now.


zovt (approved):
🪓

PR #1282, opened 2022-04-22
fix: blank to/from trip planner. use address instead of description for trip planner controls 

bfauble:
Auto complete result no longer populates a 'description' field for 'locations' type of results. Change the input to be populated w/ returned full address instead of leaving blank/null.

image


zovt (approved):
Thanks for fixing this and sorry for the churn 😅

PR #1285, opened 2022-04-26
feat: add referrer to Algolia API request

thecristen:
I was experimenting with adding some restrictions on the Algolia keys we use in our deployed applications. I'd added write and search keys to dev-blue that limit requests to those having (URL of dev-blue) as the referrer. It worked for other Algolia requests, but not this POST query.

This PR updates our request to include the `referrer` header value present on the `conn`. I will deploy it to dev-blue to see if my idea works!

### DIFF
--- a/apps/algolia/lib/algolia/api.ex
+++ b/apps/algolia/lib/algolia/api.ex
@@ -5,12 +5,13 @@ defmodule Algolia.Api do
   alias Algolia.Config
   require Logger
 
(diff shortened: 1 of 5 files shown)


thecristen:
Actually I should test this on dev-green. Will update API keys and deploy to that soon...

thecristen:
Going to table this idea for another time.

PR #1286, opened 2022-04-26
Improve trip-planner/to/ link results from /stops

i0sea:
#### Summary of changes
**Asana Ticket:** [🟡 Improve trip-planner/to/ link results from /stops](https://app.asana.com/0/385363666817452/1201994835377152/f)

Make improvements to trip planner to so that when linking from stop page map marker matches and to field text contains station name.


apps/site/lib/site_web/controllers/trip_plan_controller.ex
```diff
@@ -69,29 +69,28 @@ defmodule SiteWeb.TripPlanController do
   def to(conn, %{
         "address" => address
       }) do
-    with [latitude, longitude] when is_float(latitude) and is_float(longitude) <-
-           String.split(address, ",") do
+    if String.match?(address, ~r/^(\-?\d+(\.\d+)?),(\-?\d+(\.\d+)?),.*$/) do
+      [latitude, longitude, name] = String.split(address, ",")
```
thecristen, line 73: I think we should [use the `parts` option ](https://hexdocs.pm/elixir/1.10.3/String.html#split/3)to limit String.split to always output three items. Otherwise this errors if trying to go to a stop that happens to have a comma in its name (see /stops/111640, it's "Cobbs Corner, Canton")

thecristen (changes requested):
Aside from one edge case that needs fixing it seems to work really well! 🤩 

It would be great to also add a test in `trip_plan_controller_test.exs` to add a case when using this type of URL, and verify that the to latitude/longitude/name use the expected values.

apps/site/test/site_web/controllers/trip_plan_controller_test.exs
```diff
@@ -752,6 +752,14 @@ defmodule SiteWeb.TripPlanControllerTest do
       assert conn.assigns.query.to.name == "Geocoded Boston Common"
     end
 
+    test "uses expected values when addres is formatted latitutde,longitude,stopName" do
```
thecristen, line 755: this line needs to be edited to pass in the `conn` variable (see following test that happens to do this)

i0sea: added the conn variable and removed some excess spaces. This path doesn't make a query so should it be tested differently?

thecristen (approved):

PR #1291, opened 2022-05-02
Upgrade Font Awesome

bfauble:
#### Summary of changes
**Asana Ticket:** [Upgrade Font Awesome](https://app.asana.com/0/385363666817452/1202034816145696/f)

Note: Still some styling that needs to be tweaked slightly due to slight changes in icon sizes etc.

Todo: Fix fa-icon-regular mixin helper.


thecristen (commented):
Thanks for the nicely organized commits, makes this very sane to step through for review.

I have a few initial comments upon first skim.

I was also curious if you ran into any trouble due to the fact that the Font Awesome 4 icons were free but the Font Awesome 6 free version has some limitations? It looks like there are some specific icons that are not free (maybe we're not using any of those anyway?) and also some _styles_ for some icons are not free. For example, the solid carets are free but the regular variant is not available without the Pro plan (I'm concluding this based on [filtering this view](https://fontawesome.com/icons/caret-down?s=regular))

i0sea (changes requested):
There's a merge conflict in `package-lock.json`, otherwise looks good to me!

thecristen (approved):
Approved, pending merge conflict

PR #1293, opened 2022-05-06
Remove "old"

thecristen:
No ticket, just was mucking about our CDN configuration and wondering, "do we still need ALL this stuff?"

I offer the following things for the chopping block 🔪 

## 🔪 `OldSiteFileController`

Points these paths to corresponding files:
- `"/uploadedfiles/*"`
- `"/uploadedFiles/*"`
- `"/uploadedimages/*"`
- `"/uploadedImages/*"`
- `"/images/*"`
- `"/lib/*"`
- `"/gtfs_archive/archived_feeds.txt"`

Looking in GA I could find 18 page views to `/uploadedfiles*` since January 1 2020, most of those being incorrect attempts. The most plausibly correct request, to `beta.mbta.com/uploadedfiles/About_the_T/Fare_Proposals_2012/Potential MBTA%` returns a 400 Bad Request anyway.
These folks might be better served by our current site's search functionality, but admittedly I don't know if they are being referred to the old from elsewhere.

## 🔪 `OldSiteRedirectController`

Redirects to `"/schedules_and_maps/*` to appropriate places.

I found just 4 requests since Jan 1 2020, only one of which wasn't malformed (`/schedules_and_maps/subway/lines/stations/?stopId=26637` redirects to the JFK/Umass stop!). 

## 🔪 The "Search the old website" call to action

This is at the bottom of our "no search results" template and our JS-disabled search template. It doesn't seem to be working as intended?

image

I think these have low enough usage that we can do away with them, but will defer to @lund5000


lund5000 (approved):
This also seems safe to remove to me based on lack of recent traffic. A quick note in #ctd-website-dev might be a good idea, too.

PR #1296, opened 2022-05-11
Add "Find a Location" section

i0sea:
#### Summary of changes
**Asana Ticket:** [Add "Find a Location" section](https://app.asana.com/0/385363666817452/1202125833856567/f)

Add find a location section to homepage.


apps/site/lib/site_web/templates/page/_find_a_location.html.eex
```diff
@@ -0,0 +1,29 @@
+<%
+%>
+<div class="c-grid-buttons-right">
+<div class="c-location-header">
+  <h2>
+    Find a Location
+  </h2>
+  </div>
+  <a href="https://www.mbta.com/stops/subway" class="c-location-button">
```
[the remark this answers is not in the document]
thecristen: Enhancement! Let's use the built-in Phoenix helpers for getting these paths. 👯 

`mix phx.routes SiteWeb.Router` will display the shorthand names for all the different paths like so: 
image

This is the one for `/stops`

```
stop_path  GET   /stops                                         SiteWeb.StopController :index
```

Which means in the template we can use `stop_path(@conn, :index)` for the URL. 

Coincidentally. There's an example in the footer that does this too, using Phoenix's built in `link` helper via `<%= link "Station Info", to: stop_path(@conn, :index) %>`.

apps/site/lib/site_web/templates/page/_find_a_location.html.eex
```diff
@@ -0,0 +1,29 @@
...
+    <div class="c-grid-button__content">
+      <span class="c-grid-button__icon">
+        <%= svg("stations-parking.svg") %>
+      </span>
+      <span class="c-grid-button__name">
+        Stations
+      </span>
+    </div>
+  </a>
+  <a href="https://www.mbta.com/transit-near-me" class="c-location-button">
```
thecristen, line 19: See my comment above about using the built-in Phoenix helper

thecristen (changes requested):
Some small changes requested. Nice job adjusting to pass all the tests!

Additional design notes:

* The mocks suggest we should use the medium font weight for these buttons rather than the default
* I'm seeing different sizes for the buttons on Safari, probably because the first icon is taller? These should be the same height. 

image

* It looks like the icons should take the same amount of horizontal space too. In the mock for the smallest breakpoint the button text are expected to align.

image

i0sea:
Made most changes, just have to get the buttons side by side for small tablet. I'll re-request a review when it's all done.
Screen Shot 2022-05-19 at 4 49 04 PM

apps/site/assets/css/_grid-button.scss
```diff
@@ -9,6 +9,54 @@
...
+.c-location-button {
+  flex-grow: 1;
+  height: 50px;
+  width: 100px;
+  font-weight: $font-weight-medium;
+  display: block;
+  flex-shrink: 0;
+  margin-bottom: $space-6;
+  margin-top: $space-6;
+  padding: 0 $base-spacing * .25;
```
thecristen, line 48: nitpick: Let's remove this padding for the smallest breakpoint. that way the button width will be aligned with the header 
image

apps/site/assets/css/_grid-button.scss
```diff
@@ -9,6 +9,54 @@
...
+    flex-wrap: wrap;
+  }
+}
+
+.c-location-header {
+  flex-grow: 2;
+  @include media-breakpoint-down(sm) {
+    flex-basis: 100%;
+  }
+}
```
thecristen, line 37: Having both of these set to `flex-grow: 2` encourages these to take up the same width but on larger breakpoints we can let the buttons take up more width. Might need adjustment to prevent wrapping like this
image

apps/site/lib/site_web/templates/page/_find_a_location.html.eex
```diff
@@ -0,0 +1,29 @@
...
+        <span class="c-grid-button__name">
+          Stations and Parking
+        </span>
+      </div>
+    <% end %>
+    <%= link to: cms_static_page_path(@conn, "/transit-near-me"), class: "c-location-button" do %>
+      <div class="c-grid-button__content">
+        <span class="c-grid-button__icon">
+          <i class="fa fa-map-location-dot" aria-hidden=true></i>
+        </span>
```
thecristen, line 22:
```suggestion
        <%= fa "map-location-dot", class: "c-grid-button__icon" %>
```

I think using our `fa` helper function will work for this.

thecristen (changes requested):
A couple small changes.

Also, let's adjust the layout in the `sm` breakpoint to match the mocks by putting the header on its own line. That'll also help the button text not wrap, as seen in the screenshot here 
image

apps/site/lib/site_web/templates/page/_find_a_location.html.eex
i0sea: The sm breakpoint matches the mocks on local chrome for me 🤔 
Screen Shot 2022-05-23 at 1 06 40 PM

thecristen: Ahh maybe that's on me then. Well, we can merge that bit as is and we can always update it later if it's off somewhere.

PR #1298, opened 2022-05-13
Homepage alerts

zovt:
#### Summary of changes
**Asana Ticket:** [Alerts summary](https://app.asana.com/0/385363666817452/1202075661172620/f)

There are probably some style tokens that I could be using in the CSS, let me know if you notice anything obvious!


apps/site/lib/site_web/controllers/page_controller.ex
```diff
@@ -13,6 +13,7 @@ defmodule SiteWeb.PageController do
     WhatsHappeningItem
   }
 
+  plug(:alerts)
```
thecristen, line 16: I haven't looked into it super closely, but was just curious if you considered including and/or modifying `plug(SiteWeb.Plugs.AlertsByTimeframe)`?

zovt: 🤔 I was aware of this, but didn't consider it too deeply. It looks like that plug expects an `alerts` assign on the `conn`, which is what this `:alerts` is doing currently, and then does some timeframe filtering which I don't think is applicable to this view? Dunno, if you have more thoughts on this let me know.

apps/site/lib/site_web/views/page_view.ex
```diff
@@ -10,6 +10,143 @@ defmodule SiteWeb.PageView do
...
+      alerts
+      |> Enum.filter(&(Alerts.Priority.priority(&1) == :high))
+      |> Enum.reduce(MapSet.new(), fn alert, routes ->
+        MapSet.union(routes, Alerts.Alert.get_entity(alert, :route))
+      end)
+      |> Enum.map(&get_route/1)
+      |> Enum.filter(& &1)
+      |> Enum.group_by(&Routes.Route.type_atom(&1.type))
+      |> (&Map.merge(%{bus: [], subway: [], ferry: [], commuter_rail: []}, &1)).()
+      |> Enum.map(fn {k, routes} ->
```
bfauble, line 42: this map/sort-by can probably be simplified by moving the matching of the key type into the a function match vs the mid-pipe case statement. Something like:

```elixir
|> Enum.sort_by(& &1.name, &sort_by_route/2)

defp sort_by_route({:bus, _}), do: &sort_string_number/2
defp sort_by_route(_), do: &<=/2
```

apps/site/lib/site_web/views/page_view.ex
```diff
@@ -10,6 +10,143 @@ defmodule SiteWeb.PageView do
 
   use SiteWeb, :view
 
+  @spec sort_string_number(String.t(), String.t()) :: boolean
+  defp sort_string_number(a, b) do
```
bfauble, line 14: this could use guard clauses to remove the need for a case statement

```elixir
defp sort_string_number(a, b) when is_integer(a) and is_integer(b) do
  a <= b
end
defp sort_string_number(a, b), do: a <= b
```

zovt: Ah, yeah you're right, I missed that `Integer.parse/2` returns a tuple with the remainder. It does happen to sort properly, but I'll adjust the code to make it more obvious, thanks!

zovt:
@bfauble think I hit everything, let me know if I've missed / misunderstood anything!

bfauble (commented):
looks good, will let @thecristen sign off on her comments as well.

thecristen (commented):
more comments!

apps/site/lib/site_web/templates/page/_alerts.html.eex
```diff
@@ -0,0 +1,74 @@
...
+  <div class="m-homepage__alerts-main">
+    <div class="m-homepage__alerts-high-priority">
+      <h2>Routes With High Priority Alerts</h2>
+      <div class="m-homepage__alerts-high-priority-by-mode">
+        <%= for {mode, routes} <- @routes_with_high_priority_alerts_by_mode do %>
+          <% mode_name = Routes.Route.type_name(mode) %>
+          <div class="m-homepage__alerts-mode">
+            <h3>
+              <a href="<%= alerts_mode_url(mode) %>">
+                <%= svg(alerts_mode_icon_name(mode)) %>
```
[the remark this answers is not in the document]
zovt: The icons that get returned by that one are the circle icons, whereas we want the icons with just the mode featured

zovt:
@thecristen I think I've hit everything, let me know if I missed something :)

apps/site/lib/site_web/templates/page/_alerts.html.eex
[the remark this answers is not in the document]
thecristen: ahh my bad

thecristen (approved):
Is good!

One thing we can address after design review is how to handle longer bus numbers - right now the slash routes don't fit too well 
image

PR #1305, opened 2022-05-26
only set error message(s) once field has been focused and/or modified.

bfauble:
#### Summary of changes
**Asana Ticket:** [Trip Planner shows error message when From/To get focus](https://app.asana.com/0/555089885850811/1202299919598976/f)

- update to/from field validation to only fire after field has been either focused then blurred or changed. "Dirty" property added to field.


thecristen (commented):
How feasible would it be to write a test for this or integrate this into part of the test suite for this file?

Also -- I forgot to mark myself as `Busy` here so GitHub autoselected me for review -- feel free to manually request a review from dev that's not OOO 😉

zovt (approved):
seems reasonable to me. some sort of testing would probably be nice but I know some of the frontend code can be tricky to test

PR #1309, opened 2022-05-31
Correctly render barberpole when there are overlapping alerts

zovt:
#### Summary of changes
**Asana Ticket:** [⏱️ Line Diagram | Barberpole is inconsistent](https://app.asana.com/0/555089885850811/1200071732205207/f)

Instead of checking for exact equivalence between the effect types between stops, check if _any_ effects are shared between them before starting a new chunk.

I believe this should work for all cases. I've added a test mimicking the issue we're currently seeing with the Blue Line to verify, but let me know if there are any others you think would be worthwhile!

Before:

After:

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line/diagram_format.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line/diagram_format.ex
@@ -147,18 +147,11 @@ defmodule SiteWeb.ScheduleController.Line.DiagramFormat do
         chunk ->
           last_stop = List.last(chunk)
           last_stop_fx = effects_for_stop(last_stop)
-          has_same_effect = Enum.any?(last_stop_fx, fn effect -> effect in stop_fx end)
 
-          if has_same_effect do
-            {:cont, chunk ++ [stop]}
-          else
-            case stop_fx do
-              [] ->
-                {:cont, chunk, []}
-
-              _ ->
-                {:cont, chunk, [stop]}
-            end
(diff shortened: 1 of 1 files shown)


apps/site/lib/site_web/controllers/schedule/line/diagram_format.ex
```diff
@@ -146,8 +146,10 @@ defmodule SiteWeb.ScheduleController.Line.DiagramFormat do
 
         chunk ->
           last_stop = List.last(chunk)
+          last_stop_fx = effects_for_stop(last_stop)
+          has_same_effect = Enum.any?(last_stop_fx, fn effect -> effect in stop_fx end)
 
-          if stop_fx == effects_for_stop(last_stop) do
+          if has_same_effect do
```
bfauble, line 152: I think a way of kind of cleaning up at least a couple of the nested if/case statements here is something like this. Lets you evaluate both checks w/ the same case and saves some indents.

```elixir
case {last_stop_fx -- stop_fx, stop_fx} do
  {^last_stop_fx, []} -> {:cont, chunk, []}
  {^last_stop_fx, _} -> {:cont, chunk, [stop]}
  _ -> {:cont, chunk ++ [stop]}
end
```

```elixir
iex(15)> [1, 2, 3] -- [4, 5, 6]
[1, 2, 3]
iex(16)> [1, 2, 3] -- [4, 5, 6, 1]
[2, 3]
```

bfauble (approved):

PR #1311, opened 2022-06-01
Add GLX note

lund5000:
#### Summary of changes
Adds a paragraph explaining that GLX Medford is not included in suggested itineraries


i0sea (approved):
🟢

PR #1313, opened 2022-06-03
Access :hide_url via assigns map

zovt:
This prevents a page rendering error when the assign isn't present (e.g.
for trip planner alerts)


thecristen (approved):
👍🏼

PR #1318, opened 2022-06-13
"News" section > "Press Releases" section

bfauble:
#### Summary of changes
**Asana Ticket:** ["News" section > "Press Releases" section](https://app.asana.com/0/555089885850811/1202075661172628/f)

image
image

### DIFF
--- a/apps/site/test/site_web/views/page_view_test.exs
+++ b/apps/site/test/site_web/views/page_view_test.exs
@@ -55,7 +55,7 @@ defmodule SiteWeb.PageViewTest do
       now = Util.now()
 
       entries =
-        for idx <- 1..5 do
+        for idx <- 1..6 do
           %Teaser{
             id: idx * 1000,
             title: "News Entry #{idx}",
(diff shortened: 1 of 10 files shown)


apps/site/test/site_web/views/page_view_test.exs
```diff
@@ -72,9 +72,9 @@ defmodule SiteWeb.PageViewTest do
         |> PageView.render_news_entries()
         |> HTML.safe_to_string()
 
-      assert rendered |> Floki.find(".c-news-entry") |> Enum.count() == 5
-      assert rendered |> Floki.find(".c-news-entry--large") |> Enum.count() == 2
-      assert rendered |> Floki.find(".c-news-entry--small") |> Enum.count() == 3
+      assert rendered |> Floki.find(".c-news-entry") |> Enum.count() == 6
+      # assert rendered |> Floki.find(".c-news-entry--large") |> Enum.count() == 2
```
i0sea, line 76: Should this be uncommented or deleted?

i0sea (commented):
Looking at the mocks there's some padding on top of the icons.  xs screens may be limited to 3 press releases but I'm unsure if that's accurate. Otherwise LGTM

lund5000:
> xs screens may be limited to 3 press releases

Good catch, @i0sea! We added that detail to the events task but missed getting it into this one. 🙈 

@bfauble If it's not a quick update we can defer adding it, but it would be good to limit the list to three items at the xs breakpoint (six at higher breakpoints).

thecristen (changes requested):
I'm ok merging this if we remove the soon-to-be-extraneous "upcoming events" link.

PR #1319, opened 2022-06-13
add header to non-promoted what's happening section

bfauble:
#### Summary of changes
**Asana Ticket:** [Add "What's Happening at the MBTA" header](https://app.asana.com/0/555089885850811/1202075661172626/f)

image
image
image

### DIFF
--- a/apps/site/assets/css/_whats-happening.scss
+++ b/apps/site/assets/css/_whats-happening.scss
@@ -1,6 +1,21 @@
 .m-whats-happening__section {
   margin-bottom: $base-spacing;
   width: 100%;
+
+  h2 {
+    @include media-breakpoint-down(xxl) {
+      font-size: $font-size-base * 2;
+      margin-bottom: $base-spacing-lg;
+    }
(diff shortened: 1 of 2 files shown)


thecristen (approved):
Gets it done! 👯 

@lund5000 I want to note that the added CSS perfectly aligns this with its mocks, but also makes this header noticeably larger than the other `<h2>` headings on this page. We should probably resolve that, but that might be out of scope for this ticket.

apps/site/assets/css/_whats-happening.scss
```diff
@@ -1,6 +1,21 @@
 .m-whats-happening__section {
   margin-bottom: $base-spacing;
   width: 100%;
+
+  h2 {
```
thecristen, line 5: Sorry, can we remove this section's styling so that this heading uses the same sizing as the rest of the `<h2>` on this page?

thecristen (changes requested):

thecristen (approved):

PR #1322, opened 2022-06-14
fix(LocationService.Utils): compile with unmatched parenthesis

thecristen:
Make sure to escape any necessary characters in the text when creating the regex expression. From Sentry reports this was most often text containing parentheses or brackets.

Added a test using a search term from one of our error reports.

#### Summary of changes
**Asana Ticket:** [LocationService | Elixir.MatchError: no match of right hand side value: {:error, {'missing )', 17}}](https://app.asana.com/0/385363666817452/1202185936565802/f)


thecristen:
tagging @zovt since he wrote the original function :D

PR #1323, opened 2022-06-15
Homepage Alerts Section visual changes

zovt:
#### Summary of changes
**Asana Ticket:** [Alerts design updates](https://app.asana.com/0/555089885850811/1202384504606986/f)

Some things I noticed right off the jump:
- No spacing at the top of the container at the larger breakpoints (I think this will be solved by a later ticket maybe?)
- Some things are no longer possible if we want to rely on the container spacing. This mostly affects the footer section containing the buttons (e.g. the horizontal bar at the medium breakpoint and up can't span the full width of the container)

Screen Shot 2022-06-15 at 14 44 35
Screen Shot 2022-06-15 at 14 44 16
Screen Shot 2022-06-15 at 14 44 01
Screen Shot 2022-06-15 at 14 43 54

### DIFF
--- a/apps/alerts/lib/accessibility.ex
+++ b/apps/alerts/lib/accessibility.ex
@@ -11,5 +11,5 @@ defmodule Alerts.Accessibility do
   def is_accessibility_alert?(%Alert{effect: effect}), do: effect in effect_types()
 
   @spec effect_type_to_group_title(effect_type()) :: String.t()
-  def effect_type_to_group_title(effect), do: Alert.human_effect(%Alert{effect: effect})
+  def effect_type_to_group_title(effect), do: Alert.human_effect(%Alert{effect: effect}) <> "s"
 end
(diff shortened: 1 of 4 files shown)


lund5000 (changes requested):
Lookin' good! The only thing that jumped out to me is making sure that the font-family, font-weight, and padding values for the CR and bus pills are consistent with other parts of the site. Once that's tightened up I think this is ready to merge and for a final review with everything else.

zovt:
Screen Shot 2022-06-17 at 10 39 16

@lund5000 I'll redeploy in a sec, but how's this look? Notably the font size inside the pills doesn't change between breakpoints (to match the pills on the stops page). Do we want it to adjust to match the rest of the content?

thecristen (approved):
A couple non-blocking comments, seems to carry out the changes requested in the ticket and then some. :)

PR #1324, opened 2022-06-16
Upcoming Public Meetings and Events section

i0sea:
#### Summary of changes
**Asana Ticket:** ["Upcoming Public Meetings and Events" section](https://app.asana.com/0/555089885850811/1202075661172618/f)

Add events section to homepage
Screen Shot 2022-06-16 at 12 30 01 PM
Screen Shot 2022-06-16 at 12 30 13 PM

### DIFF
--- a/apps/site/assets/css/_events.scss
+++ b/apps/site/assets/css/_events.scss
@@ -245,6 +245,13 @@ $mobile-control-height: 73px;
     line-height: 21px;
     padding-bottom: 0;
   }
+
+  a {
+    color: inherit;
+    &:hover {
+      color: $brand-primary;
+    }
+  }
 }
 
 .m-event__status-message {
--- a/apps/site/assets/css/_homepage.scss
+++ b/apps/site/assets/css/_homepage.scss
@@ -294,18 +294,9 @@
   }
 
   .list-group-item {
-    border-top: unset;
-    border-top: solid $gray-lightest;
-    border-width: 1px 0 0 0;
-  }
-
-  .m-event__title {
-    a {
-      color: $black;
-      &:hover {
-        color: $brand-primary;
-      }
-    }
+    border-left: 0;
+    border-right: 0;
+    border-bottom: 0;
   }
 
   .m-event__date-circle {


amaisano (approved):
Looks great from a CMS API perspective. Good work!

apps/cms/lib/repo.ex
```diff
@@ -354,4 +354,21 @@ defmodule CMS.Repo do
         end
     end
   end
+
+  @doc "Get the next n events"
+  def next_n_event_teasers(n \\ []) do
```
thecristen, line 359: Since this function seems to expect a number I don't think it makes sense to have a default value of `[]`.
That said, since it's only called once, with a value for n, maybe we don't need a default value at all!

apps/site/lib/site_web/controllers/page_controller.ex
```diff
@@ -36,6 +36,7 @@ defmodule SiteWeb.PageController do
     |> async_assign_default(:promoted_items, fn -> promoted end)
     |> async_assign_default(:whats_happening_items, fn -> remainder end)
     |> async_assign_default(:alerts, fn -> Alerts.Repo.all(conn.assigns.date_time) end)
+    |> async_assign_default(:event_teasers, fn -> CMS.Repo.next_n_event_teasers(3) end, [])
```
thecristen, line 39: The acceptance criteria in the ticket mentions showing 6 events on larger screens. We should probably bump this to 6 and then reduce the events visible on smaller screens

apps/site/assets/css/_page.scss
```diff
@@ -11,6 +11,27 @@
...
+
+  .list-group-item {
+    border-top: unset;
+    border-top: solid $gray-lightest;
+    border-width: 1px 0 0 0;
+  }
+
+  .m-event__title {
+    a {
+      color: $black;
```
thecristen, line 23: I think `unset` or `inherent` might be a better value here, as it will take on whatever text color is being used in the parent elements (not strictly committing us to `$black`)

apps/site/assets/css/_page.scss
```diff
@@ -11,6 +11,27 @@
...
+      }
+    }
+  }
+
+  .m-event__date-circle {
+    @include media-breakpoint-down(sm) {
+      display: none;
+    }
+  }
 }
```
thecristen, line 35: I think we might need to add an extra rule to adjust the padding on `.m-event` on smaller breakpoints, currently it shows a bit too narrow:

image

apps/site/assets/css/_page.scss
```diff
@@ -11,6 +11,27 @@
       border-bottom-width: 0;
     }
   }
+
+  .list-group-item {
+    border-top: unset;
+    border-top: solid $gray-lightest;
+    border-width: 1px 0 0 0;
```
thecristen, line 18:
```
    border-left: 0;
    border-right: 0;
    border-bottom: 0;
```

is more readable IMO

apps/site/lib/site_web/templates/page/_homepage-events.html.eex
```diff
@@ -0,0 +1,19 @@
+<div class="container page-section">
+  <div class="c-standard-fares__heading">
```
thecristen, line 2: Now that we're reusing this class, we should probably give it a less specific name

apps/site/assets/css/_page.scss
```diff
@@ -11,6 +11,27 @@
...
+  }
+
+  .m-event__title {
+    a {
+      color: $black;
+      &:hover {
+        color: $brand-primary;
+      }
+    }
```
thecristen, line 27: There's a _tiny_ note on the mockup to update the original styling for these links to match this -- so you can move this `a {}` declaration under the `.m-event__title` in line 238 of `_events.scss`

thecristen (changes requested):
Looking good generally! The main change I want to request is showing 6 events on the homepage for larger screens.

apps/site/lib/site_web/templates/page/_homepage-events.html.eex
thecristen: I think I'll defer this since there'll be some other sections similar to this that will need to be streamlined...

zovt (approved):
📆

PR #1333, opened 2022-06-28
feat(ci): update Algolia index periodically

thecristen:
#### Summary of changes
**Asana Ticket:** [Update Algolia for new Franklin/Foxboro name](https://app.asana.com/0/385363666817452/1202421859282539/f)

The ticket implied we might want to automate this... so here's my first attempt at that.
I've gone ahead and created a new write API key in Algolia for usage with GitHub Actions, and added that key and the app ID to the repository secrets.

### DIFF
--- a/.github/workflows/algolia-update.yml
+++ b/.github/workflows/algolia-update.yml
@@ -1,0 +1,17 @@
+name: Update Algolia index
+
+on:
+  workflow_dispatch:
+  schedule: 
+    - cron: "0 3 * * *" # every day, 3am
+
+jobs:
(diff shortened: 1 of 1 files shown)


zovt (approved):
This makes sense to me. I'm not super familiar with github actions, but they run this for us periodically automagically or do we need to do any more setup?

thecristen:
@zovt re: "but they run this for us periodically automagically" - yep, that's exactly what the "schedule" bit does! Let's merge it in and see if it runs today!

PR #1335, opened 2022-06-29
Update Fare Calculator Copy

zovt:
#### Summary of changes
**Asana Ticket:** [Update Fare Calculator copy](https://app.asana.com/0/555089885850811/1202460348017141/f)

before
after


zovt:
hmmmm... my reading is that we should replace only the text in the "Fare Calculator" section 🤔 I kept the old one so that the note under the "Base Fare Estimate" header would stay the same

thecristen (commented):
Yeah I think you're right. 👍🏼 Sounds good!

thecristen (approved):

PR #1337, opened 2022-06-29
Include reduced fare price for CR/ferry trips

thecristen:
#### Summary of changes
**Asana Ticket:** [Include reduced fare price for CR/ferry trips](https://app.asana.com/0/555089885850811/1202313560724926/f)

The $30 monthly reduced fare pass is now valid for Commuter rail in zone 1A and for the inner harbor ferry.


apps/site/test/site/content_rewriters/liquid_objects/fare_test.exs
```diff
@@ -122,6 +122,20 @@ defmodule Site.ContentRewriters.LiquidObjects.FareTest do
...
+               mode: :ferry,
+               reduced: :any,
+               duration: :month
+             ]
+             |> Repo.all()
+             |> fare_result(:ferry) == "$30.00"
+
+      assert fare_request("ferry_inner_harbor:month:reduced") == {:ok, "$30.00"}
+      refute fare_request("ferry_cross_harbor:month:reduced") == {:ok, "$30.00"}
+      assert fare_request("ferry:month:reduced") == {:ok, "$30.00"}
```
thecristen, line 136: I wonder if this is reasonable output for `ferry:month:reduced` even though it's only valid on a certain type of ferry 🤔 cc @amaisano

amaisano: I see what you're saying. The alliterative would be to have either no match (_require_ a ferry_sub_type) or require additional context by the content team for the surrounding text. I would bring it up with Jess or Katrina or Avery and see what they think.

Other alternative ideas besides my suggestions above?

zovt (approved):

PR #1339, opened 2022-07-01
Add CR Monthly reduced fares

bfauble:
#### Summary of changes
**Asana Ticket:** [Add CR reduced fare monthly pass product](https://app.asana.com/0/555089885850811/1202341935986377/f)

add monthly reduced price logic for commuter rail products. mticket and normal passes.


zovt (approved):
this seems reasonable to me, do we still need the CSV or can we delete it?

bfauble:
> this seems reasonable to me, do we still need the CSV or can we delete it?

must have gotten added back in w/ the merge commit. I had originally deleted it. Will push up another commit w/ it removed.

apps/fares/lib/fare_info.ex
```diff
@@ -5,18 +5,239 @@ defmodule Fares.FareInfo do
 
   alias Fares.Fare
 
+  @fare_data [
+    %{
+      mode: :commuter,
+      zone: "zone_1a",
```
thecristen, line 11: **non-blocking, slightly nitpick:** Since these aren't coming from the CSV I think we could name these zones in a better way. Maybe in the format the `Zones.Zone` functions expect? So this could be "1A"

apps/fares/lib/fare_info.ex
```diff
@@ -5,18 +5,239 @@ defmodule Fares.FareInfo do
...
+      single_trip: "2.40",
```
thecristen, line 12: We _could_ now format these as cents from the get-go...

thecristen (changes requested):
Thanks for this and getting rid of the CSV file no one likes. :P 
Sorry CI is not working on this PR, but I'm seeing a few failing tests when I run it locally - can you look into these?

1) test mapper/1 maps the fares for a zone into one-way, round trip, monthly, mticket, and weekend prices (`Fares.FareInfoTest`)
     `apps/fares/test/fare_info_test.exs:28`
2) test fare_info/0 no duplicate fares are present (`Fares.FareInfoTes`t)
    ` apps/fares/test/fare_info_test.exs:13`
3) test reduced_pass accepts a Trip ID (`Fares.MonthTest`)
    ` apps/fares/test/month_test.exs:43`

thecristen (approved):
Sounds good! 🤑

PR #1340, opened 2022-07-01
Transfer tweaks

i0sea:
#### Summary of changes
**Asana Ticket:** [Remove transfer logic from fare calculation](https://app.asana.com/0/555089885850811/1202460348017145/f)

Remove transfer logic from fares calculation and just use default fare amounts.


apps/site/lib/site_web/views/trip_compare_view.ex
```diff
@@ -532,20 +532,8 @@ defmodule SiteWeb.TripCompareView do
...
-        two_legs = transit_legs |> Enum.slice(leg_index - 1, 2)
-        three_legs = transit_legs |> Enum.slice(leg_index - 2, 3)
-        # If this is part of a free transfer, don't add fare
-        cond do
-          Transfer.is_maybe_transfer?(three_legs) -> acc
-          Transfer.is_maybe_transfer?(two_legs) -> acc
-          true -> acc + (leg |> Fares.get_fare_by_type(fare_type) |> fare_cents())
-        end
-      end
+    |> Enum.reduce(0, fn {leg, _leg_index}, acc ->
```
thecristen, line 535: Nice! 
Since we're not using the index value, we can actually remove the previous step (`Stream.with_index()`), which is what was adding the index in the first place. Removing `Stream.with_index()` would need to alter this from `fn {leg, _leg_index}, acc ->` to `fn leg, acc ->`.

apps/site/lib/site_web/views/trip_plan_view.ex
```diff
@@ -548,20 +548,8 @@ defmodule SiteWeb.TripPlanView do
...
-        three_legs = transit_legs |> Enum.slice(leg_index - 2, 3)
-        # If this is part of a free transfer, don't add fare
-        cond do
-          Transfer.is_maybe_transfer?(three_legs) -> acc
-          Transfer.is_maybe_transfer?(two_legs) -> acc
-          true -> acc + (leg |> Fares.get_fare_by_type(fare_type) |> fare_cents())
-        end
-      end
+    |> Enum.reduce(0, fn {leg, _leg_index}, acc ->
+      acc + (leg |> Fares.get_fare_by_type(fare_type) |> fare_cents())
```
thecristen, line 552: Thanks for handling both views 👍🏼

apps/site/test/site_web/views/trip_plan_view_test.exs
```diff
@@ -840,6 +840,95 @@ closest arrival to 12:00 AM, Thursday, January 1st."
...
+          description: "BUS",
+          from: %TripPlan.NamedPosition{
+            latitude: 42.362804,
+            longitude: -71.099509,
+            name: "Massachusetts Ave @ Sidney St",
+            stop_id: "73"
+          },
+          long_name: "Harvard Square - Dudley Station",
+          mode: %TripPlan.TransitDetail{
+            fares: bus_fares,
```
thecristen, line 882: I think this `bus_fares` is triggering a compilation error, please check this.

thecristen (changes requested):
Thanks!! Just requesting a small change to fix that test and I think we should be good. 👍🏼 💸

thecristen (approved):
🎉

PR #1341, opened 2022-07-06
cleanup: remove unused plugs

thecristen:
#### Summary of changes
**Asana Ticket:** [Document the State of Predictions in Dotcom](https://app.asana.com/0/385363666817452/1202432733748699/f)

[The documentation is over on Notion](https://www.notion.so/mbta-downtown-crossing/Predictions-WIP-4a6f64cb1fa04b84a50813c8decd8a14), go review and/or comment on that. :) 

To be honest I didn't expect to open a PR for this but I found a few places where we were including plugs but no longer using the associated data. So here those are!

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/core.ex
+++ b/apps/site/lib/site_web/controllers/schedule/core.ex
@@ -11,8 +11,6 @@ defmodule SiteWeb.ScheduleController.Core do
   plug(SiteWeb.ScheduleController.ExcludedStops)
   plug(SiteWeb.ScheduleController.PreSelect)
   plug(SiteWeb.ScheduleController.VehicleLocations)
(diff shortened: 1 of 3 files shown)


zovt (commented):
Are these plugs still being used elsewhere?

thecristen:
@zovt sorta? but not really as a traditional plug. 

* The `SiteWeb.ScheduleController.Predictions` and `VehicleTooltips` `init` and `call` methods are called in a function within `SiteWeb.ScheduleController.LineApi`. 
* Some other `SiteWeb.ScheduleController.Predictions` functions are called in `SiteWeb.ScheduleController.Green`

zovt (approved):
I see, interesting. Well, if the underlying code can't be removed then this makes sense to me 👍

PR #1346, opened 2022-07-12
Homepage Alert Icon fixes

bfauble:
#### Summary of changes
**Asana Ticket:** [Recently Visited Routes alert tooltip issues](https://app.asana.com/0/555089885850811/1202570650180355/f)

- [x] Screen reader users can't see alert tooltip
- [x] Keyboard users can't access/see alert tooltip 
- [x] Should be able to tab to alert icon and open tooltip when it gets focus, then close tooltip when tabbing away and focus is lost
- [x] Recently visited schedules buttons with alert tooltip cannot be closed and is not contained by landmarks (resolved by removing old tooltips when navigating back to page)
image

image

### DIFF
--- a/apps/site/lib/site/components/icons/svg_icon/component.html.eex
+++ b/apps/site/lib/site/components/icons/svg_icon/component.html.eex
@@ -3,7 +3,7 @@
   hyphenated_class = "icon-#{CSSHelpers.atom_to_class(icon)}"
   data_toggle = if args.show_tooltip?, do: "data-toggle=tooltip", else: ""
   title = if args.show_tooltip?, do: icon_title(icon), else: ""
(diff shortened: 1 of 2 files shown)


apps/site/lib/site/components/icons/svg_icon/component.html.eex
```diff
@@ -3,7 +3,21 @@
...
+  xmlns="http://www.w3.org/2000/svg"
+  xmlns:xlink="http://www.w3.org/1999/xlink"
+  version="1.1"
+  width="30"
+  height="30"
+  viewbox="<%= viewbox(icon) %>"
+  preserveAspectRatio="xMidYMid meet"
+  role="img"
+  tabindex="0"
+  aria-label="<%= title %>"
```
bfauble, line 16: added role, tabindex, and aria-label props

thecristen (approved):
I'm not 100% sure about whether the usage of `role="img"` for this interactive element is quite proper, but let's give it a try! Works great.

PR #1351, opened 2022-07-15
Section header should be hidden when there's no content available

i0sea:
#### Summary of changes
**Asana Ticket:** [Section header should be hidden when there's no content available](https://app.asana.com/0/555089885850811/1202532801799502/f)

Hide what's happening header if no content. Note there is already a check for events section
`<%= unless Enum.empty?(@event_teasers) do %>`

### DIFF
--- a/apps/site/lib/site_web/templates/page/_whats_happening.html.eex
+++ b/apps/site/lib/site_web/templates/page/_whats_happening.html.eex
@@ -5,7 +5,7 @@
       <h2 class="sr-only">Featured MBTA Updates and Projects</h2>
     <% end %>
     <div class="container page-section">
-      <%= unless @promoted == false do %>
+      <%= unless Enum.empty?(@whats_happening_items) do %>
         <h2>What's Happening at the MBTA</h2>
       <% end %>
       <div id="whats-happening-<%= tag %>" class="row m-whats-happening__row m-whats-happening__row--<%= tag %>">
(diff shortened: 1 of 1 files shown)


apps/site/lib/site_web/templates/page/_whats_happening.html.eex
```diff
@@ -5,7 +5,7 @@
       <h2 class="sr-only">Featured MBTA Updates and Projects</h2>
     <% end %>
     <div class="container page-section">
-      <%= unless @promoted do %>
+      <%= unless @promoted == false do %>
```
bfauble, line 8: this is kind of confusing and I think is the opposite of the intended effect. There's 2 versions of this template being rendered, one for promoted and one for not promoted. This header should be present for the not promoted so `if @promoted == false` or `unless @promoted` like it already was. The `<%= if @whats_happening_items %>` at the top should cover the rendering of the section entirely if there is no content. 

From what I can tell the ticket (and possibly dev green) are outdated compared to the current state of the code.

thecristen: I agree that this is confusing; I'd expect some logic checking the `@whats_happening_items` list being rendered below.

thecristen (changes requested):
I don't think we need to change the logic around promoted content. Let's update this to check if `@whats_happening_items` has content to render.

i0sea:
Should this be closed? As bfauble mentioned the `<%= if @whats_happening_items do %>` at the top should take care of the header rendering if there are no items.

thecristen:
@i0sea @bfauble Sadly `if @whats_happening_items` will return true even for an empty list `[]`. So we should refine this to check not just for its existence but also if it's empty. :)

thecristen (approved):
✅  it's happening!

PR #1356, opened 2022-07-19
Remove Bootstrap dependencies from header JS + more

thecristen:
#### Summary of changes
**Asana Ticket:** [Remove Bootstrap dependencies from header JS](https://app.asana.com/0/555089885850811/1202533077749642/f) and [Stop bundling and transforming JS in favor of copying unmodified source](https://app.asana.com/0/555089885850811/1202533077749641/f)

Overlaps with https://github.com/mbta/dotcomchrome/pull/8
Running `mix export.header_footer` while checking out this PR and running the server on `localhost:4001` should produce the output currently in https://github.com/mbta/dotcomchrome/pull/8 

### Main changes

- Rewrote the navigation to not use Bootstrap's collapse plugin
  - Desktop: updated some classes, had to write a bit more JS
  - Mobile: had to replace our Bootstrap-dependent `_accordion_ui.html.eex` template
- Specific to the extract of the header and footer that populates https://www.github.com/mbta/dotcomchrome :
  - Simplified the Webpack configuration, stops minifying JS
  - Adds a `setupDotcomChrome` function to the `window` object that others can use to initialize the navigation JS to the document HTML element or ShadowRoot of choosing.

**Questions**
Should we rewrite `_accordion_ui.html.eex` itself rather than creating a Bootstrap-free copy?

**Area of improvement**
The opening and closing actions on the navigation are the simplest implementation, without the smoother animated transitions included in Bootstrap. Can try to tackle that later or now if we see fit.

**Surprise bugfixes no one asked for:**
- Hidden navbar search input on tablet breakpoint was still focusable - it is no longer!
- Footer social media icons in the extracted footer were positioned weirdly - they look properly placed now


zovt (approved):
This all looks reasonable to me 👍 

> Should we rewrite _accordion_ui.html.eex itself rather than creating a Bootstrap-free copy?

Is there a reason to keep both around? I feel like ideally we don't want to maintain two separate copies of this code

thecristen:
Right now this new accordion doesn't completely match the old accordion in features -- for example, with the old accordion you can open multiple sections at once.

I'll make a ticket to facilitate the wholesale replacement -- I agree we ultimately want just one accordion rather than two!

thecristen:
Lots of changes since initial review in response to lots of feedback:

* removes React JS from the exported JS bundle. I never got tree-shaking to work effectively so ended up having to take a slightly drastic measure of rewriting a few helper files to isolate the unused React-based functions - e.g. React-based functions in `ts/helpers/keyboard-events.ts` got extracted into `ts/helpers/keyboard-events-react.ts`, so that our header can import the needed function from `ts/helpers/keyboard-events.ts` without React ending up in the bundle. 🤷🏼‍♀️ 
* to avoid naming conflicts, renames all CSS classes in the exported CSS (via webpack) and exported HTML (via the custom `mix export.header_footer` task). This involved a bit more refactoring of the header JavaScript, where I stopped querying the DOM by class selectors and started using more data attributes. I'm hoping I caught everything...
* removes unused CSS using PurgeCSS webpack plugin (this removed so much!)

I'll probably have to fix a few tests/formatting - will add another commit for that here

PR #1359, opened 2022-07-21
Move social icons, add language selector

i0sea:
#### Summary of changes
**Asana Ticket:** [Move social icons, add language selector](https://app.asana.com/0/555089885850811/1201244628068091/f)

Move social links and add language selector to footer


apps/site/lib/site_web/templates/layout/_footer.html.eex
```diff
@@ -78,30 +78,35 @@
...
         <%= unless assigns[:exclude_google_translate] do %>
           <%= render "_google_translate_plugin.html", conn: @conn %>
         <% end %>
       </li>
     </ul>
+
+    <div class="lower-footer">
+        <div class="m-menu__language"></div>
+      <div class="social-icons-container">
+          <a class="social-icon" href="https://www.facebook.com/TheMBTA/">
```
thecristen, line 90: Can you please put back the `target="_blank"` attribute on these links?

thecristen (changes requested):
A few accessibility-related comments, but I think adding the hover/active/focus states might take a bit more work. Looking good! 😄

apps/site/lib/site_web/templates/layout/_footer.html.eex
```diff
@@ -78,30 +78,35 @@
...
+            <i aria-label="linkedin" class="fa-brands fa-linkedin fa-lg social-icon"></i>
+          </a>
+          <a href="https://www.youtube.com/user/mbtagm/" target="_blank">
+            <i aria-label="youtube" class="fa-brands fa-youtube fa-lg social-icon"></i>
+          </a>
+          <a href="https://twitter.com/MBTA" target="_blank">
+            <i aria-label="twitter" class="fa-brands fa-twitter fa-lg social-icon"></i>
+          </a>
+          <a href="https://www.tiktok.com/@thembta" target="_blank">
+            <i aria-label="tiktok" class="fa-brands fa-tiktok fa-lg social-icon"></i>
```
thecristen, line 106: Thanks for the `aria-label`! One thing - the old labels were formatted `MBTA [social platform name]` and I'd like these to do the same -- "MBTA tiktok" and "MBTA instagram" and so on will make much more sense to users than the "tiktok" label alone (does it go to www.tiktok.com?). Hope this makes sense.

apps/site/assets/css/_footer.scss
```diff
@@ -90,6 +90,50 @@
...
+  padding-left: 20px;
+  color: $brand-primary-light-contrast;
+  &:hover {
+    color: $brand-primary-light;
+  }
+  &:active {
+    color: $brand-primary;
+  }
+  &:focus {
+    color: $brand-primary-light;
```
thecristen, line 137: This needs the yellow outline too.

thecristen (changes requested):
A few tiny items:

thecristen (approved):
Let's go! 🚀

PR #1360, opened 2022-07-25
Reorganize group titles and links

thecristen:
#### Summary of changes
**Asana Ticket:** [Reorganize group titles and links](https://app.asana.com/0/555089885850811/1201244628068093/f)

Reorganized the footer content, removing some old CSS too. Note this removes the social icons... so let's merge #1359 first :) 
Also refined the Cypress tests, hoping they avoid race conditions and fail less unpredictably now!


thecristen:
Just added a commit to improve the Cypress testing experience a bit by logging the result to the terminal, the output would look something like this. This way we could learn about the error without having to run `cypress open` ourselves!

image

apps/site/lib/site_web/templates/layout/_footer.html.eex
```diff
@@ -1,111 +1,63 @@
...
+      <div id="footer-know-us" class="m-footer__group-title">Get to Know Us</div>
+      <%= link "Leadership", to: cms_static_page_path(@conn, "/leadership"), data: [scroll: true] %>
+      <%= link "History", to: cms_static_page_path(@conn, "/history"), data: [scroll: true] %>
+      <%= link "Financials", to: cms_static_page_path(@conn, "/financials"), data: [scroll: true]  %>
+      <%= link "Public Meetings", to: cms_static_page_path(@conn, "/events"), data: [scroll: true]  %>
+      <%= link "Press Releases", to: cms_static_page_path(@conn, "/news"), data: [scroll: true]  %>
+      <%= link "Policies & Civil Rights", to: cms_static_page_path(@conn, "/policies"), data: [scroll: true]  %>
+      <%= link "Safety", to: cms_static_page_path(@conn, "/safety") %>
+      <%= link to: "http://www.mbtabackontrack.com/performance/index.html#/home", target: "_blank" do %>
+        Performance <%= content_tag(:span, fa("external-link")) %>
```
i0sea, line 22: Didn't see these icons in the mocks. Can this be removed along with the icon on  Request Public Records?

i0sea (changes requested):
Looks good, one small comment.

apps/site/lib/site_web/templates/layout/_footer.html.eex
thecristen: Ah, good catch! Will take those out.

thecristen:
Made the requested change, also added another couple commits to try to fix broken integration tests :eyes:

i0sea (approved):
🚀

PR #1362, opened 2022-07-26
fix: add custom Street View link for Brockton

thecristen:
#### Summary of changes
**Asana Ticket:** [GTFS | update street view for Brockton Station](https://app.asana.com/0/385363666817452/1202664751368463/f)

Here's where the new link goes, which I felt was the closest I could get to an accessible station entrance:

image


i0sea (approved):
🌐

PR #1363, opened 2022-07-27
Homepage | Show six events at large breakpoints

i0sea:
#### Summary of changes
**Asana Ticket:** [Homepage | Show six events at large breakpoints](https://app.asana.com/0/385363666817452/1202641648100462/f)

Show next six events. Updated event ended check to check date rather than just the day.
Screen Shot 2022-07-27 at 12 09 36 PM


zovt (approved):

thecristen (approved):
🫣 It looks like this causes a bit of a regression on the /events page, where previous events this month aren't hidden:

image

thecristen (changes requested):
See prior comment

thecristen (approved):

PR #1366, opened 2022-08-01
adjust api timeout duration to match api side

bfauble:
#### Summary of changes
**Asana Ticket:** n/a

Adjusting timeout for api requests to match api side to avoid timeouts especially during service day switch over.

https://github.com/mbta/api/blob/master/apps/api_web/lib/api_web/api_controller_helpers.ex#L19


i0sea (approved):

thecristen:
Thanks @bfauble!!

PR #1369, opened 2022-08-11
Update "Use my location" label in header search and schedule hubs

zovt:
#### Summary of changes
**Asana Ticket:** [Update "Use my location" label](https://app.asana.com/0/385363666817452/1202299919599024/f)

Can't get screenshots, but this updates the "Use my location" text in the header search and on schedule hubs (e.g. `/schedules`) to be "Use my location to find transit near me". Any other locations (including the trip planner widget on the schedule hubs) retain the original text.

To do this, I've parameterized the templates we use for the results.

### DIFF
--- a/apps/site/assets/js/algolia-autocomplete.js
+++ b/apps/site/assets/js/algolia-autocomplete.js
@@ -6,7 +6,16 @@ import * as QueryHelpers from "../ts/helpers/query";
 /* eslint-disable class-methods-use-this */
 
 export default class AlgoliaAutocomplete {
-  constructor({ id, selectors, indices, parent, containerEl = null }) {
+  #templates;
+
+  constructor({
+    id,
+    selectors,
+    indices,
+    parent,
+    containerEl = null,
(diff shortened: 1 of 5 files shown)


i0sea (approved):

apps/site/assets/js/algolia-autocomplete.js
```diff
@@ -6,7 +6,16 @@ import * as QueryHelpers from "../ts/helpers/query";
 /* eslint-disable class-methods-use-this */
 
 export default class AlgoliaAutocomplete {
-  constructor({ id, selectors, indices, parent, containerEl = null }) {
+  #templates;
```
thecristen, line 9: What does this mean?

thecristen: Ah, declaring a private field... TIL!

zovt: yep, relatively new but useful!

PR #1371, opened 2022-08-12
Fetch Schedule PDFs from new endpoint

zovt:
#### Summary of changes
**Asana Ticket:** [Elixir integration](https://app.asana.com/0/385363666817452/1202623240995537/f)

Mostly opening this on behalf of @thecristen since all of the actual work has been done by her :) I just confirmed things worked locally.

One question that came up: From the ticket, it sounds like we are transtioning bus routes fully to this new endpoint ~soon, and other modes later on. The code is currently hitting _both_ endpoints and combining the resulting lists. @amaisano Is there any worry that we'll run into a situation where we have PDFs coming from both endpoints for modes we don't want to support yet? Or can we make the assumption that if the schedules are on the prod CMS we do indeed want to show them?

Thanks!

### DIFF
--- a/apps/cms/priv/cms/schedule-pdfs.json
+++ b/apps/cms/priv/cms/schedule-pdfs.json
@@ -1,0 +1,118 @@
+[
+  {
+    "mid": [
+      {
+        "value": 3811
+      }
+    ],
+    "uuid": [
+      {
+        "value": "96d1dadc-9bbc-436e-becf-b0bdf817df3e"
+      }
+    ],
+    "vid": [
+      {
+        "value": 6129
+      }
+    ],
+    "langcode": [
+      {
+        "value": "en"
+      }
+    ],
+    "bundle": [
+      {
+        "target_id": "schedule",
+        "target_type": "media_type",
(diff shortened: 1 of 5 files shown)


apps/cms/priv/cms/schedule-pdfs.json
```diff
@@ -0,0 +1,118 @@
+[
```
amaisano, line 1: We have an updated API payload for this. You can view it here: https://test-mbta.pantheonsite.io/cms/schedules/350?_format=json

amaisano (approved):
Looks ok from my POV.

amaisano:
@zovt Avery has told me that only bus schedules are going up this way for the next rating, no other modes. So I believe your assumption is correct - if it's on prod - regardless of mode - it's valid to show.

amaisano:
FYI we need to wait for the prod deployment of the CMS before this is ready on live. You'll get a 404 response until then.

amaisano:
FYI CMS is fully deployed.

thecristen (approved):
📝

PR #1372, opened 2022-08-15
Limit "Recently Visited Schedules" margin removal to homepage tab

zovt:
#### Summary of changes
**Asana Ticket:** [Recently visited schedules heading margin ](https://app.asana.com/0/385363666817452/1202644949290864/f)

### DIFF
--- a/apps/site/assets/css/_recently-visited.scss
+++ b/apps/site/assets/css/_recently-visited.scss
@@ -1,8 +1,4 @@
 .c-recently-visited {
-  h2 {
-    margin-top: 0;
-  }
-
   .c-grid-button {
     flex-basis: 25%;
     margin-right: 16px;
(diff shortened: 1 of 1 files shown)


apps/site/assets/css/_recently-visited.scss
```diff
@@ -1,8 +1,4 @@
 .c-recently-visited {
-  h2 {
-    margin-top: 0;
```
thecristen, line 3: This is used on a few other pages too - are we OK with changing those?

zovt: I think this was originally introduced as part of the homepage design updates, so this change should effectively revert any other pages whilst keeping the homepage updates intact (which I believe was the point of the ticket)

thecristen: oh ha! Awesome then

thecristen (approved):

PR #1373, opened 2022-08-15
Schedule pdf labelling

zovt:
#### Summary of changes
**Asana Ticket:** [Elixir integration](https://app.asana.com/0/385363666817452/1202623240995537/f)

### DIFF
--- a/apps/site/lib/site_web/views/schedule_view.ex
+++ b/apps/site/lib/site_web/views/schedule_view.ex
@@ -184,7 +184,7 @@ defmodule SiteWeb.ScheduleView do
     pdf_name =
       cond do
         RoutePdf.custom?(pdf) -> pdf.link_text_override
-        pdf.name != nil -> pdf.name
+        pdf.name != nil && pdf.name != "" -> pdf.name
         true -> [pretty_route_name(route), " schedule"]
       end
 
--- a/apps/site/lib/site_web/views/schedule_view.ex
+++ b/apps/site/lib/site_web/views/schedule_view.ex
@@ -182,10 +182,10 @@ defmodule SiteWeb.ScheduleView do
       end
 
     pdf_name =
-      if RoutePdf.custom?(pdf) do
-        pdf.link_text_override
-      else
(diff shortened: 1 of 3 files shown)


lund5000 (commented):
That's an improvement over the duplicate labels we'd get without this change (and what I asked for in Slack, so even if it wasn't I'd have no one to blame but myself).

One thing to point out, perhaps for this PR but perhaps for the next: the website does have some formatting for "current" vs. "upcoming" and effective date for files from the existing endpoint. Maybe that's easy enough to repurpose when working across the existing and new endpoints, too?

I'm also still OK with waiting to recreate something close to this once we have the real batch uploads in production.

Screen Shot 2022-08-15 at 4 34 46 PM

apps/site/lib/site_web/views/schedule_view.ex
```diff
@@ -182,10 +182,10 @@ defmodule SiteWeb.ScheduleView do
       end
 
     pdf_name =
-      if RoutePdf.custom?(pdf) do
-        pdf.link_text_override
-      else
-        [pretty_route_name(route), " schedule"]
+      cond do
+        RoutePdf.custom?(pdf) -> pdf.link_text_override
+        pdf.name != nil -> pdf.name
```
thecristen, line 187: Should we check if this isn't an empty string `""`? 🤔

zovt: Sounds like something that should be handled CMS-side but we can add an explicit check if it makes people feel better :)

zovt:
> One thing to point out, perhaps for this PR but perhaps for the next: the website does have some formatting for "current" vs. "upcoming" and effective date for files from the existing endpoint. Maybe that's easy enough to repurpose when working across the existing and new endpoints, too?

Not 100% sure what you're referring to here. The Upcoming and Current text should apply to new `name` text as well if that's what you're asking?

thecristen (approved):
Let's follow this up with nicer names soon! Thanks!

PR #1379, opened 2022-08-17
refactor(LineDiagram): extract useRealtime

thecristen:
Also changes `liveData` prop slightly to represent absent data as `undefined` rather than `null`.

no ticket, just a thing I tried to clean up while tinkering with the line diagram. This PR also adds conditional logic, so we can easily turn the line diagram's live data "off" by specifying `false` as the third argument in `useRealtime()`. This is hardcoded to true right now and thus doesn't change anything about the current line diagram state.

### DIFF
--- a/apps/site/assets/js/app.js
+++ b/apps/site/assets/js/app.js
@@ -53,7 +53,7 @@ if (window.sentry) {
     dsn: window.sentry.dsn,
     environment: window.sentry.environment,
     autoSessionTracking: false,
-    sampleRate: 0.10, // error sampling - might increase later
+    sampleRate: 0.1, // error sampling - might increase later
     initialScope: {
       tags: { "dotcom.application": "frontend" }
     },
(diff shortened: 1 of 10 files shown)


apps/site/assets/ts/hooks/useRealtime.ts
```diff
@@ -0,0 +1,28 @@
...
+  fetch(url).then(response => response.json());
+
+const useRealtime = (
+  route: Route,
+  directionId: DirectionId,
+  enabled: boolean = true
+): LiveDataByStop | undefined => {
+  const liveUrl = `/schedules/line_api/realtime?id=${route.id}&direction_id=${directionId}`;
+  const { data } = useSWR<LiveDataByStop>(
+    enabled && route.type !== 4 ? liveUrl : null,
```
thecristen, line 19: Will return `undefined` by passing in `null` in place of the URL.

apps/site/assets/js/app.js
```diff
@@ -53,7 +53,7 @@ if (window.sentry) {
     dsn: window.sentry.dsn,
     environment: window.sentry.environment,
     autoSessionTracking: false,
-    sampleRate: 0.10, // error sampling - might increase later
+    sampleRate: 0.1, // error sampling - might increase later
```
thecristen, line 56: this got caught in formatting!

thecristen:
Whoops, realized I ended up covering this in #1384

PR #1380, opened 2022-08-17
refactor: use URLSearchParams Web API

thecristen:
no ticket!

Since we're no longer supporting IE11, we are free to use [the URLSearchParams API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to handle parsing/formatting/querying the URL query string. We had logic for this in a few different places, this is an effort to streamline that.

I did need to change a few test strings, as the formatted query parameters from `URLSearchParams` are slightly different than the one from `window.encodeURIComponent` (notably the `%20` becomes simply `+` - is this right? [This StackOverflow question has some interesting answers](https://stackoverflow.com/questions/1211229/in-a-url-should-spaces-be-encoded-using-20-or) around that. I've decided this is fine.)


apps/site/assets/ts/schedule/components/ScheduleLoader.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { ReactElement } from "react";
 import { connect } from "react-redux";
-import { useQueryParams, StringParam } from "use-query-params";
```
thecristen, line 3: I initially tried updating `use-query-params`, but its updated version required wrapping the relevant React components in a `<Provider>` type component, which I was wary to do. Also our use case is very simple, so it felt like we could get away without the overhead.

apps/site/assets/ts/helpers/use-params.ts
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  window?.history.replaceState(
+    {},
+    "",
+    `${window.location.pathname}?${params.toString()}`
+  );
+};
+
+export default {
```
arkadyan, line 29:
```suggestion
export {
```
suggestion: I don't think you want `default` for named exports.

apps/site/assets/ts/helpers/__tests__/use-params-test.ts
```diff
@@ -0,0 +1,51 @@
+import { getParam, updateParams } from "./../use-params";
+
+const oldWindowLocation = window.location;
+const mockLocation = () => {
+  const location = {
+    ...window.location,
+    search: "?best_train=Red&good_bus=1"
```
arkadyan, line 7: nitpick: 2 nits to pick… (j/k 😉)

arkadyan (approved):
👍 This is cool, `URLSearchParams` is new to me but it's nice. A few suggestions, but they are all minor.

thecristen:
gonna open a new PR for this.

PR #1381, opened 2022-08-17
feat: add plug for listing active line suspensions

thecristen:
no ticket, another intermediary step on the way to an adjusted line diagram.

This heavily follows our recent example for handling the GLX opening, by [adding a plug that can interpret the current date to determine relevance](https://github.com/mbta/dotcom/pull/1219/files), and [adding a DOM element to the application template](https://github.com/mbta/dotcom/pull/1223/files#diff-6f96a0b2dce5094d2afb1aaa067dc5181bec30ad7a76013d19989fed805f6702) that we can read from in other components. 

After this we can read the suspension-related values in our React components and create logic around it.

This uses the `date_time` URL parameter to determine "now"-ness - e.g. there'll be no extra DOM element now, but if you append `?date_time=2022-08-20T00:00:00` to the URL (or any other valid ISO-formatted datetime within range), you should see the extra element appear. I'll deploy it to dev-blue.

### DIFF
--- a/apps/site/lib/site_web/plugs/line_suspensions.ex
+++ b/apps/site/lib/site_web/plugs/line_suspensions.ex
@@ -1,0 +1,46 @@
+defmodule SiteWeb.Plugs.LineSuspensions do
+  @moduledoc """
+    assigns line_suspensions map based on hard-coded values into @suspensions
+    Could potentially add more detail e.g. which stops are served by shuttle,
+    which alternative modes are available, etc.
+  """
+
+  @suspensions %{"Orange" => [~N[2022-08-19T21:00:00], ~N[2022-09-19T04:55:00]]}
+
+  @behaviour Plug
+  import Plug.Conn, only: [assign: 3]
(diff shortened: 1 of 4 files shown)


apps/site/lib/site_web/plugs/line_suspensions.ex
```diff
@@ -0,0 +1,46 @@
+defmodule SiteWeb.Plugs.LineSuspensions do
+  @moduledoc """
+    assigns line_suspensions map based on hard-coded values into @suspensions
+    Could potentially add more detail e.g. which stops are served by shuttle,
+    which alternative modes are available, etc.
+  """
+
+  @suspensions %{"Orange" => [~N[2022-08-19T21:00:00], ~N[2022-09-19T04:55:00]]}
```
thecristen, line 8: Open to other data structures here, but a benefit of this over, say, `Tuple` is that it's JSON-serializable without extra work. :P

PR #1383, opened 2022-08-18
fix(TripPlanner): Remove the header from the accordion on the trip pl…

kotva006:
…anner

#### Summary of changes
**Asana Ticket:** [Repeated heading 3 label for itinerary and content](https://app.asana.com/0/555089885850811/1202300116369146)

The `aria-role=header` attribute has been set as optional for the `Accordion` component.
This is to stop the screen read from reading each part of the trip with header title

Note: The role is optional, but it needs to be deliberately opted out of, as it defaults to setting the header given no information

---

### DIFF
--- a/apps/site/assets/js/app.js
+++ b/apps/site/assets/js/app.js
@@ -53,7 +53,7 @@ if (window.sentry) {
     dsn: window.sentry.dsn,
     environment: window.sentry.environment,
     autoSessionTracking: false,
-    sampleRate: 0.10, // error sampling - might increase later
+    sampleRate: 0.1, // error sampling - might increase later
     initialScope: {
       tags: { "dotcom.application": "frontend" }
     },
--- a/apps/site/assets/ts/components/Accordion.tsx
+++ b/apps/site/assets/ts/components/Accordion.tsx
@@ -15,7 +15,7 @@ interface Props {
 export const AccordionNoJS = ({
   id,
   children,
-  headingRole = true,
+  headingRole = true
 }: {
   id: string;
   children: ReactElement<HTMLElement>;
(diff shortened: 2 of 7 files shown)


i0sea (approved):

lund5000:
Testing with NVDA, this does prevent the accordion from being read when browsing through headings (as Ryan feared it might). That seems undesirable. 

Is it possible and advisable to set `heading="role"` and `aria-level="3"` on the `div.c-accordion-ui__heading` inside/instead of `div.panel`? Based on manual editing in dev tools, it seems to produce a more desirable result: the accordion expand/collapse button is still read as heading level 3 but the steps within are not.

apps/site/assets/js/app.js
```diff
@@ -53,7 +53,7 @@ if (window.sentry) {
     dsn: window.sentry.dsn,
     environment: window.sentry.environment,
     autoSessionTracking: false,
-    sampleRate: 0.10, // error sampling - might increase later
+    sampleRate: 0.1, // error sampling - might increase later
```
kotva006, line 56: Does this need to be `0.10`?

apps/site/assets/ts/components/Accordion.tsx
```diff
@@ -19,15 +19,13 @@ export const AccordionNoJS = ({
   children: ReactElement<HTMLElement>;
 }): ReactElement<HTMLElement> => (
   <div className="c-accordion-ui">
-    <div role="heading" aria-level={3}>
```
kotva006, line 22: I removed this div, because there doesn't appear to be a heading/title in the NoJS accordion

thecristen (approved):
:sparkles: I think this will do the trick! Thanks for this revision.

PR #1385, opened 2022-08-18
Calculate some CR legs as $0 during OL shutdown

i0sea:
#### Summary of changes
**Asana Ticket:** [Calculate some CR legs as $0 during OL shutdown](https://app.asana.com/0/555089885850811/1202802897881991/f)

If leg in itinerary is 1A,1, or 2 don't charge for that section.

### DIFF
--- a/apps/fares/lib/fare_info.ex
+++ b/apps/fares/lib/fare_info.ex
@@ -6,31 +6,62 @@ defmodule Fares.FareInfo do
   alias Fares.Fare
 
   now = DateTime.utc_now()
-  surge_start =  ~U[2022-08-11 04:55:00Z] # TEMP VALUE FOR TESTING UPDATE BEFORE  MERGE
-  surge_end =  ~U[2022-09-19 04:55:00Z]
+  # TEMP VALUE FOR TESTING UPDATE BEFORE  MERGE
+  surge_start = ~U[2022-08-11 04:55:00Z]
+  surge_end = ~U[2022-09-19 04:55:00Z]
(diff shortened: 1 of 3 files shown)


lund5000 (approved):
Looks like it's working as requested within those zones and charging appropriately for CR trips outside those zones. 👍  Before we launch we should make sure the start datetime is the equivalent of 9:00PM ET on Fri 8/19 and the end datetime is the equivalent of 2:00AM ET on Mon 9/19 (i.e., after the end of service the night before and before service begins that morning).

apps/fares/lib/fare_info.ex
```diff
@@ -5,28 +5,62 @@ defmodule Fares.FareInfo do
...
+  surge_end = ~U[2022-09-19 06:00:00Z]
+
   @fare_data [
     %{
       mode: :commuter,
       zone: "1A",
-      single_trip: "2.40",
-      single_trip_reduced: "1.10",
+      single_trip:
+        if DateTime.compare(now, surge_start) == :gt and DateTime.compare(now, surge_end) == :lt do
```
arkadyan, line 17: suggestion(non-blocking): Since this logic is used multiple times if you wanted to you could just call it once, up in line 11 or so, and assign it to a variable, say `in_surge_window?` or something. Then each of these checks could just be `if in_surge_window? do`.

arkadyan (approved):
👍 Looks good to me. I added a small suggestion, but it doesn't need to hold anything up.

PR #1388, opened 2022-08-19
fix: Fix Green Line diagram when using suspension data

arkadyan:
#### Summary of changes
No ticket, currently broken on the site.

Get line data in LineApi the way the Line plug does. I'm not completely
sure of the "why" behind this yet, and it is not properly tested, but it
seems to fix the Green Line line diagram.

Screen Shot 2022-08-19 at 11 41 58

Screen Shot 2022-08-19 at 11 42 15

---


arkadyan:
Note that this doesn't yet fixed the few remaining tests that are currently broken.

apps/site/test/site_web/controllers/schedule/line_test.exs
```diff
@@ -287,447 +282,6 @@ defmodule SiteWeb.ScheduleController.LineTest do
     end
   end
 
-  describe "build_stop_list/2 for Green Line" do
```
zovt, line 290: Are we just deleting these temporarily? If so and if we'd like to add them back in the future I'd recommend opening a ticket and adding a link to it here as a TODO so we don't forget these

zovt (approved):

apps/site/test/site_web/controllers/schedule/line_test.exs
thecristen: Or if it's temporary, we can add a `skip` tag here.

apps/site/lib/site_web/controllers/schedule/line/diagram_helpers.ex
```diff
@@ -138,6 +115,29 @@ defmodule SiteWeb.ScheduleController.Line.DiagramHelpers do
     end)
   end
 
+  @spec fix_green_connections([stop_with_bubble_info()]) :: [stop_with_bubble_info()]
```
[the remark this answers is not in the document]
arkadyan: Yes, that's correct.

apps/site/test/site_web/controllers/schedule/line_test.exs
arkadyan: These haven't actually been deleted, just moved to [diagram_helpers_test.exs](https://github.com/mbta/dotcom/pull/1388/files#diff-f5d4adc72b585b9cffdd568bb393d3890807039426729bd4b4c7c5382aaddd17R1012) (which matches the location of the implementation files.

zovt: Ah I see, GitHub decided that part of the diff should be collapsed so I missed them. Thanks!

apps/site/lib/site_web/controllers/schedule/line_api.ex
```diff
@@ -34,12 +34,7 @@ defmodule SiteWeb.ScheduleController.LineApi do
           |> assign(:direction_id, String.to_integer(direction_id))
           |> assign_alerts(filter_by_direction?: true)
 
-        line_data =
-          get_line_data(
-            route,
-            String.to_integer(direction_id),
-            conn.query_params["route_pattern"]
-          )
+        line_data = get_line_data(route, String.to_integer(direction_id))
```
thecristen, line 37: I _suspect_ it was the removal of `conn.query_params["route_pattern"]` here that caused this regression - at left is this PR - it's no longer displaying the extra stops for Inbound - Haymarket via Forest Hills. Variants on buses, oof.

image

It's maybe easiest to see on the 62/76 - https://dev-green.mbtace.com/schedules/627/line?schedule_direction%5Bdirection_id%5D=0&schedule_direction%5Bvariant%5D=627-7-0 One expects after selecting the "via Hanscom" variant to be able to see the Hanscom stop!

thecristen (commented):
Not sure how hard this would be to address, or whether it can be a known regression that we intend to fix later. cc @lund5000

arkadyan:
@thecristen Ugh, yeah. I can dig into dealing with that better, but I don't think it's going to be something small enough that it'll happen today…

PR #1389, opened 2022-08-19
fix: Use the secrets format to retrieve the token

arkadyan:
#### Summary of changes
No ticket

This is a shot in the dark, not sure if it will actually do anything to
fix the failing GitHub action or not, but found this syntax in some
[GitHub documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-2-calling-the-rest-api)
and it seemed like it wouldn't hurt to try it.

---

### DIFF
--- a/.github/workflows/deploy-pr.yml
+++ b/.github/workflows/deploy-pr.yml
@@ -43,19 +43,19 @@ jobs:
   cleanup:
     runs-on: ubuntu-latest
     steps:
-    - name: Unlabel other PRs (if needed)
-      if: ${{ env.DO_DEPLOY }}
-      run: |
-        labelled=$(curl -s \
-        "${{ github.event.repository.url }}/issues?state=all&labels=${{ env.DEPLOY_ENV }}" \
-        | jq -r '.[].url')
+      - name: Unlabel other PRs (if needed)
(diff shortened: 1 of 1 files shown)


thecristen:
AFAIK there's no way to test that without having this workflow code in the `master` branch, right? 🤔

arkadyan:
@thecristen I was hoping it might work to try deploying this branch at a time when GH action deploys of other branches were failing. But I have no idea if that would work (and maybe it's not even possible?)

thecristen:
I'm like 90% deploying this will use the old version of this work flow but it's totally worth a try!

arkadyan:
@thecristen Ah yeah, I suspect you are probably right. I'll leave it up to you if (and when) you want to try merging it and seeing if the next deployment goes any better. We can always back it out later.

thecristen:
I don't think we can assess if this PR helps or not because the failure happens beforehand - I think it's a formatting error around getting/passing the URLs -- see https://github.com/mbta/dotcom/runs/7999667770?check_suite_focus=true

thecristen:
Decided to try these out in https://github.com/mbta/dotcom/commit/707d0b627cb75900cbb6e20f184432302c6b6374, we'll see how it goes! Thanks.

PR #1393, opened 2022-08-19
Clean up OL shuttle pages

i0sea:
#### Summary of changes
**Asana Ticket:** [Clean up OL shuttle pages](https://app.asana.com/0/555089885850811/1202824647488162/f)
set OL shuttle fares to $0.00 and remove OL from connections section.
Screen Shot 2022-08-19 at 3 54 41 PM

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line_controller.ex
@@ -250,36 +250,25 @@ defmodule SiteWeb.ScheduleController.LineController do
   end
 
   defp group_connections(conn, connections) do
+    filtered_connections =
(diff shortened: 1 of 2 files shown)


lund5000:
I wasn't even thinking about the Connections accordion when Avery requested this, so good catch!

Does this also take care of the connections in the line diagram? See https://dev-blue.mbtace.com/schedules/Shuttle-CopleyForestHills/line and scroll down to Forest Hills for an example. Production will soon be showing the line diagrams like dev-blue is now. If not or if it's not an easy update I'm fine with deferring to a follow-up task next week
example-shuttle-stop-with-ol-connection
.

zovt (approved):
Seems good to me 👍

PR #1396, opened 2022-08-23
fix(test): Fix test assertions for Green Line suspension data

arkadyan:
#### Summary of changes
No ticket, fixes currently failing tests

- fix: Properly handle different valid input types
- fix(test): Improve tests to match descriptions
- fix(test): Fix test assertions for Green Line suspension data
- fix(test): Fix more test data

~~Note: I left that last commit separate because those failures seem more suspicious to me. I'm going to keep looking into them as I work through the GL diagram issues.~~
Given that the line diagram is working better with the latest GTFS data, I'm actually probably less worried about those failures than I was at first glance.

---

### DIFF
--- a/apps/site/lib/site_web/views/schedule_view.ex
+++ b/apps/site/lib/site_web/views/schedule_view.ex
@@ -362,9 +362,7 @@ defmodule SiteWeb.ScheduleView do
     "/fares/" <> route_type <> "-fares"
   end
 
-  @spec single_trip_fares(Route.t() | Route.type_int() | String.t()) :: [
-          {String.t(), String.t() | iolist}
-        ]
+  @spec single_trip_fares(Route.t()) :: [{String.t(), String.t() | iolist}]
   def single_trip_fares(route) do
     summary =
       route
@@ -379,7 +377,7 @@ defmodule SiteWeb.ScheduleView do
     end
   end
 
-  @spec to_fare_summary_atom(Route.t() | Route.type_int() | String.t()) :: atom()
+  @spec to_fare_summary_atom(Route.t()) :: atom
   def to_fare_summary_atom(%Route{type: 3, id: id}) do
     cond do
       Fares.silver_line_rapid_transit?(id) -> :subway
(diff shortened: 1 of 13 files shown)


apps/site/test/site_web/controllers/schedule/finder_api_test.exs
```diff
@@ -295,12 +295,14 @@ defmodule SiteWeb.ScheduleController.FinderApiTest do
         |> json_response(200)
 
       assert %{"times" => times} = trip
-      assert length(times) == 8
+      destination_stop_ids = Enum.map(times, &get_in(&1, ["schedule", "stop", "id"]))
```
thecristen, line 298: I always forget about `get_in`!

apps/site/lib/site_web/views/schedule_view.ex
```diff
@@ -362,22 +362,24 @@ defmodule SiteWeb.ScheduleView do
     "/fares/" <> route_type <> "-fares"
   end
 
-  @spec single_trip_fares(Route.t()) :: [{String.t(), String.t() | iolist}]
+  @spec single_trip_fares(Route.t() | Route.type_int() | String.t()) :: [
```
thecristen, line 365: Everywhere this is used seems to be passing in a `%Route{}`, I'm curious why change this?

thecristen (approved):
LGTM with one question.

apps/site/lib/site_web/views/schedule_view.ex
arkadyan: The immediate reason was the [test case](https://github.com/mbta/dotcom/blob/master/apps/site/test/site_web/views/schedule_view_test.exs#L449) for this function calls it using a string and was breaking due to a recent change. But it does seem like this function is only called from a single place, so I'm ok with changing the test instead and leaving the more restrictive spec.

arkadyan: [Fixed](https://github.com/mbta/dotcom/pull/1396/commits/0d41fa2ef55aed55e4f63c68d1ea37a51f3c2092)

PR #1397, opened 2022-08-23
feat(AlertBanner): Parse URL out of alert text and got to url on bann…

kotva006:
…er click

#### Summary of changes
**Asana Ticket:** [Navigate to linked URL when user clicks banner alert](https://app.asana.com/0/home/1201123880594717/1202824647488164)

The main goal here was to change to a desired behavior without changing other existing behavior around the banner.

The alert banners already support navigating to a url on click.  The alert banner data just needed to be parsed differently to put the url in the right location.  Adding the parsed url to the alert banner struct caused the URL to be appended to the banner a second time.  The property `url_parsed_out_of_title` was added to stop this behavior so that if we parsed a url out of the title it would not render it again.

If a url is set on the data imported for the banner it should do the same behavior as today.

---

### DIFF
--- a/apps/alerts/lib/parser.ex
+++ b/apps/alerts/lib/parser.ex
@@ -199,12 +199,13 @@ defmodule Alerts.Parser do
               "informed_entity" => informed_entity
             } = attributes
         })
-        when title != nil do
+        when title != nil and url != nil do
       [
         %Banner{
           id: id,
           title: title,
           url: url,
+          url_parsed_out_of_title: false,
           effect: Alert.effect(attributes),
           severity: Alert.severity(severity),
           informed_entity_set:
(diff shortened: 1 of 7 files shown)


apps/alerts/lib/parser.ex
```diff
@@ -199,12 +199,40 @@ defmodule Alerts.Parser do
...
+        })
+        when title != nil and url == nil do
+      parsed_url = SiteWeb.Views.Helpers.URLParsingHelpers.get_full_url(title)
+
+      [
+        %Banner{
+          id: id,
+          title: title,
+          url: parsed_url,
+          url_parsed_out_of_title: true,
```
kotva006, line 235:
```suggestion
          url_parsed_out_of_title: parsed_url != nil,
```

i0sea (changes requested):
The linking works when clicking inside `c-alert-item__top-text-container` but not for the entire banner `announcement-container alert-announcement__container`. Perhaps the onclick method should be moved. Only seems to be an issue on desktop.

kotva006:
> The linking works when clicking inside `c-alert-item__top-text-container` but not for the entire banner `announcement-container alert-announcement__container`. Perhaps the onclick method should be moved. Only seems to be an issue on desktop.

It looks like this only affects the edges of the banner, and isn't a new bug introduced in this pull request.  I would say if it is urgent to get this out the door we merge this and make a new ticket to make the whole banner clickable.  It's going to take me about a day to figure out how to hoist the `onclick` out of one of the templates.

lund5000:
> It looks like this only affects the edges of the banner, and isn't a new bug introduced in this pull request. I would say if it is urgent to get this out the door we merge this and make a new ticket to make the whole banner clickable. 

+1 to merging as-is and handling this as a follow-up if this is functioning well otherwise

i0sea (approved):

PR #1398, opened 2022-08-23
Rk husky

kotva006:
#### Summary of changes
**Asana Ticket:** None

The goal of this pull request is to add a pre-commit hook using [husky](https://typicode.github.io/husky/#/)
The goal being it can run commands that need to pass before the code can be committed.

Right now its configured to format the elixir, and typescript/js, and then lint the typescript/js

If any of the commands fail it will stop the commit from happening.  The dev can then fix the error and then commit again.  This check can also be skipped if need be using the `--no-verify` flag with git.

The main win this brings is as a devs having to remember less commands to run on every code change, and also trying to catch issues with the code before it gets to the code checks on github.

---


.husky/pre-commit
```diff
@@ -0,0 +1,21 @@
...
+# Only save the files that were stagged for the commit
+stagged_files=$(git diff --name-only --cached --diff-filter=ACMRTUB)
+
+# formatting
+mix format
+npm run --prefix apps/site/assets prettier -- -staged
+
+# linting
+npm run --prefix apps/site/assets eslint:js
+npm run --prefix apps/site/assets eslint:ts
```
arkadyan, line 13:
```suggestion
npm run ci:lint:js
npm run ci:lint:ts
```
suggestion: There are existing scripts for these in the root `package.json` file.

apps/site/assets/package.json
```diff
@@ -132,6 +133,7 @@
     "stylelint": "stylelint css/**/*.scss --ignore-path .stylelintignore",
     "prettier": "prettier --write \"{js,ts}/**/*.{js,ts,tsx}\"",
     "prettier:check": "npm run prettier -- --list-different",
+    "prepare": "cd ../../.. && husky install .husky",
```
arkadyan, line 136: question: Would it work to move this to the root `package.json` file and not need the `cd` bit?

.husky/pre-commit
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
arkadyan, line 1:
```suggestion
#!/usr/bin/env bash
```
suggestion: I think it's considered safer to get the location of `bash` from the system `env`.

arkadyan (commented):
Mostly looks good, but added a few small suggestions. Do we want to chat this over as an entire dev team to discuss / warn everyone about the change?

apps/site/assets/package.json
thecristen: to this question, might need some prefixing hackery e.g. `"$(npm bin --prefix apps/site/assets)/husky install .husky` might work

README.md
```diff
@@ -189,6 +189,18 @@ Then, visit the site at http://localhost:4001/
...
+When commiting code a bunch of checks are run using [git pre-commit hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)
+
+This is configured using the tool [husky](https://typicode.github.io/husky/#/) and should install as part of the regular `npm run install` process
+
+This allows all code changed during a commit to be linted and formatted before being commited.
+
+If you wish to modify the checks ran during the pre-commit process the [pre-commit](.husky/pre-commit) file can be modified with any command line commands/scripts
+
+These checks can be skipped by using the `--no-verify` flag when runnint the `commit` git command
+
```
thecristen, line 203: question: should devs run the `prepare` npm script that is added in this PR? Or is that not necessary?

apps/site/assets/package.json
arkadyan: Could husky instead just be installed at the root of the project? Or would you not be able to use it from the `site` package scripts? If that's the case seems like it might be fine as is.

kotva006: The issue with installing it in the root is we have no other modules installed there.  We could, definitely do it, just setting new precedents.

README.md
kotva006: I believe it is not necessary, I think the during the install of `husky` it looks for that script and runs it.  I know the devs on PickAssist did not need to run it.

thecristen (approved):
Lookin good to me.

.husky/pre-commit
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+. "$(dirname -- "$0")/_/husky.sh"
+
+# Only save the files that were stagged for the commit
```
kotva006, line 4:
```suggestion
# Only save the files that were staged for the commit
```

PR #1401, opened 2022-08-25
fix(Localization): Remove the unsupported localizations

kotva006:
#### Summary of changes
**Asana Ticket:** [Remove Nepali and Somali translation options](https://app.asana.com/0/555089885850811/1202824647488178)

The google translate plugin we use doesn't support Nepali and Somali, so they need to be removed.

---


arkadyan (approved):
👍

PR #1402, opened 2022-08-25
feat: Transfers from a shuttle should not be free

arkadyan:
#### Summary of changes
**Asana Ticket:** [Investigate making OL/GL shuttle to subway transfer not free](https://app.asana.com/0/555089885850811/1202824647488182)

Shuttles have free fares, but if a rider then transfers to a subway or
bus line they have to pay the normal fare for that route.

---

### DIFF
--- a/apps/trip_plan/lib/trip_plan/transfer.ex
+++ b/apps/trip_plan/lib/trip_plan/transfer.ex
@@ -38,8 +38,16 @@ defmodule TripPlan.Transfer do
 
   def is_subway_transfer?(_), do: false
 
-  @doc "Takes a set of legs and returns true if there might be a transfer between the legs, based on the lists in @single_ride_transfers and @multi_ride_transfers. Exception: no transfers from bus route to same bus route."
+  @doc """
+  Takes a set of legs and returns true if there might be a transfer between the legs, based on the lists in @single_ride_transfers and @multi_ride_transfers.
+
+  Exceptions:
+  - no transfers from bus route to same bus route
+  - no transfers from a shuttle to any other mode
+  """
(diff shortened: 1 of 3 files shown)


apps/trip_plan/lib/trip_plan/transfer.ex
```diff
@@ -38,8 +38,16 @@ defmodule TripPlan.Transfer do
...
-  @doc "Takes a set of legs and returns true if there might be a transfer between the legs, based on the lists in @single_ride_transfers and @multi_ride_transfers. Exception: no transfers from bus route to same bus route."
+  @doc """
+  Takes a set of legs and returns true if there might be a transfer between the legs, based on the lists in @single_ride_transfers and @multi_ride_transfers.
+
+  Exceptions:
+  - no transfers from bus route to same bus route
+  - no transfers from a shuttle to any other mode
+  """
   @spec is_maybe_transfer?([Leg.t()]) :: boolean
+  def is_maybe_transfer?([%Leg{mode: %TransitDetail{route_id: "Shuttle-" <> _}} | _]), do: false
```
thecristen, line 49: Maybe one day we will have a better way to do this than parsing the name of the route.. 🙃

thecristen (approved):
Looks very reasonable to me. Thanks for the extra tests!

apps/trip_plan/lib/trip_plan/transfer.ex
arkadyan: Yeah, I totally agree. I didn't see anything better to use, but if there's something I missed or something in the future that would definitely be better.

PR #1407, opened 2022-08-26
Add "Trains continue to" or "coming from" row to combined timetable

i0sea:
#### Summary of changes
**Asana Ticket:** [Add "Trains continue to" or "coming from" row to combined timetable](https://app.asana.com/0/555089885850811/1202792787402853/f)

Add continue to row only on OL timetable that states destination of train.
Screen Shot 2022-08-26 at 1 07 29 PM


apps/site/lib/site_web/templates/schedule/_timetable.html.eex
```diff
@@ -155,6 +155,55 @@
             <% end %>
           <% end %>
         <% end %>
+        <%= if @route.id == "Orange" and @direction_id == 0 do %>
+         <% cell_background = if rem(2, 2) == 0 do "white" else "gray" end %>
```
arkadyan, line 159: question: Will this always be true?

apps/site/lib/site_web/templates/schedule/_timetable.html.eex
```diff
@@ -155,6 +155,55 @@
...
+                      <span class="m-timetable__time" >
+                      <%= if !is_nil(trip_schedule) do%>
+                        <%= V3Api.Trips.short_name(trip_schedule.trip.headsign) %>
+                      <% else %>
+                      <%# if trip schedule is nil check other stops for non nil value %>
+                        <%
+                          all_trips = for stop_idx <- [0,1,3] do
+                            stop = Enum.at(@all_stops, stop_idx)
+                            trip_id = schedule.trip.id
+                            trip_schedule = @trip_schedules[{trip_id, stop.id} || %{}]
```
arkadyan, line 191: suggestion: These 3 lines seem essentially identical to lines 176–178 above, and in both cases, the only result we end up using is `trip_schedule`. In order to remove the repetition, I think that you could move this logic into a helper function in the `Timetable` view, and then be able to call it twice.

arkadyan: Actually, as I look further, maybe even better than extracting just those 3 lines would be to extract all of the logic from line 183 through line 200 a function in the view. (Lines 175–179 would need to move as well, and you could still put those into a second, private helper function.) That all means that you would have a single function to call that would return the value of the `span` element.

The benefit I'd see in this refactor is that the page layout in the `.html.eex` file gets cleaner and easier to read, while at the same time you more explicitly define a chunk of business logic as a pure function with clear inputs and output as well as a name that helps describe the purpose. (This function is also easily testable in isolation if you were interested in that.)

apps/v3_api/lib/trips.ex
```diff
@@ -14,4 +14,18 @@ defmodule V3Api.Trips do
     opts = put_in(opts[:route], route_id)
     get_json("/trips/", opts)
   end
+
+  def short_name(stop_name) do
```
arkadyan, line 18: suggestion(non-blocking): This is mostly a style thing, nothing wrong with what you have here so take it or leave it, but just to mention another "Elixir-y" way to accomplish this might be to define a separate function for each potential value of `stop_name` and then not need the case statement. I.e.:

		def short_name("Needham Heights") do
			"NEED"
		end

		…

		def short_name(stop_name) do
			String.upcase(String.slice(stop_name, 0..2))
		end

Completely unrelated code, but just because I was looking at it earlier today, there's a similar example in [`mailer.ex`](https://github.com/mbta/dotcom/blob/master/apps/feedback/lib/mailer.ex#L77).

arkadyan (commented):
This is looking great—exciting to see. I added a couple small questions/comments, but nothing major. I don't know if we are looking to get this out ASAP, but if we are I don't think anything needs to be blocking.

arkadyan (approved):

PR #1408, opened 2022-08-26
deps(Bootstrap): Upgrade to version 5 of bootstrap

kotva006:
#### Summary of changes
**Asana Ticket:** No Ticket

Upgrading the projects version of bootstrap from 4 alpha to 5

---


apps/site/assets/css/_mixins.scss
```diff
@@ -375,13 +398,13 @@
 }
 
 // Caret used in some expandable/collapsable blocks in the application
-@mixin caret($color) {
+@mixin caret-mbta($color) {
```
kotva006, line 401: stops the name collision with bootstraps caret

apps/site/assets/css/app.scss
```diff
@@ -7,7 +7,10 @@
 @import 'fonts';
 @import 'navbar';
 @import 'buttons-overrides';
-//
+$link-decoration: none;
```
kotva006, line 10: This was already set in the [fonts.scss](https://github.com/mbta/dotcom/blob/53cb06b630bc539e79299af4956377f3170c383c/apps/site/assets/css/_fonts.scss#L20) file.

thecristen: Ah, yeah let's consolidate those overrides.

PR #1409, opened 2022-08-31
refactor: use URLSearchParams Web API

thecristen:
no ticket!

Since we're no longer supporting IE11, we are free to use [the URLSearchParams API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to handle parsing/formatting/querying the URL query string. We had logic for this in a few different places, this is an effort to streamline that.

I did need to change a few test strings, as the formatted query parameters from `URLSearchParams` are slightly different than the one from `window.encodeURIComponent` (notably the `%20` becomes simply `+` - is this right? [This StackOverflow question has some interesting answers](https://stackoverflow.com/questions/1211229/in-a-url-should-spaces-be-encoded-using-20-or) around that. I've decided this is fine.)

### DIFF
--- a/apps/site/assets/js/algolia-autocomplete-with-geo.js
+++ b/apps/site/assets/js/algolia-autocomplete-with-geo.js
@@ -192,8 +192,7 @@ class AlgoliaAutocompleteWithGeo extends AlgoliaAutocomplete {
     params.longitude = longitude;
     params.address = this._input.value;
     const qs = QueryHelpers.paramsToString(
-      addFilterParam(params, window.location.pathname),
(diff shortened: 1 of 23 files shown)


thecristen:
I don't know why the `package-lock.json` added so many more lines 😵‍💫 
image

thecristen:
Assigning to @arkadyan since he already reviewed most of this. The only new bit here is the two last commits and the `package-lock.json` change.

Deploying to dev-blue to do some QA.

arkadyan (approved):
Code changes look good. Looks like maybe there are some build issues or something though?

PR #1412, opened 2022-09-09
feat(RouteCardHeader): Add copy that makes the card header look clickable

kotva006:
#### Summary of changes
**Asana Ticket:** [Add clearer link to schedule page in Departures](https://app.asana.com/0/home/1201123880594717/1202908385566824)
**Follow Up Ticket:** [Design Fixes](https://app.asana.com/0/555089885850811/1202967204475095/f)

Added the text - View Schedule to the card header, and removed the upper casing from the card header as well.

---

### DIFF
--- a/apps/site/assets/css/_link-block.scss
+++ b/apps/site/assets/css/_link-block.scss
@@ -34,10 +34,11 @@
   }
 
   &__view-schedule {
+    font-size: 1rem;
     position: relative;
-    bottom: 2px;
+    bottom: 0.125rem;
     font-weight: 500;
-    font-size: 14px;
-    margin-left: 25px;
+    margin-left: 0.75rem;
+    border-bottom: 0.063rem solid;
   }
 }
--- a/apps/site/assets/css/_link-block.scss
+++ b/apps/site/assets/css/_link-block.scss
@@ -32,4 +32,12 @@
   &__inner-link {
     pointer-events: all;
   }
+
+  &__view-schedule {
+    position: relative;
+    bottom: 2px;
+    font-weight: 500;
+    font-size: 14px;
+    margin-left: 25px;
+  }
 }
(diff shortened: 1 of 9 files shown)


apps/site/assets/ts/components/RouteCardHeader.tsx
```diff
@@ -21,10 +21,13 @@ const RouteCardHeader = ({
...
-        {isASilverLineRoute(route.id)
-          ? `Silver Line ${route.name}`
-          : breakTextAtSlash(route.name)}
+      <span>
+        <span className={busClass(route)}>
+          {isASilverLineRoute(route.id)
+            ? `Silver Line ${route.name}`
+            : breakTextAtSlash(route.name)}
+        </span>
+        <u className="c-link-block__view-schedule">View Schedule</u>
```
thecristen, line 30: Nice use of a `<u>` here

apps/site/assets/css/_link-block.scss
```diff
@@ -32,4 +32,12 @@
   &__inner-link {
     pointer-events: all;
   }
+
+  &__view-schedule {
+    position: relative;
+    bottom: 2px;
+    font-weight: 500;
+    font-size: 14px;
+    margin-left: 25px;
```
thecristen, line 41:
```suggestion
    margin-left: 2ch;
```

The` 25px` looks a good deal bigger than the spacing in the mock to me (I realize the mock doesn't really enumerate). But maybe this will be better? (I happen to enjoy `ch` units for spacing between letters but feel free to adjust).

thecristen (changes requested):
A minor non-blocking comment added. It looks good!

**Requesting changes** to please update the Jest snapshots -- `npm --prefix apps/site/assets run jest:update` should do it.

lund5000:
Just added the "do not merge" label to this because we want to wait until there's a week's worth of analytics data for the before state to more easily evaluate the impact afterwards. I had to fix the event yesterday after a class name change broke it at some point, so we should be good to deploy this change on Mon 9/19. Thanks!

PR #1417, opened 2022-09-16
Restore normal OL service

i0sea:
#### Summary of changes
**Asana Ticket:** [Restore normal OL service](https://app.asana.com/0/385363666817452/1202879553628842/f)

Remove changes made for OL surge.


thecristen (commented):
I'd like to add a commit to this to remove the relevant portion of #1384 :)

**Question**: Are the fares going back to normal such that we'd also want to remove the changes from [Calculate some CR legs as $0 during OL shutdown](https://github.com/mbta/dotcom/pull/1385/files)?

lund5000:
> I'd like to add a commit to this to remove the relevant portion of #1384 :)

Could we please get this deployed as-is and ASAP first and make that a follow-up PR?

> **Question**: Are the fares going back to normal such that we'd also want to remove the changes from [Calculate some CR legs as $0 during OL shutdown](https://github.com/mbta/dotcom/pull/1385/files)?

Yes! Good catch. We'll need that reverted, too (also in a separate PR, please).

thecristen (approved):
Cool, let's do this! :rocket:

PR #1418, opened 2022-09-19
Improve JS test coverage

thecristen:
Howdy! Was writing tests for an unrelated ticket and decided to try to improve that coverage number.

This is done in two ways:
1. Reducing the parts of code that are involved in the coverage calculation.
2. Writing more tests!

A few minor test fixes later and I have a passing test suite again! :tada:

Notable other things:
* Added [React Testing Library](https://testing-library.com/docs/react-testing-library/intro). We already use other parts of the Testing Library (mainly `@testing-library/hooks`), and RTL has a helpful guide to [migrating tests away from Enzyme](https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/), which I'd like to do given Enzyme's lack of maintenance. One positive aspect of RTL is its push away from testing implementation details -- instead of making assertions based on React component state and props as encouraged by Enzyme, we can instead query the fully rendered result using methods more aligned to how a user might view and interact with our page. In my opinion RTL also has simpler APIs for firing events and user interactions.

It's a different vibe!
image


zovt (approved):
Seems reasonable to me 👍

PR #1419, opened 2022-09-19
Manually remove elevator from Lynn station accessibility features

i0sea:
#### Summary of changes
**Asana Ticket:** [Manually remove elevator from Lynn station accessibility features](https://app.asana.com/0/385363666817452/1203005769296539/f)

Remove elevator for list of accessibility features on sidebar on Lynn stop page

### DIFF
--- a/apps/site/assets/ts/stop/components/sidebar/Accessibility.tsx
+++ b/apps/site/assets/ts/stop/components/sidebar/Accessibility.tsx
@@ -92,9 +92,9 @@ const renderFeaturesList = (stop: Stop): ReactElement<HTMLElement> | null =>
       has the following features:
       <ul className="m-stop-page__sidebar-list">
         {stop.name === "Lynn"
-          ? [stop.accessibility[0], stop.accessibility[2]].map(access => (
-              <Access access={access} key={access} />
-            ))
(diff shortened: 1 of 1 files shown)


apps/site/assets/ts/stop/components/sidebar/Accessibility.tsx
```diff
@@ -91,9 +91,13 @@ const renderFeaturesList = (stop: Stop): ReactElement<HTMLElement> | null =>
       {isSpecialCase(stop) ? `${stop.name} ` : "It "}
       has the following features:
       <ul className="m-stop-page__sidebar-list">
-        {stop.accessibility.map(access => (
-          <Access access={access} key={access} />
-        ))}
+        {stop.name === "Lynn"
+          ? [stop.accessibility[0], stop.accessibility[2]].map(access => (
```
thecristen, line 95: I think this shouldn't rely on the positioning - what if the order of the strings returned changes?
We can filter by looking at the value. Something more like `stop.accessibilty.filter(access => access !== "elevator").map( // and so on`

thecristen (changes requested):

thecristen (approved):
👍🏼

PR #1420, opened 2022-09-20
Schedule Finder Route Pattern Support

zovt:
#### Summary of changes
**Asana Ticket:** [Schedule Finder | Multi-Route Trips should show all possible origin stops ](https://app.asana.com/0/385363666817452/1201022904675406/f)

NOTE: WIP while I work through the test updates, but should be good to deploy / experiment with if you'd like to test the changes.

I've broken the initial commits up with additional commentary which I hope will be helpful in providing context.

Ultimately, this change set allows the schedule finder to actually load the information for stops that are present on only a route variant, rather than the primary route. Bus line 93 is a good example of this. This is accomplished by sharing state (the currently selected route variant / direction) between the Map/Line Diagram and the ScheduleFinder. The commit comments have some additional info into what this entailed.

The UI/UX for this feels not quite right to me, but I'm not sure exactly what the desired experience is either. Specifically, I think having two controls on the page (the Map/Line Diagram "Change Direction" button, and the Schedule Finder direction picker itself) for selecting the direction feels a little weird. Additionally, aside from the Map / Line Diagram changing, there is little indication that the set of stops the Schedule Finder is working with has changed as well. I don't think we need to address these right now (since it sounds like Schedule Finder may be undergoing a major rework in the near-future anyways), but I wanted to highlight them in case these are things we want to consider moving forward.

This required a bit more surgery than expected, so let me know what you think / if you have any questions or if I've missed the mark entirely. Thanks!


zovt:
To update this: I've been working through the tests here and there's a lot to go through.

Thoughts: Originally I wanted to share the state between these two components to limit the visible stops to the selected route variant. However, I _think_ we could load all of the stops for all variants in the static page data once up front, and then rely on the search functionality to help the user find the relevant stops. This should also allow the modal to display the information properly when a variant is selected in the Map/Line Diagram (note: I haven't actually tested this theory)

Pros: much much simpler implementation, fewer tests to fix (the front end should end up being mostly untouched, hopefully)
Cons: could be a lot of stops to deal with, might be overwhelming?

I'll ask about this in standup tomorrow morning, hopefully this makes sense!

apps/site/lib/site_web/controllers/schedule/finder_api.ex
```diff
@@ -33,6 +38,29 @@ defmodule SiteWeb.ScheduleController.FinderApi do
...
+      |> Map.put("0", get_stops_for_route_pattern(route, 0, pattern_id))
+      |> Map.put("1", get_stops_for_route_pattern(route, 1, pattern_id))
+
+    json(conn, stops_by_direction)
+  end
+
+  defp get_stops_for_route_pattern(route, direction_id, pattern_id) do
+    pattern = RoutePatterns.Repo.get(pattern_id)
+    trip = Schedules.Repo.trip(pattern.representative_trip_id)
+    shape = Routes.Repo.get_shape(trip.shape_id) |> List.first()
```
thecristen, line 58: When [we parse route patterns we add the shape_id and stop_ids](https://github.com/mbta/dotcom/blob/74a4703d61005cfbd2f64ea72fa8db35921215f3/apps/route_patterns/lib/route_pattern.ex#L102), couldn't we use those?

apps/site/lib/site_web/controllers/schedule/finder_api.ex
```diff
@@ -33,6 +38,29 @@ defmodule SiteWeb.ScheduleController.FinderApi do
...
+
+    stops_by_direction =
+      Map.new()
+      |> Map.put("0", get_stops_for_route_pattern(route, 0, pattern_id))
+      |> Map.put("1", get_stops_for_route_pattern(route, 1, pattern_id))
+
+    json(conn, stops_by_direction)
+  end
+
+  defp get_stops_for_route_pattern(route, direction_id, pattern_id) do
```
thecristen, line 55: we don't seem to need `route` or `direction_id` here

thecristen (commented):
Just wanted to leave some initial comments while you reconsider approach. One nice thing I noticed from this PR is that it actually respects the URL parameter now when loading the page for a non-default route pattern. :)

thecristen:
Closing in favor of #1429

PR #1421, opened 2022-09-21
feat: Streaming predictions proof-of-concept

arkadyan:
#### Summary of changes
**Asana Ticket:** [Finish streaming POC](https://app.asana.com/0/555089885850811/1202995400172235)

- feat: Streaming predictions

Use a DynamicSupervisor for API prediction streams since they need to be
per route—a "firehose" ever everything is not supported by the API.
Start streams when a client subscribes to predictions for a route.
Register streams per route to prevent duplication.

Parse data into pure Prediction structs for the moment. We can think
harder about exactly what data format we want after demonstrating this
proof-of-concept.

- feat: Predictions Phoenix channel

- WIP(DO NOT MERGE): React predictions demo

**NB:** The first 2 commits are intended to be production-ready, but the 3rd is only a quick-and-dirty visual demonstration and is expected to be discarded/transformed.

---

### DIFF
--- a/apps/predictions/lib/parser.ex
+++ b/apps/predictions/lib/parser.ex
@@ -1,9 +1,10 @@
 defmodule Predictions.Parser do
-  alias Predictions.Prediction
-
   @moduledoc """
   Functions for parsing predictions from their JSON:API format.
   """
+
+  alias JsonApi.Item
+  alias Predictions.Prediction
 
   @type record :: {
           Prediction.id_t() | nil,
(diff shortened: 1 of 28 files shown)


apps/predictions/lib/parser.ex
```diff
@@ -13,30 +14,83 @@ defmodule Predictions.Parser do
           0 | 1,
           DateTime.t() | nil,
           non_neg_integer,
-          Prediction.schedule_relationship(),
+          Prediction.schedule_relationship() | nil,
```
thecristen, line 17: Good catch here.

thecristen (commented):
Playing around with this, I notice we get predictions in the past (screenshot was taken 10:41am). Are old predictions not being replaces on later updates? Can we have the channel only output predictions that aren't in the past?

image

arkadyan:
@thecristen Huh, I thought https://github.com/mbta/dotcom/pull/1421/files#diff-29f1b86ca05547f6e307f42931f9082aaadc38ea60095cbc3c33178069801323R46 was filtering out old predictions. I'll have to look more into that…

thecristen (commented):
I noticed you added the departure and arrival times, which is cool. 👍🏼 
Other pieces that will be important for the display code come from the prediction's relationships. I noticed we have `trip_id` and `stop_id`, but in the context of a React component these are of very limited use - we can't fetch a trip `headsign` or `name`, or a stop `platform_code` without making additional requests to the backend. Any thoughts on how/where we should reconcile this?

thecristen (changes requested):
Can we update this to have the predictions channel subscribe by both `routeId` _and_ `stopId`? After looking at how predictions are used, having channels defined by both stop+route makes more sense to me.
* Line diagram - each stop along a particular route
* Stop page - every route at a particular stop
* Transit Near Me - many routes + many stops

I'm also wondering if we can figure out why we're seeing predictions prior to `now`, but I can be convinced to leave that for a future iteration.

thecristen:
Thanks Matt. Going to start a feature branch to iterate off this.

thecristen:
Covered in #1476

PR #1425, opened 2022-09-22
Rearrange subway /line pages to get riders to departures faster

i0sea:
#### Summary of changes
**Asana Ticket:** [Rearrange subway /line pages to get riders to departures faster](https://app.asana.com/0/555089885850811/1202908385566844/f)

Move map to bottom of route page and change text to view departures for subway.


arkadyan (commented):
question: Is this rearranging the order for all route types? It sounded like we only wanted to change subway pages?

i0sea:
> question: Is this rearranging the order for all route types? It sounded like we only wanted to change subway pages?

based on the ticket the rearranging is for all route types while the text change and is only for subway lines. @lund5000 is this correct?

apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
```diff
@@ -78,7 +78,12 @@ const StopCard = (props: StopCardProps): ReactElement<HTMLElement> => {
   const showPrediction = hasLivePredictions && !isDestination;
   const showDiversion =
     diversionAlert && !(hasLivePredictions && isDestination);
-
+  const showDepartures = (): string => {
```
arkadyan, line 81: suggestion (non-blocking): Personally, this function name sounds like calling it would trigger an action. I wonder if something in the ballpark of `buttonLabel` might be a closer description? (I'm terrible at coming up with names, and that's also pretty generic sounding, so take it in whatever direction you want.)

apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
```diff
@@ -78,7 +78,12 @@ const StopCard = (props: StopCardProps): ReactElement<HTMLElement> => {
...
+    if (routeStop.route?.type === 1 || routeStop.route?.type === 0) {
+      return liveData?.headsigns.length ? "View upcoming departures" : "";
```
arkadyan, line 83: suggestion: Using an empty string for a button label feels like something you'd never want. Maybe is this conditional not even necessary? I.e., can we just return "View upcoming departures" for all subway routes?

arkadyan (commented):
A couple small comments on the button label bit. The empty string one feels much more important to me than the other.

apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
arkadyan: Sorry, one additional suggestion: I think this logic could warrant a super quick test or two.

lund5000:
> > question: Is this rearranging the order for all route types? It sounded like we only wanted to change subway pages?
> 
> based on the ticket the rearranging is for all route types while the text change and is only for subway lines. @lund5000 is this correct?

Only subway /line pages should change for now. Once the changes in this PR are limited only to those pages, let's make sure this goes through our new design review step before engineers re-review. Thanks!

i0sea:
> @i0sea Sorry, I wasn't very clear in my initial comment, but unless I'm misunderstanding I don't think returning `null` here is any better than returning an empty string. Fundamentally, I don't think you'd ever want a button without a label. But can you always just return "View upcoming departures" for the subway case and therefor not need the `liveData?.headsigns.length` condition at all?

Part of the ticket was if there are no departures for a subway there shouldn't be a link. If the stop is a subway line and has no departures I suppose we wouldn't want any button at all in that case to avoid the blank button issue.

arkadyan:
@i0sea Ah I see, yes, I think you are absolutely right, not showing the button at all seems better.

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -205,6 +205,20 @@ const ScheduleDirection = ({
           <ScheduleDirectionButton dispatch={dispatch} />
         ) : null}
       </div>
+      {route.type === 1 || route.type === 0
```
arkadyan, line 208: suggestion (non-blocking): This could be pulled out into a helper function, maybe called something like `isASubwayRoute` or something along those lines. This would 1) consolidate repeated logic, 2) describe the point of the logic through the function name, and 3) be testable.

apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx
```diff
@@ -122,16 +127,19 @@ const StopCard = (props: StopCardProps): ReactElement<HTMLElement> => {
...
-        <footer className="m-schedule-diagram__footer">
-          <button
-            className="btn btn-link"
-            type="button"
-            onClick={() => onClick(routeStop)}
-          >
-            View schedule
-          </button>
-        </footer>
+        {(!liveData?.headsigns.length &&
```
arkadyan, line 130: suggestion (blocking): Between showing/hiding the footer and the button label variations, I think this file could use a few test cases, please.

arkadyan (changes requested):
Looking good. One potential refactor that you can take or leave, but I think more substantial is a need for test coverage, particularly around `StopCard`.

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -176,6 +176,8 @@ const ScheduleDirection = ({
     error: false
   });
 
+  const isSubwayRoute: boolean = route.type === 1 || route.type === 0;
```
kotva006, line 179: nitpick: Since this check is also used in other files, it might make sense to make this a function and move it to a helper

arkadyan: That's a good point. https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/models/route.ts might be a good place for that.

arkadyan (approved):
Ryan makes a good suggestion, but otherwise this looks good 👍 

Thanks for dealing with those unrelated tests.

PR #1426, opened 2022-09-22
fix: Move Union Square to the end of the line diagram

arkadyan:
#### Summary of changes
**Asana Ticket:** [Union Square displayed in middle of line diagram](https://app.asana.com/0/555089885850811/1203005769296541)

fix: Move Union Square to the end of the line diagram

Screen Shot 2022-09-22 at 16 27 45

---

### DIFF
--- a/apps/site/lib/site_web/controllers/schedule/line/diagram_helpers.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line/diagram_helpers.ex
@@ -28,6 +28,7 @@ defmodule SiteWeb.ScheduleController.Line.DiagramHelpers do
       ) do
     branches
     |> Enum.reverse()
+    |> override_green_0_order(direction_id)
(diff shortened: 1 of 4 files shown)


zovt (approved):
This seems reasonable to me as a way to get this working for now. However, in the long run I think it would be best to avoid this sort of semi-hidden-manual-override code w/r/t the GLX data (if anything, I would even advocate for just hardcoding it over additional complexity in this code). 

Hopefully some elegant solutions for handling the complex branching/stop ordering logic are discovered!

arkadyan:
@zovt Yes, I completely agree! I'm hopeful that we can refactor things to use a data structure (and algorithm) that just more natively handles all the branching without any of this extra effort and all these special cases.

arkadyan:
I believe that rebasing on top of https://github.com/mbta/dotcom/pull/1428 once that's merged should fix the remaining failing tests.

PR #1428, opened 2022-09-23
a11y: increase color contrast for SL pill

thecristen:
No ticket, this should resolve this error message that was failing some of our homepage accessibility tests:

```
Fix any of the following:\n  Element has insufficient color contrast of 3.67 (foreground color: #ffffff, background color: #7c878e, font size: 10.5pt (14px), font weight: bold). Expected contrast ratio of 4.5:1
```

# Before

image

---

# After

image

### DIFF
--- a/apps/site/assets/css/_homepage.scss
+++ b/apps/site/assets/css/_homepage.scss
@@ -372,6 +372,11 @@
     height: 24px;
   }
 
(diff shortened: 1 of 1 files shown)


arkadyan (approved):
👍🙇

i0sea (approved):

PR #1429, opened 2022-09-23
Support viewing route variant stops in Schedule Finder

zovt:
#### Summary of changes
**Asana Ticket:** [Schedule Finder | Multi-Route Trips should show all possible origin stops ](https://app.asana.com/0/385363666817452/1201022904675406/f)

- Return all stops for all route patterns in schedule page data
- Make clicking "View schedule" set the Schedule Finder direction to match the map / line diagram (an existing bug :))

This is a lot lower lift than the initial PR, and works just about as well :)

### DIFF
--- a/apps/site/test/site_web/controllers/schedule/line_test.exs
+++ b/apps/site/test/site_web/controllers/schedule/line_test.exs
@@ -309,4 +309,102 @@ defmodule SiteWeb.ScheduleController.LineTest do
       assert default_services = [%{id: "WinterWeekday1"}]
     end
   end
+
+  describe "get_all_route_stops/2" do
+    test "works with route with variants" do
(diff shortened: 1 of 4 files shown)


apps/site/test/site_web/controllers/schedule/line_test.exs
```diff
@@ -309,4 +309,102 @@ defmodule SiteWeb.ScheduleController.LineTest do
...
+                 "1523",
+                 "1524",
+                 "1525",
+                 "1526",
+                 "1777",
+                 "1778",
+                 "1994",
+                 "8993",
+                 "place-kencl"
+               ])
```
thecristen, line 407: Comparing the completed sorted lists feels kinda overkill to me -- could consider comparing the number of stops in the route, or asserting the presence of specific known stops.

PR #1431, opened 2022-09-28
remove extra faceted search queries

thecristen:
#### Summary of changes
**Asana Ticket:** [All search inputs are generating faceted search requests that aren't necessary](https://app.asana.com/0/555089885850811/1203050684098647/f)

See Asana ticket for details.
We ultimately figured this change probably won't impact our bottom line anyway. But removing this here doesn't seem to adversely impact anything, so may as well clean things up?


kotva006 (approved):
👍

PR #1432, opened 2022-09-28
fix: Miscellaneous small fixes

arkadyan:
#### Summary of changes
No ticket, just a handful of initial things I came across as I started looking into upgrading Elixir.

- fix: Remove duplicate docs
- fix: This parsing helper shouldn't be a view
- fix: Fix typespec
- fix: Remove function definition that is never reached

---

### DIFF
--- a/apps/site/lib/site_web/views/partial/svg_icon_with_circle.ex
+++ b/apps/site/lib/site_web/views/partial/svg_icon_with_circle.ex
@@ -97,7 +97,6 @@ defmodule SiteWeb.PartialView.SvgIconWithCircle do
   defp icon_name(:access), do: "accessible"
   defp icon_name(:no_access), do: "not-accessible"
   defp icon_name(:the_ride), do: "the-ride"
-  defp icon_name(:parking_lot), do: "parking"
   defp icon_name(:station), do: "circle-t"
   defp icon_name(:t_logo), do: "circle-t"
   defp icon_name(:silver_line), do: "silver-line"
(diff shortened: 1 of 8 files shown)


apps/site/lib/site_web/views/partial/svg_icon_with_circle.ex
```diff
@@ -97,7 +97,6 @@ defmodule SiteWeb.PartialView.SvgIconWithCircle do
   defp icon_name(:access), do: "accessible"
   defp icon_name(:no_access), do: "not-accessible"
   defp icon_name(:the_ride), do: "the-ride"
-  defp icon_name(:parking_lot), do: "parking"
```
amaisano, line 100: Does content know you're removing this icon?

amaisano (approved):
Generally LGTM, but please check the one comment.

apps/site/lib/site_web/views/partial/svg_icon_with_circle.ex
amaisano: I think so. The CMS uses this reference: https://github.com/mbta/dotcom/blob/39a294dc4b05e68ea0e32416206640c4f4db62a6/apps/site/lib/site/content_rewriters/liquid_objects.ex#L128

So if that's already switched over we should be good.

arkadyan: Nothing is being removed. This is a private function only being called from one place which has a guard on the input. It's not possible for this definition to ever be reached.

PR #1434, opened 2022-10-03
feat: Provide stops structured in an unrooted polytree

arkadyan:
#### Summary of changes
**Asana Ticket:** [Green Line - adjust backend generation of stop_data to add support for additional branching](https://app.asana.com/0/555089885850811/1201936157420490)

Provides the data in the LineController.show data. In the future, we'll
also need to provide this data in the LineApi.show response.

Also a few bits of cleanup:
- fix: Remove unused imports
- refactor: Move private functions below public ones

---


thecristen (commented):
Looking really promising so far! 

I think you'll find the challenging next part to be adapting where we render the diagram as overlapping(ish) lines - e.g. it displays a stop on branch A directly next to a line depicting branch B extending down the page. We piggybacked on the `stop_data` construct to deduce how many lines are shown next to a stop to inform us on how exactly to "indent" the stop to make space for the passing branches - and I'm not sure how to translate that from this polytree structure.

image

I'm not sure whether you are going to try to use this to replace the `line_diagram` portion of the `schedule_page_data`. But if you do I just wanted to note that the line diagram data contains route stops but also lumps in alerts as well as whether a stop is experiencing a disruption. Something to consider later maybe.

thecristen (approved):

PR #1436, opened 2022-10-06
fix: account for services with empty `valid_days`

thecristen:
**Problem:** instead of selecting weekday service as the default on a weekday, we were finding cases where Saturday is selected instead.

This seems to be happening when the typical weekday service that _should_ be selected happens to have an empty value for `valid_days`. One example (typical weekday service for the 1 bus): https://api-v3.mbta.com/services/BUS422-hbc42cl1-Wdy-02

**This PR:** When evaluating whether a given service is "current": for services where `valid_days` is an empty array, skip the valid day check. This falls back to instead checking whether the service is within the rating dates, which should generally produce a favorable result.

### DIFF
--- a/apps/site/assets/ts/helpers/service.ts
+++ b/apps/site/assets/ts/helpers/service.ts
@@ -74,7 +74,8 @@ export const isCurrentValidService = (
   }
 
   // check on valid_days
-  if (!isOnValidDay(service, currentDate)) {
+  // disregard if empty valid_days value
+  if (service.valid_days.length > 0 && !isOnValidDay(service, currentDate)) {
     return false;
   }
 
--- a/apps/site/assets/ts/helpers/service.ts
+++ b/apps/site/assets/ts/helpers/service.ts
@@ -74,8 +74,7 @@ export const isCurrentValidService = (
   }
 
   // check on valid_days
-  // disregard if empty valid_days value
-  if (service.valid_days.length > 0 && !isOnValidDay(service, currentDate)) {
+  if (!isOnValidDay(service, currentDate)) {
     return false;
   }
 
--- a/apps/site/lib/site_web/controllers/schedule/line_controller.ex
+++ b/apps/site/lib/site_web/controllers/schedule/line_controller.ex
@@ -99,7 +99,6 @@ defmodule SiteWeb.ScheduleController.LineController do
   @spec dedup_similar_services([Service.t()]) :: [Service.t()]
   def dedup_similar_services(services) do
     services
-    |> Enum.reject(&(&1.valid_days == [5] || &1.valid_days == [1, 2, 3, 4]))
     |> Enum.group_by(&{&1.type, &1.typicality})
     |> Enum.flat_map(fn {_, service_group} ->
       Enum.reject(service_group, &service_completely_overlapped?(&1, service_group))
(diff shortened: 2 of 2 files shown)


apps/site/assets/ts/helpers/service.ts
```diff
@@ -74,7 +74,8 @@ export const isCurrentValidService = (
   }
 
   // check on valid_days
-  if (!isOnValidDay(service, currentDate)) {
+  // disregard if empty valid_days value
+  if (service.valid_days.length > 0 && !isOnValidDay(service, currentDate)) {
     return false;
   }
```
kotva006, line 80: Minor:
Refactor to just return the boolean (I think I applied DeMorgans Law correctly)
```suggestion
  return (service.valid_days.length <= 0 || isOnValidDay(service, currentDate));
```

thecristen: Thanks. I ended up rethinking this and going with a different approach :weary:

thecristen:
Revised comment: We were filtering out too many services in some cases where there were multiple services covering the same start/end dates but differing by applicable `valid_days`. Latest commit compares the `valid_days` in order to determine whether services are _**actually**_ overlapping.

This produces successful display of the Weekday schedule for the 505 bus and successful selection of the relevant schedule (today, the "Monday-Thursday" schedule) for the 1 bus -- deploying on dev-green now so we can QA more cases.

thecristen:
Targets this Asana ticket: [Saturday service selected by default for some bus routes](https://app.asana.com/0/385363666817452/1203119019966533/f)

lund5000:
**TL;DR: Current state of this PR isn't perfect, but it would allow riders to view weekday schedules on some routes they currently can't. I think we should deploy as-is today unless there's an obvious and quick fix.** 

I tested the latest version of this on dev-green just now, comparing the top 20 most visited bus route schedules over the last week to production. I also tested the 501 and 505, which were known to be a problem before the initial fix to handle these split weekday schedules.

The current state of production for the routes listed in the [Asana task](https://app.asana.com/0/385363666817452/1203119019966533/f) is different now than it was 12 days ago. At that time the weekday schedule was an option, but not selected by default. Today, the weekday schedules are no longer presented as options for routes like 1, 47, and 66. 

From what I've tested again and the service patterns we saw two weeks ago, this implementation:
* displays different Monday-Thursday and Friday weekday options for routes that have M-R and F services but not M-F (e.g., 1, 47, 66) 
* displays a single weekday option for routes that have M-R, F, and M-F services as well as weekend services (e.g., 8, 57, 60)
* displays a single weekday option for routes that have M-R, F, and M-F services but no weekend services (e.g., 501, 505, CT2)

Ideally those routes with service patterns like 47 or 66 would display a single weekday option, but presenting multiple options is better than none. **Can we get this deployed today and come back again for another round of polish and testing later?**

Quick links:
* missing weekday on prod: https://www.mbta.com/schedules/1/line?schedule_finder%5Bdirection_id%5D=0&schedule_finder%5Borigin%5D=place-nubn
* split weekdays on dev-green: https://dev-green.mbtace.com/schedules/1/line?schedule_finder%5Bdirection_id%5D=0&schedule_finder%5Borigin%5D=place-nubn

kotva006 (approved):

PR #1437, opened 2022-10-11
refactor: format files in pre-commit hook with `lint-staged`

thecristen:
No ticket. This is proposed as an improvement to #1398. I found the prior implementation did not permit me to stage and commit certain lines of file changes - it would let me stage partial changes but ultimately commit the entire set of changes.

I found [`lint-staged`](https://github.com/okonet/lint-staged), which seems to be a more fully developed implementation of the same idea.

### DIFF
--- a/apps/site/assets/package.json
+++ b/apps/site/assets/package.json
@@ -102,7 +102,7 @@
     "mocha-jsdom": "2.0.0",
     "postcss-loader": "^6.2.1",
     "postcss-preset-env": "^7.4.3",
-    "prettier": "^1.15.3",
+    "prettier": "1.19.1",
     "purgecss-webpack-plugin": "^4.1.3",
     "react-test-renderer": "^16.8.6",
     "regenerator-runtime": "^0.13.7",
--- a/apps/site/assets/package.json
+++ b/apps/site/assets/package.json
@@ -96,6 +96,7 @@
     "focus-options-polyfill": "^1.1.1",
     "husky": "^8.0.1",
     "jest": "^27.0.1",
+    "lint-staged": "^13.0.3",
     "mini-css-extract-plugin": "^2.6.0",
     "mocha": "^9.2.2",
     "mocha-jsdom": "2.0.0",
--- a/apps/site/assets/package.json
+++ b/apps/site/assets/package.json
@@ -135,6 +135,7 @@
     "eslint:js": "git diff --name-only --diff-filter=dx origin/master... | grep js/.*\\.js | awk -F/js '{print \"js\"$2}' | xargs eslint -c .eslintrc.js",
     "eslint:js:fix": "npm run eslint:js -- --fix",
     "eslint:ts": "cd ts && eslint -c .eslintrc.js --ext .ts,.tsx --max-warnings=0 .",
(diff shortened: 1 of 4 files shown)


apps/site/assets/package.json
```diff
@@ -134,6 +135,7 @@
     "eslint:js": "git diff --name-only --diff-filter=dx origin/master... | grep js/.*\\.js | awk -F/js '{print \"js\"$2}' | xargs eslint -c .eslintrc.js",
     "eslint:js:fix": "npm run eslint:js -- --fix",
     "eslint:ts": "cd ts && eslint -c .eslintrc.js --ext .ts,.tsx --max-warnings=0 .",
+    "eslint:ts:fix": "cd ts && eslint -c .eslintrc.js --ext .ts,.tsx --fix .",
```
thecristen, line 138: I didn't keep the `max-warnings` option since I feel like that's more of an error to catch in CI, but can put it back if desired.

package.json
```diff
@@ -24,5 +24,8 @@
     "cypress:open": "PORT=4004 $(npm bin --prefix apps/site/assets)/cypress open",
     "test:cypress:run": "$(npm bin --prefix apps/site/assets)/start-server-and-test 'MIX_ENV=test PORT=4004 mix phx.server' 4004 cypress:run",
     "test:cypress:open": "$(npm bin --prefix apps/site/assets)/start-server-and-test 'MIX_ENV=test PORT=4004 mix phx.server' 4004 cypress:open"
+  },
+  "lint-staged": {
+    "**/*.{ex,exs,eex}": "mix format"
```
thecristen, line 29: `mix` only works at top level! I _think_ it would still work if I had included it in the other `package.json` file by writing it as `"../../../**/*.{ex,exs,eex}": "mix format"` but I felt like this both looked cleaner and was easier to understand.

thecristen (commented):
I debated removing the format checks for CI but didn't assuming that not all formatting failures are fixable. So leaving the CI checks as they are will make it so whatever's not corrected in the pre-commit hook can get flagged in CI.

apps/site/assets/package.json
```diff
@@ -102,7 +102,7 @@
     "mocha-jsdom": "2.0.0",
     "postcss-loader": "^6.2.1",
     "postcss-preset-env": "^7.4.3",
-    "prettier": "^1.15.3",
+    "prettier": "1.19.1",
```
arkadyan, line 105: question: Do configuration defaults really change in minor- or patch-level versions? That would seem like a pretty big breach of semvar.

arkadyan (approved):
👍 The prettier thing seems funky (but I'd also believe it). But fundamentally this looks good and I appreciate the improvement. (It was honestly unusable for me the old way.)

apps/site/assets/package.json
thecristen: Unclear, but their changelog did indicate they include new options/rename options/deprecate old options in minor versions. 🥲

We were technically probably safe anyway since Prettier is on v2 and I don't think they'll update v1, but I wanted to make sure we reflect best practice anyway. Buried somewhere in their documentation -- sorry for not linking it -- was a recommendation to use exact version.

arkadyan: 👌

PR #1438, opened 2022-10-11
Liquid replacements for monthly passes

thecristen:
#### Summary of changes
**Asana Ticket:** [Liquid replacements for monthly passes](https://app.asana.com/0/385363666817452/1203017920930544/f)

Edited fare info to add some missing information needed to generate the shortcodes used in the CMS to render certain fare information, and added a test case for each:

* Added a monthly reduced price for `:commuter_ferry` fare
* Edited the unused monthly reduced price for express buses, and added this `:express_bus` fare


i0sea (approved):
🚀

PR #1441, opened 2022-10-14
Integrate line diagram with new data structure to display branches

arkadyan:
#### Summary of changes
**Asana Ticket:** [Integrate line diagram with new data structure to display branches](https://app.asana.com/0/555089885850811/1203082255397796)

---


arkadyan:
@thecristen My condolences for getting the short straw…

apps/site/assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -248,16 +273,25 @@ const ScheduleDirection = ({
...
-          lineState.data[0] && (
-            <LineDiagramAndStopListPage
-              lineDiagram={lineState.data}
-              route={route}
-              directionId={state.directionId}
-            />
-          )
-        : null}
+
+      {/* OLD LINE DIAGRAM */}
```
arkadyan, line 277: My plan—either in this PR or a follow-up—is to remove all of the old components and rename the new ones to remove the "Tree" prefix. I thought that at least for the initial review though it might be helpful to leave them for reference since most of this is taken as directly as possible.

thecristen (changes requested):
This is monumental! I had some small comments here or there but overall good stuff 👑 

(edited: removed some notes around other routes, those weren't specific to this PR)

### And a few requests to keep this consistent with the original line diagram:

1. Can we adjust the line diagram to not label the ends of single-line routes? See example superfluous "Oak Grove Branch" label:
image

1. Example from the Green Line B showing B among connections - I don't think it should show itself in the connections list.
image

1. In the original line diagram, expanded portions of the line are actually shown in a thicker line (`stroke-width` of `16px` vs `10px`), e.g. 
image



Also, I did notice the failing Jest test coverage, but I'm under the impression that'll resolve itself in the old line diagram cleanup work.

thecristen (changes requested):
Coming back after verifying some other problems (had to make sure they are not specific to dev-blue API). This one is pretty major.

* **CR-Franklin isn't showing at all**. The data appears to be in the backend, and there's a frontend error:

image


image



* **CR-Newburyport is missing its branches and only displaying the trunk**. I did a little debugging and they seem to be in the backend including in the polytree structure - but I haven't verified if they're a correct polytree structure, or why it's not showing on the frontend.
image

apps/alerts/lib/alert.ex
```diff
@@ -255,3 +255,53 @@ defmodule Alerts.Alert do
...
+  defp alert_active_period({first, last}) do
+    [first, last] |> Enum.map(&format_time(&1))
+  end
+
+  defp format_time(t) do
+    case Timex.format(t, "{YYYY}-{M}-{D} {h24}:{m}") do
+      {:ok, formatted_time} -> formatted_time
+      _ -> nil
+    end
+  end
```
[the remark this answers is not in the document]
arkadyan: No, this was intentional—I should have made a comment. In order to do that we would have to include `site` as a dependency of `alert` which we wouldn't want. These could be moved somewhere central and shared, but I think the other usage is going to fall away upon cleanup, so that didn't seem to make sense to bother with.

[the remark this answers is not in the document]
thecristen: Ah ok-- thanks!

arkadyan:
@thecristen I think this is ready for a final review. The last 3 commits are big, but they should have no actual changes—everything removed wasn't being used any longer.

thecristen (changes requested):
Sorry to be all like "but, Foxboro!" but... https://dev-blue.mbtace.com/schedules/CR-Franklin/line this line's pretty funky! Let's fix that.

image


Another thing, not enough for me to consider a blocker, but the branch in Newburyport/Rockport is a bit farther from the main line than I'd expect:
image

The other routes I've seen look great! Will add more code comments soon, just wanted to flag the one troublesome line diagram I found first.

apps/site/lib/site_web/controllers/schedule/line_controller.ex
```diff
@@ -71,12 +69,6 @@ defmodule SiteWeb.ScheduleController.LineController do
         route_patterns: conn.assigns.route_patterns,
         stop_tree: conn.assigns.stop_tree,
         alerts: conn.assigns.alerts,
-        line_diagram:
-          update_route_stop_data(
-            conn.assigns.all_stops,
```
thecristen, line 75: **question**: does this mean we can remove `plug(SiteWeb.ScheduleController.AllStops)` from this controller?

thecristen (commented):
It turns out I didn't have many additional comments. I wonder what's behind those strange commuter rail diagrams.. 🤔

apps/site/lib/site_web/controllers/schedule/line_controller.ex
arkadyan: I don't think so, or I guess more accurately I don't know enough to be able to say that. The `AllStops` plug is a generic plug that is used in many places and actually gets all the stops for a route, just like it sounds like it would. Confusingly, just for this controller in [line.ex](https://github.com/mbta/dotcom/pull/1441/files#diff-7091bf01cdd3f25d37f063b7a084da7e63b948517b44497a6024e5de5262720aL147) we were overwriting that value with a completely different, unrelated data that had nothing to do with the more general `all_stops`. So I guess maybe? But if so, it never should have been there in the first place, which makes me uncertain because I don't know whether there would be unintended consequences.

arkadyan:
@thecristen My guess is that the issue with the Franklin line is that the overlapping "trunk" section of the two branches actually differs—one branch skips the Forest Hills stop. It looks like in production we paper over that. So I made [one small fix to the library](https://github.com/arkadyan/unrooted_polytree/pull/3) (as a side note we might want to transfer that at some point), and I'm looking into the best way to tweak the data so both branches include FH.

arkadyan:
Fixed Franklin line

Screenshot 2022-11-15 at 14 26 56
Screenshot 2022-11-15 at 14 27 08

Looking into Providence next. Hoping the same fix might work.

arkadyan:
@thecristen Fixed Providence as well!

Screenshot 2022-11-15 at 14 58 14

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -71,23 +71,79 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
+
+    min..max
+  end
+
+  @spec largest_trunk([{[RouteStop.t()], Range.t()}]) :: [RouteStop.t()]
+  defp largest_trunk(route_stop_lists_with_trunk_ranges) do
+    largest_trunk_range =
+      route_stop_lists_with_trunk_ranges
+      |> Enum.map(&elem(&1, 1))
+      |> Enum.max()
```
thecristen, line 139: Technically this isn't so much largest trunk range (I am interpreting that to mean "longest"), so much as it is range with the highest indices. Is this the intent?

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -71,23 +71,79 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
+    Enum.map(route_stop_lists_with_trunk_ranges, fn {route_stops, trunk_range} ->
+      old_trunk = Enum.slice(route_stops, trunk_range)
+
+      if old_trunk == largest_trunk do
+        # No need to replace anything
+        route_stops
+      else
+        route_stops
+        |> Enum.reject(&Enum.member?(old_trunk, &1))
+        |> List.insert_at(trunk_range.first, largest_trunk)
```
thecristen, line 115: man, this line diagram surgery is 🔥

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
```diff
@@ -71,23 +71,79 @@ defmodule SiteWeb.ScheduleController.Line.Helpers do
...
-
-    route_patterns
+    route.id
+    |> get_line_route_patterns(direction_id, route_pattern_id)
     |> filtered_by_typicality(route, route_pattern_id)
     |> Enum.map(&stops_for_route_pattern/1)
     |> maybe_use_overarching_branch()
   end
 
+  @routes_with_trunk_discrepancies ~w(CR-Franklin CR-Providence)
```
thecristen, line 90: It would be nice if we could detect trunk discrepancies instead of hard-coding the routes, but I also like this because it ensures this special case code doesn't affect other routes.

thecristen (approved):
Just a small question, but it works and is clear. Thanks for this!

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
arkadyan: I think that's correct. Let me phrase it another way to double-check…

Each trunk begins and ends at the same stops, but in between, the "largest" one hits the greatest number of stops.

I'm very happy if you have ideas for a better name for that.

apps/site/lib/site_web/controllers/schedule/line/helpers.ex
arkadyan: Yeah, in theory / eventually I definitely agree. But as you say, I was afraid of causing any unintended side effects on other routes. Maybe eventually we can build up enough test cases to be confident that we are able to detect this one form of discrepancy without any false positives.

PR #1444, opened 2022-10-20
restore Fl/vote map functionality

i0sea:
restore vote map


thecristen (approved):
Noticed the last deploy seemed to have trouble with the `npm install` step -- I think that was my bad after having updated the master branch with a version of the `package-lock.json` file that the `Dockerfile` on this branch wasn't expecting -- I'm hoping to  rectify that in #1449  

Approving anyway because the restored code was already deployed and looks good! 

I think we'll have to either (1) incorporate the changes from #1449 or (2) run `npm run install` again to update the `package-lock.json` on _this_ branch and commit the changes to get it all working. No preference which option to go with there. :)

PR #1445, opened 2022-10-20
fix: Ensure that alerts exists before checking the length

arkadyan:
#### Summary of changes
**Asana Ticket:** [Transit Near Me does not render nearby locations](https://app.asana.com/0/555089885850811/1203202409417379)

---

### DIFF
--- a/apps/site/assets/ts/tnm/components/RouteCard.tsx
+++ b/apps/site/assets/ts/tnm/components/RouteCard.tsx
@@ -44,7 +44,7 @@ const RouteCard = ({
     return null;
   }
 
-  const hasAlert = route.route.alerts.length > 0;
+  const hasAlert = route.route.alerts && route.route.alerts.length > 0;
 
   return (
     <div className="m-tnm-sidebar__route" data-mode={mode}>


lund5000 (approved):

i0sea:
Was going to comment the map still doesn't load and perhaps ` const hasAlert = route.route.alerts?.length > 0;` would work?
https://dev-green.mbtace.com/transit-near-me?_utf8=%E2%9C%93&location%5Baddress%5D=10+Park+Plz%2C+Boston%2C+MA+02116-3900%2C+United+States&location%5Blatitude%5D=&location%5Blongitude%5D=

arkadyan:
@i0sea Yeah, I mentioned in Slack that the deploying isn't working for me, so what you are seeing on dev-green is still the old code. Not sure what's up with that… As for the syntax suggestion though, that's a good suggestion, I definitely like that better. I can put up a follow-up.

apps/site/assets/ts/tnm/components/RouteCard.tsx
```diff
@@ -44,7 +44,7 @@ const RouteCard = ({
     return null;
   }
 
-  const hasAlert = route.route.alerts.length > 0;
+  const hasAlert = route.route.alerts && route.route.alerts.length > 0;
```
kotva006, line 47: nit-pick: You could use the nullish operator here `routes.route.alerts?.length > 0`

PR #1446, opened 2022-10-20
feat: Script for geocoding a polling places CSV file

arkadyan:
#### Summary of changes
**Asana Ticket:** [Geo-match polling places](https://app.asana.com/0/555089885850811/1203042792767525/f)

Attached the output CSV file to the [parent ticket](https://app.asana.com/0/555089885850811/1203042792105805/f).

---

### DIFF
--- a/scripts/geocode_addresses.py
+++ b/scripts/geocode_addresses.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+"""
+Geocode polling addresses from a CSV input file. Adds columns for the clean
+address, latitude, and longitude for each row and writes the results to a new CSV file.
+
+Usage:
+    geocode_addresses.py <infile> <outfile>
+    geocode_addresses.py -h | --help
+
+Arguments:
+    <infile>  The CSV file to use an input.
+    <outfile>  The location to write an output CSV file. If an existing file exists it will be overwritten.
+"""
+
+import csv
+import os
+import googlemaps
+from docopt import docopt
+
+gmaps = googlemaps.Client(key=os.getenv('GOOGLE_API_KEY'))
+
+
+def address_from_csv_row(row):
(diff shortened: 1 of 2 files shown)


scripts/geocode_addresses.py
```diff
@@ -0,0 +1,74 @@
...
+    <infile>  The CSV file to use an input.
+    <outfile>  The location to write an output CSV file. If an existing file exists it will be overwritten.
+"""
+
+import csv
+import os
+import googlemaps
+from docopt import docopt
+
+gmaps = googlemaps.Client(key=os.getenv('GOOGLE_API_KEY'))
```
thecristen, line 20: We've been slowly working our way away from using Google's API, and also I think it'd be nicer to use a free resource if possible and avoid having the register/handle API keys. 🙈 Have you considered the [`geocoder` Python library](https://geocoder.readthedocs.io/index.html)? I notice it happens to have an interface to 
[OpenStreetMap addresses via Nominatum](https://geocoder.readthedocs.io/providers/OpenStreetMap.html).

arkadyan: I haven't looked at that, the ticket just described using Google Maps, but I certainly can. A few things to consider at first glance:

1. It looks like we might have to run our own server to support that
2. The Google results provide us with a "formatted address" that we are relying on. At a glance, I haven't yet seen support for that in this library (but I very well just might not have come across it yet).
3. This is something we'll just be running manually once every 2 years, so the requirements about keys & free feel a little less onerous. Given that I already ran this and got the results for this year, checking in this script is basically about making it easier next time so we don't have to figure everything all out from scratch again. So we might want to consider upgrades like a new library in a separate ticket.

arkadyan:
Combined into https://github.com/mbta/dotcom/pull/1450

PR #1449, opened 2022-10-24
chore: use consistent npm version

thecristen:
No ticket. I was motivated to update this as I noticed in a [deploy from #1444](https://github.com/mbta/dotcom/actions/runs/3299840723/jobs/5443665167) that the Docker image was having trouble following along with the lockfile. In the printed messages it's shown to be using npm v8, which isn't really what we want.
image

The issue: `npm` updated how it creates/works with `package-lock.json` in version 7 - the differences prior and after are pretty substantial, leaving to large diffs and console messages about the lockfile detected as being made with an older/newer version of npm, etc. You may have run into this! 

It turns out I (and maybe also others?) had upgraded our local `npm` versions from 6 to 7 or 8, resulting in the different `lockfileVersion: 2` lockfile. And then whenever a dev using npm@6 updated, the file would go back to `lockfileVersion: 1`, and so on, and so forth. 

There's no programmatic way to "freeze" the local npm version. But to ameliorate this:

1 - Well, added a short note to the docs to discourage upgrading `npm` (and I've downgraded mine back to 6, locally 😬 )
2 - edited the `Dockerfile` so it doesn't upgrade its own `npm`
3 - regenerated/recommitted the 'right', `lockfileVersion: 1` `package-lock.json` file.

Hoping this makes sense!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,6 @@ COPY --from=elixir-builder /root/deps /root/deps
 ADD apps/site/assets /root/apps/site/assets
 
 WORKDIR /root/apps/site/assets/
-RUN npm install -g npm
 RUN npm ci
 # Create apps/site/priv/static
 RUN npm run webpack:build -- --env SENTRY_DSN=$SENTRY_DSN
--- a/README.md
+++ b/README.md
@@ -154,6 +154,8 @@ any given time, the site may not be compatible with the very latest API version 
     ```
     You won't see a `node_modules` folder at the root of this project -- this installs packages into `apps/site/assets`.
 
+    Minor note - you may see a prompt to upgrade `npm`. This isn't needed, and `"lockfileVersion": 1` in our `apps/site/assets/package-lock.json` file means it was generated with an `npm` version prior to 7.
+
 1.  Build the project:
     ```
     npm run build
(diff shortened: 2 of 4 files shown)


thecristen:
ok.. this turned out to be more of a chore than i expected. but i think i finally got all of it straightened out. this was all worked out by my testing locally.

after first review: `npm ci` was failing in the Dockerfile, it turns out that's because the `package-lock.json` file was generated using the wrong npm version.
and then after fixing that: type checks for several react components were failing. it turns out various dependencies were using different versions for the typings behind react -- `npm --prefix apps/site/assets ls @types/react` produces

```
├─┬ @types/enzyme@3.10.12
│ └── @types/react@16.14.34  deduped
├── @types/react@16.14.34 
├─┬ @types/react-dom@16.9.17
│ └── @types/react@16.14.34  deduped
├─┬ @types/react-leaflet@2.8.2
│ └── @types/react@16.14.34  deduped
├─┬ @types/react-redux@7.1.24
│ ├─┬ @types/hoist-non-react-statics@3.3.1
│ │ └── @types/react@18.0.25 <--- bad!
│ └── @types/react@18.0.25 
└─┬ @types/react-test-renderer@16.9.5
  └── @types/react@16.14.34  deduped
```

and we're not using React 18 yet, so having `@types/react@18.0.25` was checking for certain things we don't have yet. I figured out I could specify the desired version to use in the `package.json` file, and that's what the final commit does. 

sharing here so @i0sea knows what's up ;)

PR #1450, opened 2022-10-25
feat: Script to combine polling places with geo shapes

arkadyan:
#### Summary of changes

**Asana Ticket:** [Update precincts.json](https://app.asana.com/0/555089885850811/1203042792164454/f)

feat: Script to combine polling places with geo shapes

Add documentation on all three scripts.

Depends on https://github.com/mbta/dotcom/pull/1446 and https://github.com/mbta/dotcom/pull/1448. These could be combine into 1 PR if that seemed better.

---

#### General checks

### DIFF
--- a/scripts/POLLING_PLACES.md
+++ b/scripts/POLLING_PLACES.md
@@ -1,0 +1,20 @@
+# Polling Places helper scripts
+
+Some years during election season we enable a page to help riders search for their polling place and get transit directions to it. In order to support this functionality, we need to gather current data on polling places and district shapes each year. There are a few scripts to help automate this.
+
+## geocode_addresses.py
+
+Geocode polling addresses from a CSV input file. Adds columns for the clean
+address, latitude, and longitude for each row and writes the results to a new CSV file.
+
+If the polling places data comes to us as an Excel file it should be saved as a CSV file to feed into this script.
+
+## shp_to_geojson.sh
+
+Convert geo files from .shp to .geojson format.
+
+This script depends on the `ogr2ogr` library from `gdal`. Before you run it you should install gdal from homebrew via `brew install gdal`.
+
+## combine_geocoded_addresses_with_shapes.py
+
+Combine geocode polling place addresses from a CSV file with district shapes from a geojson file. This assembles everything from the previous two scripts into a single `precincts.json` file that the dotcom app can consume.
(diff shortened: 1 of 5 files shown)


lund5000 (changes requested):
148 entries are null instead of polling place objects, probably one for each of the places where we couldn't find a matching district. Those nulls will need to be removed from the final result.

arkadyan:
@lund5000 @i0sea I pushed a commit with all of the manual geocoding fixes to the issues Steve found.

scripts/POLLING_PLACES.md
```diff
@@ -0,0 +1,20 @@
+# Polling Places helper scripts
+
+Some years during election season we enable a page to help riders search for their polling place and get transit directions to it. In order to support this functionality, we need to gather current data on polling places and district shapes each year. There are a few scripts to help automate this.
```
thecristen, line 3: Can you add a note here describing the fact that the scripts don't output 100% correct info and manual adjustment is a must?

scripts/combine_geocoded_addresses_with_shapes.py
```diff
@@ -0,0 +1,141 @@
...
+    district = find_district(districts, town, ward, precinct)
+
+    if (district):
+        path = district['path']
+
+        return dict(town=town,
+                    ward=ward,
+                    precinct=precinct,
+                    lat=lat,
+                    lng=lng,
```
thecristen, line 105: question: should these be `lat=float(lat)` and `lng=float(lng)`?

scripts/shp_to_geojson.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# Convert geo files from .shp to .geojson format.
+# This script depends on the `ogr2ogr` library from `gdal`.
+# Install it from homebrew via `brew install gdal`.
+
+ogr2ogr -f GeoJSON ~/tmp/polling_places/MA_025_Statewide_Precincts_Final/MA_025_Statwide2021PrecinctsFinal.geojson ~/tmp/polling_places/MA_025_Statewide_Precincts_Final/MA_025_Statwide2021PrecinctsFinal.shp
```
thecristen, line 7: I feel kinda weird about hard-coded paths like this but can't think of a better resolution.

thecristen (approved):
it's proven to get it done! given how much manual adjustment we needed afterwards I think it'd be nice to add a note about that. otherwise looks good to me!

lund5000 (approved):

scripts/shp_to_geojson.sh
arkadyan: Yeah, it could be a full-fledged script that takes arguments, but it felt more like "documentation" than something that needed a lot of time spent on in.

PR #1451, opened 2022-10-29
feat: add custom event ‘changeddirection’

thecristen:
#### Summary of changes

**Asana Ticket:** [Schedules | Schedule Finder default direction doesn't match user selection](https://app.asana.com/0/385363666817452/1200234754629197/f)

Establishes a new custom event.

Adds a helper function for dispatching this event to communicate the changed direction across disparate react trees. 

The `useDirectionChangeEvent` hook sets up the event listener on the document and outputs the updated direction after said custom event is fired.

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/hooks/__tests__/useDirectionChangeEventTest.tsx
+++ b/apps/site/assets/ts/hooks/__tests__/useDirectionChangeEventTest.tsx
@@ -1,0 +1,19 @@
+import { fireEvent } from "@testing-library/react";
+import { renderHook, act } from "@testing-library/react-hooks";
+import useDirectionChangeEvent from "../useDirectionChangeEvent";
+
+describe("useDirectionChangeEvent", () => {
+  test("listens to the 'changeddirection' event", () => {
+    const { result } = renderHook(() => useDirectionChangeEvent(0));
+    expect(result.current).toEqual(0);
+
+    act(() => {
+      fireEvent(
+        document,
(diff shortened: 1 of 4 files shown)


arkadyan:
This feels like a lot of overhead for what—if I'm understanding properly—feels like a case of improper state management. Let me know if I'm missing important things here, but it seems like `ScheduleLoader` defines a "directionId" state (`readjustedDirectionId`), which it passes down to both `ScheduleDirection` and `ScheduleFinder`. But then `ScheduleDirection` copies that into its own state, and when a change action happens it modifies its local state. So then the problem is that `ScheduleFinder` relying on the higher-level, `ScheduleLoader` state isn't aware of the changes.

I read the solution here as (more or less) making _another_ local copy of the `directionId` state and adding extra events to synchronize from one local state to the other. What if, instead, we got rid of all local states? So the only concept of `directionId` that `ScheduleDirection` and `ScheduleFinder` have is what is passed into them. And when there's a change direction action in `ScheduleDirection`, it bubbles that up to a handler in `ScheduleLoader` which updates the single source of truth. So essentially, trying to better follow the "data down, actions up" principle. Does that seem reasonable, or am I missing important context?

arkadyan (approved):
@thecristen Thanks for explaining the context around this to me—I'd missed that these were in two separate instantiations of React. Given that, I really like this solution using plain old events! 👍

PR #1454, opened 2022-10-31
Update subway map

i0sea:
#### Summary of changes

**Asana Ticket:** [Update Subway Map](https://app.asana.com/0/385363666817452/1203186251878814/f)
Update subway map image on schedules page


thecristen (approved):
Seems reasonable to me :)

PR #1466, opened 2022-11-17
feat: Populate trip planner origin stops from URL param

arkadyan:
#### Summary of changes

**Asana Ticket:** [/from links: Pre-populate origin stop for trip planner from URL](https://app.asana.com/0/555089885850811/1203192939356021)

Populate trip planner origin stops from the URL path. Parallels how the existing "to" field works. Supports both lat-lon and geocoding addresses.

lat-lon:
Screenshot 2022-11-17 at 11 00 39

geocoding:
Screenshot 2022-11-17 at 11 03 07

---

#### General checks

### DIFF
--- a/apps/site/lib/site_web/controllers/trip_plan_controller.ex
+++ b/apps/site/lib/site_web/controllers/trip_plan_controller.ex
@@ -62,6 +62,76 @@ defmodule SiteWeb.TripPlanController do
     render(conn, :index)
   end
 
+  def from(conn, %{"plan" => _plan} = params) do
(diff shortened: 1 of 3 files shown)


thecristen (approved):
I feel like there's some potential abstractions we could make here, but I like to follow the [rule of three](https://wou.edu/las/cs/csclasses/cs161/Lectures/rulesofthumb.html) so we're good to go!

arkadyan:
@thecristen Agreed. I was thinking that when we eventually support both `from` and `to` concurrently that would probably be the best time to consolidate.

PR #1469, opened 2022-11-29
Destination header in Schedule Finder should be left-aligned

i0sea:
**Asana Ticket:** [Destination header in Schedule Finder should be left-aligned](https://app.asana.com/0/555089885850811/1203117266516157/f)

left align header in modal
Screen Shot 2022-11-29 at 2 06 36 PM

### DIFF
--- a/apps/site/assets/css/_schedule-page.scss
+++ b/apps/site/assets/css/_schedule-page.scss
@@ -400,7 +400,7 @@
   vertical-align: top;
 
   &:last-child:not(:first-child) {
-    text-align: right;
+    text-align: left;
   }
 
   &--expanded {
--- a/apps/site/assets/css/_schedule-page.scss
+++ b/apps/site/assets/css/_schedule-page.scss
@@ -430,6 +430,10 @@
   &--tiny {
     width: 1ch;
   }
+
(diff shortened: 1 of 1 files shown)


thecristen (changes requested):
I'm guessing it was previously set to `right` to place it correctly in **Upcoming Departures**, as that section now looks like 🙈 

image

Also while it does move it to the left this doesn't seem to solve the initial problem of the "Destination" and "Trip Details" showing squished against each other. Probably need a different approach.

image

apps/site/assets/css/_schedule-page.scss
```diff
@@ -563,3 +563,7 @@
     }
   }
 }
+
+.pull-left {
```
kotva006, line 567: nitpick: If this isn't going to be nested or use a BEM name I think this might want to go in a CSS file outside of schedule specific css.
Put it in a file like `utilities.scss`, or nest it under the `schedule-table__cell` class
I say this because I already see we have a class called `pull-left` nested in the `datepicker.scss` file

lund5000:
This is looking good to me in schedule finders across all modes 🎉

PR #1472, opened 2022-12-05
Emergency contact info missing or misplaced in navigation

i0sea:
#### Summary of changes

**Asana Ticket:** [Emergency contact info missing or misplaced in navigation](https://app.asana.com/0/385363666817452/1203103106365016/f)

Put emergency contacts in footer and change positioning in header.
Screen Shot 2022-12-05 at 12 04 50 PM
Screen Shot 2022-12-05 at 12 05 24 PM

### DIFF
--- a/apps/site/lib/site_web/templates/layout/_new_nav_desktop.html.eex
+++ b/apps/site/lib/site_web/templates/layout/_new_nav_desktop.html.eex
@@ -20,25 +20,24 @@
               </div>
               <div class="m-menu-desktop__submenu-section" aria-label="<%= section_name %>">
                 <%= Enum.map(links, &SiteWeb.LayoutView.render_nav_link/1) %>
+                <%= if section_name == "Transit Police" do %>
+                  <div class="m-menu__feature emergency-contacts desktop">
+                    <div class="m-menu__section-heading">Emergency Contacts</div>
+                    <small>24 hours, 7 days a week</small>
+                    <div class="m-menu__feature_phone">
+                      <strong>Transit Police:</strong> <%= tel_link "617-222-1212" %>
+                    </div>
+                    <div class="m-menu__feature_phone">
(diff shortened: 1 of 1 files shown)


apps/site/assets/css/_footer.scss
```diff
@@ -219,6 +219,10 @@ address {
         border-radius: 0;
         margin-bottom: 2px;
         margin-top: 0;
+
+        @include media-breakpoint-up(lg) {
```
thecristen, line 223: This sorta feels like it defeats the purpose of the `.mobile` class name. Do we know why that class is there? Would it make more sense to remove it from the footer?

apps/site/lib/site_web/templates/layout/_new_nav_desktop.html.eex
```diff
@@ -20,6 +20,18 @@
...
+                    <div class="m-menu__section-heading">Emergency Contacts</div>
+                    <small>24 hours, 7 days a week</small>
+                    <div class="m-menu__feature_phone">
+                      <strong>Transit Police:</strong> <%= tel_link "617-222-1212" %>
+                    </div>
+                    <div class="m-menu__feature_phone">
+                      <strong>TTY:</strong> <%= tel_link "617-222-1200" %>
+                    </div>
+                  </div>
+                <%end %>
```
thecristen, line 34: Ah, good move! ➡️ 

I notice a missed opportunity to extract this section (& the identical section in the footer) into its own file, but maybe we can do that if we put this info in a third place.

thecristen (approved):
I had a question, but it works!

apps/site/assets/css/_footer.scss
i0sea: This class is used in the footer(as well as header) so to change the styles just for footer desktop had to add this breakpoint. Might be worth separating out the footer styles but there would be a lot of duplicate css 💭

PR #1476, opened 2022-12-08
Subscribe to streaming updates by routeId and stopId

thecristen:
#### Summary of changes

**Asana Ticket:** [Subscribe to streaming updates by routeId and stopId](https://app.asana.com/0/385363666817452/1203082255397801/f)

> **Warning**
> This was substantially reworked Dec 14! Description has been updated below.

This builds upon the work done in #1421. To that end, the first three commits are directly from Matt's branch there. The fourth commit (`fixup: feat: Parse and save prediction arrival and departure times`) conceptually goes with the changes in the third commit. 

So, the fifth commit (`feat: get streaming predictions by route, stop, direction`) contains the bulk of the task to enable subscribing to a stream of predictions for a specified stop on a route, rather than all predictions for an entire route. This was further enhanced to add a direction argument. This mostly involved adding a stop and direction arguments to wherever we were using only route ID.

The code uses a channel name format of `"predictions:routeId:stopId:directionId"` e.g. `predictions:Red:place-sstat:1`.

The last two commits are additional enhancements that I feel complete the implementation. These:
* Ensure we use the _parent_ stop when we're parsing the streamed predictions.
* Push the predictions to the socket using the "data" event name, just for consistency with our other channel.

Note this is a backend-only implementation. I'll post another PR soon for the frontend piece of this, which will also build upon the frontend pieces of #1421 but will be a bit more significant of a change.

---

#### General checks

### DIFF
--- a/apps/predictions/lib/parser.ex
+++ b/apps/predictions/lib/parser.ex
@@ -1,9 +1,10 @@
 defmodule Predictions.Parser do
-  alias Predictions.Prediction
-
   @moduledoc """
   Functions for parsing predictions from their JSON:API format.
   """
+
+  alias JsonApi.Item
+  alias Predictions.Prediction
 
   @type record :: {
           Prediction.id_t() | nil,
@@ -13,30 +14,83 @@ defmodule Predictions.Parser do
           0 | 1,
           DateTime.t() | nil,
           non_neg_integer,
-          Prediction.schedule_relationship(),
+          Prediction.schedule_relationship() | nil,
           String.t() | nil,
           String.t() | nil,
           boolean
         }
 
-  @spec parse(JsonApi.Item.t()) :: record
-  def parse(%JsonApi.Item{attributes: attributes} = item) do
+  @spec parse(Item.t()) :: record
+  def parse(%Item{} = item) do
     {
       item.id,
       trip_id(item),
       stop_id(item),
       route_id(item),
-      attributes["direction_id"],
-      [attributes["arrival_time"], attributes["departure_time"]] |> first_time,
-      attributes["stop_sequence"] || 0,
-      schedule_relationship(attributes["schedule_relationship"]),
+      direction_id(item),
+      first_time(item),
+      stop_sequence(item),
+      schedule_relationship(item),
       track(item),
-      attributes["status"],
-      departing?(attributes)
+      status(item),
+      departing?(item)
     }
   end
 
-  defp first_time(times) do
+  @spec departing?(Item.t()) :: boolean()
+  def departing?(%Item{attributes: %{"departure_time" => binary}}) when is_binary(binary),
+    do: true
+
+  def departing?(%Item{attributes: %{"status" => binary}}) when is_binary(binary),
+    do: upcoming_status?(binary)
+
+  def departing?(_), do: false
+
+  @spec direction_id(Item.t()) :: 0 | 1
+  def direction_id(%Item{attributes: %{"direction_id" => direction_id}}), do: direction_id
+
+  @spec first_time(Item.t()) :: DateTime.t() | nil
+  def first_time(%Item{
+        attributes: %{"arrival_time" => arrival_time, "departure_time" => departure_time}
+      }),
+      do: first_time_from_arrival_departure([arrival_time, departure_time])
+
+  def first_time(_), do: nil
+
+  @spec schedule_relationship(Item.t()) :: Prediction.schedule_relationship() | nil
+  def schedule_relationship(%Item{attributes: %{"schedule_relationship" => "ADDED"}}), do: :added
+
+  def schedule_relationship(%Item{attributes: %{"schedule_relationship" => "UNSCHEDULED"}}),
+    do: :unscheduled
+
+  def schedule_relationship(%Item{attributes: %{"schedule_relationship" => "CANCELLED"}}),
+    do: :cancelled
+
+  def schedule_relationship(%Item{attributes: %{"schedule_relationship" => "SKIPPED"}}),
+    do: :skipped
+
+  def schedule_relationship(%Item{attributes: %{"schedule_relationship" => "NO_DATA"}}),
+    do: :no_data
+
+  def schedule_relationship(_), do: nil
+
+  @spec status(Item.t()) :: String.t() | nil
+  def status(%Item{attributes: %{"status" => status}}), do: status
+  def status(_), do: nil
+
+  @spec stop_sequence(Item.t()) :: non_neg_integer()
+  def stop_sequence(%Item{attributes: %{"stop_sequence" => stop_sequence}}), do: stop_sequence
+  def stop_sequence(_), do: 0
+
+  @spec track(Item.t()) :: String.t() | nil
+  def track(%{attributes: %{"track" => track}}), do: track
+
+  def track(%{relationships: %{"stop" => [%{attributes: %{"platform_code" => track}} | _]}}),
+    do: track
+
+  def track(_), do: nil
+
+  defp first_time_from_arrival_departure(times) do
     case times
          |> Enum.reject(&is_nil/1)
          |> List.first()
@@ -46,56 +100,28 @@ defmodule Predictions.Parser do
     end
   end
 
-  @spec track(JsonApi.Item.t()) :: String.t() | nil
-  defp track(%{attributes: %{"track" => track}}), do: track
-
-  defp track(%{relationships: %{"stop" => [%{attributes: %{"platform_code" => track}} | _]}}),
-    do: track
-
-  defp track(_), do: nil
-
-  defp departing?(%{"departure_time" => binary}) when is_binary(binary) do
-    true
-  end
-
-  defp departing?(%{"status" => binary}) when is_binary(binary) do
-    upcoming_status?(binary)
-  end
-
-  defp departing?(_) do
-    false
-  end
-
   @spec upcoming_status?(String.t()) :: boolean
   defp upcoming_status?("Approaching"), do: true
   defp upcoming_status?("Boarding"), do: true
   defp upcoming_status?(status), do: String.ends_with?(status, "away")
 
-  @spec schedule_relationship(String.t()) :: Prediction.schedule_relationship()
-  defp schedule_relationship("ADDED"), do: :added
-  defp schedule_relationship("UNSCHEDULED"), do: :unscheduled
-  defp schedule_relationship("CANCELLED"), do: :cancelled
-  defp schedule_relationship("SKIPPED"), do: :skipped
-  defp schedule_relationship("NO_DATA"), do: :no_data
-  defp schedule_relationship(_), do: nil
-
-  defp stop_id(%JsonApi.Item{relationships: %{"stop" => [%{id: id} | _]}}) do
+  defp stop_id(%Item{relationships: %{"stop" => [%{id: id} | _]}}) do
     id
   end
 
-  defp stop_id(%JsonApi.Item{relationships: %{"stop" => []}}) do
+  defp stop_id(%Item{relationships: %{"stop" => []}}) do
     nil
   end
 
-  defp trip_id(%JsonApi.Item{relationships: %{"trip" => [%{id: id} | _]}}) do
+  defp trip_id(%Item{relationships: %{"trip" => [%{id: id} | _]}}) do
     id
   end
 
-  defp trip_id(%JsonApi.Item{relationships: %{"trip" => []}}) do
+  defp trip_id(%Item{relationships: %{"trip" => []}}) do
     nil
   end
 
-  defp route_id(%JsonApi.Item{relationships: %{"route" => [%{id: id} | _]}}) do
+  defp route_id(%Item{relationships: %{"route" => [%{id: id} | _]}}) do
     id
   end
 end
--- a/apps/predictions/lib/parser.ex
+++ b/apps/predictions/lib/parser.ex
@@ -50,10 +50,13 @@ defmodule Predictions.Parser do
   def direction_id(%Item{attributes: %{"direction_id" => direction_id}}), do: direction_id
 
   @spec first_time(Item.t()) :: DateTime.t() | nil
-  def first_time(%Item{
-        attributes: %{"arrival_time" => arrival_time, "departure_time" => departure_time}
-      }),
-      do: first_time_from_arrival_departure([arrival_time, departure_time])
+  def first_time(%Item{attributes: %{"departure_time" => departure_time}})
+      when not is_nil(departure_time),
+      do: parse_time(departure_time)
+
+  def first_time(%Item{attributes: %{"arrival_time" => arrival_time}})
+      when not is_nil(arrival_time),
+      do: parse_time(arrival_time)
 
   def first_time(_), do: nil
 
@@ -90,13 +93,13 @@ defmodule Predictions.Parser do
 
   def track(_), do: nil
 
-  defp first_time_from_arrival_departure(times) do
-    case times
-         |> Enum.reject(&is_nil/1)
-         |> List.first()
-         |> Timex.parse("{ISO:Extended}") do
-      {:ok, time} -> time
-      _ -> nil
+  defp parse_time(prediction_time) do
+    case Timex.parse(prediction_time, "{ISO:Extended}") do
+      {:ok, time} ->
+        time
+
+      _ ->
+        nil
     end
   end
 
--- a/apps/predictions/lib/prediction.ex
+++ b/apps/predictions/lib/prediction.ex
@@ -22,7 +22,7 @@ defmodule Predictions.Prediction do
   @type t :: %__MODULE__{
           id: id_t,
           trip: Schedules.Trip.t() | nil,
-          stop: Stops.Stop.t(),
+          stop: Stops.Stop.t() | nil,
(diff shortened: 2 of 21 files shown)


i0sea (commented):
Does [this](https://github.com/mbta/dotcom/pull/1421) need to be merged first? See some overlap in changes here.

apps/predictions/lib/parser.ex
```diff
@@ -13,30 +14,83 @@ defmodule Predictions.Parser do
           0 | 1,
           DateTime.t() | nil,
           non_neg_integer,
-          Prediction.schedule_relationship(),
+          Prediction.schedule_relationship() | nil,
```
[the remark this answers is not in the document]
thecristen: Not sure, but `Prediction.schedule_relationship()` includes the `nil` type anyway so looks like this change isn't needed. 🤔

thecristen:
Inviting @arkadyan to review in case he has any thoughts or context. Thanks for making the foundation of this work we're finally getting around to revisiting! 😉

thecristen:
@i0sea I've added some more detail to the description and updated everything. Basically this would replace part of the #1421, so we don't expect to merge that in first.

apps/predictions/lib/parser.ex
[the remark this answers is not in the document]
arkadyan: @thecristen Yep, I must have missed that. Agreed that this change isn't needed.

apps/predictions/lib/stream_parser.ex
```diff
@@ -68,10 +68,18 @@ defmodule Predictions.StreamParser do
 
   defp trip(_), do: nil
 
+  # Predictions are generally associated in the API with child stops. Find the
```
arkadyan, line 71: 👍

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -32,12 +36,15 @@ defmodule Predictions.PredictionsPubSub do
...
-    Registry.register(:prediction_subscriptions_registry, registry_key, route_id)
+  @spec subscribe(Route.id_t(), Stop.id_t(), direction_id_t()) :: [Prediction.t()]
+  @spec subscribe(Route.id_t(), Stop.id_t(), direction_id_t(), GenServer.server()) :: [
+          Prediction.t()
+        ]
+  def subscribe(route_id, stop_id, direction_id, server \\ __MODULE__) do
+    key = prediction_key(route_id, stop_id, direction_id)
+    StreamSupervisor.ensure_stream_is_started(key)
+    {registry_key, predictions} = GenServer.call(server, {:subscribe, key})
+    Registry.register(:prediction_subscriptions_registry, registry_key, key)
```
arkadyan, line 47: suggestion (non-blocking): If you wanted to, you could store predictions by route+stop+direction, while leaving the streaming as by route. In other words, the server-client contract could be based on route+stop+direction, while the server-API client would just be based on route. The con here would be that you'd be streaming more data than you need from the API, but the pro would be less churn in starting small streams. I'm not sure what the performance implications would be in practice, but my gut instinct is that there are very likely to be multiple users looking at the same route, even if they are looking at different stops along it.

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -1,20 +1,24 @@
...
   alias Predictions.{Prediction, StreamSupervisor}
   alias Routes.Route
+  alias Stops.Stop
 
-  defstruct predictions_by_route_id: %{}
+  defstruct predictions_by_key: %{}
 
   @type t :: %__MODULE__{
-          predictions_by_route_id: predictions_by_route_id()
+          predictions_by_key: predictions_by_key()
```
arkadyan, line 16: suggestion (non-blocking): At some point (especially if we see performance issues with this), it might be worth considering using an `ets` table for this data. It could use `route`, `stop`, and `direction` as keys.

arkadyan (commented):
I did a little more skimming than going totally in-depth, but this looks pretty great to me. I threw out a couple of comments you can consider and take or leave, whether now or in the future.

apps/predictions/lib/stream_supervisor.ex
```diff
@@ -0,0 +1,100 @@
...
+
+  @spec sses_opts(PredictionsPubSub.prediction_key()) :: Keyword.t()
+  defp sses_opts(route_stop_direction) do
+    [route_id, stop_id, direction_id] = String.split(route_stop_direction, ":")
+
+    filters =
+      "filter[route]=#{route_id}&filter[stop]=#{stop_id}&filter[direction_id]=#{direction_id}"
+
+    path =
+      "/predictions?#{filters}&fields[prediction]=status,departure_time,arrival_time,direction_id,schedule_relationship,stop_sequence&include=route,trip,trip.occupancies,stop&fields[route]=long_name,short_name,type&fields[trip]=direction_id,headsign,name,bikes_allowed&fields[stop]=platform_code"
```
thecristen, line 78: I was reading the [documentation](https://api-v3.mbta.com/docs/swagger/index.html#/Prediction/ApiWeb_PredictionController_index) and realized the API already offers the `time` field where it uses `arrival_time` if available otherwise using `departure_time` (aka implementing our logic for "first" time). Maybe we should use that.

apps/predictions/lib/predictions_pub_sub.ex
thecristen: @arkadyan I ended up doing this!

apps/predictions/lib/predictions_pub_sub.ex
thecristen: We'd have to read/write to the table super frequently no? An interesting thing to keep in mind for sure! Let's see how well this works out without ETS first.

paulswartz: reading/writing super-frequently (especially from different processes) is basically the reason you'd use ETS :)

arkadyan:
Asking Eddie to review instead of me for a fresh pair of eyes since I was involved early on.

apps/predictions/lib/parser.ex
```diff
@@ -13,30 +14,83 @@ defmodule Predictions.Parser do
...
     }
   end
 
-  defp first_time(times) do
+  @spec departing?(Item.t()) :: boolean()
+  def departing?(%Item{attributes: %{"departure_time" => binary}}) when is_binary(binary),
+    do: true
+
+  def departing?(%Item{attributes: %{"status" => binary}}) when is_binary(binary),
+    do: upcoming_status?(binary)
```
lemald, line 45: **question:** could you speak a bit to how the logic in `upcoming_status?` related to whether or not a prediction corresponds to a departure? It looks like it might have something to do with Commuter Rail, but then `upcoming_status?` also checks for the string "away" in the status which looks to me like it might be related to "stops away" on the rapid transit?

apps/predictions/lib/stream_supervisor.ex
```diff
@@ -0,0 +1,84 @@
...
+  @impl DynamicSupervisor
+  def init(_) do
+    DynamicSupervisor.init(strategy: :one_for_one)
+  end
+
+  @spec ensure_stream_is_started(Route.id_t()) :: {:ok, pid()} | :bypassed
+  def ensure_stream_is_started(route_id),
+    do: ensure_stream_is_started(route_id, System.get_env("USE_SERVER_SENT_EVENTS"))
+
+  defp ensure_stream_is_started(_route_id, "false"), do: :bypassed
```
lemald: **question:** do I understand correctly that this means that the application will use SSEs unless the `USE_SERVER_SENT_EVENTS` environment variable is explicitly set to `"false"`?

apps/predictions/lib/stream_supervisor.ex
```diff
@@ -0,0 +1,84 @@
...
+  def start_link(opts), do: DynamicSupervisor.start_link(__MODULE__, opts, name: __MODULE__)
+
+  @impl DynamicSupervisor
+  def init(_) do
+    DynamicSupervisor.init(strategy: :one_for_one)
+  end
+
+  @spec ensure_stream_is_started(Route.id_t()) :: {:ok, pid()} | :bypassed
+  def ensure_stream_is_started(route_id),
+    do: ensure_stream_is_started(route_id, System.get_env("USE_SERVER_SENT_EVENTS"))
```
lemald: **suggestion (non-blocking):** I'm not sure exactly how dotcom is built and deployed and how releases factor into it (or don't), but if you want to be more Elixir-y about this it may be possible to [set the application configuration at runtime](https://hexdocs.pm/mix/1.14/Mix.Tasks.Release.html#module-runtime-configuration) based on the environment variable rather than querying it directly here.

apps/predictions/lib/stream_supervisor.ex
```diff
@@ -0,0 +1,84 @@
...
+    else
+      {:error, {:already_started, pid}} ->
+        {:ok, pid}
+    end
+  end
+
+  @spec sses_opts(Route.id_t()) :: Keyword.t()
+  defp sses_opts(route_id) do
+    path =
+      "/predictions?filter[route]=#{route_id}&fields[prediction]=status,departure_time,arrival_time,direction_id,schedule_relationship,stop_sequence&include=route,trip,trip.occupancies,stop&fields[route]=direction_destinations,direction_names,long_name,short_name,type&fields[trip]=direction_id,headsign,name,bikes_allowed&fields[stop]=platform_code"
```
lemald: **issue:** you probably want to use something like `URI.encode_query/2` here, I don't think any of our route IDs contain problematic characters in practice but per [the GTFS spec](https://gtfs.org/schedule/reference/#field-types) IDs can be arbitrary strings.

apps/predictions/test/predictions_pub_sub_test.exs
```diff
@@ -0,0 +1,168 @@
...
+
+      :sys.replace_state(pid, fn state ->
+        Map.put(state, :predictions_by_route_id, %{@route_39 => [@prediction39]})
+      end)
+
+      {:ok, pid: pid}
+    end
+
+    test "updates the predictions", %{pid: pid} do
+      send(pid, {:update, [@prediction39]})
```
lemald, line 145: **question:** am I reading this correctly that the initial state is `@prediction39` and it's being replaced with `@prediction39`? The test should probably take a modified prediction value to confirm that the update actually works.

apps/site/test/support/env_helpers.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Test.Support.EnvHelpers do
+  @moduledoc "Helpers for reassigning env variables"
+
+  defmacro reassign_env(app, var, value) do
```
lemald, line 4: **praise:** love having this macro handy for writing tests

apps/predictions/lib/prediction.ex
```diff
@@ -25,6 +27,9 @@ defmodule Predictions.Prediction do
           stop: Stops.Stop.t() | nil,
           route: Routes.Route.t(),
           direction_id: 0 | 1,
+          arrival_time: DateTime.t() | nil,
+          departure_time: DateTime.t() | nil,
+          # TODO: Deprecated, should be removed in favor of arrival_time and departure_time -- MSS 2022-09-22
```
lemald, line 32: **suggestion:** I'm not sure how your team tracks this sort of cleanup work, but if this field should be removed eventually is there a ticket to represent that?

apps/predictions/lib/parser.ex
thecristen: That was added six years ago in https://github.com/mbta/dotcom-legacy/pull/1004/ :D 
It look like our current code used it as a way to guess whether a prediction corresponds to a departure, for predictions which have no departure time.
Do you suppose it might no longer be relevant?

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -1,20 +1,24 @@
 defmodule Predictions.PredictionsPubSub do
   @moduledoc """
-  Allow channels to subscribe to prediction streams.
+  Allow channels to subscribe to prediction streams, which are collected into a
+  Map with keys representing route-stop pairs as "<route-id>@<parent-stop-id>"
```
lemald, line 4: **nitpick:** this should be updated to reflect the actual implementation of `prediction_key/3`.

apps/predictions/lib/stream_supervisor.ex
thecristen: Yes, exactly. We happen to use a similar pattern for vehicle data.

apps/predictions/lib/parser.ex
lemald: Ahh okay, I think that the "away" part at least must date from a different era when we didn't yet have minute countdowns on the Green Line but instead displayed [only the number of stops away](https://upload.wikimedia.org/wikipedia/commons/1/1f/Stops_away_countdown_signs_at_Park_Street%2C_January_2016.JPG). So that is likely no longer relevant. And that sign also shows "Approaching" and "Boarding."

lemald: So I think from a subway perspective we can probably ditch this logic. I would just want to make sure there are no unintended consequences with Commuter Rail data since we also use the `status` field there.

apps/predictions/lib/stream_supervisor.ex
lemald: Okay cool, I was sort of confused by it but as long as it's following an established pattern that's fine.

apps/site/lib/site_web/channels/predictions_channel.ex
```diff
@@ -21,7 +21,11 @@ defmodule SiteWeb.PredictionsChannel do
     [route_id, stop_id, direction_id] = String.split(route_stop_direction, ":")
     predictions = predictions_subscribe_fn.(route_id, stop_id, direction_id)
 
-    {:ok, %{predictions: predictions}, socket}
+    # remove old predictions
+    future_predictions = Enum.filter(predictions, &is_in_future?/1)
+    # immediately push current predictions to socket
+    send(self(), {:new_predictions, future_predictions})
+    {:ok, socket}
```
lemald: **question:** I'm a bit confused about this change. Returning `{:ok, %{predictions: predictions}, socket}` should already make the data available to the client as soon as it joins, though it's a little bit confusing because it's accessed in a different way from the usual results of a push. We actually had a bug in Skate caused by this issue but we were able to fix it client-side in mbta/skate#1575.

lemald:
@thecristen @arkadyan okay, it took a while but I think I'm through my initial pass of reviewing!

apps/predictions/lib/parser.ex
lemald: Also this PR has already got a lot of stuff going on and this logic isn't actually being changed as part of it, so I think it would be fine if you punted this to a tech debt task.

apps/predictions/lib/prediction.ex
thecristen: There is now! :D

apps/site/lib/site_web/channels/predictions_channel.ex
thecristen: That's from an older commit -- the current code removes the `future_predictions = Enum.filter(predictions, &is_in_future?/1)` from here and puts it in the `handle_info({:new_predictions, predictions}, socket)` clause, which is also handling some other data transformation. To clarify, we `send(self(), {:new_predictions, predictions})` so that the handle_info can do the data filtering before sending the data out. I hope this helps?

apps/predictions/test/predictions_pub_sub_test.exs
thecristen: 💯 hah, great catch! updated.

apps/predictions/lib/stream_supervisor.ex
thecristen: Oh interesting! 
I've added a commit to use `URI.encode/1` on the URL, but this change only affects URLs we send to `ServerSentEventStage`.

(Our other requests are part of a different workflow that would've been trickier to adjust. But that's using `HTTPoison.Base` which looks like it might have its own mechanism for at least checking the URL is parseable. I suppose it's yet another thing we can handle later!)

apps/predictions/lib/stream_supervisor.ex
thecristen: Yes!! Not ready to tackle this yet.

apps/site/lib/site_web/channels/predictions_channel.ex
thecristen: Ahhhh, thanks for elaborating. That makes sense now. Have added a commit to make this change, great suggestion 👍🏼

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -0,0 +1,239 @@
+defmodule Predictions.PredictionsPubSub do
+  @moduledoc """
+  Allow channels to subscribe to prediction streams, which are collected into a
+  Map with keys representing route-stop-direction triads as
+  "<route-id>-<parent-stop-id>-<direction-id>"
```
lemald, line 5: **super-duper nitpick (non-blocking):** this now matches the format from `prediction_key/3`, but the actual implementation uses colons rather than hyphens.

Also just as a thought, maybe you could just use tuples instead of strings as the map keys? Unless there's some place they need to be serialized as strings, but maybe even them you could do that as a separate step.

PR #1478, opened 2022-12-12
remove slashes from stop name in direction url

i0sea:
#### Summary of changes

**Asana Ticket:** ["Get directions" link broken for stops with slashes](https://app.asana.com/0/385363666817452/1202323182981930/f)

Removes slashes from stop name in get directions link.

### DIFF
--- a/apps/site/assets/ts/stop/components/LocationBlock.tsx
+++ b/apps/site/assets/ts/stop/components/LocationBlock.tsx
@@ -52,7 +52,7 @@ const LocationBlock = ({
         <a
           href={`/trip-planner/to/${latLngString(stop)
             .replace(/\s+/g, "-")
-            .toLowerCase()},${stop.name}`}
+            .toLowerCase()},${stop.name.replace(/\//g, "")}`}
           className="btn btn-primary"
           target="_blank"
           rel="noopener noreferrer"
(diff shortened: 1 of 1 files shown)


thecristen (changes requested):
This works and takes the Trip Planner to the right location but does show the name-sans-slash in the input.

image

I think there's a way to handle this gracefully with URL encoding. Sort of heavy-handed given that it's technically only the `/` character that trips up the Trip Planner controller, but will be more resilient overall.

The steps would be
1 - [encode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) the name that's used for the link. We have some JS code that encodes the entire set of query params via calling 
`QueryHelpers.paramsToString`, might be able to use that or see its implementation for inspiration.
2 - the backend will receive the URL encoded name, so adjust the `SiteWeb.TripPlanController.to` function to decode this back to the human-readable name, probably with one of the `URI` functions [maybe. `URI.decode_query/2`](https://hexdocs.pm/elixir/1.10.4/URI.html#decode_query/2)

thecristen (approved):
🙇🏻‍♀️ I see your super simple solution works and keeps the visual appearance in the Trip Planner! *slow clap*

PR #1481, opened 2022-12-14
Consolidate channel handling

thecristen:
**No ticket**/done in preparation of frontend implementation of prediction streaming work.

I noticed that on the frontend we join channels in two (2) different ways, with #1421 introducing a third way. 😩 I wanted to establish a singular approach to joining channels that works for all three (vehicles in timetable, vehicles in `<Map />`, and soon streaming predictions) use cases.

Because the `<Map />` component and the live vehicle displayed on commuter rail timetables are impacted by this change, I'm going to deploy this somewhere to check for regressions.

### Highlights 

- Combines the completely overlapping functionality from `channels.js` and `Channel.ts` into a singular `channels.ts` module. It works mostly the same as before, with minor differences:
  * Instead of a new Socket being created multiple times, it's created once and assigned to `window.socket` to be available globally.
  * Because we might attempt to join a channel from more than one place, we add a check to avoid joining a channel multiple times.
  - Adjusts `cr-timetable-trains.js` to work with `channels.ts` and updates related tests.
- Use those functions as the basis for a new generic `useChannel` React hook, which adds/removes event listeners to respond to the custom event created in `channels.ts`.
  - Adjusts the `<Map />` component to use that hook for tracking vehicles, and adjust its tests.

> **Warning**
> This is my first time writing generic Typescript functions! Feedback on that appreciated. :)

---

#### General checks


kotva006:
food for thought:  There is a library called [RxJS](https://rxjs.dev/guide/overview) that exists to handle async publication and subscribers of events and data.  It could help simplify things in the future as it appears this is trying to mimic what is supports.

apps/site/assets/ts/app/channels.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export type SocketEvent<DataType> = UpdateEvent<DataType> | RemoveEvent;
+
+const isVehicleChannel = (channelId: string): boolean =>
+  channelId.includes("vehicles:") && channelId !== "vehicles:remove";
+
+const joinChannel = <T>(channelId: string): void => {
+  if (!window.socket) return;
+
+  if (!window.channels[channelId]) {
```
kotva006, line 30: question: This array doesn't need to be initialised because we are using string keys to access it making it like a dict?

apps/site/assets/ts/schedule/components/Map.tsx
```diff
@@ -181,14 +148,15 @@ export default ({
...
-    markers: data.markers
-  });
-  useEffect(() => {
-    setupChannels(channel, dispatch);
-    return () => stopChannels(channel);
-  }, [channel, dispatch]);
+    reducer,
+    {
+      channel,
+      markers: []
```
kotva006, line 156: question: Is markers supposed to be a empty array and not like the old `data.markers`?

thecristen:
Adding some details on how to assess:

Ideally this change isn't noticeable on the frontend. We can compare with the current website to ensure we still

- Show and update trains on timetables (`/schedules/:any_commuter_rail_route_id/timetable`)

image

- Show and update vehicle icons on the schedule page's map (`/schedules/:any_route_id_except_ferry`)

image

apps/site/assets/ts/schedule/components/Map.tsx
thecristen: one would think so. but doing so added weird large icons to the map.

apps/site/assets/ts/app/channels.ts
```diff
@@ -0,0 +1,95 @@
...
+      .join()
+      .receive("error", ({ reason }) =>
+        /* eslint-disable no-console */
+        console.error(`failed to join ${channelId}`, reason)
+      )
+      .receive("ok", () => {
+        console.log(`success joining ${channelId}`);
+        /* eslint-enable no-console */
+        if (isVehicleChannel(channelId)) {
+          const [, route_id, direction_id] = channelId.split(":");
```
cmaddox5, line 47: I've actually not seen this syntax before. Is this the equivalent of underscoring an unused value in Elixir?

apps/site/assets/ts/schedule/components/Map.tsx
```diff
@@ -109,20 +73,20 @@ interface IdHash {
...
 }
+type DataReducerAction = DataFromSocketType & { channel: string };
+type DataReducerType = Reducer<DataReducerState, DataReducerAction>;
 
-export const reducer = (
-  state: State,
-  actionWithChannel: ActionWithChannel
-): State => {
-  const { action, channel } = actionWithChannel;
-  if (channel !== state.channel && action.event !== "setChannel") return state;
```
cmaddox5, line 122: Should this conditional have been removed (or at least the first half now that `setChannel` is not a handled event anymore)? I'm curious if this is the reason for the failing test.

apps/site/assets/ts/app/channels.ts
thecristen: Yep, exactly! It's apparently a valid way to skip assigning variables when destructuring.

apps/site/assets/ts/schedule/components/Map.tsx
thecristen: Good callout. After looking into this a lot today, I am concluding that the channel comparison is no longer needed -- with the logic handled in the `useChannel` hook, it seems the reducer will only be working with data from the specified channel anyway. 

I'm not really sure where the intermixing of data from different channels was experienced or expected, but I couldn't find any problem on the Green Line page (which shows trains from the four distinct routes), and switching directions works as expected (the map doesn't show trains going in both directions, just the selected one.)

Have added an extra commit to clean up and adjust the failing test that's no longer needed.

cmaddox5 (approved):
Looks good!

PR #1482, opened 2022-12-21
cleanup: deprecate old schedule features

thecristen:
#### Summary of changes

No ticket!

Our schedule pages used to have a feature where you could pick an origin stop and destination stop. Here's an example from 2018 (see **Departing from** and **Arriving at** drop-downs):

image

We don't use that anymore, but traces of the relevant code were everywhere. This PR removes that.

Similar story with the `all_stops` list, which is only now used to list stops in the Timetable (where it's overwritten with a different implementation anyway).

With the removal of those, `ExcludedStops` is no longer used. And so on and so forth.

We should be able to deploy this to dev-green or dev-blue and verify that nothing of value was lost.

---

#### General checks


paulswartz:
@thecristen 
Screenshot 2022-12-21 at 8 41 32 AM 
❤️

PR #1489, opened 2023-01-07
fix: use bus name instead of id in meta description

thecristen:
#### Summary of changes

**Asana Ticket:** [Google results should show short names instead of route IDs](https://app.asana.com/0/385363666817452/1202793084291449/f)

Just going through old tickets in the icebox.... this adjusts the description seen when searching Google or somesuch for certain routes that are better known by name rather than route id: CT2, SL4, etc.


kotva006 (approved):
👍

PR #1495, opened 2023-01-18
sentry: log resource 404s

thecristen:
Not much to say here, just added the recommended script.

#### Summary of changes

**Asana Ticket:** [Capture resource 404 errors using Sentry](https://app.asana.com/0/385363666817452/1202665687057796/f)

### DIFF
--- a/apps/site/assets/js/app.js
+++ b/apps/site/assets/js/app.js
@@ -99,6 +99,31 @@ if (window.sentry) {
   });
 }
 
+document.body.addEventListener(
+  "error",
+  event => {
+    if (!event.target) return;
+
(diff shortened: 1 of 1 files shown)


kotva006:
Do we have videos we need to worry about not loading?

thecristen:
@kotva006 ahhh good question. We don't host any videos ourselves but might load links to them sometimes. Will look into it real quick.

thecristen:
@kotva006 The only place I know we show videos on, the occasional event page, displays recorded meeting livestreams using some sort of hackery that doesn't even load a `<video>`... so that makes the answer to your question a No.

kotva006 (approved):
Looks good 👍

PR #1494, opened 2023-01-18
Upgrade Elixir to 1.12

thecristen:
This took me several tries, but eventually I figured out how to keep everything compiling + tests passing while only minimally updating other dependencies. 

With Elixir 1.11 there were many compilation warnings across the entirety of Dotcom - most of this PR's changes successfully address those. There are also compilation warnings coming from Phoenix 1.4, so we ignore those until we can upgrade Phoenix (hopefully soon).

I had to upgrade `phoenix_html` to 2.13.4 in order to successfully compile. 

And thanks @arkadyan for getting this started! :) 

#### Summary of changes

**Asana Ticket:** [Upgrade Elixir to 1.11](https://app.asana.com/0/385363666817452/1203355780667443/f)

UPDATE: While fixing issues decided to update Elixir to 1.12! It wasn't much more of a lift.

PR #1496, opened 2023-01-20
experiment: alternate chromedriver source

thecristen:
Goal is to avoid mismatch between local Chrome version and chromedriver version. Can be used both in CI and locally to download the correct chromedriver version that matches your Google Chrome install.

Ran into that issue while working on #1494.

This doesn't require any changes to a developer's local environment if not running the Elixir integration tests (unit tests with the "wallaby" tag).

We'll see here if CI works!

### DIFF
--- a/.github/workflows/algolia-update.yml
+++ b/.github/workflows/algolia-update.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   algolia:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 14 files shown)


thecristen:
Hm.. need to add a commit to trigger running the Elixir integration tests.

i0sea (approved):
CI errors look cleared up 🚀

PR #1498, opened 2023-01-20
Increase unit test coverage (Elixir)

i0sea:
#### Summary of changes

**Asana Ticket:** [Increase unit test coverage (Elixir)](https://app.asana.com/0/385363666817452/1203597642288053/f)

A few unit tests for elixir files


apps/v3_api/test/stream_test.exs
```diff
@@ -5,7 +5,7 @@ defmodule V3Api.StreamTest do
   describe "build_options" do
     test "includes api key" do
       opts = V3Api.Stream.build_options(path: "/vehicles")
-      assert Keyword.get(opts, :url) == "https://api-dev.mbtace.com/vehicles"
+      assert Keyword.get(opts, :url) == "https://api-dev-green.mbtace.com/vehicles"
```
thecristen, line 8: This one's failing, probably because CI isn't set up to use the dev-green API URL. Please undo this one!

apps/feedback/test/fake_date_time_test.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Feedback.FakeDateTimeTest do
+  @moduledoc false
+  use ExUnit.Case
+
+  describe "utc_now/1" do
+    test "can start the application" do
+      assert Feedback.FakeDateTime.utc_now() ==
+               Enum.at(Tuple.to_list(DateTime.from_unix(1_562_573_043)), 1)
```
thecristen, line 8: `elem(DateTime.from_unix(1_562_573_043), 1)` would be equivalent here; [`elem` can be used to get part of a `Tuple`](https://hexdocs.pm/elixir/1.12/Kernel.html#elem/2).

That said, this test is really just hard-coding the implementation details of `Feedback.FakeDateTime`, which is not so great. Since the module is only used for creating a mock value for testing, it's a good candidate for excluding from the code coverage calculation altogether. 

If wanting to exclude from coverage, it can be done using a [configuration file for `ExCoveralls`](https://github.com/parroty/excoveralls#coverallsjson)at the root of the relevant module - `/apps/alerts/coveralls.json` has a good example of that.

i0sea: Excluded some other mock/helper files from test coverage as well

thecristen (approved):
💯 Thank you!

PR #1505, opened 2023-01-24
Setup bus change page feature flag + initial WIP

thecristen:
#### Summary of changes

**Asana Ticket:** [Set up initial page structure](https://app.asana.com/0/385363666817452/1203811157466553/f)

The new draft  page is set up at `/bus-stop-changes`, hidden behind a new feature flag accessible at `/_flags` (will redirect to /alerts if the flag is not enabled).

### DIFF
--- a/apps/site/lib/site_web/controllers/bus_stop_change_controller.ex
+++ b/apps/site/lib/site_web/controllers/bus_stop_change_controller.ex
@@ -1,0 +1,56 @@
+defmodule SiteWeb.BusStopChangeController do
+  use SiteWeb, :controller
+
+  alias Alerts.Alert
+
+  def index(conn, _params) do
+    if Laboratory.enabled?(conn, :bus_changes) do
+      conn
+      # Don't let Google crawl this page
+      |> put_resp_header("x-robots-tag", "noindex")
+      # |> assign(:meta_description, "TBA")
+      |> assign(:alerts, bus_stop_alerts(conn.assigns.date_time))
+      |> render("index.html", %{})
+    else
+      redirect(conn, to: alert_path(conn, :index))
+    end
+  end
+
+  """
+  Identify the set of alerts most indicative of bus stop relocations and closures.
+  This is a WIP.
+  """
+
+  defp bus_stop_alerts(date_time) do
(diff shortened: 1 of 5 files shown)


apps/site/lib/site_web/controllers/bus_stop_change_controller.ex
```diff
@@ -0,0 +1,56 @@
...
+  defp bus_stop_alerts(date_time) do
+    Alerts.Repo.all(date_time)
+    |> Enum.filter(fn %Alert{
+                        active_period: _active_period,
+                        effect: effect,
+                        lifecycle: _lifecycle
+                      } = alert ->
+      is_closed? = effect in [:stop_closure, :stop_moved, :service_change]
+
+      other_adverse_effects? =
```
i0sea, line 33: Will this list need to be updated occasionally? Are these effects stored somewhere else?

apps/site/lib/site_web/controllers/bus_stop_change_controller.ex
```diff
@@ -0,0 +1,56 @@
...
+          :station_closure,
+          :suspension,
+          :shuttle,
+          :no_service,
+          :facility_issue
+        ]
+
+      stop_set = Alert.get_entity(alert, :stop) |> MapSet.delete(nil)
+      affects_stop? = MapSet.size(stop_set) > 0
+      # if affects_stop?, do: IO.inspect(stop_set, label: alert.id)
```
i0sea, line 52: delete?

i0sea (approved):
looks good to me. Just two non blocking questions.

apps/site/lib/site_web/controllers/bus_stop_change_controller.ex
thecristen: Will do in a future update!

apps/site/lib/site_web/controllers/bus_stop_change_controller.ex
thecristen: Yeah, this will probably change in a future iteration. Keeping everything here for now.

PR #1506, opened 2023-01-25
feat(StopPage): Show a smaller list of stops for smaller pages

kotva006:
#### Summary of changes

**Asana Ticket:** [Condensed Departure Section](https://app.asana.com/0/home/1201123880594717/1203803183465620)

---

#### General checks


apps/site/assets/css/_stop-page.scss
```diff
@@ -215,6 +215,18 @@
...
+    display: none;
+  }
+}
+
+.m-stop-page__departures--regular-screen {
+  @include media-breakpoint-down(sm) {
+    display: none;
+  }
+}
+
```
thecristen, line 229: **suggestion:** I believe the version of bootstrap we're on includes utility classes for hiding/unhiding elements based on breakpoint, in the format `"hidden-**-up"` or `"hidden-**-down"` where ** is the breakpoint name. Give them a try and you might not need this CSS!

apps/site/assets/ts/stop/components/Departures.tsx
```diff
@@ -39,18 +39,38 @@ const filteredByModes = (
...
 
   const handleModeClick = (mode: Mode): void => dispatch(clickModeAction(mode));
 
   const filteredRoutes = filteredByModes(allRoutes(routes), selectedModes);
+  // Only show the first 2 elements of the array (unless filtered or button clicked)
+  const shortFilteredRoutes = filterForSmallScreen(
+    allRoutes(routes),
+    selectedModes,
+    showMore
+  );
```
thecristen, line 73: issue: I'm noticing some weird interactions when I try to use the mode filters on mobile, I think this is causing it to show all routes even when I try to specify a mode. Maybe the logic needs to be something like `shortFilteredRoutes = selectedModes.length ? filteredRoutes : filterForSmallScreen(...)`? 

Might consider a deploy to dev-blue to get some feedback on the interaction.

apps/site/assets/ts/stop/components/Departures.tsx
```diff
@@ -64,20 +84,47 @@ const Departures = ({
...
+                routesAndAlerts
+                  ? routesAndAlerts[routeWithDirections.route.id]
+                  : []
+              }
+            />
+          ))}
+          {!showMore && filteredRoutes.length > 2 && (
+            <div className="d-flex justify-content-center mt-10">
+              <button
+                className="btn btn-secondary"
```
thecristen, line 120: Add the `btn-block` class to this (the mockup has the button as full-width). Also if you move the `mt-10` class to here you won't need the wrapper `<div>`.

thecristen (changes requested):

apps/site/assets/ts/stop/components/Departures.tsx
```diff
@@ -64,20 +84,45 @@ const Departures = ({
...
+              directions={routeWithDirections.directions}
+              stop={stop}
+              alerts={
+                routesAndAlerts
+                  ? routesAndAlerts[routeWithDirections.route.id]
+                  : []
+              }
+            />
+          ))}
+          {!showMore && filteredRoutes.length > 2 && (
```
thecristen, line 117: Maybe check for whether a mode is selected here?
I just ran into the edge case where, on South Station, I select "Rail", it gives me all the rail, BUT still shows the "show 6 more routes" button. 
image

thecristen (changes requested):
Solid! Requesting changes here because you've got some snapshots that need updating with the added `type="button"` attribute.

The only other improvement can I think of

suggestion: The `displayShowMoreButton` function is a great addition but it's a little confusing on what the `showMore` variable means and how that's different from the output of `displayShowMoreButton()`. The naming to me suggests they'd mean the same thing. Maybe it's a bit clearer to rename `showMore` -- what if it's framed as `allRoutesShown` initialized as `false`, and clicking the button makes `() => setAllRoutesShown(true)`. Just some food for thought!

thecristen (approved):

PR #1508, opened 2023-01-27
feat(StopRedesign): Added some placeholder text, and the base page in…

kotva006:
… react

#### Summary of changes
Added some text placeholders data and moved the page to react
Screen Shot 2023-01-27 at 10 53 19 AM

**Asana Ticket:** [Template: Scaffold redesigned page layout](https://app.asana.com/0/555089885850811/1203721784670198)

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/stop/components/StopPageRedesign.tsx
+++ b/apps/site/assets/ts/stop/components/StopPageRedesign.tsx
@@ -6,9 +6,9 @@ const StopPageRedesign = ({
   stopId: string;
 }): ReactElement<HTMLElement> => {
   return (
-    <div>
+    <article>
       {/* Title Bar Div */}
-      <div className="d-flex justify-content-space-between">
+      <header className="d-flex justify-content-space-between">
         <div>
           <h1>{stopId}</h1>
           {/* ICONS GO HERE */}
@@ -16,14 +16,14 @@ const StopPageRedesign = ({
         <div style={{ marginTop: "3.075rem" }}>
           Zone Information PLACEHOLDER
         </div>
-      </div>
+      </header>
       {/* Route and Map Div */}
       <div className="d-flex">
         <div style={{ minWidth: "50%" }}>
           <div>Route schedules & maps / Upcoming Trips PLACEHOLDER</div>
           <div>Filter Pills PLACEHOLDER</div>
-          <div style={{ maxHeight: "250px", overflowY: "auto" }}>
-            <div className="d-flex">
+          <ul style={{ maxHeight: "250px", overflowY: "auto" }}>
+            <li className="d-flex">
               <div className="me-8">210</div>
               <div>
                 <div>Fields Corner</div>
@@ -32,8 +32,8 @@ const StopPageRedesign = ({
                   <div>View Realtime Map</div>
                 </div>
               </div>
-            </div>
-            <div className="d-flex">
+            </li>
+            <li className="d-flex">
               <div className="me-8">210</div>
               <div>
                 <div>Fields Corner</div>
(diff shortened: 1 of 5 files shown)


apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -0,0 +1,113 @@
...
+          <h1>{stopId}</h1>
+          {/* ICONS GO HERE */}
+        </div>
+        <div style={{ marginTop: "3.075rem" }}>
+          Zone Information PLACEHOLDER
+        </div>
+      </div>
+      {/* Route and Map Div */}
+      <div className="d-flex">
+        <div style={{ minWidth: "50%" }}>
```
thecristen, line 22: This is okay for managing the trips/map layout, but there's room for improvement. 

**suggestion:** When we get small enough to hit the `sm` breakpoint, we're left with a void where the map was (pictured below) – it would be better to let the departures list take full width here.
Maybe try using CSS grid? `display: grid; grid-template-columns: 1fr auto;` on the wrapping div will show the departures and map side by side but also let departures take full width once the map is hidden.

image

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -0,0 +1,113 @@
...
+          <div>Route schedules & maps / Upcoming Trips PLACEHOLDER</div>
+          <div>Filter Pills PLACEHOLDER</div>
+          <div style={{ maxHeight: "250px", overflowY: "auto" }}>
+            <div className="d-flex">
+              <div className="me-8">210</div>
+              <div>
+                <div>Fields Corner</div>
+                <div className="d-flex">
+                  <div>Open Schedule</div>
+                  <div>View Realtime Map</div>
```
thecristen, line 32: non-blocking, but if you just want to show things side-by-side you don't need a `display: flex` wrapper, you can use `<span>` or any other [inline element](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements).

thecristen (changes requested):
Left a few comments. 

Overall I'd like to see more usage of [more specific HTML elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories). The overall layout can probably be a single [`<article>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) containing a [`<header>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header) element for the stop name/icons/zone section, the other content might comprise one or more `<section>` elements, the departures can probably be depicted in a list element, and so on. I'm not too concerned with the semantics of the placeholders themselves yet, but the layout doesn't have to exclusively use `<div>`.

apps/site/assets/ts/stop/stop-loader-redesign.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import StopPageRedesign from "./components/StopPageRedesign";
+
+const render = (): void => {
+  // TODO is there a better way to get this (from the pheonix page perhaps?)
```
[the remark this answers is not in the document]
kotva006: On [thing we did in pick assist](https://github.com/mbta/draft/blob/main/lib/draft_web/templates/layout/app.html.eex) was embed the data in `<meta>` tags.  This allows for passing of static data to the webpage itself, similar to the json blob but a much smaller scale.

I was tempted to go that route, as it seems cleanest, and allows quick data transfers.

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
kotva006: I am hesitant to spend too much time styling this.  As the styling feels really up in the air still, design and coding wise.
I think the CSS should be tweaked after more of the child components get coded.

For this specific case I think it looks like display width just needs to be set to 100%.  Also that wouldn't necessarily fix the void, as there is nothing to fill the white space (that list stacks).  With more filled in components it would be better to see how to fix that void.  (Maybe center the list idk).

thecristen (approved):
Thanks for taking into account the HTML tags -- since I know we'll iterate on this I'll approve and merge this in.

PR #1509, opened 2023-02-01
Backend: Fetch relevant routes for stop

i0sea:
#### Summary of changes

**Asana Ticket:** [Backend: Fetch relevant routes for stop](https://app.asana.com/0/385363666817452/1203721784670208/f)

Fetch routes and connecting routes for current stop. Would `stop.id` be more useful for html?

### DIFF
--- a/.github/workflows/algolia-update.yml
+++ b/.github/workflows/algolia-update.yml
@@ -1,9 +1,17 @@
 name: Update Algolia index
 
 on:
+  workflow_call:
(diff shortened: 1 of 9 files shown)


thecristen (approved):
I know we had some questions around approach but we can iterate on this -- this is fine to achieve the task as written!

thecristen:
Ended up merging this manually, thanks!

PR #1511, opened 2023-02-02
Add ESLint recommended configurations

thecristen:
#### Summary of changes

Initially went to add `eslint-plugin-jsx-a11y` then realized we already had it. Then realized we hadn't enabled many of the recommended rules that come from the many ESLint plugins we're using, so went ahead and updated that. Made some minor updates to existing files as a result.

---

#### General checks


i0sea (approved):
lgtm

PR #1512, opened 2023-02-02
feat(StopAlerts): Added endpoints to fetch alerts by stop_id and rout…

kotva006:
…e_ids

Added an endpoint to get alert information based off of a stop_id or route_ids

**Asana Ticket:** [Backend: Departure Status/Alerts](https://app.asana.com/0/555089885850811/1203721784670229)

---

#### General checks

### DIFF
--- a/apps/alerts/lib/cache/store.ex
+++ b/apps/alerts/lib/cache/store.ex
@@ -55,6 +55,11 @@ defmodule Alerts.Cache.Store do
     :ets.select(:route_id_and_type_to_alert_ids, keys)
   end
 
+  def alert_ids_for_stop_id(stop_id) do
+    keys = [{{stop_id, :"$1"}, [], [:"$1"]}]
+    :ets.select(:stop_id_to_alert_ids, keys)
+  end
+
   @doc """
   Retrieves the alert objects given a list of alert IDs. If an ID
   is passed that doesn't have a current alert, it is ignored.
@@ -116,6 +121,8 @@ defmodule Alerts.Cache.Store do
         read_concurrency: true
       ])
 
+    _ = :ets.new(:stop_id_to_alert_ids, [:bag, :protected, :named_table, read_concurrency: true])
+
     # no cover
     _ = :ets.new(:alert_banner, [:set, :protected, :named_table, read_concurrency: true])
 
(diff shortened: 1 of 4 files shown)


apps/site/lib/site_web/controllers/alert_controller.ex
```diff
@@ -30,6 +30,22 @@ defmodule SiteWeb.AlertController do
...
 
+  # TODO revisit this later to see if it is used/needed
+  @spec show_by_stop(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def show_by_stop(conn, %{"stop_id" => stop_id}) do
+    alerts = Repo.by_stop_id(stop_id)
+    json(conn, alerts)
+  end
+
+  @spec show_by_routes(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def show_by_routes(conn, _) do
```
thecristen, line 41: You can pattern match on `query_params` if you wanna -- `%Plug.Conn{query_params: query_params} = conn`

Anyway, please add a function clause for handling when the `route_ids` query param is missing.

apps/alerts/lib/cache/store.ex
```diff
@@ -133,13 +140,30 @@ defmodule Alerts.Cache.Store do
         }
       end)
 
+    stop_inserts =
+      Enum.flat_map(alerts, fn alert ->
+        non_nil_stops =
+          Enum.filter(alert.informed_entity.stop, fn stop_id ->
+            stop_id != nil
+          end)
```
thecristen, line 148: Do alerts really have `informed_entity.stop` values of mixed nil and non-nil IDs? Seems a bit odd to me.

apps/alerts/lib/cache/store.ex
```diff
@@ -116,6 +121,8 @@ defmodule Alerts.Cache.Store do
         read_concurrency: true
       ])
 
+    _ = :ets.new(:stop_id_to_alert_ids, [:bag, :protected, :named_table, read_concurrency: true])
```
thecristen, line 124: Good table type for this since you record lots of alerts for each stop!

apps/alerts/lib/repo.ex
```diff
@@ -38,6 +38,13 @@ defmodule Alerts.Repo do
     |> Store.alerts(now)
   end
 
+  @spec by_stop_id(String.t(), DateTime.t()) :: [Alert.t()]
+  def by_stop_id(stop_id, now \\ DateTime.utc_now()) do
+    stop_id
+    |> Store.alert_ids_for_stop_id()
+    |> Store.alerts(now)
```
thecristen, line 45: I was gonna ask if you'd considered inserting the whole alert into the ETS table as opposed to just the  Alert IDs, but I poked around and noticed the other functions in this module follow that pattern as well. I guess having it this way (get IDs, then use them to look up alerts relative to now) makes sure we don't accidentally display alerts prior to now? 🤔 Interesting...

thecristen (changes requested):
Assorted comments, looking pretty good. It gives me some ideas but since this is early development this will be good to start with. Please adjust/rename the `/alerts` endpoint so it doesn't error out, thanks!

apps/alerts/lib/cache/store.ex
kotva006: That is what I noticed in testing this.

apps/alerts/lib/repo.ex
kotva006: Yeah I was following the patterns here mainly.  I do think this could benefit from some sort of in memory db, then we have one alert table to query on.  Instead of generating multiple view types.

thecristen (approved):
🔔 great!

PR #1515, opened 2023-02-07
feat(StopRedesign): Route Filter Pills

kotva006:
#### Summary of changes

Added filter pills to filter the departure/routes list on the new stop redesign page

**Asana Ticket:** [Filter Routes or Departures by Mode](https://app.asana.com/0/385363666817452/1203721784670225)

---

#### General checks


apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -1,10 +1,81 @@
-import React, { ReactElement } from "react";
+import React, { ReactElement, useEffect, useState } from "react";
+import Pill from "./Pill";
+import { Mode } from "../../__v3api";
+import { uniqueId } from "lodash";
+
+export const BUS = "bus";
+export const SUBWAY = "subway";
+export const FERRY = "ferry";
+export const COMMUTER_RAIL = "commuter_rail";
```
kotva006, line 9: I was thinking of putting these in their own Mode model file.  Right now Mode typing lives in the `v3api` data file.  Those file types don't support variable declarations/exporting, so it would need to move as well.

thecristen: Would something like this in the v3api file be what you're getting at?

```typescript
export enum ModeEnum {
  commuter_rail,
  subway,
  bus,
  ferry
}
 
export type Mode = keyof typeof ModeEnum;
// export type Mode = "commuter_rail" | "subway" | "bus" | "ferry";
```

Then you can use the enum value as your constant (e.g. your `SUBWAY` would be `ModeEnum.subway` here.

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -1,10 +1,81 @@
...
+  const [hasFerryRoutes, setHasFerryRoutes] = useState<boolean>(false);
+
+  useEffect(() => {
+    setHasBusRoutes(filterDeparturesByMode(departures, BUS).length > 0);
+    setHasCommuterRailRoutes(
+      filterDeparturesByMode(departures, COMMUTER_RAIL).length > 0
+    );
+    setHasSubwayRoutes(filterDeparturesByMode(departures, SUBWAY).length > 0);
+    setHasFerryRoutes(filterDeparturesByMode(departures, FERRY).length > 0);
+  }, [departures]);
```
thecristen, line 73: suggestion: Since the logic is the same for all modes, it would be more legible to consolidate the modes into one object and update that. e.g. something like (untested, sorry)

```tsx
const hasRouteTypes, setRouteTypes = useState<Record<keyof ModeEnum, boolean>>({}) // should pre-populate with false values probably

useEffect(() => {
  const routeTypes = {};
  // if you happen to have an Enum with each mode listed you can iterate it!
  for (let type in ModeEnum) {
    routeTypes[type] = (filterDeparturesByMode(departures, type).length > 0);
  }
  setRouteTypes(routeTypes);
}, [departures]);

// get the value via hasRouteTypes["SUBWAY"] or some such
```

apps/site/assets/css/_spacing.scss
```diff
@@ -53,12 +53,20 @@
 // End Bottom
 
 // Start
+.ps-5 {
+  padding-left: .313rem;
```
thecristen, line 57: nitpick: Would be easier to understand where this value comes from if we represent it in fraction form , e.g. `calc(1rem * 5 / 16);`

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
kotva006: So the problem I was running into was `v3api` is a [declaration file](https://www.typescriptlang.org/docs/handbook/2/type-declarations.html#dts-files) which has very specific rules about what can be in them.  The only thing allowed is types.
The enum would have to be in its own typescript file (which would have to be imported inline via `import('./ModeEnum').BUS`)
It seems hacky but it can be done.

apps/site/assets/ts/stop/components/Pill.tsx
```diff
@@ -0,0 +1,36 @@
...
+  children
+}: {
+  onClick?: () => void;
+  selected?: boolean;
+  backgroundColor?: string | undefined;
+  optionalCSS?: string | undefined;
+  children: ReactElement<HTMLElement> | string;
+}): ReactElement<HTMLElement> => {
+  return (
+    <div
```
thecristen, line 20: Getting an ESLint error on this:

> Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard, and touch inputs to an interactive content element  [jsx-a11y/no-static-element-interactions](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-static-element-interactions.md)

**suggestion:** update this tag to a `<button>`.

thecristen: The other ESLint error I _think_ will also resolve itself if you change the tag to a button:

> Visible, non-interactive elements with click handlers must have at least one keyboard listener                                                                                                     [jsx-a11y/click-events-have-key-events](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/click-events-have-key-events.md)

thecristen (changes requested):
- It works!
- I think we can streamline the logic a bit.
- I know it's still a WIP, so I'm ignoring most of the ESLint errors, but there are a couple that I called out that we can address now.
- Finally, let's extract the set of filter pills into a separate component, and start adding tests for it.

The only other thing I wanted to ask about is that it feels very slow (more than I'd expect) to load everything when I try to view the page. Any idea what that's about?

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
thecristen: aahhh I see how that the only place we use enum are indeed in .ts files. So much for that. Maybe there's a less wacky way 🤔

kotva006:
I don't know why it feels slow for you.  I am seeing it load in about a second.  It shouldn't be anything other than html and the site that is loading

kotva006:
I couldn't come up with a good way to separate the pill buttons into their own component.  Since what pills display depends on the departures array data, and the state of the pills affects the display of the departures array data.
It seems like their state is too closely linked to create a clean separation from the 2 pieces of the page.

thecristen:
@kotva006 Yeah you're right, you'll need the departures and filtered departures in the top-level component. There's a pattern for that – you can pass the unfiltered departures _and_ the filtering function into the filter-pills-component as props. Something like `<DepartureFilters departures={departures} onModeChange={setFilteredDepartures}  />` should do the trick, give it a try!

thecristen (approved):
✔️ great, thanks!

PR #1516, opened 2023-02-08
Check if alerts exists

i0sea:
Check if alerts exists in `routeCardHeader`. 
Update: this is a fix for the tnm map disappearing. Going to https://dev-blue.mbtace.com/transit-near-me to confirm it functions as expected should suffice.

### DIFF
--- a/apps/site/assets/ts/components/RouteCardHeader.tsx
+++ b/apps/site/assets/ts/components/RouteCardHeader.tsx
@@ -29,24 +29,19 @@ const RouteCardHeader = ({
         </span>
         <div className="c-link-block__view-schedule">View schedule</div>
       </span>
-      {console.log(alerts)}
-      {alerts !== undefined
-        ? alerts.length > 0 && (
-            <a
-              className="c-link-block__inner-link"
-              href={`/schedules/${route.id}/alerts`}
-              title="alert"
-            >
-              {alertIcon(
-                "c-svg__icon-alerts-triangle m-tnm-sidebar__route-alert"
-              )}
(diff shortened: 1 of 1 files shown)


thecristen:
@i0sea Can you tell us reviewers which locations we should search to pull up relevant routes with alerts? Ideally on dev-blue, but if there aren't enough alerts on there we can run the code locally with the dev API too.

apps/site/assets/ts/components/RouteCardHeader.tsx
```diff
@@ -29,19 +29,24 @@ const RouteCardHeader = ({
...
-          title="alert"
-        >
-          {alertIcon("c-svg__icon-alerts-triangle m-tnm-sidebar__route-alert")}
-          <span>
-            {alerts.length}
-            {alerts.length > 1 ? " alerts" : " alert"}
-          </span>
-        </a>
-      )}
+      {console.log(alerts)}
```
kotva006, line 32: I think this was left in error

apps/site/assets/ts/components/RouteCardHeader.tsx
```diff
@@ -29,19 +29,24 @@ const RouteCardHeader = ({
...
+      {alerts !== undefined
+        ? alerts.length > 0 && (
```
kotva006, line 34: This could be simplified to just `{alerts?.length && (...}`. As is it looks like if `alerts` is null it will still try to run `alerts.length`

kotva006 (approved):

PR #1518, opened 2023-02-09
UI Bug | CR Timetable Realtime Train Arrival Icon

i0sea:
#### Summary of changes

**Asana Ticket:** [UI Bug | CR Timetable Realtime Train Arrival Icon](https://app.asana.com/0/385363666817452/1203869376953056/f)

styling changes so that train icon doesn't overlap time text
Screen Shot 2023-02-09 at 10 34 01 AM

### DIFF
--- a/apps/site/assets/css/_timetable.scss
+++ b/apps/site/assets/css/_timetable.scss
@@ -272,7 +272,7 @@
 
 .js-train-icon {
   display: inline-flex;
-  padding-left: .8rem;
+  padding-left: 1rem;
 }
 
 .m-timetable__time {
--- a/apps/site/assets/css/_timetable.scss
+++ b/apps/site/assets/css/_timetable.scss
@@ -265,9 +265,24 @@
   }
 }
 
+.m-timetable__schedule__content {
+  align-items: center;
+  display: contents;
+  justify-content: center;
+  padding: 0 calc(#{$base-spacing} / 2);
+
+  @include media-breakpoint-down(sm) {
+    padding: 0 calc(#{$base-spacing} / 4);
+  }
+}
+
 .m-timetable__vehicle-icon,
 .m-timetable__flag-icon {
   width: $base-spacing;
+}
+
+.js-train-icon {
+  padding-left: 5px;
 }
 
 .m-timetable__time {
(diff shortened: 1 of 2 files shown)


thecristen (commented):
Does it only work on flag stops? I have the updated CSS but it looks the same here (Safari) 
image

thecristen (changes requested):
Why doesn't it look like your screenshot?

This is a reasonable initial approach, but it doesn't seem to translate smoothly between desktop and mobile (screenshots below taken in Safari, showing desktop and mobile, respectively).
 
image

image

This is likely because many of our sizing/paddings across the website use [relative rather than absolute units](https://joseewouters.nl/blog/the-magic-of-relative-units/), so overall they adjust gracefully to different screen sizing.

You might want to try a different positioning approach here, and/or add a media query to let you apply different styles on mobile.

i0sea:
> Why doesn't it look like your screenshot?
> 
> This is a reasonable initial approach, but it doesn't seem to translate smoothly between desktop and mobile (screenshots below taken in Safari, showing desktop and mobile, respectively).
> 
> image
> 
> image
> 
> This is likely because many of our sizing/paddings across the website use [relative rather than absolute units](https://joseewouters.nl/blog/the-magic-of-relative-units/), so overall they adjust gracefully to different screen sizing.
> 
> You might want to try a different positioning approach here, and/or add a media query to let you apply different styles on mobile.

I'll try using rems and media query. Ideally the padding when there is a flag stop vs not would be different(the initial screenshot has less padding). But the icon seems to be injected in a file where flag presence isn't available.

thecristen:
@i0sea I took a look at this yesterday and it looks different but not better. Care to share any screenshots of what you're expecting me to see? That'll help me give better feedback.

i0sea:
@thecristen pushing to dev-blue. Screen shot from local below 
Screen Shot 2023-02-13 at 1 55 56 PM

i0sea:
Made suggested changes. Current state below.
Screen Shot 2023-02-15 at 11 19 42 AM
Screen Shot 2023-02-15 at 11 20 19 AM

thecristen (approved):
I also enjoy how the added `display: inline-flex` makes the icon have better vertical centering somehow. It looks great!!

PR #1520, opened 2023-02-13
feat(StopRedesign): Added the station address to an external maps link

kotva006:
#### Summary of changes
Added an external link to open maps to the stop

**Asana Ticket:** [Display Stop address with link to directions](https://app.asana.com/0/555089885850811/1203721784670204)

---

#### General checks


apps/site/lib/site_web/router.ex
```diff
@@ -202,6 +202,8 @@ defmodule SiteWeb.Router do
   scope "/api", SiteWeb do
     pipe_through([:secure, :browser])
     get("/alerts", AlertController, :show_by_routes)
+
+    get("/stop/:id", StopController, :get)
```
thecristen, line 206: The `/api` scope's path/module/function naming schemes are starting to get a bit disorganized here... we should consider streamlining that soon, if not now then when we start to add more /api routes.

apps/site/assets/ts/hooks/useStop.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const fetchData = async (stopIdString: string): Promise<Stop> =>
+  fetchJsonOrThrow(`/api/stop/${stopIdString}`);
+
+const useStop = (stopId: string): Stop | null => {
+  const [stop, setStop] = useState<Stop | null>(null);
+
+  useEffect(() => {
+    fetchData(stopId).then(result => setStop(result));
+  }, [stopId]);
```
thecristen, line 13: In the interest of trying new approaches I was wondering if you'd like to try using an [alternate approach to data fetching](https://beta.reactjs.org/learn/synchronizing-with-effects#what-are-good-alternatives-to-data-fetching-in-effects) here. Notably, we already use one of the open-source libraries suggested by the linked post - [useSWR](https://swr.vercel.app), which we use successfully in the `useRealtime` hook. I think it'd be a really nice way to simplify our API-calling in general!

apps/site/assets/ts/stop/components/ExternalMapLink.tsx
```diff
@@ -0,0 +1,20 @@
...
+import renderFa from "../../helpers/render-fa";
+
+const ExternalMapLink = ({
+  address
+}: {
+  address: string;
+}): ReactElement<HTMLElement> => {
+  return (
+    <a
+      href={`https://maps.google.com/?q=${encodeURIComponent(address)}`}
```
thecristen, line 11: I know you're waiting on some feedback from Anna, so I'd love for you to refine this a bit more! A few ideas:

1. I remember the team also discussed refining the displayed result to be more indicative of the station we're looking at, rather than plainly showing an address. Example for North Station below - is there anything we can add to the URL to help us here?

image

2. Have you done any research on approaches to having links open in native mobile apps? This [page for opening links in Google Maps](https://developers.google.com/maps/documentation/urls/get-started) has a slightly different approach than that used here. Some other resources suggest that [an alternate script can be used for iOS devices](https://medium.com/@colinlord/opening-native-map-apps-from-the-mobile-browser-afd66fbbb8a4) - give those a try or follow up with some more research!

apps/site/lib/site_web/router.ex
kotva006: I am curious what you have in mind?  Standardize the module function names, or a better way to organize the routes? (Perhaps putting them into their own file?)

apps/site/assets/ts/stop/components/ExternalMapLink.tsx
kotva006: 1.  TLDR: It's possible but its going to be a lift because google and apple do things differently, and it might be best broken out into a separate ticket to deal with how google does things.  It looks like both services allow us to query by lat-long to get user friendly names.  Google allows searching by address, or station name as well.  The only catch is each service requires a different query param for friendly location names.  [Apple](https://developer.apple.com/library/archive/featuredarticles/iPhoneURLScheme_Reference/MapLinks/MapLinks.html) requires the query to be the station name, which is simple.  Google requires a `place_id` in order to show user friendly names as outlined [here](https://developers.google.com/maps/documentation/urls/get-started#search-examples).  This means we need to figure out what the `place_id` is for every station on [this page](https://developers.google.com/maps/documentation/places/web-service/place-id).  
https://www.google.com/maps/search/?api=1&query=42.413827136668104,-70.99165717317734&query_place_id=ChIJSbu0PwZu44kRLxNeyhVE0oI
https://www.google.com/maps/search/?api=1&query=Wonderland&query_place_id=ChIJSbu0PwZu44kRLxNeyhVE0oI
https://maps.apple.com/?q=Wonderland&ll=42.413827136668104,-70.99165717317734

2.  Thanks for the links! They helped a ton.

thecristen: Great investigation. Agree we should definitely leave some of this for a future ticket then. Storing the `place_id` for each is a really useful idea that we should figure out at some point.

> We can try pushing a version of just google maps urls outlined in the links above and someone with an iphone can test it, and we could go from there.

Yep, let's go ahead with that.

apps/site/lib/site_web/router.ex
thecristen: Maybe just standardize function names, and/or route names that are more indicative? But that's something we can definitely think about after we have some more endpoints to work with.

apps/site/assets/ts/stop/components/StationInformation.tsx
```diff
@@ -0,0 +1,28 @@
...
+const StationInformation = ({
+  stop
+}: {
+  stop: Stop;
+}): ReactElement<HTMLElement> => {
+  return (
+    <div>
+      <div>Station information PLACEHOLDER</div>
+      {/* TODO combine station name with municiplaity (station name, municiplaity) if no address */}
+      {stop.address && (
```
thecristen, line 14: **issue:** By checking for the address here, `ExternalMapLink` never gets to exercise its fallback logic of showing the municipality. So nothing shows up at all for, for example, Southampton St @ Newmarket St on the 171 bus.

apps/site/assets/ts/hooks/useMobileDetect.ts
```diff
@@ -0,0 +1,10 @@
+import MobileDetect from "mobile-detect";
+
+const mobileDetect = new MobileDetect(window.navigator.userAgent);
+
+// This is to primarily to be used when generting android or ios specific behavior
+// For feature detection a different library/approach should be used
+const isIPhone = (): boolean => mobileDetect.is("iPhone");
+const isAndroid = (): boolean => mobileDetect.is("Android");
+
+export { isIPhone, isAndroid };
```
thecristen, line 10: This is more of a helper function than a React hook; can we put this in `ts/helpers`?

apps/site/assets/ts/stop/components/ExternalMapLink.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { ReactElement } from "react";
+import renderFa from "../../helpers/render-fa";
+import { isIPhone } from "../../hooks/useMobileDetect";
+
+// This is temporary to test out an idea
+const getPlaceID = (stationName: string) => {
```
thecristen, line 6: Might want to add the return type to this function or disable eslint for this line.

thecristen (changes requested):

thecristen (approved):

PR #1522, opened 2023-02-15
Transit Near Me | Alerts not Showing

i0sea:
#### Summary of changes

**Asana Ticket:** [Transit Near Me | Alerts not Showing](https://app.asana.com/0/385363666817452/1203879636893634/f)

Add alerts button to transit near me results
Screen Shot 2023-02-17 at 4 11 20 PM

### DIFF
--- a/apps/site/assets/ts/tnm/components/RouteCard.tsx
+++ b/apps/site/assets/ts/tnm/components/RouteCard.tsx
@@ -39,11 +39,8 @@ export const fetchAlerts = async (routeId: string): Promise<Alert[]> => {
   return alerts;
 };
 
-export const isUndefined = (arg: any): any => {
-  if (arg == undefined) {
-    return [];
-  }
-  return arg;
+export const isDefined = (arg: any): Alert[] => {
+  return arg !== undefined ? arg : [];
 };
 
 const routeIsEmpty = (route: RouteWithStopsWithDirections): boolean =>
--- a/apps/site/assets/ts/tnm/components/RouteCard.tsx
+++ b/apps/site/assets/ts/tnm/components/RouteCard.tsx
@@ -12,15 +12,39 @@ import { directionIsEmpty } from "../../components/Direction";
 import { modeByV3ModeType } from "../../components/ModeFilter";
 import RouteCardHeader from "../../components/RouteCardHeader";
 import { isABusRoute } from "../../models/route";
+import useSWR from "swr";
+import { isHighSeverityOrHighPriority } from "../../models/alert";
+import { fetchJsonOrThrow, isFetchFailed } from "../../helpers/fetch-json";
 
 interface Props {
   route: RouteWithStopsWithDirections;
-  alerts: Alert[];
   dispatch: Dispatch;
 }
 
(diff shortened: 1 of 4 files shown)


i0sea:
General question, should parameter for fetchAlerts be changed to url since that's what's usually passed by `useSWR`?

apps/site/assets/ts/tnm/components/RouteCard.tsx
```diff
@@ -46,7 +69,12 @@ const RouteCard = ({
 
   return (
     <div className="m-tnm-sidebar__route" data-mode={mode}>
-      <RouteCardHeader route={route.route} alerts={route.route.alerts} />
+      <RouteCardHeader
+        route={route.route}
+        alerts={isUndefined(
+          useSWR<Alert[]>(`${route.route.id}`, fetchAlerts).data
+        )}
```
thecristen, line 76: This doesn't run

apps/site/assets/ts/tnm/components/RouteCard.tsx
```diff
@@ -20,6 +24,25 @@ interface Props {
...
+  );
+  if (isFetchFailed(results)) {
+    throw new Error(
+      `Failed to fetch Alert information: ${results.status} ${results.statusText}`
+    );
+  }
+
+  let alerts: Alert[] = [];
+  alerts = results.filter(isHighSeverityOrHighPriority) as Alert[];
+  return alerts;
```
thecristen, line 39:
```suggestion
  return results.filter(isHighSeverityOrHighPriority);
```

This can be simplified!

apps/site/assets/ts/tnm/components/RouteCard.tsx
```diff
@@ -46,7 +69,12 @@ const RouteCard = ({
 
   return (
     <div className="m-tnm-sidebar__route" data-mode={mode}>
-      <RouteCardHeader route={route.route} alerts={route.route.alerts} />
+      <RouteCardHeader
+        route={route.route}
+        alerts={isDefined(
+          useSWR<Alert[]>(`${route.route.id}`, fetchAlerts).data
+        )}
```
thecristen, line 76: suggestion: this is nicely condensed, but sacrifices readability a bit. we can use some ES6 syntax ([object destructuring](https://javascript.info/destructuring-assignment)) to help us a bit here:

```typescript
// destructures the "data" property out of the response object and renames it to "alerts"
// alerts might be undefined here
const { data: alerts } = useSWR<Alert[]>(`${route.route.id}`, fetchAlerts);
```

Also you can assign a fallback value if the initial value is false with `||`. So you could pass in `alerts={alerts || []}` and it should work in the same manner as your `isDefined` function.

thecristen (changes requested):
It works! These comments are mostly for readability

thecristen (changes requested):
Looked at the Jest unit tests. `RoutesSidebarTest.tsx` and `RouteCardTest.tsx` are failing. You might need to update those tests to accommodate the new logic you introduced here.

thecristen (approved):

PR #1530, opened 2023-03-07
Logic: Station page vs. Bus Stop page

i0sea:
#### Summary of changes

**Asana Ticket:** [Logic: Station page vs. Bus Stop page](https://app.asana.com/0/385363666817452/1204099052193020/f)

Add is_station? that returns whether or not a given stop is a station

### DIFF
--- a/apps/site/test/site_web/views/schedule_view_test.exs
+++ b/apps/site/test/site_web/views/schedule_view_test.exs
@@ -435,4 +435,10 @@ defmodule SiteWeb.ScheduleViewTest do
              )
     end
   end
+
+  describe "is_station?" do
+    test "returns true if stop is station" do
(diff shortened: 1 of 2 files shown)


apps/site/test/site_web/views/schedule_view_test.exs
```diff
@@ -435,4 +435,10 @@ defmodule SiteWeb.ScheduleViewTest do
              )
     end
   end
+
+  describe "is_station?" do
+    test "returns true if stop is station" do
+      assert is_station?(%Stop{id: "place-north", station?: true}) == true
```
thecristen, line 441: Let's add a test case for it returning false too. That's also part of what coverage measures! ;)

thecristen (approved):

PR #1531, opened 2023-03-07
upgrade to React 18

thecristen:
#### Summary of changes

**Asana Ticket:** [Upgrade React](https://app.asana.com/0/1203354074531980/1203355780667453/f)

I wasn't planning to do this yet but I wanted to evaluate a library that only worked in React 18 so I decided to give this upgrade a try.

Turns out it wasn't too bad. The main things I needed to do:
* Change all `ReactDOM.render` calls to use the new `createRoot` function
* Fix a couple TypeScript changes

The test suite was the worst part because many tests rely on `enzyme`, which isn't updated to work with React 18. I tried a community-made adapter, but couldn't get it to work. So what I did was adjust the Jest tests to continue using React 16. It took a little hackery but I think I got it working OK.

This PR will help me evaluate whether we can still build and deploy okay.


kotva006 (approved):
👍

PR #1532, opened 2023-03-07
[Spike] Rendering React components in backend templates

thecristen:
#### Summary of changes

**Asana Ticket:** [[Spike] Rendering React components in backend templates](https://app.asana.com/0/385363666817452/1204099052193054/f)

## Background 
I had created this task about exploring rendering React components in backend templates for two reasons:

1. This redesign will reuse a visual element (example of current usage below) that's currently **only** used in our backend templates, but in this redesign will most likely be rendered in a React component. I wanted to explore how we could reduce duplication of work. How do we code this HTML markup only once, and use it everywhere?
image

2. We have already encountered some friction passing props to React components in the Elixir templates, driving us to adjust our approach by adding `<meta>` tags with useful data, and relying more on `fetch`-ing data from within React components. I wanted to put some more energy into exploring whether a smoother approach was possible.

Now that we'd stopped supporting IE, we could even take advantage of a modern library if it were to help us.

## Approach

I came across a library called [remount](https://github.com/rstacruz/remount) while [reading my 10000th article about using React with Elixir](https://betterprogramming.pub/using-react-17-with-phoenix-1-5-1b445526c739)... and it finally looked easy! Its approach is to convert your React component into a [Custom Element](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements), which can then be directly used within backend templates, passing the props into the components via usual means, e.g. `<x-application props-json='<%= @props %>' ></x-application>`

### Doesn't `<%= Site.Render.react("ComponentName", @props) %>` do the same?

In the simplest case yes. But here are some problems I found with `Site.Render.react`:

* I get weird text encoding issues when rendering content from the CMS. This was very hard to debug and understand, but something was off when passing the text between Elixir and Node and back. It wasn't even consistent (the same CMS content could render with different weird encoding issues on different pages).
image
* It's unclear how to support nesting components, so `Site.Render.react` ends up encouraging usage with a parent component only. Whereas with this new approach we could theoretically do more complex composition like
```eex
<x-parent props-json='<%= @props %>'>
  <x-child props-json='<%= @other_props %>'></x-child>
  <%= for thing <- @some_list do %>
    <x-list-item props-json='<%= Jason.encode!(%{item: thing}) %>'></x-list-item>
  <% end %>
</x-parent>
```

## Result
I didn't get to test the more complex example, but I tried these two cases to approach my initial questions.

1. I took the markup from `_descriptive_links.html.eex` and created an equivalent React component, `<DescriptiveLink />`. It renders identically to the backend templated version with no encoding issues! 
2. I adjusted `<StopPageRedesign />` to use this approach, demonstrating that we can remove a lot of boilerplate code that had previously supported it. No more `<meta>` tag needed, no hackery fetching the stop ID from the URL, no Webpack finagling, etc.

I really like it, it feels more intuitive to me. But I am curious what others think?

## Caveats

This doesn't render anything in the backend! Web components don't support server side rendering. I've resigned myself to accept this as a reasonable tradeoff [for now*] in order to promote more component reusability across the backend and frontend.

I did try a few other approaches in the hopes of achieving something that _can_ be fully rendered in the backend and hydrated in the frontend. 
* I'd forked https://github.com/revelrylabs/elixir_react_render to work with React 18, and set it up here, even incorporating outstanding PRs I thought would help, but ultimately couldn't get it to work with our application.
* Like I said earlier, CMS data sometimes resulting in text encoding issues. I tried my hand at hacking this but didn't get anywhere. While it isn't a concern for every component, but would be concern for `<DescriptiveLink />`. 

`*` I like being able to render everything without JS, but we're already inconsistent about it and not fully committed to it, so it wouldn't make sense for me to be dogmatic about it here. And I feel like I gave compatible approaches a damn good try at least.
 
#### General checks


kotva006:
1.  I agree with this point of the background.  Ideally we should have zero duplicated code in our code base (it doesn't always work in practice but this should be the goal)
2.  I personally think the friction comes phoenix templates, and react being 2 different libraries solving the same problem in different ways.  They are both presentation frameworks.  I think the smoothest way forward is to go all in on one framework or the other.

That being said, I think what is proposed here does a great job allowing us to reuse React components in existing phoenix template pages.  Thus solving the first problem.  I just dont think this is a solution to make future pages.  Id rather future pages be in phoenix templates or react instead of trying to straddle both frameworks.

Also what do you mean by render everything without JS?  I feel like I am missing something

thecristen:
Thanks for the comments and reviews both here and offline. We've learned a lot! We're not going to use this for now.

PR #1534, opened 2023-03-09
Refine headings

i0sea:
#### Summary of changes

**Asana Ticket:** [Refine headings](https://app.asana.com/0/385363666817452/1204099052193019/f)

Style headings for redesign

#### General checks

### DIFF
--- a/apps/site/assets/css/_stop-page.scss
+++ b/apps/site/assets/css/_stop-page.scss
@@ -297,3 +297,23 @@
   background-color: $gray-bordered-background;
   padding: $base-spacing * .75;
 }
+
+.route-schedules-maps,
+.bike-heading {
+  font-family: 'Helvetica Neue';
+  font-size: 21px;
+  font-style: normal;
+  font-weight: 700;
+  line-height: 27px;
+}
+
+.m-hide {
+  @include media-breakpoint-down(xs) {
+    display: none;
+  }
+}
+
+.bike-heading {
+  font-size: 15.75px;
+  line-height: 20px;
+}
--- a/apps/site/assets/css/_stop-page.scss
+++ b/apps/site/assets/css/_stop-page.scss
@@ -320,17 +320,3 @@
   background-color: aqua;
   height: 160px;
 }
-
-.route-schedules-heading,
-.bike-heading {
-  font-family: 'Helvetica Neue';
-  font-size: 1.3rem;
-  font-style: normal;
-  font-weight: 700;
-  line-height: 1.6rem;
-}
-
-.bike-heading {
-  font-size: 1rem;
-  line-height: 1.25;
-}
--- a/apps/site/assets/css/_stop-page.scss
+++ b/apps/site/assets/css/_stop-page.scss
@@ -320,3 +320,17 @@
   background-color: aqua;
   height: 160px;
 }
+
+.route-schedules-heading,
+.bike-heading {
+  font-family: 'Helvetica Neue';
+  font-size: 1.3rem;
+  font-style: normal;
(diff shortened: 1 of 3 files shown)


apps/site/assets/css/_stop-page.scss
```diff
@@ -297,3 +297,23 @@
   background-color: $gray-bordered-background;
   padding: $base-spacing * .75;
 }
+
+.route-schedules-maps,
+.bike-heading {
+  font-family: 'Helvetica Neue';
+  font-size: 21px;
```
kotva006, line 304: Try to use rem over px.  There are converts online

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
thecristen, line 1: Why not use HTML heading tags? Since the page heading is `<h1>`, the page's sections can use `<h2>` for their headings, and the subsection heading (for bringing your Car or Bike) can use `<h3>`.

apps/site/assets/css/_stop-page.scss
```diff
@@ -320,3 +320,17 @@
...
+  font-size: 1.3rem;
+  font-style: normal;
+  font-weight: 700;
+  line-height: 1.6rem;
+}
+
+.bike-heading {
+  font-size: 1rem;
+  line-height: 1.25;
+}
```
thecristen, line 336: Don't the heading tags already have the right styling as-is? Most of these properties shouldn't need customization.

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -56,7 +56,9 @@ const StopPageRedesign = ({
       {/* Route and Map Div */}
       <div className="d-flex">
         <div style={{ minWidth: "50%" }}>
-          <div>Route schedules & maps / Upcoming Trips PLACEHOLDER</div>
+          <h1 className="route-schedules-heading hidden-sm-down">
```
thecristen, line 59: [It's best practice to only use one `<h1>` on a page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page), which in this case will be the stop name at the top of the page. That's why my [previous comment](https://github.com/mbta/dotcom/pull/1534#discussion_r1136030522) suggested to use `<h2>` and `<h3>` here -- please fix.

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -84,6 +86,9 @@ const StopPageRedesign = ({
         </div>
       </div>
       {/* Station Information Div */}
+      <h1 className="route-schedules-heading">
+        {stop?.["station?"] ? "Station Information" : "Stop Information"}
```
thecristen, line 90: Good catch!

thecristen (changes requested):

apps/site/assets/css/_stop-page.scss
i0sea: @thecristen For clarity should I just remove all this or just keep the font-size? The font-size is the only difference between designs and default styling as far as I can tell.

thecristen: Using heading styles in the mockups generally implies using the existing heading styles as-is, but I've seen the mocked designs be inconsistent in which H2/3 it uses across different screens... so let's remove it all for now, and check in on this when we do our eventual design review.

thecristen (approved):
Thank you!

PR #1538, opened 2023-03-16
Disable background data fetching processes during tests

thecristen:
This is a bit of a behind-the-scenes change I made while working on a different thing.

**Issue:** Dotcom runs lots of different processes in the background, mostly fetching and parsing data. But these processes always run, even during CI, producing occasional error messages such as this one:

image

Not technically failing test, but kind of annoying, intermittent, and unnecessary.

I eventually came across a pattern used in other projects, where we can configure certain processes to not be automatically started in every environment. [Skate uses it here](https://github.com/mbta/skate/blob/master/lib/skate/application.ex#L17) to omit certain modules from being added to the supervision tree in the test environment. So I decided to do something similar here!

**Approach:** 
* Set a single configuration variable. The name and placement are a bit arbitrary, I just wanted every app in `apps/` to be able to access it. 
* Read that variable and use it to add child processes to various `Supervisor`s across the website
* Adjust tests that were written in a way that assumes those child processes were always running.... well in this case instead of rewriting the tests, I manually start whatever required process is needed for the test. I believe it gets torn down when the test is complete.

Going to deploy and make sure it works as expected.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 ###
 
 # 1.) Get the Elixir dependencies within an Elixir container
-FROM hexpm/elixir:1.10.4-erlang-22.3.4.26-debian-buster-20210902 as elixir-builder
+FROM hexpm/elixir:1.12.3-erlang-22.3.4.26-debian-buster-20210902 as elixir-builder
 
 ENV LANG="C.UTF-8" MIX_ENV="prod"
 
--- a/apps/routes/test/populate_caches_test.exs
+++ b/apps/routes/test/populate_caches_test.exs
@@ -26,7 +26,7 @@ defmodule Routes.PopulateCachesTest do
   alias __MODULE__.FakeRepo
 
   setup do
-    {:ok, _} = Agent.start_link(&MapSet.new/0, name: FakeRepo)
+    _ = Agent.start_link(&MapSet.new/0, name: FakeRepo)
     :ok
   end
 
--- a/apps/alerts/lib/supervisor.ex
+++ b/apps/alerts/lib/supervisor.ex
@@ -18,10 +18,17 @@ defmodule Alerts.Supervisor do
 
   @impl Supervisor
   def init(_arg) do
-    children = [
-      Alerts.Cache.Store,
-      {Alerts.Cache.Fetcher, api_mfa: @api_mfa}
-    ]
+    children =
+      [
+        Alerts.Cache.Store
+      ] ++
+        if Application.get_env(:elixir, :start_data_processes) do
+          [
+            {Alerts.Cache.Fetcher, api_mfa: @api_mfa}
+          ]
+        else
+          []
+        end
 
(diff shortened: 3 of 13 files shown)


thecristen:
Bleh, it didn't end up fixing the particular message captured in the screenshot 😩 . But maybe it'll be helpful in other ways.

thecristen:
I'm thinking now that the error I referenced in the screenshot is more likely arising from a try test module `async: true`.

Anyway, I still think this might be a nice change albeit incidental, and would like some feedback on it!

apps/routes/test/populate_caches_test.exs
```diff
@@ -26,7 +26,7 @@ defmodule Routes.PopulateCachesTest do
   alias __MODULE__.FakeRepo
 
   setup do
-    {:ok, _} = Agent.start_link(&MapSet.new/0, name: FakeRepo)
+    _ = Agent.start_link(&MapSet.new/0, name: FakeRepo)
```
lemald, line 29: **question:** is there a particular reason you were no longer able to match on `{:ok, _}`? At first glance it doesn't look like this is related to other changes as far as I can tell.

Dockerfile
```diff
@@ -3,7 +3,7 @@
 ###
 
 # 1.) Get the Elixir dependencies within an Elixir container
-FROM hexpm/elixir:1.10.4-erlang-22.3.4.26-debian-buster-20210902 as elixir-builder
+FROM hexpm/elixir:1.12.3-erlang-22.3.4.26-debian-buster-20210902 as elixir-builder
```
lemald, line 6: **question:** it looks like you upgraded the local version of Elixir in `.tool-versions` a couple of months ago, but I suppose this will actually update it when the site runs in the live production environment. Have you done any testing just to confirm there are no issues?

apps/routes/test/populate_caches_test.exs
thecristen: Oh, yeah I no longer remember the why behind it, it likely works fine as is.

Dockerfile
thecristen: good question! Before I had added this change, this PR failing the Docker build step was what actually tipped me off to this issue! 🤕 After adding this, I was successfuly able to deploy it to dev-green a few weeks ago.

lemald:
@thecristen sorry, one other question, is that one Elixir test failure expected? The output is kind of terse but at first glance it doesn't look related to anything you changed.

thecristen:
@lemald This branch is pretty out of date with the current master, so some of that's been fixed... but yeah, they're unrelated.

apps/routes/test/populate_caches_test.exs
lemald: I personally have a general preference towards being as strict as possible with matches, but I suppose it's test code that's part of the setup as opposed to an actual test so it doesn't actually change either application functionality or what the test suite is testing.

lemald (approved):

PR #1543, opened 2023-03-21
increase max walk distance

i0sea:
No ticket increase max walk distance to .7 miles to reveal some commuter rail suggestions.


i0sea:
changes to osm, otp, and stop locations resolved issue

PR #1549, opened 2023-03-24
Ps autocomplete stop

i0sea:
Asana Ticket: [🗺 Trip Planner | CR Trips Not Showing at Oak Grove, Malden, Braintree](https://app.asana.com/0/385363666817452/1204234745862727/f)

Fixes reversing of trips and changing inputs to non stop results.


thecristen:
These two issues from #1544 are still outstanding here:

- doesn't remove the stop ID if you select a non-stop option later
- no tests on either the Elixir or JS sides

The first one results in a very misleading plan!! And the second should be addressed too.

thecristen (changes requested):
While this did fix the problem where the `stop_id` overstayed its welcome in the input value, this seems to have broken the from/to reversing functionality. When I clic the reverse button nothing happens on the page, and a `TypeError` gets logged to the console.

There are a few Mocha tests that are failing on this branch, which might be related to the same issue.

thecristen (approved):
Works! 🙌🏼

PR #1553, opened 2023-03-30
☎️ Customer Support Form | Add ADA/Accessibility Complaint Checkmark

i0sea:
#### Summary of changes

**Asana Ticket:** [☎️ Customer Support Form | Add ADA/Accessibility Complaint Checkmark](https://app.asana.com/0/385363666817452/1204179585884118/f)
Screen Shot 2023-03-30 at 5 06 05 PM

Add checkmark for ADA/Accessibility related complaints to customer support form

#### General checks


apps/site/assets/js/support-form.js
```diff
@@ -321,6 +322,12 @@ export function setupRequestResponse($) {
   });
 }
 
+export function setupADAResponse($) {
```
arkadyan, line 325: question: What's the reason behind hiding the contact info section if the user selects "ADA complaint"? I'm not seeing anything about that in the Asana ticket, and there are no tests for this behavior.

arkadyan (commented):
I left one small suggestion that you can take or leave, and then one question about the block of javascript code, which seems more important. I can't tell if that behavior is desired, but if it is, it should be tested.

One additional suggestion: it can be helpful for the PR reviewer if you organize (squash) commits to make them logical, stand-alone units that communicate a discrete change. In this case, there could probably be a single commit. That helps the reviewer understand the change holistically. It also helps show corresponding things together, e.g. seeing a test at the same time as the implementation code it is testing.

apps/site/lib/site_web/templates/customer_support/_form.html.eex
```diff
@@ -50,6 +50,18 @@
   <section>
     <h3>Additional Details (optional)</h3>
     <div>Adding more details helps us more effectively respond to your concerns.</div>
+     <div class="ada-container form-group">
+      <%= SiteWeb.PartialView.render("_checkbox.html", %{
+        form: f,
+        field: :ada_response,
```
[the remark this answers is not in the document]
i0sea: Was following the convention of `no_request_response`. Was interpreting as the responses from the feedback provider instead of response from MBTA. `ada_complaint` makes more sense in the  latter case.

apps/site/assets/js/support-form.js
i0sea: This should just be toggling the checkmark and not the contact info

arkadyan: Is any JS at all required at this point?

i0sea: Yes, the setting of the checkbox determines wether the XML tag `<SWARELATED>` gets set as yes or no. I tested out removing the JS but in that case the form doesn't submit when the ADA complaint checkbox is not checked.

arkadyan: Do you know why the form only submits with the JS? I could easily be missing something, but as a general case including a checkbox in a form seems like a pretty trivial thing that wouldn't require JS in order to function.

i0sea: Page has to be reloaded after each form submission. Not sure how often one would file multiple submissions in one sitting. Might create a new ticket if having to reload the page is deemed an issue.

arkadyan: Huh. Is that an existing issue or something introduced by this change?

i0sea: Looks like an existing issue. Tried to back to back submissions on `dev-green` and the second doesn't go thru.

arkadyan: Gotcha, thanks. In that case, I agree that creating a new, separate ticket seems appropriate.

arkadyan (approved):
👍

PR #1570, opened 2023-04-12
feat(StopAlerts): Fixed endpoint for getting all the alerts at a stop

kotva006:
The alerts are first sorted into a table that stores which stop_ids map to which alert_ids
Create an endpoint that returns all the alerts for the given stop.
The bones of the logic was there this pull request fixes it.

**Asana Ticket:** [Departure Cards | Departure Display | Alerts Data](https://app.asana.com/0/home/1201123880594717/1204340687791888)

---

#### General checks

### DIFF
--- a/apps/alerts/lib/cache/store.ex
+++ b/apps/alerts/lib/cache/store.ex
@@ -121,6 +121,7 @@ defmodule Alerts.Cache.Store do
         read_concurrency: true
       ])
 
+    # no cover
     _ = :ets.new(:stop_id_to_alert_ids, [:bag, :protected, :named_table, read_concurrency: true])
 
     # no cover
@@ -142,12 +143,7 @@ defmodule Alerts.Cache.Store do
 
     stop_inserts =
       Enum.flat_map(alerts, fn alert ->
-        non_nil_stops =
-          Enum.filter(alert.informed_entity, fn %Alerts.InformedEntity{stop: stop_id} ->
-            stop_id != nil
-          end)
-
-        Enum.map(non_nil_stops, fn stop_id ->
+        Enum.map(alert.informed_entity.stop, fn stop_id ->
           {
             stop_id,
             alert.id
--- a/apps/alerts/lib/cache/store.ex
+++ b/apps/alerts/lib/cache/store.ex
@@ -132,23 +132,16 @@ defmodule Alerts.Cache.Store do
 
   @impl true
   def handle_call({:update, alerts, banner}, _from, state) do
-    {alert_inserts, route_inserts} =
-      Enum.reduce(alerts, {[], []}, fn alert, {alert_inserts_acc, route_inserts_acc} ->
+    {alert_inserts, route_inserts, stop_inserts} =
+      Enum.reduce(alerts, {[], [], []}, fn alert,
+                                           {alert_inserts_acc, route_inserts_acc,
+                                            stop_inserts_acc} ->
         {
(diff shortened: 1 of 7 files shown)


apps/site/lib/site_web/router.ex
```diff
@@ -207,6 +206,7 @@ defmodule SiteWeb.Router do
     pipe_through([:secure, :browser])
 
     get("/alerts", AlertController, :show_by_routes)
+    get("/stops/:stop_id/alerts", AlertController, :show_by_stop)
```
KaylaBrady, line 209: question: It makes sense to me that this is moved to be near the other routes using AlertController. Is there a reason `/stops` is still the first part of the path?

apps/site/assets/ts/hooks/useAlertsForStop.ts
```diff
@@ -0,0 +1,13 @@
+import useSWR from "swr";
+import { fetchJsonOrThrow } from "../helpers/fetch-json";
+import { Alert } from "../__v3api";
+
+const fetchData = async (url: string): Promise<Alert[]> =>
```
KaylaBrady, line 5: suggestion: Could you add a test file for this new hook?

apps/alerts/lib/cache/store.ex
```diff
@@ -121,6 +121,7 @@ defmodule Alerts.Cache.Store do
         read_concurrency: true
       ])
 
+    # no cover
```
KaylaBrady, line 124: question (non-blocking): For my own curiosity, what does this do?

KaylaBrady (commented):
Hello! A couple small pieces of feedback. One overall clarification - the title of this PR points to a fix, but I'm not noticing what was broken + fixed by this PR. Can you clarify what is being fixed?

apps/alerts/lib/cache/store.ex
kotva006: I believe this does nothing other than let the developer know that the return is deliberately ignored/not covered.  I added it just so it matched all the other calls in the function.

apps/site/lib/site_web/router.ex
KaylaBrady: Ah, I hadn't noticed this was moved to under `/api` prefix, thanks for pointing that out! Your description makes sense to me, thanks!

KaylaBrady (approved):
Looks good to me!

apps/alerts/test/cache/store_test.exs
```diff
@@ -79,6 +79,6 @@ defmodule Alerts.Cache.StoreTest do
     alert4 = %Alerts.Alert{id: "126", effect: :cancellation}
 
     Store.update([alert1, alert2, alert3, alert4], nil)
-    assert Store.alert_ids_for_stop_id("543") == ["123", "124"]
+    assert Store.alert_ids_for_stop_id("543") == ["124", "123"]
```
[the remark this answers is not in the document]
kotva006: Order doesn't matter, but I am unaware of how to compare lists in elixir ignoring the order of the values.  This swap was required to get the test to pass.  The reduce logic stores the order of alerts in descending order (new alerts get appended to the front of the list).  This matches how the routes are stored as well.  
I had debated about changing it, but read some articles about list appending and speed.  They made it seem like combining 2 lists required elixir to traverse the first list to find the end then append the second list to the end (without traversal of the second list).  Appending new alerts to the start of this list is the fast path because the new list to append should be shorter than accumulator list (in most cases)

PR #1573, opened 2023-04-19
Stop Page Redesign Landing Page | Map should show subway and commuter rail routes

thecristen:
#### Summary of changes

**Asana Ticket:** [Landing Page | Map should show subway and commuter rail routes](https://app.asana.com/0/385363666817452/1204340687791882/f)

Adds subway/trolley/commuter rail routes to the map.

Since fetching shapes is dependent on the routes, I figured we may as well fetch routes and shapes together, so I added that data to the existing endpoint we have for fetching routes for this page.

> **Note**
> This is behind a feature flag – please visit a stop page via `/stops/:stop_id?active_flag=stops_redesign`

image

---

#### General checks

### DIFF
--- a/apps/routes/lib/route.ex
+++ b/apps/routes/lib/route.ex
@@ -235,8 +235,8 @@ defmodule Routes.Route do
   def combined_route?(%{id: "627"}), do: true
   def combined_route?(_), do: false
 
-  @spec is_rail_route?(t()) :: boolean
-  def is_rail_route?(route) do
+  @spec rail?(t()) :: boolean
+  def rail?(route) do
     type_atom(route) in [:subway, :commuter_rail]
   end
 
--- a/apps/routes/lib/route.ex
+++ b/apps/routes/lib/route.ex
@@ -235,6 +235,11 @@ defmodule Routes.Route do
   def combined_route?(%{id: "627"}), do: true
   def combined_route?(_), do: false
 
+  @spec is_rail_route?(t()) :: boolean
+  def is_rail_route?(route) do
+    type_atom(route) in [:subway, :commuter_rail]
+  end
+
   @spec to_json_safe(t) :: map
   def to_json_safe(%__MODULE__{
         id: id,
--- a/apps/routes/test/repo_test.exs
+++ b/apps/routes/test/repo_test.exs
@@ -1,5 +1,5 @@
 defmodule Routes.RepoTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
   import Mock
   alias Routes.{Repo, Route}
 
--- a/apps/site/assets/ts/stop/components/StopMapRedesign.tsx
+++ b/apps/site/assets/ts/stop/components/StopMapRedesign.tsx
@@ -1,4 +1,3 @@
-import { uniqBy } from "lodash";
 import React, { ReactElement } from "react";
 import Map from "../../leaflet/components/Map";
 import { Stop } from "../../__v3api";
@@ -32,7 +31,7 @@ const StopMapRedesign = ({ stop, lines }: Props): ReactElement<HTMLElement> => {
         tooltip: null
       } as MapMarker
     ],
-    polylines: uniqBy(lines, "id"),
+    polylines: lines,
     tile_server_url: mapConfig?.tile_server_url,
     zoom: 16
   } as MapData;
--- a/apps/site/assets/ts/stop/components/StopMapRedesign.tsx
+++ b/apps/site/assets/ts/stop/components/StopMapRedesign.tsx
@@ -1,3 +1,4 @@
+import { uniqBy } from "lodash";
 import React, { ReactElement } from "react";
 import Map from "../../leaflet/components/Map";
 import { Stop } from "../../__v3api";
@@ -31,7 +32,7 @@ const StopMapRedesign = ({ stop, lines }: Props): ReactElement<HTMLElement> => {
         tooltip: null
       } as MapMarker
     ],
-    polylines: lines,
+    polylines: uniqBy(lines, "id"),
     tile_server_url: mapConfig?.tile_server_url,
     zoom: 16
   } as MapData;
--- a/apps/site/assets/ts/stop/components/StopMapRedesign.tsx
+++ b/apps/site/assets/ts/stop/components/StopMapRedesign.tsx
@@ -1,14 +1,19 @@
 import React, { ReactElement } from "react";
 import Map from "../../leaflet/components/Map";
 import { Stop } from "../../__v3api";
-import { MapData, MapMarker } from "../../leaflet/components/__mapdata";
+import {
+  MapData,
+  MapMarker,
+  Polyline
+} from "../../leaflet/components/__mapdata";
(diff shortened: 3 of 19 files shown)


thecristen:
I'm having technical issues assign someone not on vacation from @mbta/2023-april-reviewers to review, maybe whoever's available can assign themselves? :)

apps/site/lib/site_web/controllers/route_controller.ex
```diff
@@ -6,7 +6,30 @@ defmodule SiteWeb.RouteController do
   alias Routes.Repo
 
   def get_by_stop_id(conn, %{"stop_id" => stop_id} = _params) do
-    routes = Repo.by_stop(stop_id)
-    json(conn, routes)
+    routesWithPolylines =
+      stop_id
+      |> Repo.by_stop()
+      |> Enum.map(&[&1, route_polylines(&1)])
```
paulswartz, line 12: suggestion: you probably want to include the polylines underneath the route object, rather than switching this from a list of objects to a list of arrays. In this situation, a older frontend client that hits a newer backend server will crash, rather than silently ignoring the new data.
```suggestion
      |> Enum.map(&Map.put(&1, :polylines, route_polylines(&1))
```

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -6,19 +6,25 @@ import { useRoutesByStop } from "../../hooks/useRoute";
...
+  const routesWithPolylines = useRoutesByStop(stopId);
   // Return loading indicator while waiting on data fetch
-  if (!stop || !routes) {
+  if (!stop || !routesWithPolylines) {
     return <Loading />;
   }
+  const routes = routesWithPolylines.map(([route]) => route);
+  const polylines = routesWithPolylines
+    .filter(
+      ([route]) => [0, 1, 2].includes(route.type) || isASilverLineRoute(route)
```
paulswartz, line 25: question: why not do this filtering on the backend?

apps/routes/test/repo_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule Routes.RepoTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
```
paulswartz, line 2: Looks like this might have caused a test failure in `SiteWeb.ScheduleController.AlertsControllerTest`

paulswartz (commented):
Even with the filtering, it looks like this is returning polylines which don't need to be rendered.

Screenshot 2023-04-21 at 1 02 13 PM

This is http://localhost:4001/stops/place-bbsta?active_flag=stops_redesign, and it looks like some other CR routes going through South Station, but not Back Bay, are included.

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
thecristen: Yeah, why not! I've made this change.

apps/site/lib/site_web/controllers/route_controller.ex
thecristen: The suggested line didn't work as-is but I was able to make the change, I like it much better, thanks!

apps/routes/test/repo_test.exs
thecristen: Reverted this and still see these failures; I think I broke these elsewhere. Will address outside this PR.

thecristen:
Thanks @paulswartz for the great find. I've switched to using route patterns filtered by stop to narrow the shapes down to the appropriate ones 🎉 .

apps/site/assets/ts/stop/components/StopMapRedesign.tsx
```diff
@@ -26,7 +32,7 @@ const StopMapRedesign = ({ stop }: Props): ReactElement<HTMLElement> => {
         tooltip: null
       } as MapMarker
     ],
-    polylines: [],
+    polylines: uniqBy(lines, "id"),
```
paulswartz, line 35: question: should this filtering also be happening on the server side?

apps/site/lib/site_web/controllers/route_controller.ex
```diff
@@ -3,10 +3,46 @@ defmodule SiteWeb.RouteController do
...
+          weight: 4
+        }
+      end)
+    else
+      []
+    end
+  end
+
+  defp is_rail_route?(route) do
+    Route.type_atom(route) in [:subway, :commuter_rail]
```
paulswartz, line 46: question: should this live on `Route`, like `silver_line?/1` does?

apps/site/lib/site_web/controllers/route_controller.ex
```diff
@@ -3,10 +3,46 @@ defmodule SiteWeb.RouteController do
...
+      end)
+
+    json(conn, routesWithPolylines)
+  end
+
+  defp route_polylines(route, stop_id) do
+    if is_rail_route?(route) or Route.silver_line?(route) do
+      route.id
+      |> RoutePatterns.Repo.by_route_id(stop: stop_id)
+      |> Enum.map(fn %RoutePattern{shape_id: id, representative_trip_polyline: polyline} ->
```
paulswartz, line 26: suggestion: this could be a good spot to deduplicate the shape IDs:
```suggestion
      |> Enum.uniq_by(& &1.shape_id)
      |> Enum.map(fn %RoutePattern{shape_id: id, representative_trip_polyline: polyline} ->
```

paulswartz (commented):
I think you might want to use the opposite ordering for the rendering:
Screenshot 2023-04-26 at 1 49 11 PM
Shouldn't the Orange Line be rendered on top of the CR lines?

apps/site/lib/site_web/controllers/route_controller.ex
thecristen: Ah, wrong place for it - there aren't duplicate shapes within a single route (which is the context for this particular function, but it comes into play when serving several routes together).

apps/site/assets/ts/stop/components/StopMapRedesign.tsx
thecristen: The merging-shapes-from-many-routes thing only happens in the frontend (all routes are fed into this map), so it makes more sense to de-duplicate here.

apps/routes/lib/route.ex
```diff
@@ -235,6 +235,11 @@ defmodule Routes.Route do
   def combined_route?(%{id: "627"}), do: true
   def combined_route?(_), do: false
 
+  @spec is_rail_route?(t()) :: boolean
+  def is_rail_route?(route) do
```
paulswartz, line 239: nitpick: In Elixir, `is_X` are for guards, and don't have a `?` (for example, `is_nil/1`).  non-guard boolean functions don't have `is` but do have the `?` (for example, `Enum.empty?/1` or the above `combined_route?/1`)
```suggestion
  @spec rail?(t()) :: boolean
  def rail?(route) do
```

paulswartz (approved):
one nitpick, then I think this is good to go!

paulswartz (approved):

PR #1579, opened 2023-05-01
feat(ENVRC): Update the template to have a placeholder for AWS creden…

kotva006:
…tials

#### Summary of changes
Add the AWS env variable placeholder to the template file

### DIFF
--- a/.envrc.template
+++ b/.envrc.template
@@ -5,6 +5,10 @@
 # export ALGOLIA_APP_ID=
 # export ALGOLIA_SEARCH_KEY=
 # export ALGOLIA_WRITE_KEY=
+
+# export AWS_ACCESS_KEY_ID=
+# export AWS_SECRET_ACCESS_KEY=
+
 export DRUPAL_ROOT=https://live-mbta.pantheonsite.io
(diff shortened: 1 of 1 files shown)


paulswartz:
Should basic running of the Dotcom application require AWS credentials?

kotva006:
@paulswartz It looks like when Cristen added the [bus alerts storage](https://github.com/mbta/dotcom/commit/377cbd4dd1fa812fc65be7bbd78af6adc7286a4b) it made it required for startup

paulswartz:
Yes, but should that be the case? Could it be optional?

kotva006:
@paulswartz It probably shouldn't be required.  The service should fail gracefully, log a message, and allow the system to keep running.
There are other services besides the bus alerts that use the AWS credentials (Search).  Also most of the env variables in that template file aren't required to get the server up and running.
Figures adding a placeholder for those variables informs people that there are AWS requirements for some parts of the website.  It has also become a pain point for a couple uses.  Having it standardized should hopefully make it less of an issue in the future for people spinning up the site, and enabling features.

i0sea (approved):
🚀

PR #1585, opened 2023-05-08
feat: New vehicle channel

KaylaBrady:
#### Summary of changes

**Asana Ticket:** [Real-Time Tracking Page | Map](https://app.asana.com/0/385363666817452/1204506670061542/f)

This PR adds a new channel for getting live vehicle data. The existing channel returns vehicle data transformed into a map marker. Exposing a new channel that returns vehicle data with minimal transformation will make it easier to customize the way vehicle data is displayed on the map, such as for the new map on the redesigned stop page.

Overall I took the approach of keeping this new channel as similar to the existing vehicle channel as possible; for example, requiring an "init" message to be sent to reset the data rather than returning data on join. My intention was to keep the scope of this particular task small, but future work could improve how this channel sends data.

I considered an alternative approach of having the old vehicle channel simply broadcast each message to the new channel. It seemed preferable to me to keep the channels completely separate, but would be interested to see if there are differing opinions on that! 

---

#### General checks

### DIFF
--- a/apps/site/test/site_web/channels/vehicles_channel_test.exs
+++ b/apps/site/test/site_web/channels/vehicles_channel_test.exs
@@ -29,7 +29,7 @@ defmodule SiteWeb.VehicleChannelTest do
     assert {:ok, _, socket} =
              UserSocket
              |> socket("", %{some: :assign})
-             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehicleMapMarkerChannelTest")
+             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehiclesChannelTest")
 
     broadcast_from(socket, "reset", %{data: @vehicles})
 
@@ -44,7 +44,7 @@ defmodule SiteWeb.VehicleChannelTest do
     assert {:ok, _, socket} =
              UserSocket
              |> socket("", %{some: :assign})
-             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehicleMapMarkerChannelTest")
+             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehiclesChannelTest")
 
     broadcast_from(socket, "add", %{data: @vehicles})
 
@@ -59,7 +59,7 @@ defmodule SiteWeb.VehicleChannelTest do
     assert {:ok, _, socket} =
              UserSocket
              |> socket("", %{some: :assign})
-             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehicleMapMarkerChannelTest")
+             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehiclesChannelTest1")
 
     broadcast_from(socket, "update", %{data: @vehicles})
 
@@ -72,7 +72,7 @@ defmodule SiteWeb.VehicleChannelTest do
     assert {:ok, _, socket} =
              UserSocket
              |> socket("", %{some: :assign})
-             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehicleMapMarkerChannelTest2")
+             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehiclesChannelTest2")
 
     broadcast_from(socket, "remove", %{data: ["vehicle_id"]})
 
@@ -85,7 +85,7 @@ defmodule SiteWeb.VehicleChannelTest do
     assert {:ok, _, socket} =
              UserSocket
              |> socket("", %{some: :assign})
-             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehicleMapMarkerChannelTest3")
+             |> subscribe_and_join(VehicleChannel, "vehicles-v2:VehiclesChannelTest3")
 
     push(socket, "init", %{"route_id" => "route_id", "direction_id" => "0"})
 


KaylaBrady:
@arkadyan if you have the time, I would appreciate your review of this too!

apps/site/lib/site/application.ex
```diff
@@ -39,6 +39,7 @@ defmodule Site.Application do
         if Application.get_env(:elixir, :start_data_processes) do
           [
             {Site.Stream.Vehicles, name: Site.Stream.Vehicles},
+            {Site.Stream.Vehicles, name: Site.Stream.Vehicles, topic: "vehicles-v2"},
```
arkadyan, line 42: question: Do we need the second GenServer? Having separate channels makes sense to me, but could `VehicleMapMarkerChannel` and `VehicleChannel` both subscribe to the same Vehicles pub-sub stream? (I might be confusing myself about the different stages here…)

If that doesn't make sense, a more descriptive topic might be helpful. (And maybe topics for both?)

apps/site/lib/site_web/channels/vehicle_map_marker_channel.ex
```diff
@@ -1,6 +1,6 @@
-defmodule SiteWeb.VehicleChannel do
+defmodule SiteWeb.VehicleMapMarkerChannel do
   @moduledoc """
-  VehicleChannel module
+  Channel allowing clients to subscribe to streams of vehicle map markers
```
arkadyan, line 3: 👍

arkadyan (approved):
I threw out a couple of small suggestions, but in general, this looks good. I appreciate the refactoring for more descriptive names.

apps/site/lib/site_web/channels/user_socket.ex
```diff
@@ -2,7 +2,8 @@ defmodule SiteWeb.UserSocket do
   use Phoenix.Socket
 
   ## Channels
-  channel("vehicles:*", SiteWeb.VehicleChannel)
+  channel("vehicles:*", SiteWeb.VehicleMapMarkerChannel)
```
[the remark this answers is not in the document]
KaylaBrady: I agree! I was wary of changing the existing channel name as part of this PR to avoid issues during a deploy since the React code currently expects these messages via the "vehicles:" channel. I suppose I could add an additional `vehicle-map-markers-` channel and add frontend handling for that first... though I'll try to think if there is a simpler approach!

apps/site/lib/site/application.ex
KaylaBrady: Yeah good point! Instead I'll have a single gen server take a list of topics to broadcast to.

arkadyan: That certainly works too, but could there even be a single topic? Are the two topics broadcasting the same data? (Maybe they aren't and I just missed something.)

apps/site/lib/site_web/channels/user_socket.ex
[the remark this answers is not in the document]
arkadyan: Ah, that's a good point, I wasn't thinking about that. Yeah, that could work, I'll let you decide if it's even worth it or not.

apps/site/lib/site/application.ex
KaylaBrady: I think I see what you mean! This is pushing the boundaries of my current phoenix channel knowledge, but maybe something like[ subscribing to external topics](https://hexdocs.pm/phoenix/Phoenix.Channel.html#module-subscribing-to-external-topics) would support that. Though I'll have to think more about this piece

> Note: the caller must be responsible for preventing duplicate subscriptions.

arkadyan: Maybe this is me misunderstanding and confusing things. I thought that `Site.Stream.Vehicles` was publishing over Phoenix PubSub to `VehicleMapMarkerChannel`, and then `VehicleMapMarkerChannel` is publishing over the socket channel to the front-end client. Do I have that correct? If so, I was wondering if it was possible to leave the channels as they are, but feed them both from the same Phoenix PubSub feed, like:

```
Site.Stream.Vehicles  ->  VehicleMapMarkerChannel  ->  "vehicles:*"
		      ↳  VehicleChannel  ->  "vehicles-v2:*"
```

But maybe that doesn't make sense in reality.

KaylaBrady: I think I see what you mean, but perhaps the current limitation is that `Site.Stream.Vehicles` isn't using Phoenix `PubSub.broadcast/3`- it is using `Endpoint.broadcast/3` to broadcast directly to the "vehicles" channel topic. 
With that, am stuck at how to execute on the ` ↳` piece of your diagram diagram.

I think to do what you describe without modifying `Site.Stream.Vehicles`, `VehicleChannel` would have to also subscribe to the "vehicles" topic, but push messages to clients subscribed to "vehicles-v2". One way to achieve that seems to be 'subscribing to external topics', but that seems like it could result in duplicate subscriptions in the case where a single client is subscribed to both the `VehicleMapMarkerChannel` and `VehicleChannel`. 

Alternatively, I could change `Site.Stream.Vehicles` to broadcast a to an internal topic not associated with a channel, and have both channels subscribe to that topic on join. Though that feels like it could be more substantial change

arkadyan: Ah, I see. Sorry, I was definitely confused about that. In that case, I think I would just revoke the original suggestion in the first place. Sorry for all the confusion!

KaylaBrady: Sorry for not clarifying that sooner, I was still working through the details of it all myself! Do you think it would still be worth consolidating to use a single GenServer which takes a list of topics?

arkadyan: Up to you, but I might say maybe it doesn't seem worthwhile. At first, I thought the two modules might be doing the same thing, but now knowing that they aren't, I think one per channel seems totally reasonable and maybe even easier to understand.

apps/site/lib/site_web/channels/user_socket.ex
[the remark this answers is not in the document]
KaylaBrady: With the ideal that the old "vehicles:*" channel might eventually be removed if we standardize on building map markers in the view, I'm going to opt to leave this as-is and make a tech debt ticket for eventually removing it altogether.

PR #1588, opened 2023-05-09
fix(DepartureTimes): Added key to generated time row to remove errors

kotva006:
#### Summary of changes
Noticed a bunch of console errors about no keys being specified in the `DepartureTimes` component.   I thought that it had gotten dealt with, but I guess not.
This fixes the console errors from showing up about no keys on the departure times rendered.

**Asana Ticket:** No Ticket

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/stop/components/DepartureTimes.tsx
+++ b/apps/site/assets/ts/stop/components/DepartureTimes.tsx
@@ -77,7 +77,7 @@ const displayFormattedTimes = (
       {formattedTimes.map((time, index) => {
         const classes = departureTimeClasses(time, index);
         return (
-          <div className="d-flex" key={`${time.displayString}-departure-times`}>
+          <div className="d-flex" key={`${time.key}-departure-times`}>
             {time.isPrediction && (
               <div className="me-4">
                 {SVGIcon("c-svg__icon--realtime fs-10", realtimeIcon)}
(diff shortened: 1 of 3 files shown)


apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -228,16 +228,16 @@ const departureTimeRow = (
           {formattedTimes.map((time, index) => {
             const classes = departureTimeClasses(time, index);
             return (
-              <>
+              <div
+                key={`${time.displayString}-departure-times`}
```
KaylaBrady, line 232: Good catch!
question: Are the display times guaranteed to be unique? May be worth appending [useId](https://react.dev/reference/react/useId) to be sure. Maybe two buses could be predicted to arrive at the same time when there is bunching going on

KaylaBrady (approved):

apps/site/assets/ts/stop/components/DepartureTimes.tsx
kotva006: Great Question!
I didn't think about bus bunching.  Yeah this will fail if 2 buses are predicted to arrive in the same minute.  It looks like [useId should not be used to generate key ids](https://react.dev/reference/react/useId#:~:text=useId%20should%20not%20be%20used%20to%20generate%20keys%20in%20a%20list.%20Keys%20should%20be%20generated%20from%20your%20data.).  This might need some more thoughts about it.

PR #1593, opened 2023-05-15
fix(Alert): Fixed how we parsed alert active period dates

kotva006:
#### Summary of changes
Found a bug while working on alert badges for the stop redesign.
During one of the tech sprints how parsing date strings of alerts was changed.
Unfortunately this change was breaking how current alerts work.

This pull request unifies the change to use standard ISO date format parsing, and out custom format.

**Asana Ticket:** None

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/models/alert.ts
+++ b/apps/site/assets/ts/models/alert.ts
@@ -25,12 +25,28 @@ export const uniqueByEffect = (
   alerts: Alert[]
 ): boolean => alerts.findIndex(a => a.effect === alert.effect) === index;
 
-const activePeriodToDates = (activePeriod: TimePeriodPairs): Date[] => {
-  return activePeriod.map(
-    (d: string): Date => {
-      return parseISO(d);
-    }
-  );
+const withLeadingZero = (n: string): string => `0${n}`.slice(-2);
+
+const legacyDateParsing = (dateString: string): Date | null => {
+  const datePattern = /^(\d{4})-(\d{1,2})-(\d{1,2})\s(\d{1,2}):(\d{2})$/;
+  const match = datePattern.exec(dateString);
+  if (match) {
+    const [, year, rawMonth, rawDay, rawHour, min] = match;
+    return new Date(
+      `${year}-${withLeadingZero(rawMonth)}-${withLeadingZero(
+        rawDay
+      )}T${withLeadingZero(rawHour)}:${min}:00`
+    );
+  }
+  return null;
+};
+
+const activePeriodToDates = (
+  activePeriod: TimePeriodPairs
+): (Date | null)[] => {
+  return activePeriod.map((d: string): Date | null => {
+    return isValid(d) ? parseISO(d) : legacyDateParsing(d);
+  });
 };
 
 const isCurrentLifecycle = ({ lifecycle }: Alert): boolean =>
@@ -46,8 +62,10 @@ export const isCurrentAlert = (
   const dateRanges = alert.active_period.map(ap => activePeriodToDates(ap));
   const isInARange = dateRanges.some((range): boolean => {
     const [start, end] = range;
-    if (!isValid(start)) return false; // end might be null for ongoing alerts
-    return currentDate >= start && (isValid(end) ? currentDate <= end : true);
+    if (!start || !isValid(start)) return false; // end might be null for ongoing alerts
+    return (
+      currentDate >= start && (end && isValid(end) ? currentDate <= end : true)
+    );
   });
   return isCurrentLifecycle(alert) && isInARange;
 };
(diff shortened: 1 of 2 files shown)


apps/site/assets/ts/models/alert.ts
```diff
@@ -25,12 +25,28 @@ export const uniqueByEffect = (
...
+    const [, year, rawMonth, rawDay, rawHour, min] = match;
+    return new Date(
+      `${year}-${withLeadingZero(rawMonth)}-${withLeadingZero(
+        rawDay
+      )}T${withLeadingZero(rawHour)}:${min}:00`
+    );
+  }
+  return null;
+};
+
```
KaylaBrady, line 43: suggestion: Please add some tests for this new function! 

One option which might simplify the logic here could be to use the `parse` function from [date-fns](https://date-fns.org/v2.30.0/docs/parse) using the `d` and `M` format strings which don't have leading 0s.

kotva006: All of the test in `isCurrentAlert` should hit this function.  The reason is that they were passing instead of failing is that `date-fns` could figure out how to parse the date strings used.
`alert8` (which was added) covers that case, as it was the data which found this issue out to begin with (it failed to be marked as current even though it was).
I just pasted the [old code](https://github.com/mbta/dotcom/commit/b892e4e73d746c1b7a80bb55d441475348bac656#diff-516144abe4e4d9e616e676affed5458b3510723a18555335a449e271b654391f) back here.  I don't like it, but I needed it back working again.
It could definitely be updated, but I feel like that is out of scope.

KaylaBrady: Thanks for the context, I didn't realize this change was reinstating the old code.
So is the impact of this issue that no current alerts are showing up on the website? How frequently do alerts use the old format?

kotva006: I think that (yyyy-M-d HH:mm) is the only format alerts currently use.  I have no evidence to suggest they do anything otherwise.  The original code was an attempt at simplifying the date parsing.  All tests passed so the new code was assumed to work.  It doesn't.
This could be reverted back to just the old state.  I figured I would marry the forms of parsing instead.  That way incase any current behavior depends on the new parsing it should still work.
I also added ticket to the backlog to add timezones to alerts.  The goal is then the times would match the ISO standard and be parsed easier.

KaylaBrady: Can you share more about what doesn't work before this PR is merged? I'm having trouble understanding where we might look on the website to see a change in behavior once this code is merged. For example, with my current assumption of what this PR is fixing, I would expect to see no active alerts on the prod website right now. But I am seeing an active alert for the 85 on prod, so I'm not entirely sure I understand the impact of this change.
image

kotva006: Okay, did a little more deep dive.
All the tests were passing because the dates defined in them had 2 digits for month and day.  For some reason the alert I defined through the alerts tool broke it.  The [alert we return from the backend](https://dev-green.mbtace.com/api/alerts?route_ids=110) has month and day as a single digit.  None of the tests we have today test for that.  This causes the `parseISO` function to return `Invalid Date`.  If it had 2 digits (like the tests) then the date is parsed (somewhat correctly, timezones throw things off)

Right now this code is used in 3 places.  StopPage.tsx (isActiveDiversion), Line.tsx (hasAnActiveDiversion), and Merges.tsx (hasAnActiveDiversion)

1. StopPage - This looks like it is a super conditional check.  If the stop card on the schedules page [doesn't show predictions](https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/schedule/components/line-diagram/StopCard.tsx#L168), its then supposed to show the Alert.  I really don't know how to trigger the don't show predictions, but that is why the alert appears to be working on the [Route 110 page](https://dev-green.mbtace.com/schedules/110/line)

2. Line - This one looks like its actually broken.  If there is a diversion at a stop the line in the line diagram is supposed be have a [different stroke](https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/schedule/components/line-diagram/graphics/Line.tsx#L60).  Right now the stop `125 Western Ave` is closed, yet there is [no change in the line pattern](https://www.mbta.com/schedules/70/line) to indicate that it is closed.  There should be.  

3. Merges.tsx - Same thing as Line.tsx above, but this uses it for when branches merge.

KaylaBrady: Thanks for sharing that additional detail!
suggestion (non-blocking): I think it would be nice to add some test coverage for those 3 cases you outlined, or update the existing tests to use the correct alert date format. Maybe something to consider adding to the ISO parsing breakout ticket!

KaylaBrady (approved):

PR #1594, opened 2023-05-15
Ixc departure row click

i0sea:
Allow departure row click and update mobile overlay styling

**Asana Ticket:** [Real-Time Tracking Page | Layout](https://app.asana.com/0/555089885850811/1204506670061538/f)

---

#### General checks


apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -294,18 +296,40 @@ const DepartureTimes = ({
...
+          {Object.entries(schedules).map(([headsign, schs]) => {
+            const formattedTimes = toDisplayTime(
+              schs,
+              predictionsByHeadsign[headsign]
+                ? predictionsByHeadsign[headsign]
+                : [],
+              overrideDate
+            );
+            return departureTimeRow(headsign, formattedTimes);
+          })}
```
KaylaBrady, line 316: question: is the `departureTimeRow` the element that should have the `onClick` functionality? The title suggests that each row should be clickable, but it seems here like the rows are wrapped in a single clickable div.

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -269,6 +269,7 @@ interface DepartureTimesProps {
   directionId: DirectionId;
   schedulesForDirection: ScheduleWithTimestamp[] | undefined;
   overrideDate?: Date;
+  onClick?: any;
```
KaylaBrady, line 272: question: Is there a case where an `onClick` would not be passed? Wondering if it is safe for this to not be an optional prop

apps/site/assets/ts/stop/__tests__/DeparturesAndMapTest.tsx
```diff
@@ -1,14 +1,14 @@
 import React from "react";
-import StopPageOverlay from "../components/StopPageOverlay";
+import DeparturesAndMap from "../components/DeparturesAndMap";
 import { Stop } from "../../__v3api";
 import { mount } from "enzyme";
 
 const stop = {} as Stop;
 
-describe("StopPageOverlay", () => {
+describe("DeparturesAndMap", () => {
```
KaylaBrady, line 8: suggestion: Consider adding some tests for the new row click functionality here

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -294,18 +296,40 @@ const DepartureTimes = ({
...
+          {Object.entries(schedules).map(([headsign, schs]) => {
+            const formattedTimes = toDisplayTime(
+              schs,
+              predictionsByHeadsign[headsign]
+                ? predictionsByHeadsign[headsign]
+                : [],
+              overrideDate
+            );
+            return departureTimeRow(headsign, formattedTimes);
+          })}
```
KaylaBrady, line 329: suggestion (non-blocking): Consider splitting this out into a helper element on the page to avoid duplication

apps/site/assets/ts/stop/components/DepartureTimes.tsx
kotva006: Yeah the `onClick` should be on the [div](https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/stop/components/DepartureTimes.tsx#L221) returned by the departure time row.  There is currently a placeholder [button](https://github.com/mbta/dotcom/blob/master/apps/site/assets/ts/stop/components/DepartureTimes.tsx#L255) on the row around the chevron.  That button should be replaced to make the whole row clickable, as stated in the TODO above the button

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -294,18 +296,40 @@ const DepartureTimes = ({
...
-            : [],
-          overrideDate
-        );
-        return departureTimeRow(headsign, formattedTimes);
-      })}
-    </>
+    <div>
+      {onClick ? (
+        <div
+          className="testSample"
```
kotva006, line 302: question: Is this class name intentional?  If so it should be more descriptive

apps/site/assets/ts/stop/components/DepartureTimes.tsx
i0sea: It doesn't have to be optional. Thought these were used on the current page but that's not the case

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -0,0 +1,107 @@
...
+    departureSchedules: ScheduleWithTimestamp[] | null;
+  }>({
+    departureRoute: null,
+    departureDirectionId: null,
+    departureSchedules: null
+  });
+
+  const routesWithPolylines = useRoutesByStop(stop.id);
+
+  const polylines = chain(routesWithPolylines)
```
kotva006, line 42: This is already passed in to the component as `lines`.  Why is it being recalculated?

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -0,0 +1,107 @@
...
+  routes,
+  stop,
+  schedules,
+  lines
+}: /* eslint-enable @typescript-eslint/no-unused-vars */
+DeparturesAndMapProps): ReactElement<HTMLElement> => {
+  const [state, setState] = useState<{
+    email: null | string;
+    password: null | string;
+  }>({ email: null, password: null });
```
kotva006, line 28: What is this state for?

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const [departureInfo, setDepartureInfo] = useState<{
+    departureRoute: Route | null;
+    departureDirectionId: DirectionId | null;
+    departureSchedules: ScheduleWithTimestamp[] | null;
```
kotva006, line 33: This should be an array of `DepartureInfo`s as that will hold both the prediction and schedule information.

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -0,0 +1,107 @@
...
+      departureSchedules: routeSchedules
+    });
+  };
+  const clearDepartureVariables: () => void = () => {
+    setDepartureInfo({
+      departureRoute: null,
+      departureDirectionId: null,
+      departureSchedules: null
+    });
+  };
```
kotva006, line 65: Not sure these helper functions are needed.  You should just be able to pass the `setDepartureInfo` function and then each card can pass the information, or clear it.

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -0,0 +1,107 @@
...
+          <button type="button" onClick={clearDepartureVariables}>
+            {`Back to all ${stop.name}`}
+          </button>
+          <div className="placeholder-map">imagine a nap</div>
+          <div className="placeholder-departures">
+            {`Route ${departureInfo.departureRoute?.id}`}
+          </div>
+        </div>
+      )}
+      <div className="">
```
kotva006, line 100: nit-pick: className isn't needed

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
kotva006: Okay, the issue is you are using the wrong `PolyLine` import.
This is the reason why I like uniquely named structs.
The polyline type the map is expecting is from `leaflet/components/__mapdata`
You are just importing the `leaflet` one.  Changing imports should fix the issue and allow the removal of this recalculation.

Though you are probably best in calculating it here.  Since this is becoming it's own component, and nothing in the `StopPageRedesign` requires it.  Just remove the polyline calculation in the `StopPageRegesign` as well.

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
i0sea: Keeping setDepartureVariables to wrap the setting into a function. `setDepartureInfo` is a dispatch action and requires a bit more overhead then just wrapping in this function

i0sea:
Ready for another pass. Aware that all the changed files need their tests updated to have an onclick prop and will be fixing in tandem

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -292,18 +297,27 @@ const DepartureTimes = ({
...
+        const preds = predictionsByHeadsign[headsign]
+          ? predictionsByHeadsign[headsign]
+          : [];
+        const formattedTimes = toDisplayTime(schs, preds, overrideDate);
+        departures = mergeIntoDepartureInfo(schs, preds);
+        id += 1;
+        return (
+          <div
+            key={`${headsign}-${id}`}
+            onClick={() => onClick(route, directionId, departures)}
```
KaylaBrady, line 314: suggestion: It looks like the value for route here would be the same for different headsigns, but I think we would want to pass an identifier for the different headsigns so that clicking each of the rows under the 39 would result in show the departures only associated with that headsign.
image

Each headsign corresponds to a [route pattern](https://api-v3.mbta.com/docs/swagger/index.html#/RoutePattern/ApiWeb_RoutePatternController_show). 

I would recommend passing the route_pattern_id instead of the route id here. Or addressing in a separate PR if need be!

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -292,18 +297,27 @@ const DepartureTimes = ({
...
-          overrideDate
+        const preds = predictionsByHeadsign[headsign]
+          ? predictionsByHeadsign[headsign]
+          : [];
+        const formattedTimes = toDisplayTime(schs, preds, overrideDate);
+        departures = mergeIntoDepartureInfo(schs, preds);
+        id += 1;
+        return (
+          <div
+            key={`${headsign}-${id}`}
```
kotva006, line 313: This could probably just be `${headsign}`.  In this array headsign should be unique.  Also its bad practice to use [the index of a list as a key](https://react.dev/learn/rendering-lists#where-to-get-your-key)

i0sea: Interesting using index was also used in `Headsign.tsx`. `${headsign}` isn't always unique looking at the courthouse stop there are multiple `South Station` headsigns

apps/site/assets/ts/stop/__tests__/DeparturesAndMapTest.tsx
```diff
@@ -0,0 +1,34 @@
...
+    let departuresAndMap = render(
+      <DeparturesAndMap
+        routes={[]}
+        stop={stop}
+        schedules={[]}
+        routesWithPolylines={testRoutesWithPolylines}
+      />
+    );
+    const list = getByRole(departuresAndMap.baseElement, "list");
+    expect(list).toBeInTheDocument();
```
KaylaBrady, line 32: suggestion: Consider adding some tests to this file, or making this test more specific. 

For example, what should the expected list have in it - is it the DepartureCard or DepartureList? There should be a test analogous to the "opens on click" test that was a part of StopPageOverlayTest to validate that clicking a row in the DepartureCard will open a row in the DepartureList.

KaylaBrady (commented):
Looks like the checks are failing due to not meeting the target test coverage threshold. I added a few recommendations for tests to add and am happy to talk more about testing in slack!

apps/site/assets/ts/stop/components/DepartureTimes.tsx
KaylaBrady: Based on discussion in slack, I think it is OK for now not to pass route_pattern_id. This can be revisited if needed as part of future implementation work on the DepartureList + realtime map.

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -309,8 +309,8 @@ const DepartureTimes = ({
         return (
           <div
             key={`${headsign}-${route.id}`}
-            onClick={() => onClick(route, directionId, departures)}
-            onKeyDown={() => onClick(route, directionId, departures)}
+            onClick={() => onClick(route, directionId, schedules)}
+            onKeyDown={() => onClick(route, directionId, schedules)}
```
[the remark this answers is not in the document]
i0sea: could pass schs instead but will also need the headsign to fetch predictions in other components

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <StopPageDepartures
+          routes={routes}
+          stop={stop}
+          schedules={schedules}
+          onClick={setDepartureVariables}
+        />
+      ) : (
+        <div className="departures-container">
+          <button
+            type="button"
```
kotva006, line 74: The figma has this styled as a link, its currently a button

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -292,18 +297,23 @@ const DepartureTimes = ({
...
-          predictionsByHeadsign[headsign]
-            ? predictionsByHeadsign[headsign]
-            : [],
-          overrideDate
+        const preds = predictionsByHeadsign[headsign]
+          ? predictionsByHeadsign[headsign]
+          : [];
+        const formattedTimes = toDisplayTime(schs, preds, overrideDate);
+        return (
+          <div
```
kotva006, line 308: This should be styled to show the hand on hover to let the users know its clickable.

kotva006 (changes requested):
I pushed the code up to dev blue to verify its looks.  Though I think something is broken.
When I click a row I am greeted with `No upcoming trips today` even though the row in question has times displaying on the previous screen.

apps/site/assets/css/_stop-card.scss
```diff
@@ -211,3 +206,31 @@ $radius: 4px;
...
+    display: block;
+  }
+}
+
+.departures-container {
+  padding: .5rem;
+  @include media-breakpoint-down(sm) {
+    background-color: $white;
+    box-shadow: 0 0 $space-4 $space-2 $drop-shadow-color;
+    height: 100vh;
```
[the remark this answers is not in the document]
i0sea: To cover the size on the screen 100% only covers the parent container

apps/site/assets/ts/stop/components/DepartureList.tsx
```diff
@@ -0,0 +1,46 @@
...
+import usePredictionsChannel from "../../hooks/usePredictionsChannel";
+
+interface DepartureListProps {
+  route: Route;
+  stop: Stop;
+  schedules: Dictionary<ScheduleWithTimestamp[]>;
+  directionId: DirectionId;
+}
+
+const DepartureList = ({
```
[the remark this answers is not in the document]
i0sea: Yes, added this here to render predictions to make sure they're live updating on the page. This relates to why schedules is being used once again. When passing the departureInfo thru onclick it acts as a snapshot and the predictions don't update live . Open to suggestions on passing streams from child components. For now the most straightforward way seems to be creating the stream directly in the departuresLIst

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -0,0 +1,110 @@
...
+  routes,
+  stop,
+  schedules,
+  routesWithPolylines
+}: DeparturesAndMapProps): ReactElement<HTMLElement> => {
+  const [departureInfo, setDepartureInfo] = useState<{
+    departureRoute: Route | null;
+    departureDirectionId: DirectionId | null;
+    departureSchedules: Dictionary<ScheduleWithTimestamp[]> | null | undefined;
+  }>({
```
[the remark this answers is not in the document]
i0sea: Turns out this requires checking if departureInfo exists before accessing any of the fields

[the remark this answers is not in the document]
kotva006: I am not sure everywhere its checked, but in `DeparturesAndMap` you could replace the `viewAllRoutes()` function with `departureInfo !== null` test.  Then TS should understand that `departureInfo` is not null when its passed to the `StopPageDepartures` component

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -292,18 +297,24 @@ const DepartureTimes = ({
...
+            className="test"
```
KaylaBrady, line 309: suggestion: could you please update this className to something more descriptive?

apps/site/assets/ts/stop/__tests__/components/DepartureTimesTest.tsx
```diff
@@ -397,4 +400,42 @@ describe("DepartureTimes", () => {
...
+        route={route}
+        stop={stop}
+        directionId={0}
+        schedulesForDirection={schedules}
+        onClick={mockClickAction}
+      />
+    );
+
+    expect(departureTimes.container.querySelector(".test")).toBeDefined();
+    fireEvent.click(departureTimes.baseElement.querySelector(".test")!);
```
KaylaBrady, line 438: suggestion (non-blocking): consider using [useEvent](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event) over `fireEvent` since it more closely resembles user interaction

KaylaBrady (approved):
A couple small cleanup comments, but otherwise this looks good to me! 
I agree with @kotva006's comments about a possible refactor for how schedules/predictions are shared between components, could you please add a ticket to inbound to capture that idea?

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -292,18 +297,24 @@ const DepartureTimes = ({
...
-            ? predictionsByHeadsign[headsign]
-            : [],
-          overrideDate
+        const preds = predictionsByHeadsign[headsign]
+          ? predictionsByHeadsign[headsign]
+          : [];
+        const formattedTimes = toDisplayTime(schs, preds, overrideDate);
+        return (
+          <div
+            className="departure-row-click-test"
```
kotva006, line 309: In the future we should query by [test ids](https://testing-library.com/docs/queries/bytestid) instead of class names.  This should only be used if all other query methods are not working.

PR #1596, opened 2023-05-16
Add larger site icons to dotcomchrome export script

wleona3:
#### Summary of changes

Adding the larger site icons to the dotcomchrome script, so that sites using dotcomchrome can mirror the existing icon tags from mbta.com

### DIFF
--- a/apps/site/assets/webpack.config.export-headerfooter.js
+++ b/apps/site/assets/webpack.config.export-headerfooter.js
@@ -103,7 +103,7 @@ module.exports = (env, argv) => {
       // copy images and fonts
       new CopyWebpackPlugin({ patterns: [
             { from: "static/fonts/*", to: "fonts/[name][ext]" },
-            { from: "static/favicon.ico", to: "favicon.ico" },        
+            { from: "static/favicon.ico", to: "favicon.ico" },
             { from: "static/images/mbta-logo.svg", to: "images/mbta-logo.svg" },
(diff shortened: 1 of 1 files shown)


apps/site/assets/webpack.config.export-headerfooter.js
```diff
@@ -103,9 +103,11 @@ module.exports = (env, argv) => {
       // copy images and fonts
       new CopyWebpackPlugin({ patterns: [
             { from: "static/fonts/*", to: "fonts/[name][ext]" },
-            { from: "static/favicon.ico", to: "favicon.ico" },
+            { from: "static/favicon.ico", to: "favicon.ico" },
```
kotva006, line 106: nit-pick: This seems like accidental whitespace

wleona3: Fixed

kotva006:
Could you expand on what you mean by dotcom chrome export script?

wleona3:
The dot com website uses these assets for its various icon tags. In order to replicate those same tags a sister site needs these assets.

wleona3:
Ah yeah that could work, was under the impression that that base64 string would potentially be changing every time you did a build and didn't realize it could be removed. Also was a concern that hat we could also be running into CORS isssues but looks like you've got access control set to *. Does still seeem strange to me that the 16x16 favicon.ico file is included but not the other icon sizes.

thecristen (approved):
Thanks for the PR! Will use this to run the export task and update the https://github.com/mbta/dotcomchrome repo soon.

PR #1598, opened 2023-05-16
feat(DepartureTimes): Add an alert badge to the departure row

kotva006:
#### Summary of changes
Adds a Badge to display alerts in the departure cards instead of the times if an alert is present.

Screen Shot 2023-05-16 at 5 15 41 PM

The alerts effect types this works for are Suspensions, Shuttle Service, and Detours.

This pull request fetches route wide alerts for the new stop redesign page, merges them with the stop specific routes.
The alerts are then divided up by route, and then direction.  To be passed to the corresponding `DepartureTimes` component for the direction.  This is then processed to display the corresponding badge if eligible.

This also breaks out `DepartureInfo` into its own file to help readability.

NOTE: This pull request needs the alert date parsing pull request merged for the badges to display

**Asana Ticket:** [Alerts | Route Specific Badges on Departure Cards](https://app.asana.com/0/555089885850811/1204305469426565)

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/components/Badge.tsx
+++ b/apps/site/assets/ts/components/Badge.tsx
@@ -1,11 +1,19 @@
 import React from "react";
 
-const Badge = ({ text }: { text: string }): JSX.Element => {
+const Badge = ({
+  text,
+  contextText
+}: {
+  text: string;
+  contextText?: string;
+}): JSX.Element => {
   return (
     <div
       className="u-error-background font-weight-bold ps-8 pe-8 fs-14"
       style={{ borderRadius: "0.75rem" }}
     >
+      {/* The purpose of this block is to have invisble text for screen readers*/}
+      {contextText && <span className="u-display-none">{contextText}</span>}
       {text}
     </div>
   );
--- a/apps/site/assets/ts/components/Badge.tsx
+++ b/apps/site/assets/ts/components/Badge.tsx
@@ -3,7 +3,7 @@ import React from "react";
 const Badge = ({ text }: { text: string }): JSX.Element => {
   return (
     <div
-      className="u-error-background font-weight-bold ps-5 pe-5"
+      className="u-error-background font-weight-bold ps-8 pe-8 fs-14"
       style={{ borderRadius: "0.75rem" }}
     >
       {text}
--- a/apps/site/assets/ts/components/Badge.tsx
+++ b/apps/site/assets/ts/components/Badge.tsx
@@ -13,7 +13,7 @@ const Badge = ({
       style={{ borderRadius: "0.75rem" }}
     >
       {/* The purpose of this block is to have invisble text for screen readers*/}
-      {contextText && <span className="u-display-none">{contextText}</span>}
+      {contextText && <span className="sr-only">{contextText}</span>}
       {text}
     </div>
   );
--- a/apps/site/assets/ts/components/Badge.tsx
+++ b/apps/site/assets/ts/components/Badge.tsx
@@ -1,0 +1,14 @@
+import React from "react";
+
+const Badge = ({ text }: { text: string }): JSX.Element => {
+  return (
+    <div
+      className="u-error-background font-weight-bold ps-5 pe-5"
+      style={{ borderRadius: "0.75rem" }}
+    >
+      {text}
+    </div>
+  );
+};
+
+export default Badge;
--- a/apps/site/assets/ts/components/Badge.tsx
+++ b/apps/site/assets/ts/components/Badge.tsx
@@ -12,7 +12,7 @@ const Badge = ({
       className="u-error-background font-weight-bold ps-8 pe-8 fs-14"
       style={{ borderRadius: "0.75rem" }}
     >
-      {/* The purpose of this block is to have invisble text for screen readers*/}
+      {/* The purpose of this block is to have invisble text for screen readers */}
       {contextText && <span className="sr-only">{contextText}</span>}
       {text}
     </div>
--- a/apps/site/assets/ts/models/__tests__/alert-test.ts
+++ b/apps/site/assets/ts/models/__tests__/alert-test.ts
@@ -9,7 +9,8 @@ import {
   isActiveDiversion,
   hasAnActiveDiversion,
   alertsByRoute,
-  alertsByDirectionId
+  alertsByDirectionId,
+  alertsAffectingBothDirections
 } from "../alert";
 
 const zeroPadded = (num: number): string => `${num}`.padStart(2, "0");
@@ -275,7 +276,7 @@ describe("alertsByRoute", () => {
 });
 
 describe("alertsByDirectionId", () => {
-  test("it should group alerts by directionId", () => {
+  test("it should group alerts by directionId, filtering out both directions", () => {
     const alerts = [
       {
         id: "1234",
@@ -300,5 +301,47 @@ describe("alertsByDirectionId", () => {
     expect(Object.keys(result).length).toBe(2);
     expect(result[1].length).toBe(1);
     expect(result[0].length).toBe(1);
+  });
+});
+
+describe("alertsAffectingBothDirections", () => {
+  test("should return only the alerts that affect both directions", () => {
+    const alerts = [
+      {
+        id: "1234",
+        informed_entity: {
+          direction_id: [0]
+        }
+      },
+      {
+        id: "3333",
+        informed_entity: {
+          direction_id: [null]
+        }
+      },
+      {
+        id: "0987",
+        informed_entity: {
+          direction_id: [1]
+        }
+      },
+      {
+        id: "4444",
+        informed_entity: {
+          direction_id: []
+        }
+      },
+      {
+        id: "5555",
+        informed_entity: {
+          direction_id: null
+        }
+      }
+    ] as Alert[];
+    const result = alertsAffectingBothDirections(alerts);
+    expect(result.length).toBe(3);
+    expect(result[0].id).toBe("3333");
+    expect(result[1].id).toBe("4444");
+    expect(result[2].id).toBe("5555");
   });
 });
--- a/apps/site/assets/ts/models/__tests__/alert-test.ts
+++ b/apps/site/assets/ts/models/__tests__/alert-test.ts
@@ -7,7 +7,9 @@ import {
   uniqueByEffect,
   isDiversion,
   isActiveDiversion,
-  hasAnActiveDiversion
+  hasAnActiveDiversion,
+  alertsByRoute,
+  alertsByDirectionId
 } from "../alert";
 
 const zeroPadded = (num: number): string => `${num}`.padStart(2, "0");
@@ -245,5 +247,58 @@ describe("hasAnActiveDiversion", () => {
       ])
     ).toBeTruthy();
     expect(hasAnActiveDiversion(stopId, [nonActiveDiversionAlert])).toBeFalsy();
+  });
+});
+
+describe("alertsByRoute", () => {
+  test("it should group alerts across multiple routes", () => {
+    const alerts = [
+      {
+        id: "1234",
+        informed_entity: {
+          route: ["441442", "442"]
+        }
+      },
+      {
+        id: "4321",
+        informed_entity: {
+          route: ["442", "53"]
+        }
+      }
+    ] as Alert[];
+    const result = alertsByRoute(alerts);
+    expect(Object.keys(result).length).toBe(3);
+    expect(result["442"].length).toBe(2);
+    expect(result["441442"].length).toBe(1);
+    expect(result["53"].length).toBe(1);
+  });
+});
+
+describe("alertsByDirectionId", () => {
+  test("it should group alerts by directionId", () => {
+    const alerts = [
+      {
+        id: "1234",
+        informed_entity: {
+          direction_id: [0]
+        }
+      },
+      {
+        id: "4321",
+        informed_entity: {
+          direction_id: [null]
+        }
+      },
+      {
+        id: "0987",
+        informed_entity: {
+          direction_id: [1]
+        }
+      }
+    ] as Alert[];
+    const result = alertsByDirectionId(alerts);
+    expect(Object.keys(result).length).toBe(2);
+    expect(result[1].length).toBe(1);
+    expect(result[0].length).toBe(1);
   });
 });
--- a/apps/site/assets/ts/models/__tests__/alert-test.ts
+++ b/apps/site/assets/ts/models/__tests__/alert-test.ts
@@ -293,14 +293,14 @@ describe("alertsByDirectionId", () => {
       {
         id: "0987",
         informed_entity: {
-          direction_id: [1]
+          direction_id: [0, 1]
         }
       }
     ] as Alert[];
     const result = alertsByDirectionId(alerts);
     expect(Object.keys(result).length).toBe(2);
     expect(result[1].length).toBe(1);
-    expect(result[0].length).toBe(1);
+    expect(result[0].length).toBe(2);
   });
 });
 
--- a/apps/site/assets/ts/models/alert.ts
+++ b/apps/site/assets/ts/models/alert.ts
@@ -23,13 +23,13 @@ export const alertsByStop = (alerts: Alert[], stopId: StopId): Alert[] =>
 const hasEffect = (alerts: Alert[], effect: string): boolean =>
   some(alerts, alert => alert.effect === effect);
 
-export const isSuspension = (alerts: Alert[]): boolean =>
+export const hasSuspension = (alerts: Alert[]): boolean =>
   hasEffect(alerts, "suspension");
 
-export const isShuttleService = (alerts: Alert[]): boolean =>
+export const hasShuttleService = (alerts: Alert[]): boolean =>
   hasEffect(alerts, "shuttle");
 
-export const isDetour = (alerts: Alert[]): boolean =>
+export const hasDetour = (alerts: Alert[]): boolean =>
   hasEffect(alerts, "detour");
 
 export const alertsWithStop = (alerts: Alert[]): Alert[] =>
--- a/apps/site/assets/ts/models/alert.ts
+++ b/apps/site/assets/ts/models/alert.ts
@@ -42,7 +42,6 @@ export const routeWideAlerts = (alerts: Alert[]): Alert[] =>
     entities.some(entity => !entity.stop && !entity.trip)
   );
 
-// Can there be alerts that don't affect routes
 export const alertsByRoute = (alerts: Alert[]): { [key: string]: Alert[] } => {
   return reduce(
     alerts,
--- a/apps/site/assets/ts/models/alert.ts
+++ b/apps/site/assets/ts/models/alert.ts
@@ -1,4 +1,5 @@
 import { isValid, parseISO } from "date-fns";
+import { isArray, mergeWith, reduce, some } from "lodash";
 import { StopId } from "../schedule/components/__schedule";
 import { Alert, TimePeriodPairs } from "../__v3api";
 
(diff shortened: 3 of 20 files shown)


apps/site/assets/ts/components/Badge.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+const Badge = ({ text }: { text: string }): JSX.Element => {
```
KaylaBrady, line 3: praise: Appreciate this being broken out for reusability! 
suggestion: Is there a role that should be added for accessibility? Maybe the [alert](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role) role?

apps/site/assets/ts/models/__tests__/alert-test.ts
```diff
@@ -247,3 +249,56 @@ describe("hasAnActiveDiversion", () => {
...
+      {
+        id: "1234",
+        informed_entity: {
+          direction_id: [0]
+        }
+      },
+      {
+        id: "4321",
+        informed_entity: {
+          direction_id: [null]
```
KaylaBrady, line 289: question (non-blocking): It seems this test is additionally testing that alerts without direction_id are being filtered out. Is that intended? If so, may be worth mentioning in the test name?

apps/site/assets/ts/models/alert.ts
```diff
@@ -19,6 +20,93 @@ export const alertsByStop = (alerts: Alert[], stopId: StopId): Alert[] =>
       !!entities.stop && entities.stop!.some((id: StopId) => id === stopId)
   );
 
+const hasEffect = (alerts: Alert[], effect: string): boolean =>
```
KaylaBrady, line 23: praise: Appreciate all these helpers!

apps/site/assets/ts/models/alert.ts
```diff
@@ -19,6 +20,93 @@ export const alertsByStop = (alerts: Alert[], stopId: StopId): Alert[] =>
...
+  alerts.filter(
+    ({ informed_entity: entites }: Alert): boolean => !!entites.stop
+  );
+
+export const routeWideAlerts = (alerts: Alert[]): Alert[] =>
+  alerts.filter(({ informed_entity: { entities } }: Alert): boolean =>
+    entities.some(entity => !entity.stop && !entity.trip)
+  );
+
+// Can there be alerts that don't affect routes
```
KaylaBrady, line 45: question: Is this question resolved?

apps/site/assets/ts/models/alert.ts
```diff
@@ -19,6 +20,93 @@ export const alertsByStop = (alerts: Alert[], stopId: StopId): Alert[] =>
...
+export const alertsByRoute = (alerts: Alert[]): { [key: string]: Alert[] } => {
+  return reduce(
+    alerts,
+    (acc, alert) => {
+      // create a map of every route id to the alert
+      const explodedAlerts = reduce(
```
KaylaBrady, line 51: suggestion (non-blocking): `exploded` to me doesn't imply what the comment suggests this data type is. Maybe `alertByRoute` would be helpful?

Also non-blocking: consider using lodash [group_by](https://lodash.com/docs/4.17.15#groupBy) to simplify this function

apps/site/assets/ts/models/alert.ts
```diff
@@ -19,6 +20,93 @@ export const alertsByStop = (alerts: Alert[], stopId: StopId): Alert[] =>
...
+          return objValue.concat(srcValue);
+        }
+        return srcValue;
+      });
+    },
+    {}
+  );
+};
+
+// This will only return alerts with direction id specified (aka affecting a single direction)
```
KaylaBrady, line 70: question: Does null direction id imply that an alert affects both directions? If so, why would we want to exclude those alerts?

apps/site/assets/ts/models/alert.ts
```diff
@@ -19,6 +20,93 @@ export const alertsByStop = (alerts: Alert[], stopId: StopId): Alert[] =>
...
+  );
+};
+
+export const alertsAffectingBothDirections = (alerts: Alert[]): Alert[] => {
+  return alerts.filter(alert => {
+    return (
+      alert.informed_entity.direction_id === null ||
+      alert.informed_entity.direction_id.length === 0 ||
+      alert.informed_entity.direction_id[0] === null
+    );
```
KaylaBrady, line 106: question: would `direction_id` list ever look like `[0, 1]` to represent both directions?

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
```diff
@@ -48,11 +53,39 @@ describe("StopPageRedesign", () => {
...
+    });
+    jest.spyOn(useStop, "default").mockImplementation(() => {
+      return undefined;
+    });
+
+    render(<StopPageRedesign stopId="123" />);
+    expect(screen.queryByText("Loading...")).toBeDefined();
+  });
+
+  it("shows Loading without alerts", () => {
```
KaylaBrady, line 77: thought (non-blocking): Is this the expected behavior?

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
```diff
@@ -85,28 +118,103 @@ describe("StopPageRedesign", () => {
     expect(mapPolylines).toHaveLength(10);
   });
 
+  const dateFormatter = (date: Date): string => {
+    return format(date, "yyyy-M-d HH:mm");
+  };
```
KaylaBrady, line 123: suggestion (non-blocking): Perhaps this would be useful in a test helper with a more descriptive name like `alertDateFormatter`

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
```diff
@@ -85,28 +118,103 @@ describe("StopPageRedesign", () => {
...
-      .mockImplementation(() => [lowAlert]);
+      .mockImplementation(() => lowAlerts);
 
     render(<StopPageRedesign stopId="123" />);
     expect(
       screen.queryByText("There is construction at this station.")
     ).not.toBeNull();
   });
+
+  it("should only render current alerts and not future alerts", () => {
```
KaylaBrady, line 154: suggestion: Should there be a test for past alerts?

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -291,19 +173,22 @@ const DepartureTimes = ({
...
+  const schedulesForHeadsign = schedulesByHeadsign(schedulesForDirection);
 
   return (
     <>
-      {Object.entries(schedules).map(([headsign, schs]) => {
-        const formattedTimes = toDisplayTime(
+      {Object.entries(schedulesForHeadsign).map(([headsign, schs]) => {
+        const predictions = predictionsByHeadsign[headsign]
+          ? predictionsByHeadsign[headsign]
+          : [];
```
KaylaBrady, line 184: nitpick (non-blocking): Could be simplified to `predictionsByHeadsign[headsign] || []`

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -8,30 +8,68 @@ import StopPageHeaderRedesign from "./StopPageHeaderRedesign";
...
+  const alertsForRoutes = useAlertsByRoute(routeIdState);
   //
 
+  useEffect(() => {
+    setRouteState(
+      routesWithPolylines
+        ? routesWithPolylines.map(rwp => omit(rwp, "polylines"))
+        : []
+    );
+  }, [routesWithPolylines]);
```
KaylaBrady, line 40: question: Why is this needed in a useEffect hook? When would routesWithPolylines change?

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -8,30 +8,68 @@ import StopPageHeaderRedesign from "./StopPageHeaderRedesign";
...
+import { useAlertsByRoute, useAlertsByStop } from "../../hooks/useAlerts";
+import { isCurrentAlert, routeWideAlerts } from "../../models/alert";
 
 const StopPageRedesign = ({
   stopId
 }: {
   stopId: string;
 }): ReactElement<HTMLElement> => {
+  const [routeIdState, setRouteIdState] = useState<string[]>([]);
+  const [routeState, setRouteState] = useState<Route[]>([]);
```
KaylaBrady, line 23: question: Are these needed as separate pieces of state? Couldn't routeIds be calculated from routeState?

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -8,30 +8,68 @@ import StopPageHeaderRedesign from "./StopPageHeaderRedesign";
...
+    // routeWideAlertsArray are all the alerts that affect the whole route
+    // not just specific stops
+    const routeWideAlertsArray = routeWideAlerts(alertsRouteArray);
+    // Get only alerts that are current
+    const currentAlerts = filter(
+      alertsArray.concat(routeWideAlertsArray),
+      alert => isCurrentAlert(alert)
+    );
+    setAlertState(currentAlerts);
+  }, [alerts, alertsForRoutes]);
```
KaylaBrady, line 60: question: Is this needed in a useEffect hook rather than a helper function? Does this fit into a case from [You Might Not Need An Effect](https://react.dev/learn/you-might-not-need-an-effect)?

KaylaBrady (commented):
Looking good! A few minor comments

KaylaBrady (commented):
Looking good! A few minor comments

apps/site/assets/ts/components/Badge.tsx
kotva006: I am a bit hesitant to use an alert role as it sounds like screen readers just say it, regardless of where the user is looking.  So if one route is suspended, they will just hear "Suspended" upon the page loading.  Also they would hear multiple if multiple badges are present.  This is what is says on the documentation about the alert role:
`Warning: Because of its intrusive nature, the alert role must be used sparingly and only in situations where the user's immediate attention is required.`

KaylaBrady: Ah, good point! Maybe [status](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role) would be more appropriate then?

apps/site/assets/ts/models/alert.ts
KaylaBrady: Ah thanks, I missed that the alerts => multiple routes aspect. It looks to me like on line 54 a routeId is only mapped to a list containing a single alert.

Maybe [flatMap](https://lodash.com/docs/4.17.15#flatMap) could be used to get a flattened list of [{route_id, alert}], and `groupBy` could be called on that flattened list?

apps/site/assets/ts/models/alert.ts
kotva006: I believe not.  It looks like in the UI for alerts creation you are only allowed to click `Both Directions`, `Eastbound`, or `Westbound` (eastbound and westbound are route dependent wordings).  You can't select stops from both directions (IE I can't create an alert that affects stop 1 in direction 0, and stop 2 in direction 1.  That is 2 seperate alerts.  So the array should never have more than one direction id in it.  My testing indicates that dotcom transforms both direction alerts are of the form `[null]`.  The rest of this is just to cover bases.

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
kotva006: Past alerts do not get handed to us from dotcom.  I guess if someone sits on a page long enough the current alert, could become a past alert.

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
kotva006: I did some re-arranging of code, and managed to remove the state requirements

apps/site/assets/ts/models/alert.ts
kotva006: This was broken out because `obj[null]` isn't allowed by typescript.
`alertsByDirectionId` would map alerts affecting both directions to the key of "null".
The only way to access it would be to use `alertsByDirectionObj["null"]` (instead of `alertsByDirectionObj[null]`).
Typescript discourages this as it seems error prone. I broke it out into its own function instead.

apps/site/assets/ts/components/Badge.tsx
kotva006: I am leaning towards we don't need to add more context.  I think the text precisely describes that status of the route.
I imagine the screen reader saying the visible text "Wonderland Suspended" or "Wellington Shuttle Service".  Under normal circumstances it would say "Wonderland 5min 15min".  That seems like enough context to put together what's happening.

apps/site/assets/ts/components/Badge.tsx
```diff
@@ -0,0 +1,22 @@
...
+  text: string;
+  contextText?: string;
+}): JSX.Element => {
+  return (
+    <div
+      className="u-error-background font-weight-bold ps-8 pe-8 fs-14"
+      style={{ borderRadius: "0.75rem" }}
+    >
+      {/* The purpose of this block is to have invisble text for screen readers*/}
+      {contextText && <span className="u-display-none">{contextText}</span>}
```
KaylaBrady, line 16: suggestion (non-blocking): I think there is an existing `sr-only` class name that could be applied here

apps/site/assets/ts/models/alert.ts
KaylaBrady: @kotva006 is there an existing test case for when there is more than one route associated with an alert? Want to confirm this is working as expected

apps/site/assets/ts/models/alert.ts
```diff
@@ -19,6 +20,92 @@ export const alertsByStop = (alerts: Alert[], stopId: StopId): Alert[] =>
...
+): { [key: number]: Alert[] } => {
+  return reduce(
+    alerts,
+    (acc, alert) => {
+      const explodedAlerts = reduce(
+        alert.informed_entity.direction_id,
+        (innerAcc, directionId) => {
+          // only include specified directions
+          if (directionId !== null) {
+            return { ...innerAcc, [directionId]: [alert] };
```
KaylaBrady, line 81: question: similar to the comment on `alertsByRoute`, is there a test case for when there are multiple alerts in a direction?
This line reads to me like there will only ever be one alert per directionId, but I may be overlooking something.

apps/site/assets/ts/models/alert.ts
KaylaBrady: Agree that it seems error prone to have a [null] key. 
Suggestion (non-blocking): When directionId is null, I wonder if it would simplify things to add it to the list *both* directionIds so that alertsByDirectionId returns a full picture and a separate function isn't needed.

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -299,20 +181,24 @@ const DepartureTimes = ({
...
+        const predictions = predictionsByHeadsign[headsign] || [];
         return (
           <div
+            // TODO remove this class name in favor of test ids
             className="departure-row-click-test"
             key={`${headsign}-${route.id}`}
-            onClick={() => onClick(route, directionId, schs)}
-            onKeyDown={() => onClick(route, directionId, schs)}
+            onClick={() => onClick(route, directionId, schedules)}
+            onKeyDown={() => onClick(route, directionId, schedules)}
```
KaylaBrady, line 192: question: Is this change intentional? Seems like it may have been a merge conflict resolution issue

apps/site/assets/ts/models/alert.ts
```diff
@@ -19,6 +20,92 @@ export const alertsByStop = (alerts: Alert[], stopId: StopId): Alert[] =>
       !!entities.stop && entities.stop!.some((id: StopId) => id === stopId)
   );
 
+const hasEffect = (alerts: Alert[], effect: string): boolean =>
+  some(alerts, alert => alert.effect === effect);
+
+export const isSuspension = (alerts: Alert[]): boolean =>
```
KaylaBrady, line 26: nitpick(non-blocking): Consider renaming these `hasX` instead of `isX` since it is looking in a list of alerts

KaylaBrady (commented):
This is looking good to me! Only blocking comments are related to tests for when there is more than one route for an alert and more than one alert for a direction. Once those are in place this is good to go from my perspective!

apps/site/assets/ts/models/alert.ts
kotva006: @KaylaBrady yeah [this test](https://github.com/mbta/dotcom/blob/61d64317396a890dcaf344d1fc33f46983ca1b12/apps/site/assets/ts/models/__tests__/alert-test.ts#L255) tests that a single route can have multiple alerts

KaylaBrady: Thank you! I see the problem now with how I was interpreting this function.

Now I understand: `alert.informed_entity.route` is a list of unique route ids for an alert, so each route/alert combo added to the accumulator on line 53 is unique. 

Still non-blocking feedback, but perhaps some of the suggested functions above help simplify this function so it is easier to read. Alternatively, consider giving the accumulators more descriptive names or breaking out the nested reduces into helper functions.

apps/site/assets/ts/models/alert.ts
kotva006: Yeah this function (and `alertsByRoute` is a bit complicated.  What this function does is create an object with directionId as the key, and for a value an array of alerts affecting that route.
Yeah are right in thinking that this part of the code returns a single alert for directionId.

This inner `reduce` takes a single alert, and creates an object with each directionId it maps the alert to it.
Ex
`const alert = { id: 1, informed_entity: {direction_id: [0, 1]}};`
Becomes:
```
{
  0: [{id: 1, informed_entity: {direction_id: [0, 1]}],
  1: [{id: 1, informed_entity: {direction_id: [0, 1]}}]
}
  ```
 
The 2nd part of the function below is where the combination happens.  The `mergeWith` function takes the result of the inner `reduce` (explodedAlerts), and combines with the accumulator from the outer `reduce` (starts as an empty object {}).  This will combine the 2 objects and merge the array values of the matching keys.  So example:
```
{
  0: [{id: 1, informed_entity: {direction_id: [0, 1]}],
  1: [{id: 1, informed_entity: {direction_id: [0, 1]}}]
}

{
  0: [{id: 2, informed_entity: {direction_id: [0, 1]}],
  1: [{id: 2, informed_entity: {direction_id: [0, 1]}}]
}
  ```
  Becomes:
  ```
{
  0: [{id: 1, informed_entity: {direction_id: [0, 1]}, {id: 2, informed_entity: {direction_id: [0, 1]}],
  1: [{id: 1, informed_entity: {direction_id: [0, 1]}}, {id: 2, informed_entity: {direction_id: [0, 1]}]
}
```
after the merge with.
This last object is what is returned from the function

Comment (kotva006) [reply apps/site/assets/ts/models/alert.ts:81]: Also there isn't a test that has more than one alert for the direction.  I will modify the test to include one.  Though this function follows the same logic as the `alertsByRoute` which does have a test for multiple alerts for the route.

Comment (KaylaBrady) [reply apps/site/assets/ts/models/alert.ts:81]: I see, thanks for explaining! I was getting tripped up in the nested reduces and keeping track of what was being accumulated at each level. May be worth considering some of the same recommendations applied from the comment on the alertsByRoute thread here. 

For example, perhaps this function could be simplified to  (pseudocode)
```
chain(alerts)
.flatmap(alert => alert.informed_entity.direction_id.map(direction_id => {direction_id, alert))
.groupBy((alert) => alert.direction_id)
.value()
```

KaylaBrady (approved):
✅ Thanks for confirming the behavior of those functions!

PR #1602, opened 2023-05-19
Proposed changes

kotva006:
This is what I am proposing by using the departure info, instead of schedules.  It reduces the need to recalculate, and refetch a lot of the data.


i0sea:
When I tried this approach the predictions were no longer updating live. It just captured a snapshot of departureInfo on click. Out of curiosity  does the departureList component appear to update for you?

kotva006:
> When I tried this approach the predictions were no longer updating live. It just captured a snapshot of departureInfo on click. Out of curiosity does the departureList component appear to update for you?

Yeah you are right the predictions won't update when the data is passed that way.  Forgot about that requirement.  I have another thought then.  Will close this pull request.

PR #1606, opened 2023-05-22
fix(DepartureCardTest): Fix test from merge

kotva006:
It looks like a test was broken with a merge conflict.
This addresses it.


i0sea (approved):
🚀

PR #1617, opened 2023-05-26
feat(StopPageRedesign): Return error message if key data fails to load

KaylaBrady:
#### Summary of changes

**Asana Ticket:** [Stop Page | Show helpful message when there's an error fetching page-level data](https://app.asana.com/0/385363666817452/1204541822486710/f)

This PR makes use of the existing `FetchState` type to record if an error is returned from calls to `useSWR`. That type is defined for use with a different and less used hook `useFetch`. It seemed like the simplest starting point to reuse that type, but open to feedback on whether a new type should be defined!

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
+++ b/apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
@@ -1,9 +1,5 @@
 import React from "react";
-import {
-  screen,
-  waitForElementToBeRemoved,
-  within
-} from "@testing-library/dom";
+import { screen, within } from "@testing-library/dom";
 import { render } from "@testing-library/react";
 import StopPageRedesign from "../components/StopPageRedesign";
 import * as useStop from "../../hooks/useStop";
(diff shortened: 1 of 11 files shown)


apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
```diff
@@ -56,27 +52,27 @@ describe("StopPageRedesign", () => {
...
-      .mockImplementation(() => undefined);
+      .mockReturnValue({ status: FetchStatus.Data, data: undefined });
     jest
       .spyOn(useAlerts, "useAlertsByRoute")
-      .mockImplementation(() => undefined);
+      .mockReturnValue({ status: FetchStatus.Data, data: undefined });
 
     render(<StopPageRedesign stopId="123" />);
-    expect(screen.queryByText("Loading...")).toBeDefined();
+    expect(screen.getByText("Loading...")).toBeDefined();
```
KaylaBrady, line 75: This test was passing, but the loading state was not actually being triggered. 
This is because `queryBy...`[ returns null](https://testing-library.com/docs/react-testing-library/cheatsheet/#queries) if there is no match, but `toBeDefined()` is only checking if the result is undefined. so `expect(null).toBeDefined()` passes. Updating these two instances to use `getByText` instead which will throw an error if there is no match. 
Going to look for other instances of `queryBy...` in the stop page redesign tests and address in a separate PR or make a ticket to do so.

thecristen (changes requested):
Nice! Just two small things to address here.

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -10,28 +10,50 @@ import { useSchedulesByStop } from "../../hooks/useSchedules";
...
-  const alertsForStop = useAlertsByStop(stopId);
-  const alertsForRoutes = useAlertsByRoute(routeIds);
-  //
+  // Return loading indicator while waiting on data fetch
+  if (
+    !stopResult.data ||
+    !routesWithPolylinesResult.data ||
+    !schedulesResult.data ||
+    !alertsForRoutesResult.data ||
+    !alertsForStopResult.data
```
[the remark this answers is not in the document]
KaylaBrady: Ah thanks for noticing that! I've updated to do the opposite - since we are checking that data exists here, we can remove the fallback on lines 55-56. I think we would want to preserve the loading state while alerts load based on [this test case](https://github.com/mbta/dotcom/blob/b323b669f9f71131e77ee978117b03cdef2477af/apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx#L70)

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
KaylaBrady: Opened separate PR: https://github.com/mbta/dotcom/pull/1617

thecristen (approved):
LGTM!

PR #1620, opened 2023-06-01
feat(Schedules): parse arrival and departure times

thecristen:
#### Summary of changes

**Asana Ticket:** [Tech Debt - add arrival and departure times to Schedules struct](https://app.asana.com/0/385363666817452/1204737025261971/f)

In order to support proper handling of display countdown logic, we need to be able to distinguish between departure and arrival times in both predictions and schedules. This PR adds this information for schedules.

image


apps/schedules/lib/parser.ex
```diff
@@ -104,6 +108,19 @@ defmodule Schedules.Parser do
...
   end
 
+  defp arrival_time(%JsonApi.Item{attributes: %{"arrival_time" => time}}) when not is_nil(time) do
+    Timex.parse!(time, "{ISO:Extended}")
+  end
+
+  defp arrival_time(_), do: nil
+
+  defp departure_time(%JsonApi.Item{attributes: %{"departure_time" => time}})
+       when not is_nil(time) do
```
i0sea, line 118: suggestion: could this be `when time do` or `time != nil`? To me `not is_nil` is a little confusing

i0sea (approved):
lgtm, one suggestion

apps/schedules/lib/parser.ex
thecristen: yeah I can do that - added

PR #1621, opened 2023-06-01
refactor(Predictions.Parser): favor arrival over departure time

thecristen:
No ticket, a quick experiment to see if this can be a quick win that makes our displayed predictions better!


thecristen:
Was completed in #1634

PR #1622, opened 2023-06-01
Filter Out Certain Alert Banners:

i0sea:
#### Summary of changes

**Asana Ticket:** [Filter Out Certain Alert Banners:](https://app.asana.com/0/385363666817452/1204716014042468/f)

Filter Alerts shown on stops landing page to just stop/station closures, shuttles, and suspensions happening now or within the next 7 days. Consolidated isCurrent with nextXDays to function since there was overlap. Open to suggestions on cleaner refactor.
---

#### General checks

### DIFF
--- a/apps/site/assets/ts/models/__tests__/alert-test.ts
+++ b/apps/site/assets/ts/models/__tests__/alert-test.ts
@@ -2,7 +2,7 @@ import { StopId } from "../../schedule/components/__schedule";
 import { Alert, InformedEntitySet, TimePeriodPairs } from "../../__v3api";
 import {
   isHighSeverityOrHighPriority,
-  isCurrentAlert,
+  isInNextXDays,
   alertsByStop,
   uniqueByEffect,
   isDiversion,
@@ -12,6 +12,7 @@ import {
   alertsByDirectionId,
   alertsAffectingBothDirections
 } from "../alert";
+import { add } from "date-fns";
 
 const zeroPadded = (num: number): string => `${num}`.padStart(2, "0");
 
@@ -40,6 +41,17 @@ export const beforeNow = () => {
     [
       activePeriodDateFormatted(now - 2 * five_minutes),
       activePeriodDateFormatted(now - five_minutes)
+    ] as TimePeriodPairs
+  ];
+};
+
+export const threeDaysFromNow = () => {
+  const now = Date.now();
+  const five_minutes = 300_000;
+  return [
+    [
+      activePeriodDateFormatted(now - five_minutes),
+      activePeriodDateFormatted(add(now, { days: 3 }).getMilliseconds())
     ] as TimePeriodPairs
   ];
 };
@@ -125,7 +137,7 @@ describe("isDiversion", () => {
   });
 });
 
-describe("isCurrentAlert", () => {
+describe("isInNextXDays", () => {
   test.each`
     alert     | isCurrent
     ${alert1} | ${true}
@@ -137,15 +149,25 @@ describe("isCurrentAlert", () => {
     ${alert7} | ${true}
     ${alert8} | ${true}
   `(
-    "isCurrentAlert returns whether alert is current based on lifecycle and active period",
+    "isInNextXDays returns whether alert is current based on lifecycle and active period given days from now is 0",
     ({ alert, isCurrent }) => {
       if (isCurrent) {
-        expect(isCurrentAlert(alert, testDate)).toBeTruthy();
+        expect(isInNextXDays(alert, 0, testDate)).toBeTruthy();
       } else {
-        expect(isCurrentAlert(alert, testDate)).toBeFalsy();
+        expect(isInNextXDays(alert, 0, testDate)).toBeFalsy();
       }
     }
   );
+
+  test("true if the alert is in the next 7 days", () => {
+    const alert: Alert = {
+      active_period: threeDaysFromNow(),
+      lifecycle: "new",
+      effect: "shuttle"
+    } as Alert;
+
+    expect(isInNextXDays(alert, 7)).toBeTruthy();
+  });
 });
 
 describe("alertsByStop", () => {
--- a/apps/site/assets/ts/models/alert.ts
+++ b/apps/site/assets/ts/models/alert.ts
@@ -1,4 +1,4 @@
-import { isValid, parseISO } from "date-fns";
+import { isValid, parseISO, add } from "date-fns";
 import { isArray, mergeWith, reduce, some } from "lodash";
 import { StopId } from "../schedule/components/__schedule";
 import { Alert, TimePeriodPairs } from "../__v3api";
@@ -141,24 +141,34 @@ const isCurrentLifecycle = ({ lifecycle }: Alert): boolean =>
   lifecycle === "ongoing" ||
   lifecycle === "ongoing_upcoming";
 
-export const isCurrentAlert = (
+export const isInNextXDays = (
   alert: Alert,
+  days: number,
   currentDate: Date = new Date()
 ): boolean => {
+  const xDays = add(Date.now(), { days });
+  xDays.setHours(24, 0, 0, 0); // set to midnight
   if (!alert.active_period) return false;
   const dateRanges = alert.active_period.map(ap => activePeriodToDates(ap));
   const isInARange = dateRanges.some((range): boolean => {
     const [start, end] = range;
     if (!start || !isValid(start)) return false; // end might be null for ongoing alerts
+
     return (
-      currentDate >= start && (end && isValid(end) ? currentDate <= end : true)
+      currentDate >= start &&
+      // eslint-disable-next-line
+      (end && isValid(end)
+        ? days === 0
+          ? currentDate <= end
+          : currentDate <= xDays
+        : true)
     );
   });
-  return isCurrentLifecycle(alert) && isInARange;
+  return days === 0 ? isCurrentLifecycle(alert) && isInARange : isInARange;
 };
 
 export const isActiveDiversion = (alert: Alert): boolean =>
-  isDiversion(alert) && isCurrentAlert(alert);
+  isDiversion(alert) && isInNextXDays(alert, 0);
 
 export const hasAnActiveDiversion = (
   stopId: StopId,
--- a/apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
+++ b/apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
@@ -130,7 +130,7 @@ describe("StopPageRedesign", () => {
     return format(date, "yyyy-M-d HH:mm");
   };
 
-  it("should render alerts", () => {
+  it("should render shuttle alerts", () => {
     const now = new Date();
     const future1 = add(now, { days: 1 });
     const lowAlerts: Alert[] = [
@@ -143,7 +143,7 @@ describe("StopPageRedesign", () => {
         informed_entity: {} as InformedEntitySet,
         id: "00005",
         header: "There is construction at this station.",
-        effect: "other",
+        effect: "shuttle",
         description: "",
         url: "https://www.mbta.com"
       }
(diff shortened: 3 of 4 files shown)


apps/site/assets/ts/models/__tests__/alert-test.ts
```diff
@@ -44,6 +45,17 @@ export const beforeNow = () => {
   ];
 };
 
+export const threeDaysFromNow = () => {
+  const now = Date.now();
+  const five_minutes = 300_000;
+  return [
+    [
+      activePeriodDateFormatted(now - five_minutes),
```
kotva006, line 53: This can be simplified as `activePeriodDateFormatted(add(now, { minutes: -5}).getMilliseconds())`

apps/site/assets/ts/models/__tests__/alert-test.ts
```diff
@@ -44,6 +45,17 @@ export const beforeNow = () => {
...
+      activePeriodDateFormatted(add(now, { days: 3 }).getMilliseconds())
```
kotva006, line 54: Wondering if it is worth making `activerPeriodDateFormatted` take in Dates and milliseconds.  The first thing it does is cast the milliseconds to a `Date` object.  Then no need to call `getMilliseconds` on these calls

apps/site/assets/ts/models/alert.ts
```diff
@@ -141,24 +141,34 @@ const isCurrentLifecycle = ({ lifecycle }: Alert): boolean =>
...
   lifecycle === "ongoing_upcoming";
 
-export const isCurrentAlert = (
+export const isInNextXDays = (
   alert: Alert,
+  days: number,
   currentDate: Date = new Date()
 ): boolean => {
+  const xDays = add(Date.now(), { days });
+  xDays.setHours(24, 0, 0, 0); // set to midnight
```
kotva006, line 150: question: Is the intent here to get the start of the day?

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -58,7 +58,11 @@ const StopPageRedesign = ({
   // Get only alerts that are current
   const currentAlerts = filter(
     alertsForStopResult.data.concat(routeWideAlertsArray),
-    alert => isCurrentAlert(alert)
+    alert =>
+      isInNextXDays(alert, 7) &&
+      ["suspension", "stop_closure", "station_closure", "shuttle"].includes(
```
kotva006, line 63: I'd break this filtering out into a separate line/step. Its doing to much filtering by date and effect

kotva006 (commented):
Have we cleared some of this behavior with Anna?
The ticket specifies that this filtering only affects baner alerts.  This looks like the filtering is affecting all alerts used by the page.  As is this seems like if there is an upcoming suspension we won't display any of the times for the next 7 days which seems wrong.
The ticket has different bullet points for Active, and Upcoming.  I'd ask Anna if she has any differences she wants to display.

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -56,16 +60,21 @@ const StopPageRedesign = ({
   // not just specific stops or trips
   const routeWideAlertsArray = routeWideAlerts(alertsForRoutesResult.data);
   // Get only alerts that are current
+  const alertsWithinSevenDays = filter(
+    alertsForStopResult.data.concat(routeWideAlertsArray),
```
kotva006, line 64: nit-pick: This could be set as its own variable (`allAlerts`) for reuse in the filter below as well.

i0sea:
> Is this meant to be a draft?

Yeah, resolving some test issues from merge. Will request your review when it's ready

apps/site/assets/ts/models/alert.ts
```diff
@@ -159,24 +159,33 @@ const isCurrentLifecycle = ({ lifecycle }: Alert): boolean =>
   lifecycle === "ongoing" ||
   lifecycle === "ongoing_upcoming";
 
-export const isCurrentAlert = (
+export const isBannerAlert = ({ effect }: Alert): boolean =>
+  ["suspension", "stop_closure", "station_closure", "shuttle"].includes(effect);
```
kotva006, line 163: suggestion(nit-pick): This is for testing if its a banner alert for the stop page. I'd move this logic to the stop page as that is where this code makes sense.  Other pages may have different alerts for the banners.  Also we have a website wide banner alert this may be confused with for testing.

apps/site/assets/ts/models/alert.ts
```diff
@@ -159,24 +159,33 @@ const isCurrentLifecycle = ({ lifecycle }: Alert): boolean =>
...
-    return (
-      currentDate >= start && (end && isValid(end) ? currentDate <= end : true)
-    );
+    if (!start || !isValid(start)) return false; // start might be null for ongoing alerts
+    if (!end || !isValid(end)) return true; // end might be null for ongoing alerts
+
+    if (days === 0) {
+      return start <= currentDate && end >= currentDate;
+    }
+    return start <= xDays && end >= currentDate && end <= xDays;
```
kotva006, line 182: I don't think start can be null for ongoing alerts.  Based off the previous logic (and here) false is returned if there is not a start, so it shouldn't have a comment about being an ongoing alert if start is null.

I think the `currentDate` variable can be removed.  You can make days optional, and set it to 0 by default (maybe that is too confusing).  I say that because it looks like `currentDate` equals `xDays` when days is `0`

The logic at the end needs some rethinking.  It looks like this will return false if the alert ends after `xDays` even though it started before `xDays`. That doesn't seem right.  I think you just want to replication the `start <= xDays && end >= xDays`

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
```diff
@@ -161,10 +161,70 @@ describe("StopPageRedesign", () => {
     ).not.toBeNull();
   });
 
-  it("should only render current alerts and not future alerts", () => {
+  it("should only render closures, shuttles, and supensions", () => {
```
kotva006, line 164: This test doesn't test for suspensions.  One should be added

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
```diff
@@ -161,10 +161,70 @@ describe("StopPageRedesign", () => {
...
+
+    render(<StopPageRedesign stopId="Test 1" />);
+
+    expect(screen.getByText(/Road Closed/)).toBeInTheDocument();
+    expect(screen.getByText(/Stop Closed/)).toBeInTheDocument();
+    expect(screen.getByText(/Station Closed/)).toBeInTheDocument();
+    expect(screen.queryByText(/The Walkway has spillage/)).toBeNull();
+  });
+
+  it("should only render current stop alerts and route alerts within 7 days", () => {
```
kotva006, line 222: We should add a test for the 0 days, as that was the original behavior.

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
```diff
@@ -161,10 +161,70 @@ describe("StopPageRedesign", () => {
...
     const now = new Date();
     const future1 = add(now, { days: 1 });
     const future2 = add(now, { days: 3 });
+    const future3 = add(now, { days: 8 });
+    const future4 = add(now, { days: 9 });
```
kotva006, line 227: This test is missing an alert who starts before the next x days, and then ends after the next x days

apps/site/assets/ts/stop/components/DepartureList.tsx
```diff
@@ -55,7 +55,7 @@ const DepartureList = ({
   const routeAlerts = allAlertsForDirection(alertsForRoute, directionId);
   const stopAlerts = alertsByStop(alerts, stop.id);
   const allAlerts = concat(routeAlerts, stopAlerts).filter(alert => {
```
kotva006, line 57: nit-pick: Can this be renamed to describe that its a filter of the alerts?

kotva006 (changes requested):
Some tests and logic need to be addressed

apps/site/assets/ts/models/alert.ts
i0sea: I agree this function defaulting to 0 days when no xDays isn't intuitive, will leave `xDays` as required (although saying 0 days from now == currently happening is also odd). 

Will return `return start <= xDays && end >= currentDate` since the end of the alert could be before xDays but after the current date

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
i0sea: The first Alert in `alertsForStop` tests for this as it's start date is now

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
kotva006: This name needs to be changed, or an alert with the `suspension` status should be added to this test

apps/site/assets/ts/stop/__tests__/StopPageRedesignTest.tsx
kotva006: I am confused then because that is not what I am seeing.  The first alert in this tests uses `past1` and `future1` for its times, after having remove the use of `now`.
Screen Shot 2023-06-08 at 7 35 06 AM

i0sea: 0 days means a currently active alert. Having the active period be now(which it was originally) or in the past would mean it's currently active. Will switch back to now but am confused what the original comment was requesting.

apps/site/assets/ts/models/alert.ts
```diff
@@ -159,24 +159,30 @@ const isCurrentLifecycle = ({ lifecycle }: Alert): boolean =>
   lifecycle === "ongoing" ||
   lifecycle === "ongoing_upcoming";
 
-export const isCurrentAlert = (
+export const isInNextXDays = (
   alert: Alert,
+  days: number,
   currentDate: Date = new Date()
 ): boolean => {
+  const xDays = add(Date.now(), { days });
```
thecristen, line 167: question: Should this be based on the `currentDate` argument instead of using `Date.now()`?

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -48,24 +49,33 @@ const StopPageRedesign = ({
     return <Loading />;
   }
 
+  const isBannerAlert = ({ effect }: Alert): boolean =>
```
thecristen, line 52: A bit of a misleading name since it technically has nothing to do with the banner.

thecristen (commented):
Just a small question, otherwise is looking good.

apps/site/assets/ts/models/alert.ts
i0sea: They're interchangeable `currentDate` is Date object and `Date.now()` is the number representation

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
i0sea: True will change to `isStopPageAlert` but open to other suggestions

apps/site/assets/ts/models/alert.ts
thecristen: But if we call `isInNextXDays(alert, 3, new Date(2023, 10, 10))` it'll return alerts that are within three days from whenever I call the function, as opposed to within three days from the date I specified.

i0sea: @thecristen  Thinking currentDate doesn't need to be a parameter but a variable instead. We only ever want alerts from now(the day the function is called) into sometime in the future

thecristen: @i0sea Rather true for our use case; but having it as a parameter can be quite powerful! it makes a few things possible:

- it makes the function easier to test: we can write tests for a specific `currentDate` value rather than its internal `Date.now()` changing based on whatever day the test runs.
- it lets us display the page according to a specific date.
  - we don't really exercise this in the frontend much... but one example where we support it is in the timetable backend - if you visit a `/timetable?date=2023-06-30`, the controller passes that `date` into the functions that ultimately generate the data needed, and it'll show the timetable for that date in lieu of defaulting to today.

PR #1625, opened 2023-06-05
feat(DepartureList): format display times on the realtime tracking page

thecristen:
#### Summary of changes

**Asana Ticket:** [Departure List | Format Display Time](https://app.asana.com/0/385363666817452/1204591650723169/f)

I tried first reusing `<DepartureTimes />` then just `infoToDisplayTime`, which was almost okay, but I couldn't really figure out how to best adjust that function to accommodate the many differences in UI display logic needed for the realtime tracking page as opposed to the times displayed in the landing page. So this PR creates a new abstraction that I wonder could later be adapted elsewhere.

I basically have a wrapper component for each "row", `<DisplayTime />`, that establishes a new Context containing information on the departure that's conceptually similar to the output of the existing `infoToDisplayTime` function, with slightly different properties. Those properties are accessed by a variety of child components that are responsible for testing distinct smaller pieces of the UI in each row.

image

---

#### General checks
* [x] **Tech debt.** Have you checked for tech debt you can address in the area you're working in? This can be a good time to address small issues, or create Asana tickets for larger issues. (have checked but not addressed here - have spun out a few other tickets and PRs)

### DIFF
--- a/apps/site/assets/ts/helpers/date.ts
+++ b/apps/site/assets/ts/helpers/date.ts
@@ -88,9 +88,6 @@ export const formatRelativeTime = (
   targetDate: Date = new Date()
 ): string => {
   const seconds = differenceInSeconds(time, targetDate);
-  if (seconds <= 30) {
-    return "Arriving";
-  }
   if (seconds <= 60) {
     return "<1 minute away";
   }
--- a/apps/site/assets/ts/stop/__tests__/DepartureListTest.tsx
+++ b/apps/site/assets/ts/stop/__tests__/DepartureListTest.tsx
@@ -92,7 +92,7 @@ describe("DepartureList", () => {
     );
 
     expect(screen.queryAllByRole("listitem")).toHaveLength(schedules.length);
-    expect(screen.queryByText("Arriving")).toBeTruthy();
+    expect(screen.queryByText("<1 minute away")).toBeTruthy();
     expect(screen.queryByText("10 min")).toBeTruthy();
     expect(screen.queryByText("19 min")).toBeTruthy();
   });
--- a/apps/site/assets/ts/stop/components/DisplayTime.tsx
+++ b/apps/site/assets/ts/stop/components/DisplayTime.tsx
@@ -152,7 +152,7 @@ const DisplayTime = ({
       <div className="stop-routes__departures-time">
         {isDelayed ? <DelayedTimeCountdown /> : <TimeCountdown />}
       </div>
-      <div className="stop-routes__departures-details fs-12">
+      <div className="stop-routes__departures-details fs-14">
         {isDelayed && <DelayedTimeDetails />} <BaseTimeDetails />
       </div>
     </DepartureContext.Provider>
--- a/apps/site/assets/ts/helpers/__tests__/date-test.ts
+++ b/apps/site/assets/ts/helpers/__tests__/date-test.ts
@@ -32,9 +32,6 @@ describe("formatRelativeTime", () => {
   it("should format a time relative to another time to indicate arrival", () => {
     const baseDate = new Date("2020-02-28T10:00:00");
 
-    const arrivingDate = addSeconds(baseDate, 20);
-    expect(formatRelativeTime(arrivingDate, baseDate)).toBe("Arriving");
-
     const approachingDate = addSeconds(baseDate, 45);
     expect(formatRelativeTime(approachingDate, baseDate)).toBe(
       "<1 minute away"


apps/site/assets/ts/helpers/date.ts
```diff
@@ -77,4 +83,22 @@ export const formatToBostonTime = (
...
+export const formatRelativeTime = (
+  time: Date,
+  targetDate: Date = new Date()
+): string => {
+  const seconds = differenceInSeconds(time, targetDate);
+  if (seconds <= 30) {
+    return "Arriving";
+  }
+  if (seconds <= 60) {
+    return "Approaching";
```
i0sea, line 95: For schedules this should be `<1 minute away`

thecristen: Thanks, fixed.

apps/site/assets/ts/helpers/date.ts
```diff
@@ -77,4 +83,22 @@ export const formatToBostonTime = (
...
+export const formatRelativeTime = (
+  time: Date,
+  targetDate: Date = new Date()
+): string => {
+  const seconds = differenceInSeconds(time, targetDate);
+  if (seconds <= 30) {
+    return "Arriving";
+  }
+  if (seconds <= 60) {
+    return "<1 minute away";
```
i0sea, line 95: Adding clarity to previous comment, based on the ticket I think this needs a paramter to distinguish if this is a schedule or a prediction. If it's a schedule the text should be "<1 minute away" and if it's a prediction it should be "approaching"

i0sea (changes requested):

apps/site/assets/ts/stop/models/displayTimeConfig.ts
```diff
@@ -20,7 +20,7 @@ interface DisplayTimeConfig {
   trackName?: string | null;
   isTomorrow?: boolean;
   isBolded?: boolean;
-  isStikethrough?: boolean;
+  isStrikethrough?: boolean;
```
kotva006, line 23: good catch

apps/site/assets/ts/stop/__tests__/DepartureListTest.tsx
```diff
@@ -70,11 +70,15 @@ describe("DepartureList", () => {
       />
     );
 
-    expect(screen.getByText(schedules[2].time.toString())).toBeDefined();
+    expect(container.querySelector("time")).toBeDefined();
+    expect(container.querySelectorAll("time")[2]).toHaveAttribute(
+      "datetime",
+      schedules[2].time.toISOString()
+    );
```
kotva006, line 77: Is there a specific reason that `container` is used here to parse the dom instead of the `screen.get/queryBy()` functions?  I think the test would be better if it queried what the display string should be.

apps/site/assets/ts/stop/__tests__/DepartureListTest.tsx
```diff
@@ -83,7 +87,11 @@ describe("DepartureList", () => {
         alerts={[]}
       />
     );
-    expect(screen.getByText(predictionTime.toString())).toBeDefined();
+    expect(container.querySelector("time")).toBeDefined();
+    const datetimes = Array.from(container.querySelectorAll("time")).map(el =>
+      el.getAttribute("datetime")
+    );
+    expect(datetimes).toContain(predictionTime.toISOString());
```
kotva006, line 94: Same here.  `container` should only be used if all other avenues failed.  Such as checking if a specific color gets applied to a background.

apps/site/assets/ts/stop/components/DisplayTime.tsx
```diff
@@ -0,0 +1,162 @@
...
+  isCR: false,
+  isLessThanHourAway: false,
+  targetDate: new Date()
+});
+
+function TimeCountdown(): JSX.Element {
+  const { time, isCR, targetDate } = useContext(DepartureContext);
+  return (
+    <BasicTime
+      displayType={isCR ? "absolute" : "relative"}
```
kotva006, line 63: What is the purpose of relative?  We don't display timezones anywhere and I wonder if this could lead to confusion.  Also talking to Anna in the past we have favored displaying time in the local timezone of the MBTA.

thecristen: relative is stuff like `3 min`, `Arriving`, etc... as opposed to `4:23 PM`.

apps/site/assets/ts/stop/__tests__/DepartureListTest.tsx
thecristen: oh!!! yeah I first tried get/queryByRole and it didn't work; I don't think the library supports the `<time>` role yet. and in some cases, I didn't want to query on specific text displayed, because the text might be displayed differently based on many things, whereas the `datetime` attribute will be consistently formatted... so to me that ruled out the get/queryByText family of functions. should I add a `data-testid` for this case maybe?

apps/site/assets/ts/stop/components/DisplayTime.tsx
kotva006: Okay, I mis-understood what this purpose was.

apps/site/assets/ts/helpers/date.ts
thecristen: Thanks for raising -- after talking it over with Anna we've decided to keep both prediction and schedule as "<1 minute away" for now.

apps/site/assets/ts/stop/__tests__/DepartureListTest.tsx
thecristen: I just pushed a change taking a slightly different approach with this test to avoid needing to use `container`.

i0sea (approved):
lgtm, after resolving merge conflict 🚀

PR #1628, opened 2023-06-06
refactor(StopPageDepartures): filter schedules in backend

thecristen:
I saw the comment `// This filtering should be done on the backend` for the listing current schedules, so I went ahead and did that, and added a condition to remove schedules with a `last_stop?` value of `true`.

#### Summary of changes

**Asana Ticket:** [For terminal stops, don't list routes that end at that stop](https://app.asana.com/0/385363666817452/1204678909646879/f)

### Wonderland station:

| Before      | After |
| ----------- | ----------- |
|  dev mbtace com_stops_place-wondl   |  localhost_4001_stops_place-wondl (3)   |

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/hooks/useSchedules.ts
+++ b/apps/site/assets/ts/hooks/useSchedules.ts
@@ -37,7 +37,7 @@ const useSchedulesByStop = (
   stopId: string
 ): FetchState<ScheduleWithTimestamp[]> => {
   const { data, error } = useSWR<ScheduleData[]>(
-    `/api/stops/${stopId}/schedules`,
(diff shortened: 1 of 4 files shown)


kotva006:
Could these filters be put behind query params?  That way if the front end doesn't need these filters we don't need to make a new endpoint to get the schedules.

thecristen:
> Could these filters be put behind query params? That way if the front end doesn't need these filters we don't need to make a new endpoint to get the schedules.

That's a cool idea, I've added a commit to try this. For now `useSchedule` hard-codes the particular query parameters this use case needs, but if needed later we can modify it to add support for configuring those query params.

i0sea (approved):

PR #1629, opened 2023-06-06
refactor: add stop_id directly to container element

thecristen:
...and stop reading the stop ID from the window location.

The way our current loading worked, whatever URL params are on the page ended up getting captured in the stop ID, e.g. visiting `/stops/place-whatever?something=foo` would result in a `stopId` of `place-whatever?something=foo`, which caused multiple subsequent requests to fail.

This fixes that by passing in the stop ID via the HTML.

#### Summary of changes

**Asana Ticket:** [Stop Page | Load page when directed from search result](https://app.asana.com/0/385363666817452/1204659277085822/f)

### DIFF
--- a/apps/site/lib/site_web/templates/stop/show-redesign.html.eex
+++ b/apps/site/lib/site_web/templates/stop/show-redesign.html.eex
@@ -1,6 +1,6 @@
 <link rel="stylesheet" href="<%= static_url(@conn, "/css/map.css") %>" data-turbolinks-track="reload">
 <div>
-  <div id="react-stop-redesign-root"></div>
+  <div id="react-stop-redesign-root" data-stop-id="<%= @stop_id %>"></div>
 </div>
 
 <%= if Application.get_env(:site, :dev_server?) do %>
--- a/apps/site/lib/site_web/templates/stop/show-redesign.html.eex
+++ b/apps/site/lib/site_web/templates/stop/show-redesign.html.eex
@@ -1,6 +1,6 @@
 <link rel="stylesheet" href="<%= static_url(@conn, "/css/map.css") %>" data-turbolinks-track="reload">
 <div>
-  <div id="react-stop-redesign-root" data-stop-id="<%= @stop_id %>"></div>
+  <div id="react-stop-redesign-root" data-mbta-stop-id="<%= @stop_id %>"></div>
 </div>
 
 <%= if Application.get_env(:site, :dev_server?) do %>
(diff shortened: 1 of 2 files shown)


apps/site/lib/site_web/templates/stop/show-redesign.html.eex
```diff
@@ -1,6 +1,6 @@
 <link rel="stylesheet" href="<%= static_url(@conn, "/css/map.css") %>" data-turbolinks-track="reload">
 <div>
-  <div id="react-stop-redesign-root"></div>
+  <div id="react-stop-redesign-root" data-stop-id="<%= @stop_id %>"></div>
```
kotva006, line 3: question (non-blocking): Is this the way we want to embed data on the webpage?  In the root div?  Have we explored meta tags?  It looks like we already [embed the stop id](https://github.com/mbta/dotcom/blob/master/apps/site/lib/site_web/templates/layout/app.html.eex#L17) when available in a meta tag.
I ask these questions to try and get a standard way of doing things going forward.

thecristen: Ah yeah, forgot about that meta tag that we already have! :sparkles: Storing data in custom `data-*` attribute values is super common, custom meta tags less so from what I can tell.
* Meta tags contain metadata! And are generally used to describe the HTML document, usually used by external agents (interpretation by search engines, display on social media, among others). There are a set of known standardized tags but it's ok to make our own - but new tags can be added to the standard in the future (so there's a mild risk of naming conflict).
* Data attributes don't inherently mean anything, and are a very common way to embed arbitrary data for use by JavaScript libraries. For example, [Bootstrap's plugins use data attributes](https://getbootstrap.com/docs/5.3/getting-started/javascript/#data-attributes) as a means of configuration.

I'd say an advantage of using a meta tag is that it can be consumed by anything on the page. By that same token, an advantage of using a data attribute is that you're not limited to one usage per page.

One nice thing about data-attributes, when placed on the relevant HTML, it's a lot more explicit about what's going on, as opposed to reliant on an attribute that's defined elsewhere in the codebase.

If this were a reusable thing I'd say a custom data attribute makes more sense, but this happens to not be the case for `<StopsPageRedesign />` so  I think the meta tag value would be fine. I'm honestly not very convinced one way or the other in terms of a standard way going forward. Curious if @i0sea or @kotva006 have any additional opinions on the matter!

kotva006: My only hang up about the code reuse is that we should only ever have one react tree/root on any given page.  So root reuse shouldn't really be a thing.  Based off the links/description you shared I would lean more towards using the `data-` properties.  Since meta tags seem more for use by the browser/crawlers.

i0sea (approved):
Agree that data attribute would be best

PR #1634, opened 2023-06-12
Retrofit Departure Cards + List To Show Arrivals

thecristen:
#### Summary of changes

**Asana Ticket:** [Retrofit Departure Cards + List To Show Arrivals](https://app.asana.com/0/385363666817452/1204795517302694/f)

In our predictions and schedules, we have a field called `time` that was meant to abstract the idea of which of departure or arrival time to send to the frontend, but this became murky so we added `departure_time` and `arrival_time` back into the predictions and schedules so that it was easier to understand what we're displaying. But alas, now came time to implement logic to:

- For subway and bus, prefer arrival times
- For everything else, prefer departure times

So I replaced `Predictions.Parser.first_time/1` with `Schedules.Parser.display_time/3`, which implements this mode-specific logic with a slightly more legible name and comment. And put the result in the `time` field. Ideally, all front-end will just use this `time` value.

This affects schedules, predictions and predictions parsed from the stream.

---

#### General checks

### DIFF
--- a/apps/alerts/test/cache/store_test.exs
+++ b/apps/alerts/test/cache/store_test.exs
@@ -5,6 +5,11 @@ defmodule Alerts.Cache.StoreTest do
 
   @now Timex.parse!("2017-06-08T10:00:00-05:00", "{ISO:Extended}")
 
+  setup_all do
+    _ = start_supervised(Alerts.Cache.Store)
+    :ok
+  end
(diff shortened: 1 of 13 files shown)


kotva006 (approved):
Looks good to me 👍 
I like that we have all 3 times available.

PR #1639, opened 2023-06-16
Fetch Facilities

i0sea:
#### Summary of changes

**Asana Ticket:** [Fetch Facilities](https://app.asana.com/0/555089885850811/1204843838122708/f)

---

Fetch facilities by stop Id to be passed in to amenity cards. For testing purposes log `facilitiesByType` in `StationInformation.tsx` to view result

#### General checks


apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -480,4 +480,12 @@ defmodule SiteWeb.StopController do
 
     %{conn | assigns: assigns}
   end
+
+  def get_facilities(conn, %{"stop_id" => stop_id}) do
+    res =
+      [{"stop", stop_id}]
+      |> V3Api.Facilities.filter_by()
+
+    json(conn, res.data)
```
thecristen, line 489: **suggestion:** because `V3Api.Facilities.filter_by/1` can potentially return `{:error, any}`, we should check for that before assuming there's data to return.

Also please add unit tests for this function. :)

thecristen (changes requested):

apps/site/assets/ts/stop/components/StationInformation.tsx
```diff
@@ -12,12 +12,26 @@ import FareSalesAmenityCard from "./amenities/FareSalesAmenityCard";
...
+
+  facilities.forEach((facility: Facility) => {
+    const { type } = facility.attributes;
+    if (facilitiesByType.has(type)) {
+      facilitiesByType.get(type)?.push(facility);
+    } else {
+      facilitiesByType.set(type, [facility]);
+    }
+  });
+
```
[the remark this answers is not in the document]
i0sea: Do you mean `useFacilitiesByStop` returns this map or takes another parameter of facility type? I thought about the latter but didn't want to have to make.a call for each type. Leaving as is for now just curious.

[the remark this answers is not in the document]
thecristen: I was thinking returning this map, and agree with not wanting to make a call for each type. This is fine for now.

apps/site/test/site_web/controllers/stop_controller_test.exs
```diff
@@ -143,4 +144,58 @@ defmodule SiteWeb.StopControllerTest do
       end
     end
   end
+
+  describe "get_facilities/1" do
+    setup_with_mocks([
+      {SiteWeb.StopController, [:passthrough],
```
thecristen, line 150: **suggestion:** instead of mocking the function you're testing (which makes the unit test rather superficial), we should mock the underlying functions that the tested function depends on -- in this case, the `V3Api.Facilities.filter_by([{"stop", stop_id}])` call.

apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -480,4 +480,18 @@ defmodule SiteWeb.StopController do
...
   end
+
+  def get_facilities(conn, %{"stop_id" => stop_id}) do
+    res =
+      [{"stop", stop_id}]
+      |> V3Api.Facilities.filter_by()
+
+    case res do
+      {:error, _x} ->
+        json(conn, res)
```
thecristen, line 491: **suggestion:** Elixir Tuples are not JSON-serializable, so this will break. And are you sure you want to send the error to the frontend? Could return a reasonable fallback value instead.

thecristen (changes requested):

apps/site/lib/site_web/controllers/stop_controller.ex
i0sea: The pattern I've noticed is sending the response whether it's an error or not and handling in the frontend

kotva006: I added an [encoder](https://github.com/mbta/dotcom/blob/master/apps/site/lib/site_web/encoding/tuple_encoder.ex) a few months back that should handle tuples.  That being said I think the ideal case (for production) would be just to return a 500 status code.  Then in dev it could return a status code and an error describing what happened

apps/site/assets/ts/hooks/useStop.ts
```diff
@@ -1,9 +1,11 @@
 import useSWR from "swr";
 import { fetchJsonOrThrow } from "../helpers/fetch-json";
-import { Stop } from "../__v3api";
+import { Facility, Stop } from "../__v3api";
 import { FetchState, FetchStatus } from "../helpers/use-fetch";
 
 const fetchData = async (url: string): Promise<Stop> => fetchJsonOrThrow(url);
+const fetchFacilityData = async (url: string): Promise<Facility[]> =>
+  fetchJsonOrThrow(url);
```
kotva006, line 8: nit-pick: Since these functions do the same thing we could try generics to combine them into one:
`const fetchData = async <T>(url: string): Promise<T> => fetchJsonOrThrow(url);`

apps/site/assets/ts/hooks/useStop.ts
```diff
@@ -13,4 +15,15 @@ const useStop = (stopId: string): FetchState<Stop> => {
   return { status: FetchStatus.Data, data };
 };
 
-export default useStop;
+const useFacilitiesByStop = (stopId: string): FetchState<Facility[]> => {
+  const { data, error } = useSWR<Facility[]>(
+    `/api/facilities/${stopId}`,
```
kotva006, line 20: nit-pick:  I think the url should contain the word stop, since that is the id we are using to fetch the information, and the controller this function is in.  Without this context the current url would look like `/api/facilities/342` which would make me think we are grabbing a specific facility.
This end point should be something more like: `/api/stop/{stopId}/facilities`
That seems more to indicate this endpoint returns all the facilities for a given stop.

apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -27,6 +27,7 @@ const StopPageRedesign = ({
   const routesWithPolylinesResult = useRoutesByStop(stopId);
   const schedulesResult = useSchedulesByStop(stopId);
   const alertsForStopResult = useAlertsByStop(stopId);
+  const facilities = useFacilitiesByStop(stopId);
```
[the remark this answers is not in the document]
kotva006: Id say it depends if this data is need for the loading/error state check.  That is all handled in this component.

thecristen (approved):
I think there are a few worthy improvements mentioned in the comments, but this is good to go with and we can advance it later.

PR #1640, opened 2023-06-16
fix(ServiceOptGroup): improve (now) logic

thecristen:
Because of the way the backend logic works, if there's only one service available this rating, that service might get labelled as "today's", even if it's not within the valid service dates. To improve this, 

- actually check whether the service is in the current service before labelling with "(now)"
- if it's not in the current service, adjust the text to indicate start date

#### Summary of changes

**Asana Ticket:** [Schedule finder should not present future services as current](https://app.asana.com/0/385363666817452/1204794505976013/f)

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/ServiceOptGroup.tsx
+++ b/apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/ServiceOptGroup.tsx
@@ -4,7 +4,8 @@ import { shortDate, stringToDateObject } from "../../../../helpers/date";
 import {
   ServiceGroupNames,
   serviceDays,
-  dedupeServices
+  dedupeServices,
+  isInCurrentService
 } from "../../../../helpers/service";
 
 interface Props {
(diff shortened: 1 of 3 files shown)


apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/ServiceOptGroup.tsx
```diff
@@ -67,7 +68,19 @@ const ServiceOptGroup = ({
         }
 
         if (service.id === todayServiceId) {
-          optionText += " (now)";
+          // if it's the only service this rating, this service might not
+          // technically be now - adjust text for that
+          const now = Date.now();
+          if (isInCurrentService(service, new Date(now))) {
```
kotva006, line 74: nit-pick: Since I don't see now being used anywhere else this could just be `new Date()`

thecristen: true, though it made it easier to mock the date in Jest, via 
```javascript
jest
      .spyOn(Date, "now")
      .mockImplementation(() => new Date("2019-09-07").getTime());
```

do you know if there's a straightforward way to mock `new Date()`?

apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/ServiceOptGroup.tsx
```diff
@@ -67,7 +68,19 @@ const ServiceOptGroup = ({
...
+            optionText += " (now)";
+          } else {
+            optionText += ` (Starting ${startDate.toLocaleDateString("en-US", {
```
kotva006, line 77: nit-pick(non-blocking): I think we should use a single library for formatting time to keep things consistent (`date-fns`)

apps/site/assets/ts/schedule/components/schedule-finder/daily-schedule/ServiceOptGroup.tsx
kotva006: Oh, I see your conundrum.  What we have been doing else where is make it an optional prop/param that defaults to `new Date()`.  Then you can pass the desired date into the prop for testing.

kotva006 (approved):
Looks good!

PR #1642, opened 2023-06-20
Predictions Refactor | Flexible Subscription IDs

thecristen:
#### Summary of changes

**Asana Ticket:** [Predictions Refactor | Flexible Subscription IDs](https://app.asana.com/0/1204716173461809/1204668141429887/f)

The goal of this PR was to make the backend changes needed to enable predictions streaming for more than a route-direction-stop triad. In total,

- Instead of managing lists of predictions in a `Map` keyed by `route:stop:direction`, predictions are put into an ETS table. The single ETS table is created on startup and each new subscription starts a new streaming connection to the API (same as before), each connection having the ability to remove/update/add entries to the table. Each entry is a tuple of `{prediction ID, route ID, stop ID, direction ID, trip ID, vehicle ID, prediction (the whole object)}` - making it easy to fetch predictions by any and all of those keys.
- The frontend update was just to make this change non-breaking. This will have to be followed up by a PR that makes `usePredictionsChannel` take more flexible arguments, so that we can, say, subscribe to all predictions for a stop.
- I don't love the channel naming scheme I came up with to enable this. Here the format is `predictions:<key>=<value>:<key2>=<value2>` and so on. The other idea I have is to just make it more constrained: `predictions:stop:<stopid>`, `predictions:route:<routeid>-<directionid>`, `predictions:trip:<tripid>` and so forth.

---

#### General checks

### DIFF
--- a/apps/predictions/lib/predictions_pub_sub.ex
+++ b/apps/predictions/lib/predictions_pub_sub.ex
@@ -1,26 +1,22 @@
 defmodule Predictions.PredictionsPubSub do
   @moduledoc """
-  Allow channels to subscribe to prediction streams, which are collected into a
-  Map with keys representing route-stop-direction triads as
-  "<route-id>:<parent-stop-id>:<direction-id>"
+  Allow channels to subscribe to prediction streams, which are collected into an
+  ETS table keyed by prediction ID, route ID, stop ID, direction ID, trip ID,
+  and vehicle ID for easy retrieval.
   """
 
   use GenServer
 
   alias Predictions.{Prediction, StreamSupervisor}
   alias Routes.Route
+  alias Schedules.Trip
   alias Stops.Stop
 
-  defstruct predictions_by_key: %{}
-
-  @type t :: %__MODULE__{
-          predictions_by_key: predictions_by_key()
-        }
+  @valid_filters ["route", "stop", "direction", "trip"]
 
   @type direction_id_t :: 0 | 1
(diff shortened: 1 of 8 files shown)


apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -229,11 +123,15 @@ defmodule Predictions.PredictionsPubSub do
...
-    new_predictions = predictions_for_key(predictions_by_key, route_stop_direction)
+  @spec send_data({pid, String.t()}, state) :: broadcast_message()
+  defp send_data({pid, keys}, %{ets: table}) do
+    new_predictions = predictions_for_keys(table, keys)
     send(pid, {:new_predictions, new_predictions})
   end
+
+  def to_record(prediction) do
+    {prediction.id, prediction.route.id, prediction.stop.id,
+     if(is_integer(prediction.direction_id), do: Integer.to_string(prediction.direction_id)),
```
kotva006, line 134: question: does this if need an else?  In the off chance that the `direction_id` is not an integer?

thecristen: These are coming directly from our parsing of the API, so they definitely should be integers by this point. FWIW if there's no `else` defined, the `if` statement in that case won't error out but will default to `nil` (which will work fine for inserting into the ETS table).

kotva006 (approved):

PR #1645, opened 2023-06-27
feat(AmenityCard): add badge prop, improve styles

thecristen:
#### Summary of changes

No ticket, just a few improvements on the original `AmenityCard` component we'll all use:

* Some CSS and layout improvements, mostly around the card's heading. This allows the badge to be on the same line as the header text, and wrap nicely without interfering with the width of the right arrow.
* Adds disabled state
  * Grey background and black header text and black icon
  * No click enabled / hides arrow
  * Adjusts badge to support additional CSS class - e.g. you can do `<Badge bgClass="u-bg--gray-lighter" />` to replace the default red background with a grey one
* Added an `<AmenityLink />`, there isn't much to it but just standardizing our rendering of links
* Improves the typing used for children components

### DIFF
--- a/apps/site/assets/ts/components/Badge.tsx
+++ b/apps/site/assets/ts/components/Badge.tsx
@@ -2,14 +2,17 @@ import React from "react";
 
 const Badge = ({
   text,
-  contextText
+  contextText,
+  bgClass
 }: {
   text: string;
   contextText?: string;
+  bgClass?: string;
 }): JSX.Element => {
   return (
     <div
-      className="u-error-background font-weight-bold ps-8 pe-8 fs-14"
+      className={`${bgClass ||
+        "u-error-background"} font-weight-bold ps-8 pe-8 fs-14`}
       style={{ borderRadius: "0.75rem" }}
     >
       {/* The purpose of this block is to have invisble text for screen readers */}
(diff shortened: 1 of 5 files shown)


apps/site/assets/ts/components/Badge.tsx
```diff
@@ -2,14 +2,17 @@ import React from "react";
 
 const Badge = ({
   text,
-  contextText
+  contextText,
+  bgClass
 }: {
   text: string;
   contextText?: string;
+  bgClass?: string;
```
i0sea, line 10: is this just to change background color? Assuming any class can be passed here

apps/site/assets/ts/stop/components/amenities/AmenityCard.tsx
```diff
@@ -29,16 +29,32 @@ export const AmenityModal = ({
   );
 };
 
+export const AmenityLink = ({
```
i0sea, line 32: Like the standardization of links. Would be nice to have for use everywhere

i0sea (approved):
lgtm

apps/site/assets/ts/components/Badge.tsx
thecristen: yeah technically any class can be passed. but the way it's used in the `Badge`, it will replace/remove the default background color.

PR #1650, opened 2023-06-27
feat(StopPageDepartures): fetch predictions at top level

thecristen:
#### Summary of changes

No ticket. It moves our call to predictions higher up in the tree, making one websocket connection for all the predictions, instead of one per route/direction.

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/hooks/__tests__/usePredictionsChannelTest.tsx
+++ b/apps/site/assets/ts/hooks/__tests__/usePredictionsChannelTest.tsx
@@ -1,7 +1,6 @@
 import { act, renderHook } from "@testing-library/react-hooks";
 import usePredictionsChannel, {
   StreamPrediction,
-  groupByHeadsigns,
   parsePrediction
 } from "../usePredictionsChannel";
 import {
@@ -9,7 +8,6 @@ import {
   makeMockSocket
 } from "../../helpers/socketTestHelpers";
 import { Route, Stop, Trip } from "../../__v3api";
-import { PredictionWithTimestamp } from "../../models/perdictions";
 
 const predictionsFromStream = [
   {
@@ -62,7 +60,7 @@ describe("usePredictionsChannel hook", () => {
         directionId: 0
       })
     );
-    expect(result.current).toEqual({});
+    expect(result.current).toEqual([]);
     expect(
       window.channels[
         "predictions:route=Purple:stop=place-somewhere:direction_id=0"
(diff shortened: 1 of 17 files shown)


apps/site/assets/ts/stop/components/StopPageDepartures.tsx
```diff
@@ -14,7 +17,7 @@ interface StopPageDeparturesProps {
   onClick: (
     route: Route,
     directionId: DirectionId,
-    departures: ScheduleWithTimestamp[] | null | undefined
+    departures: DepartureInfo[] | null | undefined
```
kotva006, line 20: So the one downside to this approach is that the predictions will not update on the realtime list.  The value of this property `departures` is set when the user clicks on a row.  When the `usePredictionsChannel` updates with new predictions it never gets a chance to pass them to the realtime list.  The user is on a different component, and this predictions values are never passed to the list directly.

apps/site/assets/ts/hooks/__tests__/usePredictionsChannelTest.tsx
```diff
@@ -68,7 +70,11 @@ describe("usePredictionsChannel hook", () => {
 
   test("gets and outputs data", () => {
     const { result } = renderHook(() =>
-      usePredictionsChannel("Purple", "place-somewhere", 0)
+      usePredictionsChannel({
```
i0sea, line 73: nitpick: can a test be added that only uses `stopId`?

apps/site/assets/ts/stop/__tests__/DepartureListTest.tsx
```diff
@@ -42,40 +42,26 @@ const schedules = [
...
+  }
+] as PredictionWithTimestamp[];
 
-  afterAll(() => {
-    jest.restoreAllMocks();
-  });
+const departures = mergeIntoDepartureInfo(schedules, predictions);
 
+describe("DepartureList", () => {
   it("should render a schedule when no predictions available", () => {
```
i0sea, line 59: add test for rendering prediction if there are no schedules

i0sea (changes requested):

apps/site/assets/ts/hooks/usePredictionsChannel.ts
```diff
@@ -66,45 +65,46 @@ export const parsePrediction = (
...
-  );
-};
+interface PredictionsChannelArgs {
+  routeId?: string;
+  stopId?: string;
+  directionId?: 0 | 1;
+}
+
+function channelFromArgs(channelArgs: PredictionsChannelArgs): string {
+  const keysWithValues = Object.entries({
```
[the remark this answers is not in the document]
thecristen: It does! Nice catch, using an empty string is probably suboptimal... will update to not try to join channels..

i0sea (approved):
🚀

PR #1654, opened 2023-06-30
feat(FareSalesAmenityCard): add info and modal (backend)

thecristen:
#### Summary of changes

The corresponding front-end work is in #1658 

**Asana Ticket:** [Amenity Modal | Fare Sales](https://app.asana.com/0/385363666817452/1204415037587451/f)
(There was no separate Asana ticket for the card content, but that work is in this PR too).

Inadvertently covers:
**Asana Ticket:** [Create list of all fare classes as shown in API](https://app.asana.com/0/385363666817452/1204827964810515/f)

* Adds support for parsing a route's `fare_class`. I deduced the valid values for this field by finding [the source code in `gtfs_creator`](https://github.com/mbta/gtfs_creator/blob/e031d3618840c4f33f62c7c4541140b621efa547/gtfs_creator2/validate.py#L504) 🙃 
  * Adds support for this field in the front-end too, but makes it optional so that I don't have to change 40+ TypeScript files... looks like that's we did for the route's `color` too.
* Adds a function to `Fares.Repo` to make it easier to retrieve the relevant `%Fares.Fare{}` objects for a specific `fare_class` value.
* In our Route controller's `get_by_stop_id/2` function, I added the route-derived fares information to the response, so that I could show those values in the front-end table.
  * Instead of using the same routes shown in the departures list, I also include routes at connecting stops. This produced better results at larger multi-modal stations, e.g. at Haymarket station we'll see local bus and express bus fares that technically apply to the adjacent bus stop; at Aquarium station we'll see the ferry fare at the adjacent terminal, and so on.
image

  * I have some logic masking the effect of routes with free or special fares. This produced better results because we don't have prices for special fares anyway, and for the moment don't want to show "Free fare $0.00" in our table. For these cases, we inspect the route and try to determine its appropriate "typical" fare based on mode. (The stop below is only served by the currently-free 23 bus):
image

---

#### General checks

### DIFF
--- a/apps/fares/lib/repo.ex
+++ b/apps/fares/lib/repo.ex
@@ -38,4 +38,24 @@ defmodule Fares.Repo do
       (!media || media in fare.media) && match?(^opts, Map.take(fare, Map.keys(opts)))
     end)
   end
+
+  @spec for_fare_class(Routes.Route.gtfs_fare_class(), Keyword.t()) :: [Fare.t()]
+  def for_fare_class(fare_class, opts \\ []) do
+    opts = Keyword.merge(opts, fare_class_opts(fare_class))
+    all(opts)
+  end
+
+  def fare_class_opts(:free_fare), do: [name: :free_fare]
+
+  def fare_class_opts(:express_bus_fare), do: [name: :express_bus]
+
+  def fare_class_opts(:local_bus_fare), do: [name: :local_bus]
+
+  def fare_class_opts(:rapid_transit_fare), do: [mode: :subway]
+
+  def fare_class_opts(:ferry_fare), do: [mode: :ferry]
+
+  def fare_class_opts(:commuter_rail_fare), do: [mode: :commuter_rail]
+
+  def fare_class_opts(_), do: []
 end
--- a/apps/site/lib/site_web/controllers/route_controller.ex
+++ b/apps/site/lib/site_web/controllers/route_controller.ex
@@ -6,18 +6,42 @@ defmodule SiteWeb.RouteController do
   alias Leaflet.MapData.Polyline
   alias Routes.{Repo, Route}
 
+  @display_fare_classes [
+    :local_bus_fare,
+    :express_bus_fare,
+    :rapid_transit_fare,
+    :commuter_rail_fare,
+    :ferry_fare
+  ]
+
   @spec get_by_stop_id(Plug.Conn.t(), map) :: Plug.Conn.t()
   def get_by_stop_id(conn, %{"stop_id" => stop_id} = _params) do
+    routes = stop_id |> Repo.by_stop()
+
     routesWithPolylines =
-      stop_id
-      |> Repo.by_stop()
+      routes
       |> Enum.map(fn route ->
         route
         |> Route.to_json_safe()
         |> Map.put(:polylines, route_polylines(route, stop_id))
       end)
 
-    json(conn, routesWithPolylines)
+    fares =
+      stop_id
+      # Get fare_class for all routes at this stop and at connecting stops
+      |> Repo.by_stop(include: "stop.connecting_stops")
+      |> Enum.map(&display_fare_class/1)
+      |> Enum.uniq()
+      # Sort in same order as @display_fare_classes
+      |> Enum.sort_by(&Enum.find_index(@display_fare_classes, fn fc -> fc == &1 end))
+      |> Enum.map(&Fares.Repo.for_fare_class(&1, duration: :single_trip, reduced: nil))
+      |> Enum.map(fn fares_for_class ->
+        price = price_range(fares_for_class)
+        name = List.first(fares_for_class) |> fare_name()
+        {name, price}
+      end)
+
+    json(conn, %{routes: routesWithPolylines, fares: fares})
   end
 
   defp route_polylines(route, stop_id) do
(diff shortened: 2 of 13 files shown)


apps/fares/lib/repo.ex
```diff
@@ -38,4 +38,24 @@ defmodule Fares.Repo do
...
+
+  def fare_class_opts(:local_bus_fare), do: [name: :local_bus]
+
+  def fare_class_opts(:rapid_transit_fare), do: [mode: :subway]
+
+  def fare_class_opts(:ferry_fare), do: [mode: :ferry]
+
+  def fare_class_opts(:commuter_rail_fare), do: [mode: :commuter_rail]
+
+  def fare_class_opts(_), do: []
```
kotva006, line 60: question(non-blocking): This seems like a hack to map the existing mode fares map (`fare_info`) to the new `fare_class` property.  Is that right?  Is there a plan to move this towards reworking the `fare_info` map to use `fare_class` instead of mode/name/zone filters?

apps/site/assets/package.json
```diff
@@ -49,9 +49,9 @@
   },
   "devDependencies": {
     "@babel/core": "^7.17.4",
-    "@babel/preset-env": "^7.16.11",
+    "@babel/preset-env": "^7.22.5",
     "@babel/register": "^7.17.0",
-    "@faker-js/faker": "^7.6.0",
+    "@faker-js/faker": "^8.0.2",
```
kotva006, line 54: nit-pick: These updates should be their own pull request

apps/site/lib/site_web/controllers/route_controller.ex
```diff
@@ -6,18 +6,42 @@ defmodule SiteWeb.RouteController do
...
+      # Sort in same order as @display_fare_classes
+      |> Enum.sort_by(&Enum.find_index(@display_fare_classes, fn fc -> fc == &1 end))
+      |> Enum.map(&Fares.Repo.for_fare_class(&1, duration: :single_trip, reduced: nil))
+      |> Enum.map(fn fares_for_class ->
+        price = price_range(fares_for_class)
+        name = List.first(fares_for_class) |> fare_name()
+        {name, price}
+      end)
+
+    json(conn, %{routes: routesWithPolylines, fares: fares})
```
kotva006, line 44: This is a pattern I am not a huge fan of.  We keep appending new data to existing models instead of making it a new endpoint.  In this case I think it would benefit both flexibility and reuse for making fares its own endpoint to fetch all the fares by stop id, instead of modifying the routes endpoint to return an array of routes and a separate array of fares

apps/site/lib/site_web/controllers/route_controller.ex
```diff
@@ -6,18 +6,42 @@ defmodule SiteWeb.RouteController do
...
         |> Route.to_json_safe()
         |> Map.put(:polylines, route_polylines(route, stop_id))
       end)
 
-    json(conn, routesWithPolylines)
+    fares =
+      stop_id
+      # Get fare_class for all routes at this stop and at connecting stops
+      |> Repo.by_stop(include: "stop.connecting_stops")
+      |> Enum.map(&display_fare_class/1)
```
kotva006, line 33: This function confuses me.  Don't all the routes already have a `fare_class` property?  Why does the mode need to be converted into a fare class atom?  Also doesn't this defeat the purpose of using the `fare_class` property on a route in the first place?

apps/site/lib/site_web/controllers/route_controller.ex
```diff
@@ -26,4 +50,57 @@ defmodule SiteWeb.RouteController do
...
+  end
+
+  defp fare_name(%Fares.Fare{mode: :ferry}) do
+    "Ferry one-way"
+  end
+
+  defp fare_name(fare) do
+    Fares.Format.name(fare)
+    |> String.capitalize()
+    |> String.replace_suffix("", " one-way")
```
kotva006, line 104: I feel like these name functions should also be in the `Fares.Format` module.  Trying to keep everything in one place

apps/site/lib/site_web/controllers/route_controller.ex
thecristen: Maybe I could name it something better; the intent was to translate the actual fare class into the one we ultimately wanted to display on the page. See comment on `display_fare_class/1` function.
It's basically compensating for the fact that we don't have any thought out logic or design for displaying special fares or free fares right now - if we remedy this later on, we could deprecate this function.
Hope that makes some sense.

apps/site/lib/site_web/controllers/route_controller.ex
thecristen: Have reverted this and made an endpoint for it.

thecristen:
@kotva006 Have broken things out into another PR and made requested changes here, thanks for bearing with me!

apps/fares/lib/repo.ex
thecristen: Yeah, you've got it exactly right. We can definitely use `fare_class` to improve our `fare_info` maps, but I figured that could wait, especially if we're gonna eventually incorporate fares info from the API down the line (whenever that happens).

apps/site/assets/ts/hooks/useRoute.ts
```diff
@@ -4,13 +4,18 @@ import { Route } from "../__v3api";
 import { Polyline } from "../leaflet/components/__mapdata";
 import { FetchState, FetchStatus } from "../helpers/use-fetch";
 
+export type RouteFarePrices = [string, string][];
 export type RouteWithPolylines = Route & { polylines: Polyline[] };
+export type RoutesAndFares = {
+  routes: RouteWithPolylines[];
+  fares: RouteFarePrices;
+};
```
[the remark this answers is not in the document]
thecristen: Ended up reverting this anyway.

apps/site/lib/site_web/router.ex
```diff
@@ -217,6 +217,7 @@ defmodule SiteWeb.Router do
     get("/stop/:id", StopController, :get)
     get("/map-config", MapConfigController, :get)
     get("/routes/by-stop/:stop_id", RouteController, :get_by_stop_id)
+    get("/fares/one-way/by-stop/:stop_id", FareController, :one_way_by_stop_id)
```
kotva006, line 220: not-pick (non-blocking):  I think it would make sense for this endpoint to be `/stop/:stop_id/fares/on-way` instead of using the `by-stop` path value.  An equally REST-ful design would be something like `/fares/one-way?stop_id=123`.  They'd both return the same data just 2 different approaches (that could coexist)

kotva006 (approved):
This looks good overall.  I am curious about the impact that changing fare clase into fare info queries will affect the whole site.  Especially the Blue Line.  I think this will make it show $0.00 for the Blue line.

apps/site/lib/site_web/router.ex
thecristen: Oh I like that, lemme see what I can do

PR #1658, opened 2023-07-07
feat(FareSalesAmenityCard): add info and modal (frontend)

thecristen:
#### Summary of changes

The front-end piece of #1654 

**Asana Ticket:** [Amenity Modal | Fare Sales](https://app.asana.com/0/385363666817452/1204415037587451/f)

- Some refactoring to account for the changed data structure now including fares
- Made a helper method for scaffolding a Stop for our tests.
- Added the modal content and amenity card text. Shows either information about fare vending machines or a link to view retail sales locations.
- Used styling from `.responsive-table` as a reference for the CSS (but couldn't copy it outright... too many differences)

### DIFF
--- a/apps/site/assets/ts/hooks/useFetch.ts
+++ b/apps/site/assets/ts/hooks/useFetch.ts
@@ -1,0 +1,25 @@
+import useSWR from "swr";
+import { fetchJsonOrThrow } from "../helpers/fetch-json";
+import { FetchState, FetchStatus } from "../helpers/use-fetch";
+
+/**
+ * Generic helper function for fetching JSON from a URL. Specify the expected
+ * return type as the type parameter, e.g
+ * `useFetch<Schedule[]>("/api/schedules/or/something")`
+ * @param {string} backendUrl - The path, usually starting with /api/, that our
+ * backend will respond with data from.
+ * @returns {{ status: FetchStatus, data: T }} An object containing the fetch
+ * status, and the returned data.
+ */
+const useFetch = <T>(backendPath: string): FetchState<T> => {
+  const { data, error } = useSWR<T>(
+    backendPath,
+    async (url: string): Promise<T> => fetchJsonOrThrow(url)
+  );
+  if (error) {
+    return { status: FetchStatus.Error };
+  }
+  return { status: FetchStatus.Data, data };
+};
+
+export default useFetch;
--- a/apps/site/assets/ts/stop/__tests__/helpers.ts
+++ b/apps/site/assets/ts/stop/__tests__/helpers.ts
@@ -1,7 +1,15 @@
 import { faker } from "@faker-js/faker";
 import { uniqueId } from "lodash";
 import { Polyline } from "../../leaflet/components/__mapdata";
-import { Route, RouteType } from "../../__v3api";
+import {
+  AccessibilityType,
+  BikeStorageType,
+  FareFacilityType,
+  ParkingLot,
+  Route,
+  RouteType,
+  Stop
+} from "../../__v3api";
 import { RouteWithPolylines } from "../../hooks/useRoute";
 
 export const newLatOrLon = (): number => +faker.string.numeric(2);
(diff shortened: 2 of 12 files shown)


thecristen:
This is finally cleaned up and synced with the changes in #1654 for review.

thecristen:
Just added a commit for:

**Asana Ticket:** [Amenity Modal | Fares Content Changes](https://app.asana.com/0/385363666817452/1205040071698674/f)

apps/site/assets/ts/hooks/useFetch.ts
```diff
@@ -0,0 +1,25 @@
...
+const useFetch = <T>(backendPath: string): FetchState<T> => {
+  const { data, error } = useSWR<T>(
+    backendPath,
+    async (url: string): Promise<T> => fetchJsonOrThrow(url)
+  );
+  if (error) {
+    return { status: FetchStatus.Error };
+  }
+  return { status: FetchStatus.Data, data };
+};
```
kotva006, line 23: I like the idea of doing this.  I feel like this pull request is the wrong place for it.
It would make sense to do this update as its own pull request, and update the usage across all of our web requests.  At least for the stop page redesign.  We could also discuss things about error state handling.  Something like this would allow a global way handle errors (show a toast or banner) or loading (display the dots) so the components don't need to worry about it.

apps/site/assets/ts/stop/__tests__/helpers.ts
```diff
@@ -36,3 +44,25 @@ export const routeWithPolylines = (
     polylines
   };
 };
+
+const defaultStop: Stop = {
```
kotva006, line 48: 👍 Great idea

apps/site/assets/ts/stop/components/amenities/FareSalesAmenityCard.tsx
```diff
@@ -1,13 +1,120 @@
...
+const renderFareTable = ({
+  status,
+  data
+}: FetchState<[string, string][]>): React.ReactElement | null => {
+  switch (status) {
+    case FetchStatus.Loading:
+      return <Loading />;
+    default:
+      if (data && data.length > 0) return fareTable(data);
+      return null;
```
kotva006, line 84: I feel like there should be a error state here instead of null, if we go this route

kotva006: As in there should be a state for success, and then the default is the error state

apps/site/assets/ts/hooks/useFetch.ts
thecristen: I'm happy to follow up with a separate PR migrating all the fetches in the stop page redesign to this, am less happy about removing this and having to scaffold files and tests for a useOneWayFares hook... 😂

apps/site/assets/ts/stop/components/amenities/FareSalesAmenityCard.tsx
thecristen: After consulting with Anna on this, we've decided that hiding the table and heading on error would be best.
So, this `default: return null` is in fact the default error state. :grin:

apps/site/assets/ts/hooks/useFetch.ts
kotva006: We should add a follow up ticket to try and have this handle the error and loading states so each page/component doesn't have to handle it separately and help towards a more consistent user experience

thecristen: @kotva006 About to write ticket, have WIP migration over in #1667 - lots to think about!

thecristen: I ended up separating this out into this PR, which also migrates the stop page redesign into using it - #1667

apps/site/assets/ts/stop/components/amenities/FareSalesAmenityCard.tsx
```diff
@@ -1,13 +1,120 @@
...
+
+const FareVendingMachines = (): JSX.Element => (
+  <>
+    <h2 className="h3 mb-8">Fare Vending Machines</h2>
+    <img
+      className="img-fluid mt-1 mb-1"
+      src="https://cdn.mbta.com/sites/default/files/styles/max_2600x2600/public/media/2021-11/Fare-Transformation-new-fvms-Nov2021.jpg"
+      alt="MBTA fare vending machines"
+      style={{ maxHeight: "250px" }}
+    />
```
kotva006, line 33: nit-pick (non-blocking): This could be converted to a reusable icon.

kotva006 (approved):

PR #1660, opened 2023-07-07
Switch AWS Location Service provider to Esri

EmmaSimon:
#### Summary of changes

**Asana Ticket:** [Switch geocoding search from HERE to Esri](https://app.asana.com/0/385363666817452/1204974196349553/f)

AWS Location Services requests can use two different geodata sources. We were previously using a mix of the two for different purposes, but one of the sources, HERE, was giving incorrect results in many different cases (many based around incorrect zip codes). This switches all searches to instead use the other source, Esri. If we start seeing more problems, we can switch back, but for now, Esri looks more reliable.

---

#### General checks

### DIFF
--- a/apps/location_service/lib/aws_location/request.ex
+++ b/apps/location_service/lib/aws_location/request.ex
@@ -56,9 +56,9 @@ defmodule AWSLocation.Request do
 
   defp place_index_path(:text), do: place_index_base("esri") <> "/search/text"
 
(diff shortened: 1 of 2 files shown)


thecristen (approved):
I'm glad I already got through https://github.com/mbta/devops/pull/1026 so we won't run into permissions issues when we deploy this! Let's give it a try!

PR #1666, opened 2023-07-12
feat(useSchedules): update time, filter schedules

thecristen:
This PR is based on the changes from #1650, which is needed for this PR to update the displayed schedules as expected.

#### Summary of changes

**Asana Ticket:** [Departure data refresh investigation](https://app.asana.com/0/385363666817452/1204573540717218/f)

### Highlights

* The `useSchedules` hook now has an internal state, that updates a timestamp every 15 seconds. The timestamp is used to re-filter the list of schedules to ensure no stale schedules are returned.
  * I learned a lot about how timers in Jest work while trying to test this
* I adjusted the schedules endpoint to set a cache header with a `max_age` of an hour. This means the client will get a refreshed version after an hour has elapsed, and prevents the client from fetching the same ~1.35MB of schedules over and over whenever the component state updates.
  * I felt like caching for an hour would be better than for a day (which could cause clients to fetch yesterday's cached data on a given morning, which would be problematic).

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/hooks/useSchedules.ts
+++ b/apps/site/assets/ts/hooks/useSchedules.ts
@@ -1,4 +1,6 @@
 import useSWR from "swr";
+import useInterval from "use-interval";
+import { useState } from "react";
 import { pick } from "lodash";
 import { fetchJsonOrThrow } from "../helpers/fetch-json";
 import { ScheduleWithTimestamp } from "../models/schedules";
@@ -36,6 +38,11 @@ const fetchData = async (url: string): Promise<ScheduleData[]> =>
 const useSchedulesByStop = (
   stopId: string
 ): FetchState<ScheduleWithTimestamp[]> => {
+  const [updateDate, setUpdateDate] = useState(Date.now());
+  useInterval(() => {
+    setUpdateDate(Date.now());
+  }, 15000);
+
   const { data, error } = useSWR<ScheduleData[]>(
     `/api/stops/${stopId}/schedules?last_stop_departures=false&future_departures=true`,
     fetchData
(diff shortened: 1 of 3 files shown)


apps/site/assets/ts/hooks/useSchedules.ts
```diff
@@ -36,14 +38,21 @@ const fetchData = async (url: string): Promise<ScheduleData[]> =>
...
     `/api/stops/${stopId}/schedules?last_stop_departures=false&future_departures=true`,
     fetchData
   );
   if (error) {
     return { status: FetchStatus.Error };
   }
-  const parsedData = data?.map(d => parse(d));
+  const parsedData = data
+    ?.map(d => parse(d))
+    .filter(d => d.time.getTime() >= updateDate);
```
kotva006, line 55: So this is filtering all the schedules that are before the `updateDate`?

thecristen: Yeah, excluding all schedules before the ever-changing `updateDate`.

apps/site/assets/ts/hooks/useSchedules.ts
```diff
@@ -36,14 +38,21 @@ const fetchData = async (url: string): Promise<ScheduleData[]> =>
...
+
   const { data, error } = useSWR<ScheduleData[]>(
     `/api/stops/${stopId}/schedules?last_stop_departures=false&future_departures=true`,
     fetchData
   );
   if (error) {
     return { status: FetchStatus.Error };
   }
-  const parsedData = data?.map(d => parse(d));
+  const parsedData = data
```
i0sea, line 53: question: The number of schedules is no longer an issue because there's no longer a react hook for each right?

i0sea (commented):
Any particular reason 15 seconds was chosen for refresh?

apps/site/assets/ts/hooks/useSchedules.ts
thecristen: @i0sea yes...
(also I had no special reason for choosing 15 seconds, just felt nice and frequent and avoids waiting 59 seconds for a "<1 minute" to go away 😅 )

PR #1667, opened 2023-07-13
wip: consolidate logic to `useFetch`

thecristen:
#### Summary of changes

I introduced `useFetch` in https://github.com/mbta/dotcom/pull/1658/commits/05fee508f836976ff48579b6f58bfedee94ddb5e, this is the attempt to use it across all our fetches in the stops page redesign.

**I considered deleting all the hooks** and using `useFetch` directly and repeatedly, which would produce code something like this

```typescript
  const stopResult = useFetch<Stop>(`/api/stop/${stopId}`);
  const routesWithPolylinesResult = useFetch<
    (Route & { polylines: Polyline[] })[]
  >(`/api/routes/by-stop/${stopId}`);
  const alertsForStopResult = useFetch<Alert[]>(`/api/stops/${stopId}/alerts`);
  const facilities = useFetch<Facility[]>(`/api/stop/${stopId}/facilities`);
  const alertsForRoutesResult = useFetch<Alert[]>(`/api/alerts?route_ids=${
    (routesWithPolylinesResult.data || []).map(r => r.id).join(",")}`);
```

But it felt messy, would've made refactoring all the tests and hook mocking a heavy lift, and `useSchedulesByStop` had some additional logic (parsing the data before returning).

Welcome feedback, visions, comments.

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/hooks/useFetch.ts
+++ b/apps/site/assets/ts/hooks/useFetch.ts
@@ -1,0 +1,27 @@
+import useSWR from "swr";
+import { fetchJsonOrThrow } from "../helpers/fetch-json";
+import { FetchState, FetchStatus } from "../helpers/use-fetch";
+
+/**
+ * Generic helper function for fetching JSON from a URL. Specify the expected
+ * return type as the type parameter, e.g
+ * `useFetch<Schedule[]>("/api/schedules/or/something")`
+ * @param {string | null} backendUrl - The path, usually starting with /api/,
+ * that our backend will respond with data from. If this value is null, useSWR's
+ * behavior will in effect disable the fetch, and the returned data will simply
+ * be undefined.
+ * @returns {FetchState<T>} An object containing the fetch status, and
+ * (optionally) the fetched data.
+ */
+const useFetch = <T>(backendPath: string | null): FetchState<T> => {
+  const { data, error } = useSWR<T>(
+    backendPath,
+    async (url: string): Promise<T> => fetchJsonOrThrow(url)
+  );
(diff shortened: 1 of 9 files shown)


apps/site/assets/ts/hooks/useFetch.ts
```diff
@@ -0,0 +1,27 @@
...
+ * return type as the type parameter, e.g
+ * `useFetch<Schedule[]>("/api/schedules/or/something")`
+ * @param {string | null} backendUrl - The path, usually starting with /api/,
+ * that our backend will respond with data from. If this value is null, useSWR's
+ * behavior will in effect disable the fetch, and the returned data will simply
+ * be undefined.
+ * @returns {FetchState<T>} An object containing the fetch status, and
+ * (optionally) the fetched data.
+ */
+const useFetch = <T>(backendPath: string | null): FetchState<T> => {
```
kotva006, line 16: Could this be named something else?  We already have a function called `useFetch` from the `helpers/use-fetch` file.  It would be nice to avoid name collisions to help with readability.

thecristen: I really missed that that file has, among other things, it's old overlapping `useFetch` 🤦🏼 ... yeah, will resolve.

thecristen: `helpers/use-fetch` is only used in two spots, so I think I'm gonna refactor those and make the `helpers/use-fetch` file obsolete.

thecristen:
Can revisit later, it's a bit much for now

PR #1669, opened 2023-07-13
Page should Display No Upcoming trips when there is a route with no schedules or predictions

i0sea:
#### Summary of changes

**Asana Ticket:** [Page should Display No Upcoming trips when there is a route with no schedules or predictions](https://app.asana.com/0/385363666817452/1204553726528494/f)

---

Display no upcoming trips in row when there are no schedules or predictions in either direction.

#### General checks


apps/site/assets/ts/stop/components/DepartureCard.tsx
```diff
@@ -34,22 +34,28 @@ const DepartureCard = ({
...
-      />
-      <DepartureTimes
-        key={`${route.id}-1`}
-        route={route}
-        directionId={1}
-        departuresForDirection={departuresByDirection[1]}
-        onClick={onClick}
-        alertsForDirection={allAlertsForDirection(alertsForRoute, 1)}
-      />
+      {!departuresByDirection[0] && !departuresByDirection[1] ? (
```
thecristen, line 37: I think it'd be better to treat each direction independently, so that if (for example) direction `1` has departures and `0` doesn't, we don't hide the departure times for `1`

thecristen (changes requested):
Don't we want to show a destination name for each direction, then the no trips text?
image

Seems like it would make more sense to adjust the logic inside the `<DepartureTimes />` component, to display this scenario.

apps/site/assets/ts/stop/components/DepartureCard.tsx
i0sea: @thecristen do you have a link to the designs for that screenshot?
I think this is what's currently happening. The `No upcoming trips` only shows if both directions have no upcoming trips. If either has trips it goes to render departure times

apps/site/assets/ts/helpers/__tests__/departureInfoTest.ts
```diff
@@ -1,10 +1,13 @@
+import { Route } from "@sentry/react/types/reactrouterv3";
```
thecristen, line 1: Please fix

apps/site/assets/ts/stop/components/DepartureCard.tsx
```diff
@@ -34,28 +36,26 @@ const DepartureCard = ({
...
-            key={`${route.id}-1`}
-            route={route}
-            directionId={1}
-            departuresForDirection={departuresByDirection[1]}
-            onClick={onClick}
-            alertsForDirection={allAlertsForDirection(alertsForRoute, 1)}
-          />
-        </>
-      )}
+      <>
```
thecristen, line 39: suggestion: The logic in `<DepartureTimes>` is growing quite complex and I wonder if it makes sense to make this problem `<DepartureCard>`'s responsibility instead. Maybe something like 
```
{
  departuresByDirection[0].length > 0 
  ? <DepartureTimes {/* existing stuff */} /> 
  : !isAtDestination(stopId, route, 0) && <> 
      <div className="departure-card__headsign-name">{route.direction_destinations[0]}</div>
      <div>No upcoming trips</div>
    </>
}
```

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -169,23 +174,47 @@ const DepartureTimes = ({
...
+          {!isAtDestination(stopId, route, directionId) ? (
+            departureTimeRow(
+              route.direction_destinations[directionId]
+                ? route.direction_destinations[directionId]!
+                : "",
+              [],
+              route.type === 2
+            )
+          ) : (
+            <></>
```
thecristen, line 214: Looks like this would create a clickable row that's empty, is that your intention?

apps/site/assets/ts/stop/components/DepartureTimes.tsx
```diff
@@ -169,23 +174,47 @@ const DepartureTimes = ({
...
+            );
+          })}
+        </>
+      ) : (
+        <div
+          key={`${route.direction_destinations[directionId]}-${route.id}`}
+          onClick={() => onClick(route, directionId)}
+          onKeyDown={() => onClick(route, directionId)}
+          role="presentation"
+        >
```
thecristen, line 204: opportunity: since they all do the same thing onClick, making a reusable component for clickable rows would reduce the need to retype this div and these attributes

apps/site/assets/ts/helpers/departureInfo.ts
```diff
@@ -107,9 +107,21 @@ const mergeIntoDepartureInfo = (
...
 };
 
+const isAtDestination = (
+  stopName: string,
+  route: Route,
+  directionId: DirectionId
+): boolean => {
+  return (
+    stopName ===
+    route.direction_destinations[directionId]!.replace(" Station", "")
```
thecristen, line 117: suggestion: some direction destinations might be named "Stop 1 or Stop 2", so I think `route.direction_destinations[directionId].includes(stopName)` will be more flexible and work ok here. though I do feel like a more robust solution might be needed eventually.

i0sea: I didn't do includes since some headsigns have `or` . For example 450 headsign could be "Wonderland or Haymarket" and didn't want to drop incase its going to Haymarket. Open to suggestions for this

apps/site/assets/ts/stop/components/DepartureTimes.tsx
i0sea: made a component but it looks pretty similar to typing this div 😅 
```
<DepartureRow
              route={route}
              directionId={directionId}
              onClick={() => onClick(route, directionId)}
              >
                {getRow(headsign, departures, alertsForDirection, overrideDate)}
              </DepartureRow>
       ```

apps/site/assets/ts/helpers/departureInfo.ts
thecristen: ohh yeah I see what you're saying, that's a good point. 
though I'm not sure you'll actually hit that condition given how the logic leading to this call is structured, because when you're checking this in the departure rows you're in the not condition of `Object.keys(groupedDepartures).length > 0` (which is derived from departures filtered for this direction)... so if I'm interpreting this right you're already in the situation where there are 0 departures for the entire direction anyway.

thecristen (changes requested):
can you add tests showing:
- that the route destination is being displayed along with the No upcoming trips text
- that, when the `stopName` is the same as the destination, neither route destination nor No upcoming trips is shown.

thecristen (approved):

PR #1673, opened 2023-07-18
fix(DeparturesAndMap): add headsign to departure filters

thecristen:
Most of this PR is just renaming, but it fixes a major issue where, for directions having multiple destinations, clicking on each of them led to the same view (e.g. clicking Ashmont or Braintree each lead to a "Red Line to Ashmont" list with the same list of departures, instead of showing different lists).

#### Summary of changes

**Asana Ticket**: [Departures | Different destinations lead to the same realtime tracking page](https://app.asana.com/0/385363666817452/1205084471790526/f)

- remove repetition by defining a type for the filter and the function that handles it
- use headsign to help filter departures
- pass in the headsign as a prop to DepartureList
- simplify the filter type a bit by having a null state instead of a state where each object key can have a null value.

### DIFF
--- a/apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
+++ b/apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement, useLayoutEffect, useRef, useState } from "react";
-import { chain, isNull, some } from "lodash";
+import { chain } from "lodash";
 import { clearAllBodyScrollLocks, disableBodyScroll } from "body-scroll-lock";
 import { Alert, DirectionId, Route, Stop } from "../../__v3api";
 import StopPageDepartures from "./StopPageDepartures";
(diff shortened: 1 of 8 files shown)


apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -25,6 +25,13 @@ interface DeparturesAndMapProps {
   alerts: Alert[];
 }
 
+type DepartureFilter = {
+  route: Route;
+  directionId: DirectionId;
+  headsign: string;
+};
+export type DepartureFilterFn = (filters: DepartureFilter) => void;
```
i0sea, line 33: nitpick: Should this be defined in a helper file?

i0sea (approved):

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
thecristen: I'd say it should stay close to the departure filtering functionality it's related to (the state, the set departure filters function, the unset departure filters function, etc). If it goes in a helper file, all the other related bits should join it.

PR #1678, opened 2023-07-21
fix(DepartureList): (subway only) remove schedules when showing predictions

thecristen:
#### Summary of changes

**Asana Ticket:** [Departures | Inaccurate subway times](https://app.asana.com/0/385363666817452/1205064463145400/f)

~~Adds another flag to the schedules endpoint, to adjust the returned schedules to exclude subway schedules unless they're from a terminus on the Green Line. Deploying to dev-blue for testing.~~

**The change in this PR applies to the `<DepartureList />` component only.** Another PR applying this logic to the landing page `<DepartureTimes />` and its rows of times... will come after.

This covers a revised approach where, **for subways only**, we suppress schedule-only departures in favor of predictions. When there are no remaining predictions to display (usually after an hour or two away), the list allows showing schedules again through the remainder of the day.

---

#### General checks


kotva006:
This makes sense for what it is doing.  I thought we were using predictions over schedules on the display logic already.

apps/site/assets/ts/helpers/departureInfo.tsx
```diff
@@ -123,11 +135,54 @@ const isCommuterRail = (departureInfo: DepartureInfo): boolean => {
...
+ * This function can optionally
+ * - wrap each `<DisplayTime />` in a specified wrapper element, `WrapperEl`
+ * - truncate the list to a desired `listLength` value
+ */
+const departuresListFromInfos = (
+  departureInfos: DepartureInfo[],
+  isCR: boolean,
+  targetDate?: Date,
+  listLength?: number,
+  WrapperEl: typeof DefaultWrapper = DefaultWrapper
```
kotva006, line 154: Why is this needed?

thecristen: @kotva006 Does the comment on the line above this function declaration help answer this at all? What details are you looking for?

kotva006: I understand what it does, I just don't understand the point.  As this functionality doesn't appear to be used on this pull request

thecristen: @kotva006 looks like we commented at the same time - see my second comment

apps/site/assets/ts/helpers/departureInfo.tsx
```diff
@@ -123,11 +135,54 @@ const isCommuterRail = (departureInfo: DepartureInfo): boolean => {
...
+): React.ReactElement[] => {
+  // optional cutoff time, before which we won't show schedules.
+  // just used with subway for now.
+  const predictionTimeCutoff = chain(departureInfos)
+    .filter(d => d.routeMode === SUBWAY)
+    .maxBy("prediction.time")
+    .value()?.prediction?.time;
```
kotva006, line 161: nit-pick(non-blocking): Since we are maxing by `"prediction.time"` I would think the `prediction` and `time` field would be defined, so they could probably be `prediction!.time`;  Though not sure if an `!` is allowed after a `?`

apps/site/assets/ts/helpers/departureInfo.tsx
```diff
@@ -123,11 +135,54 @@ const isCommuterRail = (departureInfo: DepartureInfo): boolean => {
...
+        schedule &&
+        schedule.time <= predictionTimeCutoff
+    )
+    .map(d => (
+      <WrapperEl key={getInfoKey(d)}>
+        <DisplayTime departure={d} isCR={isCR} targetDate={targetDate} />
+      </WrapperEl>
+    ))
+    .slice(0, listLength)
+    .value();
```
kotva006, line 177: Why is this complex `chain` logic needed over just using the `prediction` from the `departureInfos`?  Based of the ticket it seems like the goal is just to show predictions if they exist, which I thought we were doing already.
The `departureInfoToTime` function returns the prediction time in favor of the schedule time.
Maybe I am having a hard time following this logic.

thecristen: @kotva006 Here's a non-chained translation, if it's helpful 🤔 

```javascript
const departuresListWithoutIntermingledSchedules = omitBy(departureInfos, departure => {
  if (predictionTimeCutoff) {
    // omit schedule-only departures that are before the latest prediction time
    !prediction && schedule && schedule.time <= predictionTimeCutoff;
  }
  return false;
});
const departureElements = departuresListWithoutIntermingledSchedules.map(d => /* render as <DisplayTime /> and wrap in wrapper element */)
return departureElements.slice(0, listLength);
```

apps/site/assets/ts/helpers/departureInfo.tsx
thecristen: I think you're right, and I just tried it and it works.

kotva006 (approved):
Looks fine

PR #1689, opened 2023-08-01
background data refresh improvements

thecristen:
#### Summary of changes

**Asana Ticket:** [Data Refresh when Stop Page in Background](https://app.asana.com/0/385363666817452/1205161605826762/f)

The three commits do different things, but they're all in the realm of improving the client experience of receiving data from the predictions channel.

1. ### `feat(channels): add socket close event listener`
This is the fundamental commit that prevents stale predictions from sticking around. 

**The problem:** In our case, after leaving a tab open in the background for ~10 minutes, the browser closes the WebSocket connection. With no more predictions data being received until a user refreshes the page. 

> [!NOTE]
> [Phoenix.js](https://hexdocs.pm/phoenix/js/) is _supposed_ to handle this situation (it's not uncommon or unexpected, after all) and automatically reconnect, but for whatever reason it doesn't. Perhaps it's fixed in a future version, but we're not in a great position to upgrade.

**The improvement:** Add an event listener to the socket's `onClose` event. If this is triggered, force a page reload via `window.location.reload()`. This is essentially what Glides and Skate do, and Anna and I figured it was an okay mitigation to take in this case, because this problem is generally only encountered on inactive tabs (so refreshing won't interrupt any user actions).

2. ### `feat(useChannel and channels): on visibilitychange`

**The problem:** The client's always receiving data, even when not actively using the tab! This will probably waste energy and cellular data.

**The improvement:** Be a bit more judicious, and leave channels when the page is not active, and rejoin when resuming activity. We use the widely-supported [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) for this, and pretty much copy Glide's approach. 

3.  ### `feat(PredictionsPubSub): batch updates`

**The problem:** The channel broadcasts many updates at once containing the same data, instead of sending the client all the data in one. It turns out that the events sent from the V3 API's ServerSentEvents stream are themselves not batched.

**The improvement:** Instead of broadcasting with _every_ event, only broadcast periodically. This PR sets this to 10 seconds.

## How to preview

It turns out the Phoenix socket supports a `logger` argument which we can use to turn on logging for all websocket activity. So adding this change in `channels.ts`  will print out every event in the DevTools console.

```javascript
  // replace window.socket = new Socket("/socket", {});
  window.socket = new Socket("/socket", {
    logger(kind, message, data) {
      // eslint-disable-next-line no-console
      console.log(
        new Date().toLocaleTimeString(),
        `[${kind}] ${message}`,
        data
      );
    }
  });

```

---

#### General checks

### DIFF
--- a/apps/predictions/config/config.exs
+++ b/apps/predictions/config/config.exs
@@ -1,30 +1,7 @@
-# This file is responsible for configuring your application
-# and its dependencies with the aid of the Mix.Config module.
-use Mix.Config
+import Config
 
-# This configuration is loaded before any dependency and is restricted
-# to this project. If another project depends on this project, this
-# file won't be loaded nor affect the parent project. For this reason,
-# if you want to provide default values for your application for
-# 3rd-party users, it should be done in your "mix.exs" file.
-
-# You can configure for your application as:
-#
-#     config :predictions, key: :value
-#
-# And access this configuration in your application as:
-#
-#     Application.get_env(:predictions, :key)
-#
-# Or configure a 3rd-party app:
-#
-#     config :logger, level: :info
-#
-
-# It is also possible to import configuration files, relative to this
-# directory. For example, you can emulate configuration per environment
-# by uncommenting the line below and defining dev.exs, test.exs and such.
-# Configuration from the imported file will override the ones defined
-# here (which is why it is important to import them last).
-#
-#     import_config "#{Mix.env}.exs"
+if config_env() == :test do
+  config :predictions, broadcast_interval_ms: 50
+else
+  config :predictions, broadcast_interval_ms: 10_000
+end
--- a/apps/site/assets/ts/app/channels.ts
+++ b/apps/site/assets/ts/app/channels.ts
@@ -96,12 +96,14 @@ const setupChannels = (): void => {
   window.socket.connect();
   window.channels = {};
 
-  document.addEventListener("turbolinks:load", () => {
+  const joinAllChannels = (): void => {
     document.querySelectorAll("[data-channel]").forEach(el => {
       const channelId = el.getAttribute("data-channel");
       if (channelId) joinChannel(channelId);
     });
-  });
+  };
+
+  document.addEventListener("turbolinks:load", joinAllChannels);
 
   // leave subscribed channels when navigating away from a page.
   const leaveAllChannels = (): void => {
@@ -109,6 +111,14 @@ const setupChannels = (): void => {
   };
   document.addEventListener("turbolinks:before-render", leaveAllChannels);
   window.addEventListener("beforeunload", leaveAllChannels);
+
+  document.addEventListener("visibilitychange", () => {
+    if (document.hidden) {
+      leaveAllChannels();
+    } else {
+      joinAllChannels();
+    }
+  });
 };
 
 export { joinChannel, leaveChannel };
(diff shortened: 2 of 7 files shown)


apps/predictions/config/config.exs
```diff
@@ -1,30 +1,7 @@
...
-
-# It is also possible to import configuration files, relative to this
-# directory. For example, you can emulate configuration per environment
-# by uncommenting the line below and defining dev.exs, test.exs and such.
-# Configuration from the imported file will override the ones defined
-# here (which is why it is important to import them last).
-#
-#     import_config "#{Mix.env}.exs"
+if config_env() == :test do
+  config :predictions, broadcast_interval_ms: 50
```
i0sea, line 4: question: why every half second on test env?

apps/site/assets/ts/app/channels.ts
```diff
@@ -84,22 +84,41 @@ const leaveChannel = (id: string): void => {
 
 const setupChannels = (): void => {
   window.socket = new Socket("/socket", {});
+  window.socket.onClose(event => {
```
i0sea, line 87: question: Is the `close` event type only ever caused by the socket being closed by the browser?

apps/site/assets/ts/hooks/useChannel.ts
```diff
@@ -45,6 +46,22 @@ function useChannel<
...
+  const isVisible = usePageVisibility();
+  useLayoutEffect(() => {
+    if (channelId !== null) {
+      if (!isVisible) {
+        leaveChannel(channelId);
+      } else {
+        joinChannel<OnDataEventType[]>(channelId, dispatch);
+      }
+    }
+    return () => {
```
i0sea, line 58: nitpick: could this return occur after `!isVisible` or does useLayoutEffect require a return?

apps/site/assets/ts/hooks/usePageVisibility.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export const usePageVisibility = (): boolean => {
+  const [hidden, setHidden] = useState<boolean>(false);
+  const callback = (): void => {
+    setHidden(document.hidden);
+  };
+  useLayoutEffect(() => {
+    document.addEventListener("visibilitychange", callback);
+    return () => {
+      document.removeEventListener("visibilitychange", callback);
```
i0sea, line 15: question: Don't we always want to be listening for visibility changes to know wether to leave or join channels?

i0sea (commented):
Some clarifying questions. Looks good overall

apps/predictions/config/config.exs
thecristen: I wanted to be able to pass the tests but the tests have a default timeout, I think 100ms? I wanted to be able to broadcast predictions before the test time ran out 😓

apps/site/assets/ts/app/channels.ts
thecristen: probably not! I would guess it _could_ also be closed programmatically (we don't do it as far as I'm aware). and honestly, I'm not sure where/how else this event could be caused.

apps/site/assets/ts/hooks/useChannel.ts
thecristen: like `useEffect`, `useLayoutEffect` calls the (optional!) return function when the component is destroyed. so it's an appropriate place for any sort of cleanup tasks. Some more documentation [here](https://react.dev/learn/synchronizing-with-effects#how-to-write-an-effect).

apps/site/assets/ts/hooks/usePageVisibility.ts
thecristen: yeah, probably!

i0sea (approved):
🚀

PR #1694, opened 2023-08-02
feat(StopPageRedesign): surface error message if predictions aren't fetched

thecristen:
## Summary of changes

**Asana Ticket:** [Bubble up prediction fetching error](https://app.asana.com/0/385363666817452/1205144855393526/f) and [Departures | Predictions aren't loading consistently](https://app.asana.com/0/385363666817452/1205093855734953/f)

1. #### Bubble up prediction fetching error

Use a `null` return type of `usePredictionsChannel` to signify the error state. Makes use of the fact that the underlying `useChannel` function is designed to return the `initialState` argument if the `channelName` argument happens to be `null`.

2. #### Add error message to top of page if predictions aren't fetched

If predictions are `null`, show the message at top. This uses some of the styling of the Transit Near Me error banner, but (a) doesn't use heading styling, just bolds the top text, and (b) removes the button to dismiss the message. 

I decided not to make a standalone reusable version of the error message component, figuring we could wait until we need to use this a second or third time...

## Future

I'm considering adding an `onError` callback to the `useChannel` hook and underlying `joinChannel` function, so that we can more properly respond to other types of errors in the channel.

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/stop/components/StopPageRedesign.tsx
+++ b/apps/site/assets/ts/stop/components/StopPageRedesign.tsx
@@ -20,7 +20,7 @@ import { Alert } from "../../__v3api";
 const isStopPageAlert = ({ effect }: Alert): boolean =>
   ["suspension", "stop_closure", "station_closure", "shuttle"].includes(effect);
 
-const FullwidthErrorMessage = (
+const FullwidthErrorMessage = (): JSX.Element => (
   <div className="c-fullscreen-error__container">
     <div className="container">
       <p className="c-fullscreen-error__heading u-bold">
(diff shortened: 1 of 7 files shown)


apps/site/assets/ts/stop/components/StopPageRedesign.tsx
```diff
@@ -20,11 +20,22 @@ import { Alert } from "../../__v3api";
 const isStopPageAlert = ({ effect }: Alert): boolean =>
   ["suspension", "stop_closure", "station_closure", "shuttle"].includes(effect);
 
+const FullwidthErrorMessage = (
```
kotva006, line 23: nit-pick: For consistency this should be a function returning a react component.  We don't really use names with capital first letters

kotva006 (changes requested):
It looks like a typescript unit test is failing

PR #1709, opened 2023-08-11
feat: enable stops redesign

thecristen:
Switches the feature flag: now we can use it to re-enable the old design.

- fix: enable redirecting child stop IDs to parent stop pages, to match the original stop page
- stop checking for zone number or routes, the new stop page doesn't assign those variables
- the new stop page doesn't render any react content on the server, so remove that test


apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -66,19 +66,26 @@ defmodule SiteWeb.StopController do
...
+          |> halt()
+        else
+          routes_by_stop = Routes.Repo.by_stop(stop_id, include: "stop.connecting_stops")
+
+          conn
+          |> assign(:breadcrumbs, breadcrumbs(stop, routes_by_stop))
+          |> meta_description(stop, routes_by_stop)
+          |> render("show-redesign.html", %{
+            stop_id: stop_id,
+            routes_by_stop: routes_by_stop
```
kotva006, line 81: nit-pick(non-blocking): This can be addressed in later clean up, but I don't think we use the `routes_by_stop` property on the template page (`show-redesign.html`).

i0sea (approved):
🚀

kotva006 (approved):
Looks good!

PR #1710, opened 2023-08-14
fix: check current lifecycle when counting out of order amenities

thecristen:
#### Summary of changes

**Asana Ticket:** [Elevator Status Affected by Upcoming Alerts](https://app.asana.com/0/385363666817452/1205268550058218/f)

The first commit fixes the table, the second fixes the badge, both should now display the correct number of current outages.

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/models/__tests__/alert-test.ts
+++ b/apps/site/assets/ts/models/__tests__/alert-test.ts
@@ -4,6 +4,7 @@ import {
   Alert,
   Facility,
   InformedEntitySet,
+  Lifecycle,
   TimePeriodPairs
 } from "../../__v3api";
 import {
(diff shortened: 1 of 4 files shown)


apps/site/assets/ts/models/__tests__/alert-test.ts
```diff
@@ -398,7 +398,7 @@ describe("alertsAffectingBothDirections", () => {
   });
 });
 
-describe("hasFacilityAlert", () => {
+describe("hasCurrentFacilityAlert", () => {
```
kotva006, line 401: nit-pick: could there be a test to ensure that non current alerts are filtered out from the `hasCurrentFacilityAlert` function.

PR #1714, opened 2023-08-16
feat: pull stop/trip information in with Vehicle stream

paulswartz:
This avoids making a bunch of extra API calls for data that the streaming API can already provide.

#### Summary of changes

**Asana Ticket:** No ticket

---

#### General checks

* [x] **Have we load-tested** any new pages or internal API endpoints that will receive significant traffic? [See load testing docs](./../apps/site/load_tests/README.md) (this reduces the number of API calls)

### DIFF
--- a/apps/vehicles/lib/parser.ex
+++ b/apps/vehicles/lib/parser.ex
@@ -3,13 +3,16 @@ defmodule Vehicles.Parser do
   alias Vehicles.Vehicle
 
   @spec parse(JsonApi.Item.t()) :: Vehicle.t()
-  def parse(%JsonApi.Item{id: id, attributes: attributes, relationships: relationships}) do
+  def parse(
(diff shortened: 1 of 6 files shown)


thecristen:
One distinction I noticed between this implementation and what we might do for predictions, is that our `%Vehicle{}` only needs the IDs to populate `shape_id` and `stop_id`, so no `attributes` were needed to be fetched/saved/otherwise handled. But our `%Prediction{}` expects full parsed `%Stop{}`, `%Trip{}`, and `%Route{}`, so yeah, ingesting and parsing would have to be handled differently on our end.. 🤔

paulswartz:
@thecristen that's true, but all the data you need to parse into a Stop, Trip, or Route struct is present. This PR only pulls out the data that it needs, but you should be seeing stops/trips/routes with all of the data in the feed.

thecristen:
Closing as obsolete.

PR #1715, opened 2023-08-17
Adds new priority rules for delay alerts on bus routes

ErinLMoore:
#### Summary of changes

**Asana Ticket:** [Make most bus delay alerts due to traffic low priority on MBTA.com](https://app.asana.com/0/385363666817452/1204624899415346/f)

Alerts of type delay are now set to low priority if at least one of their informed entities has a type of "3" (bus) and their severity is less than 5. All other delay alerts are unchanged. This change is being made in response to rider feedback.

#### General checks
* [X] **Are the changes organized into self-contained commits with descriptive and well-formatted commit messages?** This is a good practice that can facilitate easier reviews.
* [X] **Testing.** Do the changes include relevant passing updates to tests? This includes updating screenshots. Preferably tests are run locally to verify that there are no test failures created by these changes, before opening a PR.
* [X] **Tech debt.** Have you checked for tech debt you can address in the area you're working in? This can be a good time to address small issues, or create Asana tickets for larger issues.


apps/alerts/lib/priority.ex
```diff
@@ -16,7 +16,19 @@ defmodule Alerts.Priority do
     :low
   end
 
-  def priority(%{effect: :delay}, _) do
+  def priority(
+        %{effect: :delay, severity: severity, informed_entity: informed_entities},
```
i0sea, line 20: question: don't all delay alerts have a severity and informed_entity? If so wouldn't `def priority(%{effect: :delay}, _params)` be unreachable?

ErinLMoore: I was a little unsure if we were guaranteed to have an informed entity here or not. If we do always have one, then this would be unreachable. I can go ahead and remove it if that's the case.

ErinLMoore: This one only applies to delay alerts with a severity under 5 (line 23), the rest will all still use `def priority(%{effect: :delay}, _params)`.

apps/alerts/lib/priority.ex
```diff
@@ -16,7 +16,19 @@ defmodule Alerts.Priority do
...
+        _time
+      )
+      when severity < 5 do
+    # delay alerts for bus routes under severity 5 are low priority
```
lund5000, line 24: This logic doesn't quite seem to capture what's written in the task: "Bus delay alerts due to traffic with a severity of 5 or less are low priority."

Expanding on that, it should also be the case that:
- Bus delay alerts with a `severity` of 6 or more are high priority, regardless of `cause`.
- Bus delay alerts with any `cause` other than "TRAFFIC" and a `severity` of 5 or less are high priority, including alerts without a `cause` value.

So I think we'd need to change the `<` to `<=` and add logic that checks the `cause` of the alert. (I'm no Elixir expert, but at a glance it also looks like we'd need to start adding `cause` to the Alert struct to do so?)

lund5000 (changes requested):

apps/alerts/lib/priority.ex
ErinLMoore: Sorry about that, I somehow misread "5 or under" as "less than 5".

I will look into adding the `cause` logic.

ErinLMoore: @lund5000  If I'm reading the GTFS docs right, TRAFFIC isn't a possible cause that can be listed for an alert. It can be listed on a stop time update, but I don't think we currently have access to those in DotCom. 

https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#enum-cause

lund5000: That's true, but we regularly publish alerts with all sorts of `cause` values that aren't compliant with that spec: "TRAFFIC", "SPECIAL_EVENT", etc.

ErinLMoore: I wonder if we're using the cause_description there

apps/schedules/test/hours_of_operation_test.exs
```diff
@@ -17,8 +17,8 @@ defmodule Schedules.HoursOfOperationTest do
       assert %Departures{} = week_1
       assert %Departures{} = saturday_0
       assert %Departures{} = saturday_1
-      assert %Departures{} = sunday_0
-      assert %Departures{} = sunday_1
+      assert :no_service = sunday_0
```
ErinLMoore, line 20: This is unrelated to my code, but this test started failling. I don't really understand the implications of this, maybe the service on this route changed?

apps/site/lib/site/realtime_schedule.ex
```diff
@@ -210,6 +210,7 @@ defmodule Site.RealtimeSchedule do
           map
   defp do_get_schedules(route_id, stop_id, route_patterns, now, schedules_fn) do
     route_pattern_dictionary = make_route_pattern_dictionary(route_patterns, stop_id)
+
```
ErinLMoore, line 213: The formatter added this

apps/schedules/test/hours_of_operation_test.exs
thecristen: We can disregard this, I've noticed this recently failing on other branches as well and I haven't gotten around to investigating yet.

ErinLMoore: Gotcha, so just take out these changes then?

thecristen: Yes please 😅

lund5000 (approved):
The alert changes look good after a little end-to-end testing on dev-green: a 20-minute bus delay due to traffic gets low priority treatment while other more severe bus delays are still given high priority.

I was able to reproduce some 500 errors when navigating to a couple Silver Line schedule pages, which I think is unrelated but felt I should mention nevertheless:
- https://dev-green.mbtace.com/schedules/743
- https://dev-green.mbtace.com/schedules/749


image

image

image

ErinLMoore:
> The alert changes look good after a little end-to-end testing on dev-green: a 20-minute bus delay due to traffic gets low priority treatment while other more severe bus delays are still given high priority.
> 
> I was able to reproduce some 500 errors when navigating to a couple Silver Line schedule pages, which I think is unrelated but felt I should mention nevertheless:
> 
>     * https://dev-green.mbtace.com/schedules/743
> 
>     * https://dev-green.mbtace.com/schedules/749
> 
> 
> image
> 
> image
> 
> image

I don't see those errors locally or in production, just dev-green. Wasn't able to find the error message in Splunk, I agree it doesn't seem like this code.

thecristen:
> I don't see those errors locally

Just wanted to confirm: is this locally using the dev-green API URL? Very curious... 🤔

ErinLMoore:
> Just wanted to confirm: is this locally using the dev-green API URL?

Oh, sorry, I meant I don't get errors going to that route when I run the code locally. I do see them on dev-green.

thecristen:
I'm able to reproduce the broken Silver Line pages locally by changing the `V3_API` environment variable to the dev-green one. At first glance it looks unrelated to the changes here... so carry on (I'll continue to investigate that). :eyes:

thecristen:
Since this is approved I'll merge this in! Thanks for your contribution! ✨

ErinLMoore:
> 

Thank you!

PR #1719, opened 2023-08-22
feat(Facilities): Added a cached repo for the facilities info from th…

kotva006:
…e API

#### Summary of changes
Most of these changes are scaffolding to create a facilities application in an attempt to align access pattern similar to the routes/stops/schedules.

The important part of these changes are that the results from the API are cached for an hour.  The ticket didn't specify the time to keep the cache valid, but figured that was a good start.

**Asana Ticket:** [Cache Facility Info](https://app.asana.com/0/555089885850811/1205293745657617)

---

#### General checks


apps/facilities/lib/repo.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Facilities.Repo do
+  @moduledoc """
+  Repo to get facilitiy information.
+  """
+  use RepoCache, ttl: :timer.hours(1)
+
+  def get_for_stop(stop_id) do
+    cache({stop_id}, fn {stop_id} ->
```
thecristen, line 8: comment(non-blocking): since it's just one argument, you don't need to wrap it in a Tuple.

apps/facilities/test/repo_test.exs
```diff
@@ -0,0 +1,15 @@
...
+  use ExUnit.Case
+  alias Facilities.Repo
+
+  import Mock
+
+  describe "get_for_stop/1" do
+    test "should call the api" do
+      with_mock V3Api.Facilities, filter_by: fn _stop_id -> [] end do
+        assert [] = Repo.get_for_stop("test-id")
+        assert_called(V3Api.Facilities.filter_by([{"stop", "test-id"}]))
```
thecristen, line 11: TIL `assert_called`, awesome.

thecristen: I'm curious if calling it again would test the caching functionality (e.g. it shouldn't call the API again if calling the Repo less than an hour later). But I don't think we test anything else in that manner (I suppose we assume the caching functionality is tested elsewhere and works), so this isn't a blocking comment so much as a curiosity 😅

apps/facilities/test/test_helper.exs
```diff
@@ -0,0 +1,6 @@
+# Ensure the deps are all started
+Application.ensure_all_started(:facilities)
+
+# Report warnings as errors
+Code.compiler_options(warnings_as_errors: true)
+ExUnit.start(max_cases: 1)
```
thecristen, line 6: question: Why the `max_cases`?

apps/site/lib/site_web/router.ex
```diff
@@ -211,7 +211,7 @@ defmodule SiteWeb.Router do
 
     get("/alerts", AlertController, :show_by_routes)
     get("/stops/:stop_id/alerts", AlertController, :show_by_stop)
-    get("/stop/:stop_id/facilities", StopController, :get_facilities)
```
thecristen, line 214: question: I didn't notice any code removing `StopController.get_facilities/2`, did you mean to include that in this PR?

apps/site/lib/site_web/controllers/facilities_controller.ex
```diff
@@ -0,0 +1,18 @@
...
+  Page for display of information about facilities.
+  """
+  use SiteWeb, :controller
+
+  def get_facilities(conn, %{"stop_id" => stop_id}) do
+    res = Facilities.Repo.get_for_stop(stop_id)
+
+    case res do
+      {:error, _x} ->
+        json(conn, res)
```
thecristen, line 12: suggestion(non-blocking): we have some helper functions in `SiteWeb.ControllerHelpers` to return error status, they're not widely used but I think `return_error/3` might be a good fit here.

apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -494,18 +494,4 @@ defmodule SiteWeb.StopController do
 
     %{conn | assigns: assigns}
   end
-
-  def get_facilities(conn, %{"stop_id" => stop_id}) do
```
thecristen, line 498: removing this breaks one of the unit tests, can you please remove that test too?

thecristen (approved):
👍🏼

PR #1721, opened 2023-08-23
WIP: Pick between OTP1 and 2 based on mbta_id cookie

EmmaSimon:
This generates a random value, seeded by the mbta_id cookie, which stays static per user (until cookies are cleared), to determine which instance of OTP they should be served. The percentage of traffic being served OTP2 can be configured with a new env var.

Two new feature flags are also included, which can be used to force a specific version (OTP2 is picked if both are selected).

#### Summary of changes

**Asana Ticket:** [Make dotcom changes to perform traffic splitting between OTP1 and OTP2](https://app.asana.com/0/1204355099788517/1205204806153400/f)

---

#### General checks

### DIFF
--- a/apps/site/lib/site/trip_plan/query.ex
+++ b/apps/site/lib/site/trip_plan/query.ex
@@ -23,15 +23,15 @@ defmodule Site.TripPlan.Query do
           itineraries: TripPlan.Api.t() | nil
         }
 
-  @spec from_query(map, Keyword.t()) :: t
-  def from_query(params, date_opts) do
+  @spec from_query(map, TripPlan.Api.connection_opts(), Keyword.t()) :: t
+  def from_query(params, connection_opts, date_opts) do
     opts = get_query_options(params)
 
     %__MODULE__{}
     |> Site.TripPlan.DateTime.validate(params, date_opts)
     |> Site.TripPlan.Location.validate(params)
     |> include_options(opts)
-    |> maybe_fetch_itineraries(opts)
+    |> maybe_fetch_itineraries(connection_opts, opts)
   end
 
   @spec get_query_options(map) :: keyword()
@@ -42,17 +42,18 @@ defmodule Site.TripPlan.Query do
     |> opts_from_query
   end
 
-  @spec maybe_fetch_itineraries(t, Keyword.t()) :: t
+  @spec maybe_fetch_itineraries(t, TripPlan.Api.connection_opts(), Keyword.t()) :: t
   defp maybe_fetch_itineraries(
          %__MODULE__{
            to: %NamedPosition{},
            from: %NamedPosition{}
          } = query,
+         connection_opts,
          opts
        ) do
     if Enum.empty?(query.errors) do
       query
-      |> fetch_itineraries([query.time | opts])
+      |> fetch_itineraries(connection_opts, [query.time | opts])
       |> parse_itinerary_result(query)
     else
       query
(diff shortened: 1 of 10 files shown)


apps/trip_plan/lib/trip_plan/personal_detail.ex
```diff
@@ -64,4 +67,13 @@ defmodule TripPlan.PersonalDetail.Step do
   def human_relative_direction(:elevator), do: "Take the elevator"
   def human_relative_direction(:uturn_left), do: "Make a U-turn"
   def human_relative_direction(:uturn_right), do: "Make a U-turn"
+  def human_relative_direction(:enter_station), do: "Enter the station"
+  def human_relative_direction(:exit_station), do: "Exit the station"
+  def human_relative_direction(:follow_signs), do: "Follow signs"
+
+  @spec human_relative_preposition(relative_direction) :: binary
```
thecristen, line 74: what a neat addition!

thecristen (approved):
Just a few minor questions/suggestions, generally speaking this seems like a good approach!

Also I know this wasn't the main point of this PR, but it's really interesting being able to run trip plans with either version, and interesting comparing the results! I'm sure we'll be learning a lot in the days to come :grin:

<figure>
image
<figcaption>Trip plan output using OPT1 on left, OPT2 on right.</figcaption>
</figure>

apps/trip_plan/lib/trip_plan/personal_detail.ex
EmmaSimon: I didn't even realize this had been implemented until the json started failing to parse because there were unexpected values! I think it's a quite new feature, from within the last couple weeks. I hope they make grammatical sense in most cases, I'm not certain exactly where the location strings are being pulled from, I assume it's GTFS pathways? And I really don't like when it does the thing like in your screenshot where it says to follow the same signs multiple times in a row... I'll try bringing that up with them, but we might need to filter out consecutive dupes on our side.

apps/site/lib/site/trip_plan/query.ex
```diff
@@ -63,23 +64,24 @@ defmodule Site.TripPlan.Query do
     query
   end
 
-  @spec fetch_itineraries(t, Keyword.t()) :: TripPlan.Api.t()
+  @spec fetch_itineraries(t, String.t(), Keyword.t()) :: TripPlan.Api.t()
```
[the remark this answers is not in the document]
EmmaSimon: Good catch 😅

apps/trip_plan/lib/trip_plan/api.ex
```diff
@@ -16,6 +16,12 @@ defmodule TripPlan.Api do
           | {:optimize_for, :less_walking | :fewest_transfers}
           | {:max_walk_distance, float}
   @type plan_opts :: [plan_opt]
+  @type connection_opt ::
+          {:user_id, integer}
+          | {:force_otp1, boolean}
+          | {:force_otp2, boolean}
+  @type connection_opts :: [connection_opt]
```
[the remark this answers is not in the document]
EmmaSimon: Oooh, yes that makes sense, I just assumed that was how keyword lists were represented generally.

apps/trip_plan/lib/trip_plan/api/open_trip_planner.ex
```diff
@@ -26,12 +26,54 @@ defmodule TripPlan.Api.OpenTripPlanner do
...
+      true ->
+        user_id = connection_opts[:user_id]
+        percent_threshold = get_otp2_percentage()
+
+        :rand.seed(:exsss, user_id)
+        placement = :rand.uniform()
+        use_otp2 = placement < percent_threshold / 100
+
+        Logger.info(fn ->
+          "#{__MODULE__}.pick_url placement=#{placement} otp2_percentage=#{percent_threshold}% mbta_id=#{user_id} use_otp2=#{use_otp2}"
```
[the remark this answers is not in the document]
EmmaSimon: I thought about it, but wasn't sure if it would be worth it, since it would only be used internally, but I'd be happy to, it might be useful if we're trying to test one or the other to make sure we're hitting what we expect.

apps/trip_plan/lib/trip_plan/personal_detail.ex
EmmaSimon: https://github.com/opentripplanner/OpenTripPlanner/issues/5322
Made an issue in the OTP repo on the dupes

EmmaSimon:
Ok, I'm pretty sure I fixed all the tests effected by the changes and the only ones failing now are unrelated. Are the existing failures expected @thecristen ? And did you want to take another pass? The only changes from the initial approval are in the tests.

apps/trip_plan/test/api/open_trip_planner/http_test.exs
```diff
@@ -12,10 +12,11 @@ defmodule TripPlan.Api.OpenTripPlanner.HttpTest do
   alias TripPlan.NamedPosition
   import Plug.Conn, only: [send_resp: 3]
 
-  @tag :external
-  @tag :skip
-  describe "plan/3" do
-    test "can make a basic plan" do
+  describe "plan/4" do
+    @describetag :external
+    @describetag :skip
```
thecristen, line 17: TIL about `@describetag`! 😆

thecristen (approved):
Yep, tests look good! Thanks!

PR #1725, opened 2023-08-24
feat: add mechanism for closing streams of prediction data

thecristen:
#### Summary of changes

**Asana Ticket:** [Close open connections to the API when there are no longer subscriptions to the stream](https://app.asana.com/0/385363666817452/1205293745657616/f)

I ended up not implementing an explicit `unsubscribe` function as originally written in the attached Asana ticket, as I came to realize:

* Subscribers listed in the `:prediction_subscriptions_registry` were already being removed automatically,
* When subscribers were removed, this wasn't via an `:unsubscribe` event. The processes that were subscribed were simply being shut down whenever a channel was closed.

I considered adding some code to a `terminate/2` callback in `SiteWeb.PredictionsChannel`, but ran into some race conditions around handling its `pid` and lookup in the registry, then I came across this note in the `Registry` docs:

> Looking up, dispatching and registering are efficient and immediate at the cost of delayed unsubscription. For example, if a process crashes, its keys are automatically removed from the registry but the change may not propagate immediately. 

and this note in the Phoenix `Channel` docs:

> Generally speaking, if you want to clean something up, it is better to monitor your channel process and do the clean up from another process.
> Consider `Process.monitor(self)` on subscribe instead

And through a lot of trial and error, I eventually realized that the calling process was what's being somewhat magically registered every time `PredictionsPubSub.subscribe/1` is called. So ultimately I decided to use `Process.monitor/1` on it, and added some functionality to respond to the calling process being terminated.
* Checks the subscribers registry to see if any other subscribers remain for the given key
* If not, then `DynamicSupervisor.terminate_child/2` it

In the meantime, I also refactored `Predictions.StreamSupervisor` slightly. Instead of starting three interconnected (via `GenStage`) child processes for each key, it wraps the children in a new supervising worker process (`Predictions.StreamSupervisor.Worker`) that can more cleanly be started/stopped. It also uses a `:via` tuple for the worker name, so that these workers can automatically be registered/unregistered to the registry tracking streams by key, instead of manually registering the `pid` of the process running the `ServerSentEventStage` stream.

---

#### General checks

### DIFF
--- a/apps/predictions/lib/predictions_pub_sub.ex
+++ b/apps/predictions/lib/predictions_pub_sub.ex
@@ -100,7 +100,7 @@ defmodule Predictions.PredictionsPubSub do
     pattern = {registry_key, :"$2", key}
 
     guards = [
-      {:"=/=", :"$2", pid_to_omit},
+      {:"=/=", :"$2", pid_to_omit}
     ]
 
     body = [true]
--- a/apps/predictions/lib/predictions_pub_sub.ex
+++ b/apps/predictions/lib/predictions_pub_sub.ex
@@ -11,6 +11,7 @@ defmodule Predictions.PredictionsPubSub do
 
   @valid_filters ["route", "stop", "direction", "trip"]
   @broadcast_interval_ms Application.compile_env!(:predictions, [:broadcast_interval_ms])
+  @subscribers :prediction_subscriptions_registry
 
   @type broadcast_message :: {:new_predictions, [Prediction.t()]}
 
@@ -33,7 +34,7 @@ defmodule Predictions.PredictionsPubSub do
   def subscribe(key, server \\ __MODULE__) do
     StreamSupervisor.ensure_stream_is_started(key)
     {registry_key, predictions} = GenServer.call(server, {:subscribe, key})
-    Registry.register(:prediction_subscriptions_registry, registry_key, key)
+    Registry.register(@subscribers, registry_key, key)
     predictions
   end
 
@@ -48,17 +49,36 @@ defmodule Predictions.PredictionsPubSub do
   end
 
   @impl GenServer
-  def handle_call({:subscribe, keys}, _from, state) do
+  def handle_call({:subscribe, keys}, {from_pid, _ref}, state) do
     registry_key = self()
+
+    # Let us detect when the calling process goes down, and save the associated
+    # keys for easier lookup
+    Process.monitor(from_pid)
+    new_state = Map.put_new(state, from_pid, keys)
+
     predictions = keys |> table_keys() |> Store.fetch()
-    {:reply, {registry_key, predictions}, state}
+    {:reply, {registry_key, predictions}, new_state}
+  end
+
+  @impl GenServer
+  def handle_cast({:stop_stream, key}, state) do
+    # By this point, the caller_pid is no longer alive or registered/subscribed.
+    # Here we can check if there are other subscribers for the associated key.
+    # If there are no other subscribers remaining, we stop the associated
+    # predictions data stream.
+    if Registry.count_match(@subscribers, self(), key) == 0 do
+      StreamSupervisor.stop_stream(key)
+    end
+
+    {:noreply, state}
   end
 
   @impl GenServer
   def handle_info(:broadcast, state) do
     registry_key = self()
 
-    Registry.dispatch(:prediction_subscriptions_registry, registry_key, fn entries ->
+    Registry.dispatch(@subscribers, registry_key, fn entries ->
       Enum.each(entries, &send_data(&1))
     end)
 
@@ -69,6 +89,16 @@ defmodule Predictions.PredictionsPubSub do
   def handle_info({event, predictions}, state) do
     :ok = Store.update(event, predictions)
     {:noreply, state}
+  end
+
+  def handle_info(
+        {:DOWN, parent_ref, :process, caller_pid, _reason},
+        state
+      ) do
+    Process.demonitor(parent_ref, [:flush])
+    {key, new_state} = Map.pop(state, caller_pid)
+    GenServer.cast(__MODULE__, {:stop_stream, key})
+    {:noreply, new_state}
   end
 
   @spec table_keys(String.t()) :: [
--- a/apps/predictions/lib/predictions_pub_sub.ex
+++ b/apps/predictions/lib/predictions_pub_sub.ex
@@ -62,19 +62,6 @@ defmodule Predictions.PredictionsPubSub do
   end
 
   @impl GenServer
-  def handle_cast({:stop_stream, key}, state) do
-    # By this point, the caller_pid is no longer alive or registered/subscribed.
-    # Here we can check if there are other subscribers for the associated key.
-    # If there are no other subscribers remaining, we stop the associated
-    # predictions data stream.
-    if Registry.count_match(@subscribers, self(), key) == 0 do
-      StreamSupervisor.stop_stream(key)
-    end
-
-    {:noreply, state}
-  end
-
-  @impl GenServer
   def handle_info(:broadcast, state) do
     registry_key = self()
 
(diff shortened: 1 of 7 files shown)


apps/predictions/test/predictions_pub_sub_test.exs
```diff
@@ -110,4 +110,61 @@ defmodule Predictions.PredictionsPubSubTest do
...
+      # The caller process is removed from the state
+      assert task not in Map.keys(:sys.get_state(pid))
+
+      assert %{
+               ^p1 => "stop=1",
+               ^p2 => "stop=2",
+               ^p3 => "stop=3"
+             } = :sys.get_state(pid)
+    end
+  end
```
paulswartz, line 169: question: is there a way to test this that's looking at the externally visible behavior? something like:
- client A registers, a new worker is created
- client B registers, no workers are created
- client A disconnects, worker is still alive
- client B disconnects, worker is terminated

paulswartz (approved):
I tried this locally, and I can verify that it's terminating the prediction stream when the websocket disconnects. If anything, it's too aggressive; Firefox seems to disconnect the websocket if I switch tabs, but that seems outside the scope of this PR.

paulswartz:
suggestion: it might be helpful to periodically log the open predictions, and the number of clients?

apps/predictions/lib/stream_supervisor.ex
```diff
@@ -24,81 +26,47 @@ defmodule Predictions.StreamSupervisor do
...
+      stream_pid ->
+        {:ok, stream_pid}
     end
   end
 
   @spec lookup(String.t()) :: pid() | nil
   defp lookup(key) do
-    case Registry.lookup(:prediction_streams_registry, key) do
-      [{_pid, sses_pid}] ->
-        if Process.alive?(sses_pid), do: sses_pid
```
arkadyan, line 36: question: Is this `Process.alive?` check no longer needed?

My memory is shaky, so this might all be wrong, but here's the situation I thought this was covering: 

A opens a socket and needs a new stream, so we start one and register it. Then B opens a socket for the same route, and we can look up the existing stream in the registry, and B can use that. Then, A and B both close their sockets, so we no longer need the stream, and it can be closed. Then C opens a socket, we go to the registry lookup and find the existing stream pid, but we don't want to actually give that to C because it no longer actually exists.

Am I getting that wrong, and/or is something else covering the duties of what `Process.alive?` was previously doing?

arkadyan (commented):
I just had one additional question, but thanks for the helpful writeup and the refactor --> feature commit structure.

apps/predictions/lib/stream_supervisor.ex
arkadyan: In any case, a test for getting a new stream spun up after a previous one was stopped might not be a bad addition.

thecristen: I'm still thinking on this one and will add a test.

My impression is that, now that the stream is wrapped up in a supervisor process that is now tied to the registry via `:via`, that process + registration now live/die together. This change also means the pid being registered/retrieved here is no longer the SSES one directly, but the parent supervisor one.

Now of course, if that child SSES process happens to die for whatever reason, it should be restarted by that supervisor... and I guess if _that_ fails, there are bigger problems to catch here (that maybe I should be checking for here?) 🤔

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -33,7 +34,7 @@ defmodule Predictions.PredictionsPubSub do
   def subscribe(key, server \\ __MODULE__) do
     StreamSupervisor.ensure_stream_is_started(key)
     {registry_key, predictions} = GenServer.call(server, {:subscribe, key})
-    Registry.register(:prediction_subscriptions_registry, registry_key, key)
+    Registry.register(@subscribers, registry_key, key)
```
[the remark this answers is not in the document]
thecristen: It's true, this will return predictions before a newly-started stream has a chance to populate the table.
 
This whole sequence is a weird thing; registering, starting the stream (which will populate the table in `Predictions.Store`) and retrieving predictions (queries the current state of that table), are all independent enough things that probably look more interconnected in these few lines of code than they actually are :D 

It is a bit noticeable on the frontend, but I think I will leave a more thoughtful implementation of this for a future PR.

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -71,6 +91,16 @@ defmodule Predictions.PredictionsPubSub do
     {:noreply, state}
   end
 
+  def handle_info(
+        {:DOWN, parent_ref, :process, caller_pid, _reason},
+        state
+      ) do
+    Process.demonitor(parent_ref, [:flush])
+    {key, new_state} = Map.pop(state, caller_pid)
+    GenServer.cast(__MODULE__, {:stop_stream, key})
```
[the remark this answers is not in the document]
thecristen: Nah, will remove

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -48,17 +49,36 @@ defmodule Predictions.PredictionsPubSub do
...
+    new_state = Map.put_new(state, from_pid, keys)
+
     predictions = keys |> table_keys() |> Store.fetch()
-    {:reply, {registry_key, predictions}, state}
+    {:reply, {registry_key, predictions}, new_state}
+  end
+
+  @impl GenServer
+  def handle_cast({:stop_stream, key}, state) do
+    # By this point, the caller_pid is no longer alive or registered/subscribed.
```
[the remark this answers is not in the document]
thecristen: not enough certainty on this... I ended up refactoring in https://github.com/mbta/dotcom/pull/1725/commits/1df42d8b552916dee74f6b71d728bb8c55873d93, to explicitly search for registrations that don't match the `caller_pid`.

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -71,6 +78,37 @@ defmodule Predictions.PredictionsPubSub do
...
+
+    guards = [
+      {:==, :"$1", registry_key},
+      {:"=/=", :"$2", pid_to_omit},
+      {:==, :"$3", key}
+    ]
+
+    body = [{{:"$2", :"$3"}}]
+    Registry.select(@subscribers, [{pattern, guards, body}])
+  end
```
paulswartz, line 110: suggestion(non-blocking):
```suggestion
  defp no_other_subscribers?(key, pid_to_omit) do
    registry_key = self()
    pattern = {registry_key, :"$2", key}

    guards = [
      {:"=/=", :"$2", pid_to_omit},
    ]

    body = [true]
    Registry.select(@subscribers, [{pattern, guards, body}]) == []
  end
```

Since you don't care about the specific values from the Registry, you can:
- move registry and key matches into the pattern
- return a random value as the body, rather than any particular values

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -71,6 +78,37 @@ defmodule Predictions.PredictionsPubSub do
...
+  def handle_info(
+        {:DOWN, parent_ref, :process, caller_pid, _reason},
+        state
+      ) do
+    Process.demonitor(parent_ref, [:flush])
+    {key, new_state} = Map.pop(state, caller_pid)
+    # Here we can check if there are other subscribers for the associated key.
+    # If there are no other subscribers remaining, we stop the associated
+    # predictions data stream.
+    if other_subscribers(key, caller_pid) == [] do
```
thecristen, line 90:
```suggestion
    if no_other_subscribers?(key, caller_pid) do
```

apps/predictions/test/predictions_pub_sub_test.exs
thecristen: @paulswartz That works. Added some tests that sometimes work. 

They're flaky -- how might they be able to be improved? I ended up structuring that way because if I register multiple subscribers from the same process, and then exit one subscriber, it'll exit all the other subscribers and leave me unable to verify my hypothesis. But at the same time I'm not sure how to create multiple calling processes that's kept alive for the duration of the test  / how to shut down individual calling processes in the best manner. 🤔

paulswartz: [Task](https://hexdocs.pm/elixir/Task.html) and [Agent](https://hexdocs.pm/elixir/Agent.html) are my go-tos for easily making separate processes. (Agent is better if you need to make calls to the separate process in between other work).

PR #1738, opened 2023-09-05
refactor: remove premature optimization and stop leaving joined channels

thecristen:
> [!Note]
> This PR can be previewed on **dev-green** with other changes from #1736 

#### Summary of changes

**Asana Ticket:** [Investigate why multiple API connections are being open for a given stop](https://app.asana.com/0/385363666817452/1205293745657618/f)

This PR removes the following "features" that were intended to avoid sending too much data to the client:
- Triggering channel joining/leaving when the page visibility changes
- Leaving channels on `beforepageunload`

My observation was that, with this PR combined with upgrading Phoenix, the connections to the streaming predictions are maintained with the open V3 API streams as long as the server is open, instead of being closed & re-started (under a new process) many times over a session. Additionally, thanks to the Phoenix upgrade, I no longer observe the socket being closed after a stop page is left in a background tab for extended time (!).

**Won't we send too much data to the client then?** Maybe, but I'd rather try this, and if we find problems associated with sending too much data to the client, then we can figure out a different way to tackle that.

---

#### General checks


paulswartz (approved):

kotva006 (approved):
👍

PR #1739, opened 2023-09-05
refactor: connect to predictions stream using route+direction

thecristen:
#### Summary of changes

**Asana Ticket:** [Refactor the predictions stream to always connect to predictions for a route+direction instead of allowing connections per-stop](https://app.asana.com/0/385363666817452/1205293745657620/f)

`SiteWeb.PredictionsChannel` defines a topic for a stop, and a new abstraction, `Predictions.StreamTopic`, translates this into one or more route/direction combinations, in accordance to the route patterns in service for each stop.

Each route/direction is associated with a distinct V3 API stream. Each stream is registered separately, so that we can handle leaving topics/closing streams in a granular way. For example, if both `stop:place-state` and `stop:place-wondl` are joined, then `stop:place-state` is left, the streams associated with predictions for the Orange line would be closed, but the ones for the Blue line would remain open.

---

#### General checks

### DIFF
--- a/apps/predictions/lib/predictions_pub_sub.ex
+++ b/apps/predictions/lib/predictions_pub_sub.ex
@@ -7,12 +7,12 @@ defmodule Predictions.PredictionsPubSub do
 
   use GenServer
 
-  alias Predictions.{Prediction, Store, StreamSupervisor}
+  alias Predictions.{Prediction, Store, StreamTopic, StreamSupervisor}
 
-  @valid_filters ["route", "stop", "direction", "trip"]
   @broadcast_interval_ms Application.compile_env!(:predictions, [:broadcast_interval_ms])
   @subscribers :prediction_subscriptions_registry
 
+  @type registry_value :: {Store.fetch_keys(), binary()}
   @type broadcast_message :: {:new_predictions, [Prediction.t()]}
 
   # Client
@@ -30,12 +30,23 @@ defmodule Predictions.PredictionsPubSub do
   end
 
   @spec subscribe(String.t()) :: [Prediction.t()]
-  @spec subscribe(String.t(), GenServer.server()) :: [Prediction.t()]
-  def subscribe(key, server \\ __MODULE__) do
-    StreamSupervisor.ensure_stream_is_started(key)
-    {registry_key, predictions} = GenServer.call(server, {:subscribe, key})
-    Registry.register(@subscribers, registry_key, key)
-    predictions
+  @spec subscribe(String.t(), GenServer.server()) :: [Prediction.t()] | {:error, term()}
+  def subscribe(topic, server \\ __MODULE__) do
+    with %StreamTopic{streams: streams, fetch_keys: fetch_keys} = stream_filter <-
+           StreamTopic.new(topic) do
+      Enum.each(streams, &StreamSupervisor.ensure_stream_is_started/1)
+
+      {registry_key, predictions} = GenServer.call(server, {:subscribe, stream_filter})
+
+      streams
+      |> Enum.map(&{fetch_keys, &1})
+      |> Enum.each(&Registry.register(@subscribers, registry_key, &1))
+
+      predictions
+    else
+      {:error, _} = error ->
+        error
+    end
   end
 
   # Server
@@ -49,15 +60,23 @@ defmodule Predictions.PredictionsPubSub do
   end
 
   @impl GenServer
-  def handle_call({:subscribe, keys}, {from_pid, _ref}, state) do
+  def handle_call(
+        {:subscribe, stream_filter},
+        {from_pid, _ref},
+        state
+      ) do
     registry_key = self()
 
     # Let us detect when the calling process goes down, and save the associated
-    # keys for easier lookup
+    # API params for easier lookup
     Process.monitor(from_pid)
-    new_state = Map.put_new(state, from_pid, keys)
+    new_state = Map.put_new(state, from_pid, stream_filter)
 
-    predictions = keys |> table_keys() |> Store.fetch()
+    predictions =
+      stream_filter
+      |> Map.get(:fetch_keys)
+      |> Store.fetch()
+
     {:reply, {registry_key, predictions}, new_state}
   end
 
@@ -66,7 +85,11 @@ defmodule Predictions.PredictionsPubSub do
     registry_key = self()
 
     Registry.dispatch(@subscribers, registry_key, fn entries ->
-      Enum.each(entries, &send_data(&1))
+      entries
+      |> Enum.uniq_by(fn {pid, {fetch_keys, _}} ->
+        {pid, fetch_keys}
+      end)
+      |> Enum.each(&send_data/1)
     end)
 
     broadcast_timer(@broadcast_interval_ms)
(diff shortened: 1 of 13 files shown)


apps/predictions/lib/stream_topic.ex
```diff
@@ -0,0 +1,67 @@
...
+
+  @type api_filter_t :: %{
+          required(:route) => Route.id_t(),
+          required(:direction) => 0 | 1
+        }
+
+  @spec new(String.t()) :: t() | {:error, term()}
+  def new(topic) do
+    with "stop:" <> stop_id when stop_id != "" <- topic,
+         filters when filters != [] <- filters(topic) do
```
paulswartz, line 38: suggestion: you could pass `stop_id` in directly since you already matched it on line 37
```suggestion
         filters when filters != [] <- filters(stop_id) do
```
or alternatively, pull this into the function head:
```elixir
def new(topic)
def new("stop:" <> stop_id) when stop_id != "" do
  filters = stop_id |> RoutePatterns.Repo.by_stop_id() |> ...
  ...
end
def new(_) do
  {:error, :unsupported_topic}
end
```

apps/predictions/lib/stream_topic.ex
```diff
@@ -0,0 +1,67 @@
...
+        direction: &1.direction_id
+      }
+    )
+    |> Enum.uniq()
+  end
+
+  @spec filter_to_string(api_filter_t()) :: String.t()
+  defp filter_to_string(%{route: r, direction: d}) do
+    "filter[route]=#{r}&filter[direction_id]=#{d}"
+  end
```
paulswartz, line 66: suggestion(non-blocking):
```suggestion
  defp filter_to_string(filters) do
    filters
    |> Enum.map(fn {filter, value} -> "filter[#{filter}]=#{value}")
    |> Enum.join("&")
  end
```

apps/predictions/test/predictions_pub_sub_test.exs
```diff
@@ -171,6 +170,7 @@ defmodule Predictions.PredictionsPubSubTest do
   end
 
   describe "handle_info/2 - :DOWN" do
+    @describetag :skip
```
paulswartz, line 173: praise: TIL!

question: but should this still be here?

apps/predictions/lib/stream_worker.ex
```diff
@@ -18,44 +17,38 @@ defmodule Predictions.StreamSupervisor.Worker do
...
-  defp sses_opts(key) do
-    filters =
-      table_keys(key)
-      |> Enum.map(fn {k, v} -> "filter[#{k}]=#{v}&" end)
-      |> Enum.join()
-
+  defp sses_opts(filters) do
     path =
-      "/predictions?#{filters}fields[prediction]=status,departure_time,arrival_time,direction_id,schedule_relationship,stop_sequence&include=route,trip,trip.occupancies,stop&fields[route]=long_name,short_name,type&fields[trip]=direction_id,headsign,name,bikes_allowed&fields[stop]=platform_code"
+      "/predictions?#{filters}&fields[prediction]=status,departure_time,arrival_time,direction_id,schedule_relationship,stop_sequence&include=route,trip,trip.occupancies,stop&fields[route]=long_name,short_name,type&fields[trip]=direction_id,headsign,name,bikes_allowed&fields[stop]=platform_code"
```
paulswartz, line 40: question: while we're in here, are the `include` values here actually being used?

apps/predictions/lib/stream_worker.ex
```diff
@@ -18,44 +17,38 @@ defmodule Predictions.StreamSupervisor.Worker do
...
-  def init(key) do
-    sses_stream_name = sses_stream_name(key)
-    api_stream_name = :"predictions_api_stream_#{key}"
-    prediction_stream_name = :"predictions_data_stream_#{key}"
+  @doc "The `filters` argument is a string representing the filter parameters of
+  the resultant API call e.g. 'filter[route]=CR-Foxboro&filter[direction_id]=0'"
+  def init(filters) do
+    sses_stream_name = sses_stream_name(filters)
+    api_stream_name = :"predictions_api_stream_#{filters}"
+    prediction_stream_name = :"predictions_data_stream_#{filters}"
```
paulswartz, line 25: question(blocking): are these filters validated before they're used here? Since both of these items are creating new atoms, we need to be careful that users/attackers can't generate whichever ones they want.

apps/predictions/lib/predictions_pub_sub.ex
```diff
@@ -30,12 +30,23 @@ defmodule Predictions.PredictionsPubSub do
...
-  def subscribe(key, server \\ __MODULE__) do
-    StreamSupervisor.ensure_stream_is_started(key)
-    {registry_key, predictions} = GenServer.call(server, {:subscribe, key})
-    Registry.register(@subscribers, registry_key, key)
-    predictions
+  @spec subscribe(String.t(), GenServer.server()) :: [Prediction.t()] | {:error, term()}
+  def subscribe(topic, server \\ __MODULE__) do
+    with %StreamTopic{streams: streams, fetch_keys: fetch_keys} = stream_filter <-
+           StreamTopic.new(topic) do
+      Enum.each(streams, &StreamSupervisor.ensure_stream_is_started/1)
```
paulswartz, line 37: suggestion: rather than having this module pull the contents out of `StreamTopic`, what about making this (and other functionality which works with data from this struct) a function in the `StreamTopic` module?

apps/route_patterns/lib/repo.ex
```diff
@@ -43,6 +43,12 @@ defmodule RoutePatterns.Repo do
     |> Enum.sort(&reorder_mrts(&1, &2, route_id))
   end
 
+  def by_stop_id(stop_id, opts) do
```
paulswartz, line 46: suggestion: I don't think `opts` is used, so you could remove it:
```suggestion
  def by_stop_id(stop_id) do
```

apps/predictions/lib/stream_worker.ex
thecristen: clarification: are we concerned about new attack atoms being part of these names, or getting into the `/predictions?#{filters}` request? (or both for separate reasons?)

paulswartz: I'm thinking about whether clients can subscribe to arbitrary filters. Looking back at the code though, I think that the functionality works like this:
- client goes to stops page
- client subscribes to `predictions:stop:<id>`
- server looks up the route patterns for the routes to subscribe to
- stream is created for the given routes

As long as it's not possible for clients to subscribe to an arbitrary client-controlled route, then this should be okay. But it's something to keep an eye on.

thecristen: Yeah, part of StreamTopic's purpose is absolutely to guard against that sorta outcome, and the tests should demonstrate that to an extent. I think it'll be ok, and I can add a few more tests to validate.

apps/predictions/lib/stream_topic.ex
thecristen: I am also trying to leave it structured so that we could expand it later (say, if we start to enable subscribing to prediction streams by a trip ID, or a route ID, etc.). I did end up adjusting this slightly.

paulswartz (approved):

PR #1751, opened 2023-09-19
feat(StopPageRedesign): use URL search parameters for interacting with the realtime tracking view

thecristen:
#### Summary of changes

**Asana Ticket:** [Navigation | Browser back button takes you two pages back](https://app.asana.com/0/385363666817452/1205175614190102/f)

On dev-green for review.

## Highlights

- This PR introduces [React Router](https://reactrouter.com/en/main) to our codebase. 
  - I chose this library because it's well known and popular! And has clean built-in functionality for [mucking about with URL query strings](https://reactrouter.com/en/main/hooks/use-search-params). In my opinion, it's a _little_ overkill for the stop page (there aren't lots of routes involved here), but the end result is deliciously performant, and seems to work perfectly well with the <kbd>Back</kbd>/<kbd>Forward</kbd> buttons
    - (once I disabled Turbolinks on every contained `<a>`, at least).
  - I tried implementing this entirely with our already-in-use `use-query-params` dependency, which is what we use on the schedules page to synchronize the URL to the chosen direction and chosen schedule finder origin and direction. The dependency was pretty out of date, and using it exposed some buggy behavior, and transitions were really slow to render. I also tried updating the dependency, but there were lots of breaking changes requiring more refactoring on our part, and even then it didn't work great. So perhaps React Router might replace `use-query-params` in the future.
- Lots of test refactoring purely to account for needing a router component wrapping any component using a React Router feature
- Extracted out the logic for setting/unsetting the `departureFilters` into a custom hook, also...
  - Renaming the concept to `departureRow`, which I'm hoping is a bit more descriptive?
  - Including the logic needed to set and unset the corresponding URL search params.
- The arbitrarily chosen params are in the format of `?route=<routeId>&direction_id=<directionId>&headsign=<headsign>`... open to other suggestions.

---

#### General checks

* [x] **Are interactive elements accessible?** This includes at minimum having [relevant keyboard interactions](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#keyboard-interaction) and [visible focus](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#visible-focus), but can also include verification with screen reader testing.


apps/site/assets/package.json
```diff
@@ -37,6 +37,7 @@
     "react-dom": "^16.8.6",
     "react-leaflet": "^2.7.0",
     "react-redux": "^7.2.0",
+    "react-router-dom": "^6.16.0",
```
kotva006, line 40: ❤️

apps/site/assets/ts/hooks/useDepartureRow.ts
```diff
@@ -0,0 +1,99 @@
...
+} => {
+  const [params, setSearchParams] = useSearchParams();
+
+  const activeRow = useMemo(() => {
+    const { routeId, directionId, headsign } = getRowParams(params);
+    const matchedRoute = routes.find(r => r.id === routeId);
+    if (
+      some([routeId, directionId, headsign], isNil) ||
+      !matchedRoute ||
+      !(directionId! in ["0", "1"])
```
kotva006, line 69: nit-pick(nonblocking): Instead of telling the typer that `directionId` exists we could just add a check for it `..|| !directionId ...`.  Would also be a bit clearer

thecristen: @kotva006 wouldn't that evaluate `false` when `directionId` is `0`? which is not what I'd want, since 0 is a valid value for it

apps/site/assets/ts/hooks/usePredictionsChannel.ts
```diff
@@ -91,7 +91,7 @@ const usePredictionsChannel = (
     PredictionWithTimestamp[] | null,
     ChannelPredictionResponse
   > = (_oldGroupedPredictions, detail) => {
-    if (detail.error) {
+    if (detail.error || !detail.predictions) {
```
thecristen, line 94: Disregard, I ended up separating this into a hotfix. But noticed it while working on this because the router's very unforgiving of errors...

apps/site/assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -139,23 +124,20 @@ const DeparturesAndMap = ({
...
-  const unsetDepartureInfo = (): void => setDepartureFilters(null);
-
   const BackToRoutes = (
     <div className="back-to-routes">
       <button
         className="btn"
-        onClick={unsetDepartureInfo}
-        onKeyDown={unsetDepartureInfo}
+        onClick={() => resetRow()}
+        onKeyDown={() => resetRow()}
```
kotva006, line 140: nit-pick(non-blocking): I think this could just be `{resetRow}`

apps/site/assets/ts/hooks/useDepartureRow.ts
kotva006: I forgot about that quirk.  Maybe [isNil](https://lodash.com/docs/4.17.15#isNil) could help here.

kotva006 (approved):
Overall looks good!  Glad to have this dependency.  I think it could help a lot with consolidating the pages on dotcom

apps/site/assets/ts/hooks/useDepartureRow.ts
thecristen: yeah, I kinda went overboard and decided to do this to be strict, so if someone tries to go to an otherwise correct URL but with `direction_id=3` or something, it'll redirect to the main view instead of a broken realtime tracking view :grin:

PR #1754, opened 2023-09-21
feat(StopPageRedesign): add an error boundary

thecristen:
#### Summary of changes

**Asana Ticket:** No ticket

This PR adds an error boundary, because React Router highly recommends it, and their default display is super ugly! I used Sentry's built-in `ErrorBoundary` component for this, which has the bonus of automagically sending the error to Sentry.

### Highlights

- As the fallback display, I wrote a new React component, `ErrorPage`, that's meant to imitate our current error page.

image

> [!Note]
> There are some small deviations from our normal error pages:

- Our error pages typically show an error code at the top, but we don't technically know the error code as this error boundary could catch anything, so I've omitted that part.
- I couldn't use our current blinky bus SVG image because those have the error codes written on them. See?
image
But I found a cute GIF in our `/images/` folder with the more cartoony, less-specified blinky bus.

### Bonus change

- Removed a few files from being counted in the Javascript test coverage.

### DIFF
--- a/apps/site/assets/ts/components/ErrorPage.tsx
+++ b/apps/site/assets/ts/components/ErrorPage.tsx
@@ -1,0 +1,37 @@
+import React from "react";
+import { FallbackRender } from "@sentry/react";
+
+// Adapted from the templates in SiteWeb.ErrorView
+// To be used in an ErrorBoundary
+const ErrorPage: FallbackRender = errorData => {
+  // eslint-disable-next-line no-console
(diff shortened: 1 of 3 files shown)


kotva006 (approved):
Looks good
Also if we upgrade react we might be able to get the error code using something like [this](https://docs.sentry.io/platforms/javascript/guides/react/features/error-boundary/#linked-errors)

PR #1755, opened 2023-09-21
refactor(RoutePattern): get `stop_ids` from trip ++ related improvements

thecristen:
#### Summary of changes

The line diagrams for all rail routes now display stops from the canonical route patterns! Because bus routes and ferry routes have no canonical information for the moment, those remain unchanged.

**Asana Ticket:** [Update the line diagram to use canonical route information](https://app.asana.com/0/385363666817452/1205263426482070/f), part of but not all of  [Actually finish removing the dependency on shape.stop_ids](https://app.asana.com/0/1203354074531980/1199671224586821/f)

So... in support of [Stops showing Headsign for Route Even if One Direction of Travel](https://app.asana.com/0/385363666817452/1205268760549362/f) I figured out how to parse the route pattern's stops directly **from the representative trip, as opposed to from the related _shape_** (which is technically deprecated and slated to be removed in [Actually finish removing the dependency on shape.stop_ids](https://app.asana.com/0/1203354074531980/1199671224586821/f)).

Having this in hand, I realized it was relatively easy to adapt the line diagram code to use these corrected route pattern `stop_ids`, so that's what this PR does. Notably, the corrected `stop_ids` are mapped to child stops instead of parent stops, so some adjustments elsewhere were needed to account for that.

I also noticed a comment in the timetable code pertaining to canonical routes, so was able to address that here too.

---

#### General checks

### DIFF
--- a/apps/route_patterns/lib/repo.ex
+++ b/apps/route_patterns/lib/repo.ex
@@ -38,7 +38,6 @@ defmodule RoutePatterns.Repo do
     opts
     |> Keyword.put(:route, route_id)
     |> Keyword.put(:sort, "typicality,sort_order")
-    |> Keyword.put(:include, "representative_trip.shape")
     |> cache(&api_all/1)
     |> Enum.sort(&reorder_mrts(&1, &2, route_id))
   end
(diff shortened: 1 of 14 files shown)


kotva006 (approved):
👍

PR #1759, opened 2023-09-26
refactor(StopPageRedesign): use route patterns

thecristen:
#### Summary of changes

**Asana Ticket:** [Stops showing Headsign for Route Even if One Direction of Travel](https://app.asana.com/0/385363666817452/1205268760549362/f)

## Backend highlights
- Adds `/api/stop/:stop_id/route-patterns` to get a JSON object of augmented route patterns grouped by headsign and route ID.
- Replaces the `/routes/by-stop/:stop_id` with `/routes/:route_ids` that returns a JSON response containing a simple list of routes.
- The polyline information continues to be computed in the backend, but is now part of the route pattern response as opposed to appending the routes response.

## Frontend highlights
- Adds a `loader` to the React Router's configuration to fetch the route patterns, which will be available from within child components via React Router's `useLoaderData()` hook. Disables revalidation because this data is not going to change.
- Adds a helper function, `departureInfoInRoutePatterns`, to filter the list of departures against a list of route patterns.
- Lots of restructuring:
  - `<StopPageRedesign />`
    - Get route patterns, and use the route IDs from that to fetch routes
    - Don't wait for routes to return a defined response, since they never will for stops that don't have any route patterns (South Attleboro)
  - `<DeparturesAndMap />`
    - Remove no longer needed `routesWithPolylines` prop
    - Get route patterns, use the included polylines
  - `<StopPageDepartures />`
    - Get route patterns, figure out route patterns for the filtered routes
  - `<DepartureCard />`
    - Add `routePatternsByHeadsign` prop
    - Refactors to render a `<DepartureTimes />` per each headsign
    - Sets an `onClick` variable to pass to child
  - `<DepartureTimes />`
    - Adds `headsign`, `isCR`, and `onClick` props
    - Removes check of stop name against route's direction_destination as it should no longer be needed. Removes no longer needed `route`, `directionId` and `stopName` props.
    - Renames `departuresForDirection` prop to just `departures` (they were filtered by `departureInfoInRoutePatterns` earlier)
---

#### General checks

### DIFF
--- a/apps/site/lib/site_web/controllers/route_controller.ex
+++ b/apps/site/lib/site_web/controllers/route_controller.ex
@@ -3,27 +3,17 @@ defmodule SiteWeb.RouteController do
   Endpoints for getting route data.
   """
   use SiteWeb, :controller
-  alias Leaflet.MapData.Polyline
   alias Routes.{Repo, Route}
 
-  @spec get_by_stop_id(Plug.Conn.t(), map) :: Plug.Conn.t()
-  def get_by_stop_id(conn, %{"stop_id" => stop_id} = _params) do
-    routesWithPolylines =
-      stop_id
-      |> Repo.by_stop()
-      |> Enum.map(fn route ->
-        route
-        |> Route.to_json_safe()
-        |> Map.put(:polylines, route_polylines(route, stop_id))
-      end)
+  @spec get(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def get(conn, %{"route_ids" => route_ids} = _params) do
+    routes =
+      route_ids
+      |> String.split(",")
+      |> Stream.map(&Repo.get/1)
+      |> Stream.reject(&is_nil/1)
+      |> Enum.map(&Route.to_json_safe/1)
 
-    json(conn, routesWithPolylines)
-  end
-
-  defp route_polylines(route, stop_id) do
-    route.id
-    |> RoutePatterns.Repo.by_route_id(stop: stop_id, include: "representative_trip.shape")
-    |> Enum.filter(&(!is_nil(&1.representative_trip_polyline)))
-    |> Enum.map(&Polyline.new(&1, color: "#" <> route.color, weight: 4))
+    json(conn, routes)
   end
 end
--- a/apps/site/lib/site_web/controllers/route_controller.ex
+++ b/apps/site/lib/site_web/controllers/route_controller.ex
@@ -5,8 +5,8 @@ defmodule SiteWeb.RouteController do
   use SiteWeb, :controller
   alias Routes.{Repo, Route}
 
-  @spec get(Plug.Conn.t(), map) :: Plug.Conn.t()
-  def get(conn, %{"route_ids" => route_ids} = _params) do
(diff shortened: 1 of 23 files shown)


apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -177,6 +179,44 @@ defmodule SiteWeb.StopController do
...
+  def grouped_route_patterns(conn, %{"id" => stop_id}) do
+    json(conn, route_patterns_by_route_and_headsign(stop_id))
+  end
+
+  @type by_route_and_headsign :: %{Route.id_t() => %{String.t() => [RoutePattern.t()]}}
+  @spec route_patterns_by_route_and_headsign(Stop.id_t()) :: by_route_and_headsign()
+  defp route_patterns_by_route_and_headsign(stop_id) do
+    stop_id
+    |> RoutePatterns.Repo.by_stop_id()
+    |> Enum.reject(&ends_at?(&1, stop_id))
```
i0sea, line 192: I believe there was a check for "is at last stop" on the JS side. Has that been removed since it's happening here?

apps/site/lib/site_web/controllers/route_controller.ex
```diff
@@ -3,27 +3,17 @@ defmodule SiteWeb.RouteController do
...
-    routesWithPolylines =
-      stop_id
-      |> Repo.by_stop()
-      |> Enum.map(fn route ->
-        route
-        |> Route.to_json_safe()
-        |> Map.put(:polylines, route_polylines(route, stop_id))
-      end)
+  @spec get(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def get(conn, %{"route_ids" => route_ids} = _params) do
```
i0sea, line 9: nitpick: rename to `get_by_route_ids` for clarity

apps/site/lib/site_web/controllers/stop_controller.ex
thecristen: @i0sea Yep, it's been removed!

i0sea (commented):
Not sure if this was occuring before these changes but am noticing routes that never stop at a given stop being listed with a headsign -> no upcoming trips. Example the 32 never stops [here](https://dev-blue.mbtace.com/stops/383) but is listed anyway. The oneway headsigns do appear to be dropped

i0sea (approved):
Edit: noticing above issue on dev green as well so unrelated to these changes 🚀

thecristen:
@i0sea good looking out -- looks like the 32 does stop there, but under [one of the uncommon route variants](https://dev-blue.mbtace.com/schedules/32/line?schedule_direction%5Bdirection_id%5D=1&schedule_direction%5Bvariant%5D=22-_-1).

Can also see a [couple schedules from this route at that stop here](https://dev-blue.mbtace.com/schedules/32/line?schedule_direction%5Bdirection_id%5D=1&schedule_direction%5Bvariant%5D=32-9-1&schedule_finder%5Bdirection_id%5D=1&schedule_finder%5Borigin%5D=383):
image

PR #1760, opened 2023-09-27
hotfix(ScheduleDirection): fix line diagram for routes with repeated stops

thecristen:
No ticket - fixing a bug on dev with the 217 & 120 bus route schedule pages.

## Background

- #1755 

The recent PR updated how stops were being fetched for the line diagram, but came with corrected data that broke the line diagram code on at least two pages. It turns out the recursive code that parses the stops to draw the branches on the line diagram assumes that each stop ID is unique, which is assuredly not the case for the 217 and 120 routes, each of which repeat at least one stop along its route.

## Highlighted changes

The approach I took here was to just not compute the stop tree for pages for a ferry or bus route. We don't need it - these either don't have line diagrams or have simple lines that won't branch.

### Backend

- Consolidated the two places that were computing a stop tree into one function
- Adjusted that function to return different data depending on the route type - allowing us to send `nil` stop tree 

### Frontend

- Refactored many components to work with a `null` stop tree.
- Use the list of stops from the backend if the stop tree is `null`

#### Fixed 217 route

image

---

#### General checks

### DIFF
--- a/apps/site/assets/ts/schedule/components/line-diagram/graphics/Diagram.tsx
+++ b/apps/site/assets/ts/schedule/components/line-diagram/graphics/Diagram.tsx
@@ -13,7 +13,12 @@ import {
 } from "../../../../helpers/stop-tree";
 import { isAGreenLineRoute } from "../../../../models/route";
 import { Alert, DirectionId, Route } from "../../../../__v3api";
-import { RouteStop, StopId, StopTree } from "../../__schedule";
+import {
+  IndexedRouteStop,
+  RouteStop,
+  StopId,
+  StopTree
+} from "../../__schedule";
 import { diagramWidth } from "../line-diagram-helpers";
 import VehicleIcons from "../VehicleIcons";
 import { LiveDataByStop } from "../__line-diagram";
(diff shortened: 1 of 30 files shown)


apps/site/assets/ts/schedule/components/line-diagram/graphics/Diagram.tsx
```diff
@@ -71,30 +71,93 @@ const branchingDescription = (stopTree: StopTree): string => {
...
+            toId={nextStop.id}
+            alerts={alerts}
+          />
+        );
+      })}
+
+      {/* Draw circles for each stop */
+      routeStopList.map((routeStop, index) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <Stop key={`${index}-${routeStop.id}`} stopId={routeStop.id} />
```
kotva006, line 155: We really shouldn't be using indexes as keys, for the underlying list items could change, but the index does not.  Is the id not unique here?  Is there some other way to get a unique id? (Like name-id-zone)

thecristen: The `routeStop.id` is not guaranteed to be unique amongst the list of stops. Because of this edge case where stops are repeated along the route. But maybe I can add a number to each item in the list and use that as the key...

thecristen: @kotva006 I've pushed an update to change this.

kotva006 (approved):
Looks good

PR #1762, opened 2023-10-03
fix(SiteWeb.StopController): don't show route patterns for services that aren't running today

thecristen:
Instead of using all route patterns from the whole rating, this PR adjusts the stop page's backend response for route patterns to remove those associated with services that aren't applicable today. 

On the stop page, this removes headsigns associated with route patterns for this weekend's shuttle service, later disruptions, and so on.

---

#### General checks

### DIFF
--- a/apps/route_patterns/lib/route_pattern.ex
+++ b/apps/route_patterns/lib/route_pattern.ex
@@ -37,6 +37,7 @@ defmodule RoutePatterns.RoutePattern do
     :route_id,
     :time_desc,
     :typicality,
+    :service_id,
     sort_order: 0
   ]
 
@@ -55,7 +56,8 @@ defmodule RoutePatterns.RoutePattern do
           route_id: Route.id_t(),
           time_desc: String.t(),
           typicality: typicality_t(),
-          sort_order: integer()
+          sort_order: integer(),
+          service_id: String.t()
         }
 
   def new(%Item{
@@ -92,7 +94,8 @@ defmodule RoutePatterns.RoutePattern do
       route_id: route_id,
       time_desc: time_desc,
       typicality: typicality,
-      sort_order: sort_order
+      sort_order: sort_order,
+      service_id: service_id(trip_relationships)
     }
   end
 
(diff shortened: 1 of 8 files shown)


apps/route_patterns/lib/route_pattern.ex
```diff
@@ -152,4 +155,15 @@ defmodule RoutePatterns.RoutePattern do
   end
 
   defp stop_ids(_), do: nil
+
+  defp service_id(%{
+         "service" => [
+           %Item{
+             id: service_id
+           }
+         ]
```
kotva006, line 164: This arraying of relationships is exactly what is happening in the `line` parsing I am working on.  Should we keep this behavior around?

apps/services/lib/service.ex
```diff
@@ -153,4 +154,57 @@ defmodule Services.Service do
...
+      end
+    end)
+    |> Enum.map(&Timex.to_date/1)
+    |> Enum.uniq()
+  end
+
+  @spec parse_listed_dates([String.t()]) :: [NaiveDateTime.t()]
+  defp parse_listed_dates(date_strings) do
+    date_strings
+    |> Enum.map(&Timex.parse(&1, "{YYYY}-{0M}-{D}"))
```
kotva006, line 206: question: Does the `{D}` need the leading zero?

apps/route_patterns/lib/route_pattern.ex
thecristen: Yeah, I see what you mean.

> Should we keep this behavior around?

Changing this (via removing `list()` in the relevant part of the `%JsonApi.Item{}` struct [here](https://github.com/mbta/dotcom/blob/master/apps/json_api/lib/json_api.ex#L8)) would require a baffling amount of downstream changes in all of our V3 API parsing. I don't see the benefit in changing it.

kotva006 (approved):
Some minor questions but it looks good

apps/services/lib/service.ex
thecristen: ..yes. I will update

PR #1769, opened 2023-10-16
feat(ShuttleHeadsigns): Merge the replacement shuttle bus cards with …

kotva006:
…the commuter rail cards

#### Summary of changes
*Working on display tests, wanted to get some eyes on this*

Hide Shuttle Cards from displaying on the new Subway stop page
Merge Shuttle times and headsign into the Commuter Rail Cards on the new Commuter Rail stop page.

**Asana Ticket:** [Shuttle Service is Shown in Separate Departure Cards](https://app.asana.com/0/home/1201123880594717/1205631015693375)

---

#### General checks

### DIFF
--- a/apps/route_patterns/lib/repo.ex
+++ b/apps/route_patterns/lib/repo.ex
@@ -8,7 +8,6 @@ defmodule RoutePatterns.Repo do
 
   alias RoutePatterns.RoutePattern
   alias V3Api.RoutePatterns, as: RoutePatternsApi
-  alias Routes.Repo, as: RoutesRepo
 
   @doc """
   Returns a single route pattern by ID
@@ -44,10 +43,7 @@ defmodule RoutePatterns.Repo do
   end
 
   def by_stop_id(stop_id) do
-    routes = RoutesRepo.by_stop(stop_id)
-    route_ids = Enum.join(Enum.map(routes, fn r -> r.id end), ",")
-
-    [route: route_ids]
+    [stop: stop_id]
     |> Keyword.put(:include, "representative_trip.shape,representative_trip.stops")
     |> cache(&api_all/1)
   end
--- a/apps/route_patterns/lib/repo.ex
+++ b/apps/route_patterns/lib/repo.ex
@@ -8,6 +8,7 @@ defmodule RoutePatterns.Repo do
 
   alias RoutePatterns.RoutePattern
   alias V3Api.RoutePatterns, as: RoutePatternsApi
+  alias Routes.Repo, as: RoutesRepo
 
   @doc """
   Returns a single route pattern by ID
@@ -43,7 +44,10 @@ defmodule RoutePatterns.Repo do
   end
 
   def by_stop_id(stop_id) do
-    [stop: stop_id]
+    routes = RoutesRepo.by_stop(stop_id)
+    route_ids = Enum.join(Enum.map(routes, fn r -> r.id end), ",")
+
+    [route: route_ids]
     |> Keyword.put(:include, "representative_trip.shape,representative_trip.stops")
     |> cache(&api_all/1)
   end
--- a/apps/routes/lib/parser.ex
+++ b/apps/routes/lib/parser.ex
@@ -19,9 +19,18 @@ defmodule Routes.Parser do
       direction_destinations:
         direction_attrs(attributes["direction_destinations"], parse_route_patterns(relationships)),
       description: parse_gtfs_desc(attributes["description"]),
-      fare_class: parse_gtfs_fare_class(attributes["fare_class"])
+      fare_class: parse_gtfs_fare_class(attributes["fare_class"]),
+      line: parse_line(relationships)
     }
   end
+
+  defp parse_line(%{"line" => [head | _] = _lines}) do
+    # TODO figure out how to make this not an array on parsing
+    # Testing shows this as always a single object from the API that gets parsed into an array for some reason
+    head
+  end
+
+  defp parse_line(_), do: nil
 
   def parse_route_with_route_pattern(%Item{relationships: relationships} = item) do
     {parse_route(item), parse_route_patterns(relationships)}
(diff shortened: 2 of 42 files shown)


apps/routes/lib/route.ex
```diff
@@ -29,7 +30,8 @@ defmodule Routes.Route do
           direction_destinations: %{0 => String.t(), 1 => String.t()} | :unknown,
           description: gtfs_route_desc,
           fare_class: gtfs_fare_class,
-          custom_route?: boolean
+          custom_route?: boolean,
+          line: []
```
thecristen, line 34: FYI you end up adding the item to this property (as opposed to a list as suggested by `line: []`), so the resulting struct ends up being like:

```
%Routes.Route{
  color: "DA291C",
  custom_route?: false,
  description: :rapid_transit,
  direction_destinations: %{
    0 => "Ashmont/Braintree", 
    1 => "Alewife"
  },
  direction_names: %{
    0 => "Southbound",
    1 => "Northbound"
  },
  fare_class: :rapid_transit_fare,
  id: "Red",
  line: %JsonApi.Item{
    attributes: nil,
    id: "line-Red",
    relationships: nil,
    type: "line"
  },
  long_name: "Red Line",
  name: "Red Line",
  sort_order: 10010,
  type: 1
}
```

thecristen: **suggestion:** change to `line_id: String.t()` and assign it the line's ID instead of including the entire `%JsonApi.Item{}`.

apps/routes/lib/parser.ex
```diff
@@ -19,10 +19,19 @@ defmodule Routes.Parser do
...
       description: parse_gtfs_desc(attributes["description"]),
-      fare_class: parse_gtfs_fare_class(attributes["fare_class"])
+      fare_class: parse_gtfs_fare_class(attributes["fare_class"]),
+      line: parse_line(relationships)
     }
   end
 
+  defp parse_line(%{"line" => [head | _] = _lines}) do
+    # TODO figure out how to make this not an array on parsing
+    # Testing shows this as always a single object from the API that gets parsed into an array for some reason
```
thecristen, line 29: Plenty of similar functions in our parsers, [example here](https://github.com/mbta/dotcom/blob/5f2f723b3e028c5e4988e0da16324e3f017350f2/apps/schedules/lib/parser.ex#L42). It's typical.

apps/routes/lib/route.ex
thecristen: that'll also fix a backend error I'm noticing because the application doesn't know how to encode the `%JsonApi.Item{}` struct to return it in a JSON format to the frontend. better to extract the information we need from that struct into a more concise format.

image

apps/route_patterns/lib/repo.ex
```diff
@@ -43,7 +44,10 @@ defmodule RoutePatterns.Repo do
   end
 
   def by_stop_id(stop_id) do
-    [stop: stop_id]
+    routes = RoutesRepo.by_stop(stop_id)
+    route_ids = Enum.join(Enum.map(routes, fn r -> r.id end), ",")
+
+    [route: route_ids]
```
thecristen, line 50: **issue**: I am noticing a lot more headsigns shown, and I am thinking that this change might be reintroducing the problem we had where we're showing headsigns that aren't applicable to this stop.

can you try re-adding the stop argument here? e.g. use `[route: route_ids, stop: stop_id]` for the API request, and we can see if that helps.

there are likely many examples but if you'd like a reference I am using stop `1260`,  in this PR having three headsigns, but on dev it shows Route 22 to **Ashmont** only:
image

thecristen (changes requested):
Thanks!! This review is incomplete, so these comments are just for the back-end piece - will circle back soon.

I think the overall approach is sound, though I had some comments.

Another (non-blocking) question I had - I notice you're doing a lot of manipulation of the grouped route patterns in the frontend, based on the newly obtained route line information - which is fine, but might it be simpler to modify those grouped route patterns before they hit the frontend, [when they're pulled together and grouped in the controller](https://github.com/mbta/dotcom/blob/5f2f723b3e028c5e4988e0da16324e3f017350f2/apps/site/lib/site_web/controllers/stop_controller.ex#L189)?

apps/site/assets/ts/stop/__tests__/DeparturesAndMapTest.tsx
```diff
@@ -48,12 +48,6 @@ beforeEach(() => {
     .spyOn(useSchedules, "useSchedulesByStop")
     .mockReturnValue({ status: FetchStatus.Data, data: schedules });
   jest.spyOn(useVehiclesChannel, "default").mockReturnValue([]);
-  jest.mock("react-router-dom", () => ({
-    __esModule: true,
-    useLoaderData: () => {
-      return TEST_LOADER_VALUE;
-    }
-  }));
```
kotva006, line 56: note: This was redundant because the data is already passed into the render function inside the helper function `renderWithRouter`

kotva006:
> Thanks!! This review is incomplete, so these comments are just for the back-end piece - will circle back soon.
> 
> I think the overall approach is sound, though I had some comments.
> 
> Another (non-blocking) question I had - I notice you're doing a lot of manipulation of the grouped route patterns in the frontend, based on the newly obtained route line information - which is fine, but might it be simpler to modify those grouped route patterns before they hit the frontend, [when they're pulled together and grouped in the controller](https://github.com/mbta/dotcom/blob/5f2f723b3e028c5e4988e0da16324e3f017350f2/apps/site/lib/site_web/controllers/stop_controller.ex#L189)?

I ended up putting the fix on the front end because I viewed this issue as a display problem.  I was trying to avoid curating the data returned by the route pattern endpoint to fit this one use case.  I thought there may be other times where we want these route patterns separate.  If this needs to be changed everywhere route patterns are consumed, then it would makes sense to change it on the backend.

PR #1770, opened 2023-10-16
Chore: Update GitHub actions to use assumed role

krisrjohnson21:
#### Summary of changes

**Asana Ticket:** [Update workflows repo to point to V2 actions and use assumed role instead of keys](https://app.asana.com/0/0/1205570959197530/f)

This PR updates the `use-deploy-ecs` workflow to use v2 of the shared `build-push-ecr` and `deploy-ecs` actions. These updated actions assume an IAM role instead of relying on long-lasting AWS keys for an IAM user. 

Note: Other workflows that rely on the `deploy-ecs` [workflow](https://github.com/mbta/workflows/blob/main/.github/workflows/deploy-ecs.yml) will be updated at a later date.

### DIFF
--- a/.github/workflows/use-deploy-ecs.yml
+++ b/.github/workflows/use-deploy-ecs.yml
@@ -13,18 +13,12 @@ on:
         required: false
         description: "Additional tag to apply to the docker image"
     secrets:
-      aws-access-key-id:
+      aws-role-arn:
         required: true
-        description: "AWS_ACCESS_KEY_ID value"
-      aws-secret-access-key:
-        required: true
-        description: "AWS_SECRET_ACCESS_KEY value"
+        description: "AWS_ROLE_ARN value"
       docker-repo:
         required: true
         description: "The Docker repository to upload the docker image to, formatted '....amazonaws.com/<app>'"
-      role-to-assume:
-        required: true
-        description: "AWS_ROLE_ARN value"
     outputs: 
       deployment-env:
         value: ${{ jobs.deploy.outputs.deployment-env }}
@@ -46,6 +40,11 @@ jobs:
     outputs:
       deployment-env: ${{ inputs.deployment-env }}
     steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.aws-role-arn }}
+          aws-region: us-east-1
       - run: |
           echo "inputs: ${{ toJson(inputs) }}"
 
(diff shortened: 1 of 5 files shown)


paulswartz (commented):
@krisrjohnson21 looks like there was an error deploying: https://github.com/mbta/dotcom/actions/runs/6540317730/job/17760018293

.github/workflows/use-deploy-ecs.yml
```diff
@@ -16,6 +16,9 @@ on:
       aws-access-key-id:
         required: true
         description: "AWS_ACCESS_KEY_ID value"
+      aws-role-arn:
```
paulswartz, line 19: If you're adding a required field, you need to make sure it's passed in everywhere: https://github.com/mbta/dotcom/actions/runs/6549188295

You can also probably get rid of the `aws-access-key-id` and `aws-secret-access-key` parameters.

krisrjohnson21: Thanks! I'd assumed that removing the access key and secret key will affect other steps, like [logging into ECR](https://github.com/mbta/dotcom/blob/3c2993d7d51521762a0be65c89b88315961a459a/.github/workflows/use-deploy-ecs.yml#L63C18-L63C18) and [uploading assets into S3](https://github.com/mbta/dotcom/blob/3c2993d7d51521762a0be65c89b88315961a459a/.github/workflows/use-deploy-ecs.yml#L79C2-L79C2). If that's not the case, I can certainly remove these required fields

.github/workflows/use-deploy-ecs.yml
```diff
@@ -16,15 +16,15 @@ on:
...
-      aws-role-arn:
-        required: true
-        description: "AWS_ROLE_ARN value"
       aws-secret-access-key:
         required: true
         description: "AWS_SECRET_ACCESS_KEY value"
       docker-repo:
         required: true
         description: "The Docker repository to upload the docker image to, formatted '....amazonaws.com/<app>'"
+      role-to-assume:
```
paulswartz, line 25: It looks like `aws-role-arn` is the expected value here: https://github.com/mbta/dotcom/pull/1770/files/f0b7e698648251412d857c24aae4a9fe9c9949ff..acb445b39f312cbfee90586b3e09f34cfa824c23#diff-ae570e673d878f853f5159ea079c8f6ae8663fc448b98158726df4762c16596dL104 but you could update it in both places if you wanted.

I think the remaining steps are:
- remove `aws-access-key-id` and `aws-secret-access-key` from the variables here
- remove any use of those values from this file, replacing them with role ARN usage
- remove those values from the workflows which call this one

krisrjohnson21: Thanks for these steps--they were a big help, I think I've got it straightened out now

paulswartz (approved):

thecristen (approved):
Thank you! ✨

PR #1771, opened 2023-10-17
feat(Algolia.Api): cache requests

thecristen:
#### Summary of changes

**Asana Ticket:** No ticket

While working on the autocomplete widget I noticed an opportunity for us to cache our requests to the Algolia REST API!

This PR uses our existing `RepoCache` module to cache the POST requests we make to Algolia, based on two parameters
1. The **request `body`**. It's a gnarly string of text looking something like this (yes it's a query for the letter `r`).
```
"{\"requests\":[{\"query\":\"r\",\"params\":\"analytics=false&clickAnalytics=true&facetFilters=%5B%5B%5D%5D&facets=%5B%22*%22%5D&hitsPerPage=5\",\"indexName\":\"routes\"},{\"query\":\"r\",\"params\":\"analytics=false&clickAnalytics=true&facetFilters=%5B%5B%5D%5D&facets=%5B%22*%22%5D&hitsPerPage=2\",\"indexName\":\"stops\"},{\"query\":\"r\",\"params\":\"analytics=false&clickAnalytics=true&facetFilters=%5B%5B%22_content_type%3Apage%22%2C%22_content_type%3Asearch_result%22%2C%22_content_type%3Adiversion%22%2C%22_content_type%3Alanding_page%22%2C%22_content_type%3Aperson%22%2C%22_content_type%3Aproject%22%2C%22_content_type%3Aproject_update%22%5D%5D&facets=%5B%22*%22%5D&hitsPerPage=2\",\"indexName\":\"drupal\"}]}"
```
It'll have different params depending on the type of search (e.g. the request body for searching `r` in the trip planner will look different as it'll just search for 3 stops instead of 2 stops, 5 routes, and results from the CMS). So the results for one search box is not necessarily cached for use by a different search box in the website -- that'll only happen if both perform exactly the same query with exactly the same sets of parameters.
2. The Algolia **configuration** - this is just a struct that captures the API keys in use. I figure caching should only be valid if the request is made with the same API keys.

---

#### General checks

### DIFF
--- a/apps/algolia/lib/algolia/api.ex
+++ b/apps/algolia/lib/algolia/api.ex
@@ -4,6 +4,7 @@ defmodule Algolia.Api do
   """
   alias Algolia.Config
   require Logger
+  use RepoCache, ttl: :timer.hours(12)
 
   defstruct [:host, :index, :action, :body]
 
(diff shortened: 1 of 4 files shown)


kotva006 (approved):
LGTM

PR #1773, opened 2023-10-18
fix(V3Api): encode url for v3 api requests

thecristen:
#### Summary of changes

We discovered sometimes a service ID might contain a space character. And we were not able to get that service information from the API because we weren't encoding the URL. This fixes that, along with updating related tests.

It also appears the V3 API now doesn't allow for fetching all services without any filter parameters, so another test has been updated to account for that.

---

#### General checks


kotva006 (commented):
Are there tests that ensure the space in the URL still work?  It looks like this pull request is removing the tests.

kotva006 (approved):

PR #1785, opened 2023-10-30
feat(StopPageRedesign): "no service" via route patterns

thecristen:
#### Summary of changes

**Asana Ticket:** No ticket

This gave me flashbacks to earlier in this page's development when we were struggling to distinguish "doesn't serve this stop" vs "no schedules or predictions for this stop." But I think I figured out an okay approach.

### Background

The changes from #1762 went too far, and ends up leaving us unable to display any route card when no service is running. It turns out that produces a really jarring display on certain bus stop pages. 

For example, there are several bus routes which only operate on weekdays -- visiting the stop pages serving these routes on the weekends will greet you with a cryptic blank display:

image

image

(screens from a technically unrelated bug report, but that's also the blank display you get in the scenario described above – I failed to take a screenshot of a relevant stop over the weekend 😅)

So I wanted to make another attempt at showing route cards in the case where transit service isn't available today. 

<details>
<summary>It was somewhat straightforward – we needed to keep the list of route patterns intact, instead of prematurely removing them before grouping them by route and headsign. Click for the pseudocode description:</summary>

```elixir
# Before - a headsign for a route pattern associated with 
# a service not running today will be removed entirely
all_route_patterns
|> remove_not_todays
|> group_by_route_id_and_headsign

# After - a headsign for a route pattern associated with a service not
# running today will be associated with a route pattern list that's empty
all_route_patterns
|> group_by_route_id_and_headsign
|> Enum.map(fn group ->
    group 
    |> remove_not_todays
end)
```

</details>

### Highlights

I tried to split this into smaller, incremental commits that each describe an incremental change. But this goal inspired modest changes on both the backend and frontend.

#### Backend

- Moved the `Enum(&not_serving_today?/1)` operation in the grouped route patterns endpoint to operate over a list of route patterns for each route, instead of removing route patterns before we group the list by route+headsign. That way we could keep all routes+headsigns as part of the response, even if they're not associated with today's service.
- Modified the grouped route patterns response to include, for each route+headsign, the relevant direction ID, instead of parsing the direction ID from the list of route patterns in the frontend. That way the direction ID won't be missing if that list of route patterns is empty.

Less relevant changes:

- Using [Stream](https://hexdocs.pm/elixir/1.12.3/Stream.html) in favor of [Enum](https://hexdocs.pm/elixir/1.12.3/Enum.html) when chaining operations to group the route patterns, hoping for a minor performance boost. Not a big deal, but the grouped route patterns endpoint is relatively slow.
- Removed services with the ID "canonical" for being considered as today's valid service. Those are meant for reference and are not associated with any day.

#### Frontend

Most changes are adapting to the changed backend.

- Adds a "No service today" text if the route+headsign's list of route patterns is an empty array
- Elaborates this to "No service today, Oct 30" (for example) when clicking the headsign to view the `<DepartureList />`.
- Adds a helper function to generate that "Oct 30" part.

### Reviewing the result

- It'll be easier to spot on weekends when Weekday-only buses go out of service. But we can also preview this on stops along the Foxboro Event Service route, and at stops served by the CapeFLYER Hyannis route -- both of which are not in service currently. These routes can both be seen at South Station.

image

image

image

image

- Otherwise, we can manually emulate the no-service conditions by using React Dev Tools to set the `hasService` prop to `false`. For example:

image

---

#### General checks

---
  - https://app.asana.com/0/0/1205864682150635

### DIFF
--- a/apps/site/assets/ts/helpers/__tests__/date-test.ts
+++ b/apps/site/assets/ts/helpers/__tests__/date-test.ts
@@ -2,8 +2,7 @@ import { addMinutes, addSeconds } from "date-fns";
 import {
   compareStringTimes,
   formatRelativeTime,
-  formatToBostonTime,
-  todayDateString
+  formatToBostonTime
 } from "../date";
 
 describe("compareStringTimes", () => {
@@ -43,20 +42,5 @@ describe("formatRelativeTime", () => {
 
     const laterDate = addMinutes(baseDate, 70);
     expect(formatRelativeTime(laterDate, baseDate)).toBe("11:10 AM");
-  });
-});
-
-describe("todayDateString", () => {
-  beforeAll(() => {
-    jest.useFakeTimers();
-    jest.setSystemTime(new Date("2012-10-10T09:00:00.000+04:00"));
-  });
-
-  it("returns formatted today date", () => {
-    expect(todayDateString()).toEqual("Oct 10");
-  });
-
-  afterAll(() => {
-    jest.useRealTimers();
   });
 });
--- a/apps/site/assets/ts/helpers/__tests__/date-test.ts
+++ b/apps/site/assets/ts/helpers/__tests__/date-test.ts
@@ -2,7 +2,8 @@ import { addMinutes, addSeconds } from "date-fns";
 import {
   compareStringTimes,
   formatRelativeTime,
-  formatToBostonTime
+  formatToBostonTime,
+  todayDateString
 } from "../date";
 
 describe("compareStringTimes", () => {
@@ -42,5 +43,20 @@ describe("formatRelativeTime", () => {
 
     const laterDate = addMinutes(baseDate, 70);
     expect(formatRelativeTime(laterDate, baseDate)).toBe("11:10 AM");
+  });
+});
+
+describe("todayDateString", () => {
+  beforeAll(() => {
+    jest.useFakeTimers();
+    jest.setSystemTime(new Date("2012-10-10T09:00:00.000+04:00"));
+  });
+
+  it("returns formatted today date", () => {
+    expect(todayDateString()).toEqual("Oct 10");
+  });
+
+  afterAll(() => {
+    jest.useRealTimers();
   });
 });
--- a/apps/site/assets/ts/stop/components/DepartureCard.tsx
+++ b/apps/site/assets/ts/stop/components/DepartureCard.tsx
@@ -73,6 +73,7 @@ const DepartureCard = ({
               )}
               onClick={onClick}
               isCR={isACommuterRailRoute(route)}
+              hasService={routePatterns.length !== 0}
             />
           );
         }
--- a/apps/site/assets/ts/stop/components/DepartureCard.tsx
+++ b/apps/site/assets/ts/stop/components/DepartureCard.tsx
@@ -9,7 +9,7 @@ import { allAlertsForDirection } from "../../models/alert";
 import { departureInfoInRoutePatterns } from "../../helpers/departureInfo";
 import { isACommuterRailRoute } from "../../models/route";
 import DepartureTimes from "./DepartureTimes";
-import { RoutePatternWithPolyline } from "../stop-redesign-loader";
+import { GroupedRoutePatterns } from "../stop-redesign-loader";
 
 const DepartureCard = ({
   alertsForRoute,
@@ -19,7 +19,7 @@ const DepartureCard = ({
 }: {
   alertsForRoute: Alert[];
   departuresForRoute: DepartureInfo[];
-  routePatternsByHeadsign: Record<string, RoutePatternWithPolyline[]>;
+  routePatternsByHeadsign: GroupedRoutePatterns[keyof GroupedRoutePatterns];
   route: Route;
 }): ReactElement<HTMLElement> => {
   const { setRow } = useDepartureRow([route]);
@@ -27,7 +27,10 @@ const DepartureCard = ({
   const sortedRoutePatternsByHeadsign = useMemo(
     () =>
       Object.entries(routePatternsByHeadsign).sort((entryA, entryB) => {
-        const [orderA, orderB] = [entryA, entryB].map(([, routePatterns]) =>
+        const [orderA, orderB] = [
+          entryA,
+          entryB
+        ].map(([, { route_patterns: routePatterns }]) =>
           Math.min(...routePatterns.map(rp => rp.sort_order))
         );
         return orderA - orderB;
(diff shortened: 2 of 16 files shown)


apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -192,19 +192,44 @@ defmodule SiteWeb.StopController do
...
-    |> Enum.reject(&ends_at?(&1, stop_id))
-    |> Enum.map(&add_polyline/1)
-    |> Enum.group_by(& &1.route_id)
-    |> Enum.map(fn {k, v} -> {k, Enum.group_by(v, & &1.headsign)} end)
+    |> Stream.reject(&ends_at?(&1, stop_id))
+    |> Stream.chunk_by(& &1.route_id)
+    |> Stream.map(fn [%RoutePattern{route_id: route_id} | _] = route_route_patterns ->
+      by_headsign =
+        route_route_patterns
+        |> Stream.chunk_by(& &1.headsign)
```
KaylaBrady, line 209: question: What is guaranteeing the sorted order?

apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -192,19 +192,44 @@ defmodule SiteWeb.StopController do
...
+          {headsign, with_headsign_info(headsign_route_patterns)}
+        end)
+        |> Map.new()
+
+      {route_id, by_headsign}
+    end)
     |> Map.new()
   end
 
+  @spec with_headsign_info([RoutePattern.t()]) :: headsign_info
```
KaylaBrady, line 220: suggestion: Consider renaming this - `with_headsign_info` reads to me like it is only adding headsign information to the route patterns, but this is also filtering & adding polylines

apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -214,6 +239,7 @@ defmodule SiteWeb.StopController do
     case Services.Repo.by_route_id(route_id) do
       [%Service{} | _] = services ->
         services
+        |> Enum.reject(&(&1.id === "canonical"))
```
KaylaBrady, line 242: question: Could these be filtered upfront using `Services.Repo.by_route_id` params?

apps/site/assets/ts/stop/components/DepartureCard.tsx
```diff
@@ -19,15 +19,18 @@ const DepartureCard = ({
 }: {
   alertsForRoute: Alert[];
   departuresForRoute: DepartureInfo[];
-  routePatternsByHeadsign: Record<string, RoutePatternWithPolyline[]>;
+  routePatternsByHeadsign: GroupedRoutePatterns[keyof GroupedRoutePatterns];
```
KaylaBrady, line 22: praise: Appreciate the more specific typing here!

KaylaBrady (commented):
Excited for this change! 🎉 Added a few questions & comments. Looks like some checks are failing too. 
Seeing [282 elixir unit test failures](https://github.com/mbta/dotcom/actions/runs/6737592188/job/18417815363?pr=1785), which is 1 more failing than the last merge to main, but didn't dig into which one is the difference.

apps/site/lib/site_web/controllers/stop_controller.ex
thecristen: it's a negation, so I don't think so?

apps/site/lib/site_web/controllers/stop_controller.ex
thecristen: nothing here, this gets sorted in the frontend

apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -192,19 +192,44 @@ defmodule SiteWeb.StopController do
     json(conn, route_patterns_by_route_and_headsign(stop_id))
   end
 
-  @type by_route_and_headsign :: %{Route.id_t() => %{String.t() => [RoutePattern.t()]}}
+  @type headsign_info :: %{
+          required(:direction_id) => 0 | 1,
+          required(:route_patterns) => [RoutePattern.t()]
+        }
+  @type by_route_and_headsign :: %{Route.id_t() => %{String.t() => headsign_info}}
```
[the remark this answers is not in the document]
thecristen: we don't do separately versioned backends and frontends here, they'll be deployed together 😉

apps/site/lib/site_web/controllers/stop_controller.ex
KaylaBrady: Ah, I was thinking of the `filter[canonical]=false` filter supported on the [/route_patterns](https://api-v3.mbta.com/docs/swagger/index.html#/RoutePattern) endpoint.

apps/site/lib/site_web/controllers/stop_controller.ex
KaylaBrady: I think it would need to be sorted for `chunk_by` to work, looking at [this example](https://hexdocs.pm/elixir/1.12.3/Stream.html#chunk_by/2), the chunks are only formed for elements that are immediately next to each other. If the stream was ["Nubian", "Ashmont", "Nubian"], then the two Nubians wouldn't be grouped.

apps/site/lib/site_web/controllers/stop_controller.ex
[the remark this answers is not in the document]
KaylaBrady: I may be missing a nuance here, but my previous understanding on Skate was that this would be considered a breaking change. Even though the frontend and backends are deployed together, there will be a period of time where some ecs instances have the old backend and some instances have the new backend, during which someone using the stop page has no control over whether the new or old instances are hit, so there could be a mismatch between the data that is returned and what the frontend expects. Is that situation avoided with how dotcom is set up?

apps/site/lib/site_web/controllers/stop_controller.ex
```diff
@@ -192,19 +192,44 @@ defmodule SiteWeb.StopController do
...
   @spec route_patterns_by_route_and_headsign(Stop.id_t()) :: by_route_and_headsign()
   defp route_patterns_by_route_and_headsign(stop_id) do
     stop_id
     |> RoutePatterns.Repo.by_stop_id()
-    |> Enum.reject(&not_serving_today?/1)
-    |> Enum.reject(&ends_at?(&1, stop_id))
-    |> Enum.map(&add_polyline/1)
-    |> Enum.group_by(& &1.route_id)
-    |> Enum.map(fn {k, v} -> {k, Enum.group_by(v, & &1.headsign)} end)
+    |> Stream.reject(&ends_at?(&1, stop_id))
```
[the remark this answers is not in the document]
thecristen: ooh I forgot about `Enum.group_by/3`, I'll give this a try

apps/site/lib/site_web/controllers/stop_controller.ex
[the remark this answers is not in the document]
thecristen: we've never thought to mitigate this before! but when you explain it here it makes sense...aah 🫠 🫠 🫠 🫠

apps/site/lib/site_web/controllers/stop_controller.ex
thecristen: ah shoot good point. building another good case for going back to `Enum.group_by`...

apps/site/lib/site_web/controllers/stop_controller.ex
thecristen: Frankly it's too weird that there is even any service with an ID of "canonical", but I did see it 😬

apps/site/assets/ts/helpers/__tests__/date-test.ts
```diff
@@ -44,3 +45,18 @@ describe("formatRelativeTime", () => {
     expect(formatRelativeTime(laterDate, baseDate)).toBe("11:10 AM");
   });
 });
+
+describe("todayDateString", () => {
```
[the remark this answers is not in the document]
thecristen: maybe! but it's removed now because I no longer need it 😅

apps/site/assets/ts/stop/components/DepartureList.tsx
```diff
@@ -35,6 +39,7 @@ const DepartureList = ({
   directionId,
   headsign,
   alerts,
+  hasService,
```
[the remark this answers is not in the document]
thecristen: not without more refactoring of the components (since the no service message & the departure times render in the same wrapping HTML) (this goes for above too)

apps/site/lib/site_web/controllers/stop_controller.ex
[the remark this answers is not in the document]
thecristen: I do think that problem might be less frequently encountered on Dotcom, because folks don't have it open very long per session (just a few minutes -- I presume folks working in Skate maintain a much longer working session during typical use). I'll have to figure out how to handle that more systematically but for the sake of this PR will consider it out of scope. But thanks so much for raising this point!!

thecristen:
Sorry about the failing tests -- those particular fails have been popping up more here, I think it's from a race condition with how the workflow steps are building the application before running the tests 😵‍💫 Anyway, thanks @KaylaBrady for the many great points, I've added some revisions here.

KaylaBrady (approved):
Approved with offline discussion about preventing breaking changes in the future!

PR #1787, opened 2023-11-03
hotfix(autocomplete): use separate panelContainer

thecristen:
this fixes a Safari bug where the tab order traps keyboard focus in between the autocomplete input and submit buttons. the bug only happened on Safari, when using the React renderer instead of the default Preact renderer.

some CSS refactoring was needed to target elements that are inside the new panelContainer element.

#### Summary of changes

Asana Ticket: [Can't tab past navbar search on macOS/Safari](https://app.asana.com/0/385363666817452/1205862370279606/f)


kotva006 (approved):
Looks good

PR #1802, opened 2023-11-22
Don't translate the names of lines and stops

anthonyshull:
I can't check Algolia search results because I don't have an account setup. But, I think I got them anyway.

Screenshot Capture - 2023-11-22 - 07-38-16

https://app.asana.com/0/385363666817452/1201924917341675/f


thecristen (commented):
question: Were you intending to cover all of the sub-tasks, or just certain ones? I thought maybe all of them since these changes cover a lot, but I do notice we're missing the attributes on the trip planner itineraries, on the stop page, and just part of transit near me. 🙈 

Transit near me: on the link to the stop page
image

anthonyshull:
> question: Were you intending to cover all of the sub-tasks, or just certain ones? I thought maybe all of them since these changes cover a lot, but I do notice we're missing the attributes on the trip planner itineraries, on the stop page, and just part of transit near me. 🙈

My intention was to get everything, but the subtasks didn't have pictures or URLs so I was just going based on names and looking at pages. If you can add screenshots of wherever else we're not wanting translations, I can add those in.

thecristen:
Sounds good, here's a few places:

# Trip planner

Open the "map and trip details", stop and route names are mentioned throughout:
 
image

And also in the links at bottom:

image

# Stop pages

These are the pages rendered for each individual stop, e.g. /stops/place-north for North Station

image

image

apps/site/lib/site_web/templates/trip_plan/_itinerary_row_step.html.eex
```diff
@@ -1,7 +1,7 @@
 <div class="route-branch-stop-list route-branch-stop <%= [@mode_class] %>">
   <%= SiteWeb.PartialView.StopBubbles.render_stop_bubbles(@bubble_params) %>
   <div class="itinerary-step">
-  <%= @step %>
+  <span class="notranslate"><%= @step %></span>
```
thecristen, line 4: This one might be too heavy-handed as it prevents the entire step's description from being translated. After poking around a bit I can't figure a way to reliably find the portion of text that contains a stop name, so let's revert this change.

image

thecristen (changes requested):
Just a tiny change :)

apps/site/lib/site_web/templates/trip_plan/_itinerary_row_step.html.eex
anthonyshull: @thecristen I asked Paul and he said to not translate it for now. I added that detail to the Asana task.

thecristen (approved):
Thanks for the undertaking!

PR #1809, opened 2023-11-28
Wrap text in alert banners

anthonyshull:
https://app.asana.com/0/555089885850811/1205466554569824/f

Screenshot Capture - 2023-11-28 - 12-51-51


thecristen (changes requested):
Looks like there are some overflow-related side-effects in other places displaying alerts, as I can't reproduce these on prod:

From `/alerts`

image

And `/schedules/<route>/alerts`


image

image

anthonyshull:
Thanks for finding that. I'll just have to learn all of the places to manually check. Who would have thought this same class would be re-used in such wildly different places: top banner, in page banner, and some kind of accordion that all look very different.

thecristen (approved):

PR #1812, opened 2023-12-01
Control what data gets logged to remove headers

anthonyshull:
https://app.asana.com/0/555089885850811/1206042318802983/f


apps/site/test/site_web/controllers/search_controller_test.exs
```diff
@@ -74,27 +77,27 @@ defmodule SiteWeb.SearchControllerTest do
...
-          assert SiteWeb.SearchController.log_error({:error, %HTTPoison.Error{}}) == :ok
+        capture_log(fn ->
+          SiteWeb.SearchController.log_error({:error, %HTTPoison.Error{reason: :econnrefused}})
         end)
 
-      assert log =~ "bad response"
+      assert log =~ "reason: :econnrefused"
     end
 
     test "does not log other types of errors" do
```
kotva006, line 94: nitpick: This test name doesn't seem to match what is being tested/

apps/site/lib/site_web/controllers/search_controller.ex
```diff
@@ -72,27 +73,29 @@ defmodule SiteWeb.SearchController do
...
-          :ok | {:error, any}
   def log_error({:ok, %HTTPoison.Response{} = response}) do
-    do_log_error(response)
-  end
-
-  def log_error({:error, %HTTPoison.Error{} = response}) do
-    do_log_error(response)
+    response
+    |> Map.from_struct()
+    |> Map.take([:body, :headers, :status_code])
```
thecristen, line 84: Might want to omit the `:headers` here. Looking at one of the old `"Received bad response from Algolia"` errors, the response headers are just
```
[{"Server", "nginx"}, {"Date", "Tue, 21 Nov 2023 12:52:06 GMT"}, {"Content-Type", "application/json; charset=UTF-8"}, {"Content-Length", "104"}, {"Connection", "keep-alive"}, {"X-Alg-PT", "1"}, {"Accept-Encoding", "deflate, gzip"}, {"Cache-Control", "no-store"}, {"Access-Control-Allow-Origin", "*"}, {"Timing-Allow-Origin", "*"}, {"X-Content-Type-Options", "nosniff"}, {"Strict-Transport-Security", "max-age=31536000; includeSubDomains; preload"}, {"Content-Disposition", "inline; filename=a.txt"}]
```

apps/site/lib/site_web/controllers/search_controller.ex
```diff
@@ -72,27 +73,29 @@ defmodule SiteWeb.SearchController do
...
+    |> Keyword.new(fn {k, v} -> {k, v} end)
```
thecristen, line 85: For the `%HTTPoison.Response{}` `body` it appears to be JSON-encoded, example `"{\"message\":\"Invalid value for \\\"query\\\" parameter, expected string shorter than 512 bytes\",\"status\":400}"`...maybe we should just decode and log this instead?

apps/site/lib/site_web/controllers/search_controller.ex
anthonyshull: The issue before was that we were just using `inspect()` to turn the map into a string. Logster will remove any headers that aren't explicitly named in: https://github.com/mbta/dotcom/blob/master/apps/site/config/config.exs#L27

apps/site/lib/site_web/controllers/search_controller.ex
```diff
@@ -72,27 +73,29 @@ defmodule SiteWeb.SearchController do
...
+    |> Map.take([:reason])
+    |> Keyword.new(fn {k, v} -> {k, v} end)
+    |> Logger.warn()
   end
 
-  @spec do_log_error(HTTPoison.Response.t() | HTTPoison.Error.t()) :: :ok | {:error, any}
-  defp do_log_error(error) do
-    Logger.warn("Received bad response from Algolia: #{inspect(error)}")
+  def log_error({:error, error}) do
+    Logger.warn(error: error)
```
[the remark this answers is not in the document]
anthonyshull: You can give Logster a keyword list and it will format it however you have setup. https://github.com/navinpeiris/logster#using-standalone-logger

apps/site/lib/site_web/controllers/search_controller.ex
anthonyshull: I'm not entirely sure how logging a map would work. But, I think it will just flatten out it like this.

apps/site/lib/site_web/controllers/search_controller.ex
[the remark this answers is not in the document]
thecristen: Ohh didn't realize this, super cool!

anthonyshull:
Putting this back to draft because I realized we're not actually using the Logster logger for most logging.

apps/site/lib/site_web/controllers/search_controller.ex
```diff
@@ -76,21 +76,18 @@ defmodule SiteWeb.SearchController do
...
+    do_log_error(response.reason)
   end
 
   def log_error(_) do
     :ok
   end
 
-  @spec do_log_error(HTTPoison.Response.t() | HTTPoison.Error.t()) :: :ok | {:error, any}
   defp do_log_error(error) do
     Logger.warn("Received bad response from Algolia: #{inspect(error)}")
```
anthonyshull, line 92: This inspect was the problem because it was printing out the headers of the HTTPoison.Response which is where the Algolia keys were. So, I just pulled the body out of the response so we have some information on what was sent. For an HTTPoison.Error we grab the reason.

kotva006 (approved):

PR #1817, opened 2023-12-04
Check the accessibility score of an itinerary to determine whether or not it is accessible

anthonyshull:
We weren't requesting the accessibility score. We get that and check it against an arbitrary value (0.5) to determine whether or not an itinerary is accessible.

Screenshot Capture - 2023-12-05 - 09-01-49


apps/trip_plan/lib/trip_plan/api/open_trip_planner/parser.ex
```diff
@@ -68,15 +68,26 @@ defmodule TripPlan.Api.OpenTripPlanner.Parser do
...
   defp error_message_atom(_, _opts), do: :unknown
 
-  defp parse_itinerary(json, request_params) do
+  defp parse_itinerary(json) do
     %Itinerary{
       start: parse_time(json["startTime"]),
       stop: parse_time(json["endTime"]),
       legs: Enum.map(json["legs"], &parse_leg/1),
-      accessible?: request_params["wheelchair"] == "true"
+      accessible?: parse_float(json["accessibilityScore"]) > 0.5
```
anthonyshull, line 76: 👁️ I just arbitrarily picked `0.5`. I'm not sure what constitutes an accessible itinerary.

kotva006: So I checked and accessible route and it returned 1.  I am wondering if anything less than 1 means part of the trip might not be accessible.

anthonyshull: The docs say it is a score between 0 and 1. If you check the Orange Line it comes back with 1 because all stops are wheelchair accessible. I imagine we will have to check a bunch of different itineraries to figure out what a "good" score looks like. But, for now, this gets itineraries showing as accessible.

kotva006: What I am trying to get at is if it returns .75 and one part of the trip isn't accessible are we mis leading the riders.  It might be safe to just keep the check at 1, until more research can be done

kotva006 (approved):

PR #1818, opened 2023-12-05
feat(Algolia): Clear old stops and routes on upload

kotva006:
#### Summary of changes
This affects both stops and routes.
This pull request adds a support for a `GET`  request to query the algolia endpoint during the update.
The order of operations

1. Grab all of the current objects off of Algolia
2. Generate a list of new objects to send to Algolia
3. Compare the current object list and new object list to get a list of no longer supported ids
4. Send a `batch` request job off to Algolia to delete the no longer supported objects
5. Send a `batch` request job off to Algolia to upload the new objects

**Asana Ticket:** [Update Algolia Scripts to remove stops](https://app.asana.com/0/home/1201123880594717/1204170419316251)

---

#### General checks

### DIFF
--- a/apps/algolia/lib/algolia/api.ex
+++ b/apps/algolia/lib/algolia/api.ex
@@ -8,6 +8,8 @@ defmodule Algolia.Api do
 
   defstruct [:host, :index, :action, :body]
 
+  @type action :: :post | :delete
+
   @http_pool Application.get_env(:algolia, :http_pool)
 
   @type t :: %__MODULE__{
@@ -17,24 +19,33 @@ defmodule Algolia.Api do
           body: String.t() | nil
         }
 
-  @spec post(t) :: {:ok, HTTPoison.Response.t()} | {:error, HTTPoison.Error.t() | :bad_config}
-  def post(opts, config \\ nil)
+  @spec action(action, t) ::
+          {:ok, HTTPoison.Response.t()} | {:error, HTTPoison.Error.t() | :bad_config}
+  def action(action, opts, config \\ nil)
 
-  def post(%__MODULE__{} = opts, nil) do
-    post(opts, Config.config())
+  def action(action, %__MODULE__{} = opts, nil) do
+    action(action, opts, Config.config())
   end
 
-  def post(%__MODULE__{} = opts, %Config{write: <<_::binary>>, app_id: <<_::binary>>} = config) do
-    do_post(opts, config)
+  def action(
+        action,
+        %__MODULE__{} = opts,
+        %Config{write: <<_::binary>>, app_id: <<_::binary>>} = config
+      ) do
+    do_action(action, opts, config)
   end
 
-  def post(%__MODULE__{}, %Config{} = config) do
+  def action(_action, %__MODULE__{}, %Config{} = config) do
     _ = Logger.warn("module=#{__MODULE__} missing Algolia config keys: #{inspect(config)}")
     {:error, :bad_config}
   end
 
-  defp do_post(%__MODULE__{index: index, action: action, body: body} = opts, %Config{} = config)
-       when is_binary(index) and is_binary(action) and is_binary(body) do
+  defp do_action(
+         action,
+         %__MODULE__{index: index, action: opts_action, body: body} = opts,
+         %Config{} = config
+       )
+       when is_binary(index) and is_binary(opts_action) and is_binary(body) do
     hackney =
       opts
       |> hackney_opts()
@@ -44,7 +55,7 @@ defmodule Algolia.Api do
       response =
         opts
         |> generate_url(config)
-        |> HTTPoison.post(body, headers(config), hackney: hackney)
+        |> send_request(action, body, config, hackney)
 
       case response do
         {:ok, %HTTPoison.Response{status_code: 200}} -> response
@@ -61,6 +72,12 @@ defmodule Algolia.Api do
       send_post_request.({body, config})
     end
   end
+
+  def send_request(url, :post, body, config, hackney),
+    do: HTTPoison.post(url, body, headers(config), hackney: hackney)
+
+  def send_request(url, :delete, _body, config, hackney),
+    do: HTTPoison.delete(url, headers(config), hackney: hackney)
 
   @spec generate_url(t, Config.t()) :: String.t()
   defp generate_url(%__MODULE__{} = opts, %Config{} = config) do
--- a/apps/algolia/lib/algolia/api.ex
+++ b/apps/algolia/lib/algolia/api.ex
@@ -66,7 +66,7 @@ defmodule Algolia.Api do
     # If we're making a query for results using the same request body AND same
     # %Algolia.Config{}, cache the response instead of making extra calls to the
     # Algolia REST API
-    if action == "queries" do
+    if opts_action == "queries" do
       cache({body, config}, send_post_request)
     else
       send_post_request.({body, config})
--- a/apps/algolia/lib/algolia/api.ex
+++ b/apps/algolia/lib/algolia/api.ex
@@ -8,7 +8,7 @@ defmodule Algolia.Api do
 
   defstruct [:host, :index, :action, :body]
 
-  @type action :: :post | :delete
+  @type action :: :post | :get
 
   @http_pool Application.get_env(:algolia, :http_pool)
 
@@ -76,8 +76,8 @@ defmodule Algolia.Api do
   def send_request(url, :post, body, config, hackney),
     do: HTTPoison.post(url, body, headers(config), hackney: hackney)
 
-  def send_request(url, :delete, _body, config, hackney),
-    do: HTTPoison.delete(url, headers(config), hackney: hackney)
+  def send_request(url, :get, _body, config, hackney),
+    do: HTTPoison.get(url, headers(config), hackney: hackney)
 
   @spec generate_url(t, Config.t()) :: String.t()
   defp generate_url(%__MODULE__{} = opts, %Config{} = config) do
--- a/apps/algolia/lib/algolia/api.ex
+++ b/apps/algolia/lib/algolia/api.ex
@@ -6,7 +6,7 @@ defmodule Algolia.Api do
   require Logger
   use RepoCache, ttl: :timer.hours(12)
 
-  defstruct [:host, :index, :action, :body]
+  defstruct [:host, :index, :action, :body, :query_params]
 
   @type action :: :post | :get
 
@@ -16,7 +16,8 @@ defmodule Algolia.Api do
           host: String.t() | nil,
           index: String.t() | nil,
           action: String.t() | nil,
-          body: String.t() | nil
+          body: String.t() | nil,
+          query_params: map | nil
         }
 
   @spec action(action, t) ::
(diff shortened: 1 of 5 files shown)


apps/algolia/lib/algolia/update.ex
```diff
@@ -32,6 +32,12 @@ defmodule Algolia.Update do
     |> do_has_routes?(data)
   end
 
+  defp send_delete_and_update(update_object, base_url, index_module) do
+    with :ok <- clear_index(update_object, base_url, index_module) do
```
kotva006, line 36: question: Does it make sense to only send the update if the `clear` command is successful?

apps/algolia/lib/algolia/api.ex
```diff
@@ -8,6 +8,8 @@ defmodule Algolia.Api do
 
   defstruct [:host, :index, :action, :body]
 
+  @type action :: :post | :delete
+
```
kotva006, line 12: Changes to this file aren't required for the code change.  I modified this file under the impression that I would have to make other network calls outside of `POST`.  I decided to leave it in as it allows us to easily add support for `GET` in the future if we need it

thecristen: I like the sentiment but this implementation strikes me as incomplete so I'm leaning towards removing. It's a good idea, if we do change this to make other types of Algolia REST API calls it should be changed more substantively to include more verbs and cover more endpoints.

apps/algolia/lib/algolia/update.ex
thecristen: It doesn't seem to make sense to _not_ do that. But I do think the `clear` endpoint is overkill here.

thecristen (changes requested):
It was mentioned with the initial proposal that deleting-everything-then-adding-records is risky in that we could theoretically end up with an empty index if the deletion succeeds and the addition fails, and that still holds here.

I think there's a simpler approach you can take here, where you can compare the stop/route IDs in the records-to-be-added against those in the records which currently exist in the Algolia indexes, and use that to do the individual updates (if needed) and deletes (if needed) in a single request.

apps/algolia/lib/algolia/update.ex
```diff
@@ -32,10 +61,28 @@ defmodule Algolia.Update do
     |> do_has_routes?(data)
   end
 
+  defp current_objects(base_url, index_module) do
+    with {:ok, body} <- get_current_objects(base_url, index_module) do
+      response_to_objects(body)
```
thecristen, line 66: Looks like we'll need to do some additional processing here (or in `response_to_objects/1`) to get _all_ the objects.

The results returned from Algolia are paginated - so, requesting the stops returns 50 pages of results, each listing 20 hits. This can be read directly from the response object, which includes in addition to `hits`:

```
"nbHits": 8074,
"page": 0,
"nbPages": 50,
"hitsPerPage": 20,
```

Navigating this is handled by query parameters, e.g. `hitsPerPage=1000` can be used to change the default (1000 is the max) and `&page=3` will go to page 3 of the results.

thecristen (changes requested):

apps/algolia/lib/algolia/update.ex
thecristen: Are you sure? It returns all the records but is still paginated. Improvement - it appears to automatically set it to the max page size.

```
"nbHits": 8074,
"page": 0,
"nbPages": 9,
"hitsPerPage": 1000,
```

anthonyshull:
Coming into this work not knowing what it was doing. It would be really great to moduledocs and docstrings for all public functions.

apps/algolia/lib/algolia/update.ex
```diff
@@ -13,14 +13,43 @@ defmodule Algolia.Update do
...
+
+    with {:ok, _body} <- clean_index(new_objects, index_module, base_url),
+         {:ok, _body} <- update_index(new_objects, index_module, base_url) do
+      :ok
+    end
+  end
+
+  @spec clean_index(Enum.t(), atom, String.t()) :: success | error
+  def clean_index(new_objects, index_module, base_url) do
+    current_objs = current_objects(base_url, index_module)
```
anthonyshull, line 34: Small thing, but above we call them new_objects and here current_objs. I'm guessing that is because current_objects is a function. But, it indicates we need some better naming maybe:

```
current_objects = get_current_objects(base_url, index_module)
```

But, I see there is also a get_current_objects. So, maybe combine what the two are doing. Currently, it's hard to read.

apps/algolia/lib/algolia/update.ex
```diff
@@ -13,14 +13,43 @@ defmodule Algolia.Update do
...
+
+    current_objs
+    |> filter_objects(new_objects)
+    |> Enum.map(&build_action_object(&1, "deleteObject"))
+    |> build_request_object()
+    |> send_update(base_url, index_module)
+  end
+
+  defp filter_objects(current_objects, new_objects) do
+    Enum.filter(current_objects, fn co ->
```
anthonyshull, line 44: I don't think we should ever use non-descriptive names...`co` for `current_object`.

apps/algolia/lib/algolia/update.ex
```diff
@@ -13,14 +13,43 @@ defmodule Algolia.Update do
...
+      Enum.find_index(new_objects, fn no -> no.objectID == co.objectID end) == nil
```
anthonyshull, line 45: Same with `no` obviously.

apps/algolia/lib/algolia/update.ex
```diff
@@ -49,35 +96,38 @@ defmodule Algolia.Update do
...
+    {:ok, json} = Jason.decode(body)
+    Enum.map(json["hits"], fn obj -> %{objectID: obj["objectID"]} end)
+  end
+
   @spec parse_response({:ok, HTTPoison.Response.t()} | {:error, HTTPoison.Error.t()}) ::
           success | error
-  defp parse_response({:ok, %HTTPoison.Response{status_code: 200}}), do: :ok
+  defp parse_response({:ok, %HTTPoison.Response{status_code: 200, body: body}}), do: {:ok, body}
   defp parse_response({:ok, %HTTPoison.Response{} = response}), do: {:error, response}
   defp parse_response({:error, %HTTPoison.Error{} = error}), do: {:error, error}
```
anthonyshull, line 115: Shouldn't this go in the Api module? Seems to be more related to requests in general than specific to updating indexes.

apps/algolia/lib/algolia/update.ex
```diff
@@ -49,35 +96,38 @@ defmodule Algolia.Update do
...
+    :post |> Api.action(opts) |> parse_response()
   end
 
   defp send_update({:error, error}, _, _) do
     {:error, {:json_error, error}}
   end
 
+  defp response_to_objects(body) do
+    {:ok, json} = Jason.decode(body)
+    Enum.map(json["hits"], fn obj -> %{objectID: obj["objectID"]} end)
```
anthonyshull, line 108: Same with naming something `obj`.

apps/algolia/lib/algolia/update.ex
```diff
@@ -49,35 +96,38 @@ defmodule Algolia.Update do
...
   defp parse_response({:error, %HTTPoison.Error{} = error}), do: {:error, error}
 
   defp build_request_object(data) do
     Poison.encode(%{requests: data})
   end
 
-  @spec build_data_object(Algolia.Object.t()) :: map
-  def build_data_object(data) do
+  @spec build_action_object(Algolia.Object.t(), String.t()) :: map
+  def build_action_object(data, action) do
```
anthonyshull, line 122: This might not be the best naming. It's not building an `Algolia.Object.t()`. It takes one as an argument and returns a map. The name of the method seems to indicate that it is going to take an object and an action and return me a special type of object.

apps/algolia/lib/algolia/update.ex
```diff
@@ -49,35 +96,38 @@ defmodule Algolia.Update do
...
-      body: do_build_data_object(data)
+      action: action,
+      body: data
     }
   end
 
-  @spec do_build_data_object(Algolia.Object.t()) :: map
-  defp do_build_data_object(data) do
+  @spec to_data_object(Algolia.Object.t()) :: map
+  def to_data_object(data) do
```
anthonyshull, line 130: Same as above.

anthonyshull (commented):
There seem to be a lot of patterns going on in this file. Should we clean some of them up? I think it is coming from complecting code that is general to using any kind of HTTP service and code that is specific to updating this index. Functions like `has_routes?` and `do_has_routes?`. Assuming `has_routes?` is doing something.

anthonyshull:
Another note is that there are a lot of functions in this module that are public, but I don't think they have any external callers.

apps/algolia/lib/algolia/update.ex
kotva006: I will work on renaming.  `get_current_objects` is the name for the function that gets all the objects from algiolas endpoint

kotva006:
> tterns going on in this file. Should we clean some of them up? I think it is coming from complecting code that is general to using any kind of HTTP service

The functions in question were existing logic.  I have not modified them.  I can change their name if they are confusing.

apps/algolia/lib/algolia/update.ex
kotva006: I will change them.  The names were a singular simplification of the enumberables name.  I have always used them when it comes to simple one line loops, as the the enumerable in question is very close to the logic that consumes it.  I agree that non-descriptive names shouldn't be used on more complex loops.

apps/algolia/lib/algolia/update.ex
thecristen: I wonder if it was that way because we'd wanted to parse responses differently when we're searching for results vs. here where we're manipulating the data. 🤔

apps/algolia/lib/algolia/update.ex
kotva006: I believe the typing is this way on purpose.  This function returns an Algolia object.  The route algolia object and stop algolia object both have different fields.  Since both share a common type of Map, that was used.  I don't think this justifies renaming the function, or changing the typing.

apps/algolia/lib/algolia/update.ex
kotva006: I believe the typing is this way on purpose. This function returns an Algolia object. The route algolia object and stop algolia object both have different fields. Since both share a common type of Map, that was used. I don't think this justifies renaming the function, or changing the typing.

apps/algolia/lib/algolia/api.ex
```diff
@@ -6,45 +6,59 @@ defmodule Algolia.Api do
...
+         %Config{} = config
+       )
+       when is_binary(index) and is_binary(opts_action) and is_binary(body) do
     hackney =
       opts
       |> hackney_opts()
       |> Keyword.put(:pool, @http_pool)
 
     send_post_request = fn {body, config} ->
+      query_param_string = generate_query_param_string(opts)
```
thecristen, line 56: nit: I feel like this should be done inside the `generate_url` function, as it's part of generating the URL!

thecristen (approved):
let's give this a try! 🚀

PR #1823, opened 2023-12-13
Centralized CMS Caching

anthonyshull:
We're using a centralized Redis cache. We use the Nebulex library with the Redis adapter. We also use Telemetry to get cache stats directly from the adapter.

We have two endpoints, protected by basic auth, that allow us to expire cache keys.


apps/cms/lib/repo.ex
```diff
@@ -182,6 +189,28 @@ defmodule CMS.Repo do
...
   end
 
+  # BEGIN PAGE CACHING #
+
+  @behaviour Nebulex.Caching.KeyGenerator
+
+  @impl true
+  def generate(_, _, [path, %{}]) do
+    "/cms/#{String.trim(path, "/")}"
+  end
```
anthonyshull, line 199: Creates a key from the path by preceding it with `/cms`.

apps/cms/lib/repo.ex
```diff
@@ -182,6 +189,28 @@ defmodule CMS.Repo do
...
+
+  @impl true
+  def generate(_, _, [path, params]) do
+    "/cms/#{String.trim(path, "/")}" <> params_to_string(params)
+  end
```
anthonyshull, line 204: Creates a key like above, but also handles parameters so you can have something like `/cms/foo?bar=baz&bop=bam`

apps/cms/mix.exs
```diff
@@ -47,16 +47,22 @@ defmodule CMS.Mixfile do
...
       {:phoenix_html, "~> 3.0"},
-      {:repo_cache, in_umbrella: true},
+      {:plug, "~> 1.14.2"},
+      {:poison, ">= 0.0.0", override: true},
+      {:quixir, "~> 0.9", only: :test},
+      {:telemetry, "0.4.3"},
+      {:telemetry_metrics, "0.6.1"},
+      {:telemetry_metrics_statsd, "0.7.0"},
+      {:telemetry_poller, "0.5.1"},
+      {:timex, ">= 0.0.0"},
```
anthonyshull, line 65: Alphabetized these.

apps/cms/test/repo_test.exs
```diff
@@ -61,30 +67,15 @@ defmodule CMS.RepoTest do
...
-      case ConCache.get(Repo, cache_key) do
-        nil ->
-          :ok
-
-        {:ok, %{"type" => [%{"target_id" => "news_entry"}]}} ->
-          ConCache.dirty_delete(Repo, cache_key)
-      end
-
-      assert %NewsEntry{} = Repo.get_page(path, params)
-      assert {:ok, %{"type" => [%{"target_id" => "news_entry"}]}} = ConCache.get(Repo, cache_key)
```
anthonyshull, line 78: I just removed this for now. The final PR will have added tests.

apps/cms/test/repo_test.exs
```diff
@@ -61,30 +67,9 @@ defmodule CMS.RepoTest do
...
     end
 
     test "does not cache previews" do
-      path = "/basic_page_no_sidebar"
-      params = %{"preview" => "", "vid" => "112", "nid" => "6"}
-      cache_key = {:view_or_preview, path: path, params: params}
-      assert ConCache.get(Repo, cache_key) == nil
-      assert %Basic{} = Repo.get_page(path, params)
-      assert ConCache.get(Repo, cache_key) == nil
```
anthonyshull, line 87: Same here. This test will be back later.

apps/cms/lib/repo.ex
```diff
@@ -29,6 +33,8 @@ defmodule CMS.Repo do
 
   @cms_api Application.get_env(:cms, :cms_api)
 
+  @ttl :timer.hours(1)
```
anthonyshull, line 36: We set a TTL because we don't have a sophisticated mechanism to handle failed requests from Drupal to Dotcom.

paulswartz: question: If you're increasing the TTL to an hour, do need any of the Nebulex changes? It seems like that alone would make a big difference in the number of cache hits.

apps/cms/lib/telemetry/reporter.ex
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  @impl true
+  def init(metrics) do
+    Process.flag(:trap_exit, true)
+
+    groups = Enum.group_by(metrics, & &1.event_name)
+
+    for {event, metrics} <- groups do
+      :telemetry.attach({__MODULE__, event, self()}, event, &__MODULE__.handle_event/4, metrics)
```
paulswartz, line 23: question: do you need a GenServer for this? I don't see it maintaining any state, so the metric source could do the logging directly?

apps/site/lib/site_web/controllers/cms_controller.ex
```diff
@@ -43,6 +43,18 @@ defmodule SiteWeb.CMSController do
     |> handle_page_response(conn)
   end
 
+  def reset_cache_key(conn, %{"object" => object, "id" => id}) do
+    CMS.Repo.delete("/cms/#{object}/#{id}")
```
paulswartz, line 47: question: is there a CMS PR for this, or is this built-in Drupal functionality that we're using?

apps/site/lib/site_web/router.ex
```diff
@@ -264,6 +271,13 @@ defmodule SiteWeb.Router do
     get("/*path", CMSController, :page)
   end
 
+  defp basic_auth(conn, _) do
+    username = System.fetch_env!("BASIC_AUTH_USERNAME")
+    password = System.fetch_env!("BASIC_AUTH_PASSWORD")
+
+    Plug.BasicAuth.basic_auth(conn, username: username, password: password)
```
paulswartz, line 278: suggestion: generally we put the environment variable values into the application config (`config/runtime.exs`) and then read it as necessary.
```elixir
# config/runtime.exs
config :site, SiteWeb.Router,
  cms_basic_auth: [
    username: System.fetch_env!("BASIC_AUTH_USERNAME"),
    password: System.fetch_env!("BASIC_AUTH_PASSWORD")
  ]

# router.ex
opts = Application.get_env(:site, SiteWeb.Endpoint)[:cms_basic_auth]
Plug.BasicAuth.basic_auth(conn, opts)
```

paulswartz (commented):
general question: how does Nebulex handle things like Redis going down? That isn't something we've needed to address with the in-memory cache.
general question: what's the performance like? we're adding both a network round trip, as well as a serialize/deserialize step.

apps/cms/lib/repo.ex
anthonyshull: it will, but the content creators need the ability to reset a key so that content gets pushed more quickly. the ttl has been one minute so that they see the content change quickly. this lets us actually cache the content, but also update it quickly...something we can't do without significant work using ets and elixir nodes that don't talk to each other currently.

anthonyshull:
> general question: how does Nebulex handle things like Redis going down? That isn't something we've needed to address with the in-memory cache. general question: what's the performance like? we're adding both a network round trip, as well as a serialize/deserialize step.

redix handles reconnections. https://github.com/whatyouhide/redix?tab=readme-ov-file#features if there is some kind of serious error, the site wouldn't function. but, the same could be said about ec2 going down or postgres going down, or route 53, etc.

we won't have performance numbers until we're actually pushing data through the system. but i highly suspect that it's going to be much faster given the fact that it current cache hit rate is 50%.

apps/site/lib/site_web/controllers/cms_controller.ex
anthonyshull: https://github.com/mbta/cms/pull/409

apps/site/lib/site_web/router.ex
anthonyshull: yeah, I have to make another PR into devops to get these environment variables. i just wanted to make sure everything was actually going to work before finalizing on even using basic auth.

apps/cms/lib/telemetry/reporter.ex
anthonyshull: afaik, this is how you're supposed to write custom telemetry reporters. it's how the other reporter we're using telemetry_metrics_statsd works: https://github.com/beam-telemetry/telemetry_metrics_statsd/blob/main/lib/telemetry_metrics_statsd.ex. and the console reporter that telemetry metrics ships with: https://github.com/beam-telemetry/telemetry_metrics/blob/main/lib/telemetry_metrics/console_reporter.ex. the idea behind telemetry is to separate out the metric emitters from reporters.

apps/cms/config/config.exs
```diff
@@ -37,4 +37,12 @@ config :cms,
...
 config :cms, :cms_api, CMS.API.HTTPClient
 
-import_config "#{Mix.env()}.exs"
+config :cms, CMS.Repo,
+  conn_opts: [
+    host: "127.0.0.1",
+    port: 6379
+  ],
+  stats: true,
+  telemetry: true
```
anthonyshull, line 46: this will be different for prod. i just don't know how until we can get something running on aws.

apps/cms/lib/repo.ex
anthonyshull: the ideal state would actually to use a message queue instead of sending patch requests. now that i know terraform...lookout. ultimately, there is no way around the two generals problem. so setting some ttl is a safety net.

paulswartz:
> if there is some kind of serious error, the site wouldn't function. but, the same could be said about ec2 going down or postgres going down, or route 53, etc.

That's a great point, and why it's good to avoid introducing new external dependencies to the critical path. Something to think about!

PR #1830, opened 2023-12-16
Centralized CMS caching 1

anthonyshull:
This PR aims to prove that a Redis connection has been made. It accepts requests to `/cms/foo/bar` (which will be made by the CMS). If it can connect to Redis, it will delete the corresponding key and log a notice. If it cannot connect to Redis, it will log a warning. It returns a 202 to the requestor regardless. Ergo, we'll know whether or not Redis is running, but incur no risk.

https://app.asana.com/0/555089885850811/1206129599937934/f


apps/cms/mix.exs
```diff
@@ -47,16 +47,17 @@ defmodule CMS.Mixfile do
...
-      {:timex, ">= 0.0.0"},
-      {:plug, "~> 1.14.2"},
-      {:html_sanitize_ex, "1.3.0"},
       {:bypass, "~> 1.0", only: :test},
-      {:quixir, "~> 0.9", only: :test},
+      {:html_sanitize_ex, "1.3.0"},
+      {:httpoison, ">= 0.0.0"},
       {:mock, "~> 0.3.3", only: :test},
+      {:nebulex, "2.5.2"},
+      {:nebulex_redis_adapter, "2.3.1"},
```
anthonyshull, line 55: These are the only two new deps. Everything else is just alphabetization.

apps/site/lib/site_web/controllers/cms_controller.ex
```diff
@@ -43,6 +43,22 @@ defmodule SiteWeb.CMSController do
     |> handle_page_response(conn)
   end
 
+  def reset_cache_key(conn, %{"object" => object, "id" => id}) do
+    CMS.RedisRepo.delete("/cms/#{object}/#{id}")
+
+    Logger.info("cms.redisrepo.delete path=#{object}/#{id}")
```
anthonyshull, line 49: We'll know that this endpoint is being hit by Drupal once this PR and that PR are promoted.

apps/site/config/test.exs
```diff
@@ -14,7 +14,7 @@ config :site, :secure_pipeline,
   ]
 
 # Print only warnings and errors during test
-config :logger, level: :warn
+config :logger, level: :notice
```
kotva006, line 17: If this change was intentional the comment above should be updated to reflect that

apps/site/lib/site_web/controllers/cms_controller.ex
```diff
@@ -43,6 +45,22 @@ defmodule SiteWeb.CMSController do
     |> handle_page_response(conn)
   end
 
+  def reset_cache_key(conn, %{"object" => object, "id" => id}) do
+    @cache.delete("/cms/#{object}/#{id}")
```
kotva006, line 49: Does that status of these requests need to be checked?  It seems this could open silent errors if the cache request failed, but we logged it and returned it as a success anyway

apps/site/lib/site_web/router.ex
```diff
@@ -51,6 +51,13 @@ defmodule SiteWeb.Router do
     get("/_health", HealthController, :index)
   end
 
+  scope "/cms", SiteWeb do
```
kotva006, line 54: We have been trying to put new data endpoints behind the `/api` path for organization purposes.

config/runtime.exs
```diff
@@ -0,0 +1,23 @@
+import Config
+
+config :cms, CMS.Cache,
+  conn_opts: [
+    host: System.get_env("REDIS_HOST", "127.0.0.1"),
```
kotva006, line 5: nit-pick: This should environment variable should be added to the `.envrc.template` file

config/runtime.exs
```diff
@@ -0,0 +1,23 @@
...
+  config :site, SiteWeb.Router,
+    cms_basic_auth: [
+      username: "username",
+      password: "password"
+    ]
+else
+  config :site, SiteWeb.Router,
+    cms_basic_auth: [
+      username: System.fetch_env!("CMS_BASIC_AUTH_USERNAME"),
+      password: System.fetch_env!("CMS_BASIC_AUTH_PASSWORD")
```
kotva006, line 21: Same as above

apps/site/lib/site_web/controllers/cms_controller.ex
anthonyshull: No, the reason we're returning a 202 is to indicate that we accepted the request. We aren't indicating whether or not it was a success (200).

apps/site/lib/site_web/router.ex
anthonyshull: I didn't do that because we're using basic auth and this route isn't for API consumers.

kotva006 (approved):

config/runtime.exs
```diff
@@ -0,0 +1,15 @@
...
+  conn_opts: [
+    host: System.get_env("REDIS_HOST", "127.0.0.1"),
+    port: 6379
+  ],
+  stats: false,
+  telemetry: false
+
+config :site, SiteWeb.Router,
+  cms_basic_auth: [
+    username: System.get_env("CMS_BASIC_AUTH_USERNAME", "username"),
```
paulswartz, line 13: suggestion: can we make this a safer default? Since this is an open-source project, anyone not setting these variables (including us) will have an easily guessable password.

My vote would be to return `403 Unauthorized` for everything if the credentials are unset, but you could also generate a random username/password at startup and log it if you wanted.

paulswartz (changes requested):

config/runtime.exs
anthonyshull: @paulswartz I just changed this so I can test it in dev-blue because I am still waiting for these to be set in my devops PR. In the second PR, it fails if these aren't set.

https://github.com/mbta/devops/pull/1434

https://github.com/mbta/dotcom/pull/1832/files#diff-e43c5cbf91db7a8062b6cb860cbf118296c1b4c7ee32fdcf702e54234ba38092

paulswartz (commented):
question: do the new environment variables need to be added to ENVIRONMENT.md since they're required?

PR #1832, opened 2023-12-18
Centralized CMS caching 2

anthonyshull:
https://app.asana.com/0/555089885850811/1206129599937934/f


apps/cms/lib/repo.ex
```diff
@@ -50,20 +56,19 @@ defmodule CMS.Repo do
 
   @spec news_entry_by(Keyword.t()) :: NewsEntry.t() | :not_found
   def news_entry_by(opts) do
```
kotva006, line 58: Is there a reason this page isn't cached?

anthonyshull: Yeah, this goes to a "view" in Drupal. We don't have a mechanism for the CMS to tell us that a view has been updated. Though, I think it's possible. So, for now we're removing the caching.

anthonyshull:
> The changes here look fine. Though it would be nice if this was pointing at the first pull request seeing as the changes are duplicated here (and aren't reflecting the feedback from the first pull request)

The differences will shake out when the first PR is merged into master.

thecristen:
Complete in #1856

PR #1833, opened 2023-12-18
chore: upgrade erlang to OTP 24

thecristen:
I can finally compile this with [wxWidgets](https://formulae.brew.sh/formula/wxwidgets) and run `:observer` woohoo!!!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 ###
 
 # 1.) Get the Elixir dependencies within an Elixir container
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -61,7 +61,7 @@ RUN mix distillery.release --verbose
 
 # 4) Use the nodejs container for the runtime environment
 # Since we're server-rendering the React templates, we need a Javascript engine running inside the container.
-FROM node:14.20.0-buster-slim
+FROM node:18.17.1-buster-slim
```
kotva006, line 64: good catch

kotva006 (approved):

PR #1834, opened 2023-12-18
feat: enable LiveDashboard

thecristen:
Sets up a working [LiveDashboard](https://hexdocs.pm/phoenix_live_dashboard/Phoenix.LiveDashboard.html).

I needed to upgrade Poison from version 1 to 3 to get it to be compatible.

Also includes a barebones telemetry module.

It's set to only be accessible in `MIX_ENV=dev`. I considered setting up some basic authentication so we can stand it up in prod, but ran into an issue with `Plug.BasicAuth` plug and ran into a weird issue around module namespacing... let's just add that later.


apps/site/lib/site_web/telemetry.ex
```diff
@@ -0,0 +1,65 @@
...
+      # every 10_000ms. Learn more here: https://hexdocs.pm/telemetry_metrics
+      {:telemetry_poller, measurements: periodic_measurements(), period: 10_000}
+      # Add reporters as children of your supervision tree.
+      # {Telemetry.Metrics.ConsoleReporter, metrics: metrics()}
+    ]
+
+    Supervisor.init(children, strategy: :one_for_one)
+  end
+
+  def metrics do
```
anthonyshull, line 22: This isn't being called anywhere. Do we need it?

apps/site/lib/site_web/telemetry.ex
```diff
@@ -0,0 +1,65 @@
...
+  def start_link(arg) do
+    Supervisor.start_link(__MODULE__, arg, name: __MODULE__)
+  end
+
+  @impl true
+  def init(_arg) do
+    children = [
+      # Telemetry poller will execute the given period measurements
+      # every 10_000ms. Learn more here: https://hexdocs.pm/telemetry_metrics
+      {:telemetry_poller, measurements: periodic_measurements(), period: 10_000}
```
anthonyshull, line 14: This is just an empty list. Do we need it?

apps/site/lib/site_web/telemetry.ex
```diff
@@ -0,0 +1,65 @@
+defmodule SiteWeb.Telemetry do
```
anthonyshull, line 1: We should have a `@moduledoc` for every module explaining what it does.

apps/site/lib/site_web/telemetry.ex
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  @impl true
+  def init(_arg) do
+    children = [
+      # Telemetry poller will execute the given period measurements
+      # every 10_000ms. Learn more here: https://hexdocs.pm/telemetry_metrics
+      {:telemetry_poller, measurements: periodic_measurements(), period: 10_000}
+      # Add reporters as children of your supervision tree.
+      # {Telemetry.Metrics.ConsoleReporter, metrics: metrics()}
```
anthonyshull, line 16: Without the reporter, this module isn't outputing anything.

anthonyshull (changes requested):

anthonyshull (approved):

PR #1838, opened 2023-12-20
fix(Predictions.Stream): handle reset and remove events better

thecristen:
#### Summary of changes

**Asana Ticket:** [Predictions.Store | remove predictions on :reset event](https://app.asana.com/0/385363666817452/1206160446331599/f) 

- Handles reset events by removing the relevant predictions before adding new predictions to `Predictions.Store`
- Fixes parsing to correctly handle the `:remove` event, as discovered during  [Stop page displaying mystery extra Red Line prediction](https://app.asana.com/0/385363666817452/1205862370279611/f)

---

#### General checks

### DIFF
--- a/apps/predictions/lib/stream.ex
+++ b/apps/predictions/lib/stream.ex
@@ -8,6 +8,7 @@ defmodule Predictions.Stream do
 
   alias Phoenix.PubSub
   alias Predictions.{Repo, Store, StreamParser}
+  alias V3Api.Stream.Event
 
   @type event_type :: :reset | :add | :update | :remove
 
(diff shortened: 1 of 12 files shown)


apps/predictions/lib/stream.ex
```diff
@@ -29,17 +30,47 @@ defmodule Predictions.Stream do
...
+      |> Enum.flat_map(fn %Event{data: %JsonApi{data: data}} ->
+        data
+        |> Enum.filter(&(&1.type == "prediction"))
+        |> Enum.map(& &1.id)
+      end)
+
+    Store.update({:remove, prediction_ids})
+  end
+
+  defp handle_by_type({event_type, events}, _) when event_type in [:add, :update] do
```
thecristen, line 69: note: because of the type of ETS table the `Predictions.Store` is, both `:add` and `:update` events will have the same result (if you add a record that already exists, it gets updated)

apps/predictions/lib/stream.ex
```diff
@@ -29,17 +30,47 @@ defmodule Predictions.Stream do
...
+    Store.clear(clear_keys)
+    handle_by_type({:add, events}, clear_keys)
+  end
+
+  defp handle_by_type({:remove, events}, _) do
+    prediction_ids =
+      events
+      |> Enum.flat_map(fn %Event{data: %JsonApi{data: data}} ->
+        data
+        |> Enum.filter(&(&1.type == "prediction"))
```
thecristen, line 62: note: Because the V3 API request has `included` data, the event data here can sometimes be of other types (trip, stop). We don't need to those in order to remove the prediction from the table.

PR #1854, opened 2024-01-17
update(Elixir): Update Elixir to 1.16 and deps upgrade

kotva006:
#### Summary of changes
Updates Elixir to 1.16
Upgrades Various dependencies to remove some warnings and a handful of errors.

Notes about existing dependency warnings.  None of these should affect the code:

`:rstar` - hasn’t been updated in years.  Library is used in 1 location.  Possible candidate for removal

`:phoenix_live_reload` - on latest version, warning about wrong logger function used.  Should have no impact

`:nebulex_redis_adapter` - on latest version, warning about unused alias.  Should have no impact

`:server_sent_event_stage` - on latest version, warning about wrong logger function used.  Should have no impact.

`:tesla` - generates warning for unused import and for using `Application.get_env` instead of `Application.compile_env`.  This library is used by :wallaby -> :web_driver_client.  Latest version of :wallaby does not update :web_driver_client so warning will still exist with update.  This warning should have no impact

`:finch` - uses wrong logger warn function, used by :tesla so has same issue as above.  Warning should have no impact

`:polyline` - uses `use Bitwise` which is deprecated.  Should be `import Bitwise`.  This is on the latest version.  This warning should have no impact

`:html_sanitize_ex` - our custom scrubber throws a warning of an unused function (scrub_css).  Not sure if the `.scrub` function might use it.  Should affect nothing

`:credo` - uses `use Bitwise` instead of import, and uses a deprecated function `regex?` instead of `is_struct`.  These warnings should have no impact.

`:recaptcha` - uses `get_env` in a module body which throws a warning.  This library hasn’t been updated in 3 years, and the issue still persists in the latest release.  This warning should affect nothing

`:mint` - has unused function.  It is used by :mint_web_socket which is used by `:slipstream` which is used by `:absinthe_client`.  `:absinthe_client` is on the latest version.  This warning should have no affect

`:dialyixr` - there is a ton of failings on CI still.  Keeping update but future task.

`:benchfella` - warnings about ambiguous float match, and deprecated functions used.  This is on the latest version.  This should not affect anything  :poison - uses deprecated `use` instead of `import`. This is the latest version.  This should not affect anything.

`:sizable` - uses the deprecated `Logger.warn`, and ambiguous float match.  This is the latest version.  It should not affect anything

`:diskusage_logger` - uses a deprecated `worker` function.  It is on the latest version.  MBTA owns this dependency so we could update it.  This should affect nothing.

`:bunt` - part of :credo.  Uses `get_env` instead of `compile_env`.  This warning should affect nothing

`:poison` - latest version removes a function relied upon Phoenix for json encoding.  Existing warnings should have no effect

**Asana Ticket:** [Upgrade Elixir to 1.13 or higher](https://app.asana.com/0/555089885850811/1205695443846917)

---

#### General checks


lib/dotcom_web/plugs/uri_checker.ex
kotva006, line 1: This plug has been removed as Elixir version 13 no longer supports the behavior this plug relied on.
This plug ran the URI parts into the `URI.decode/1` function.  The intent of this function is to convert the percent escaped characters into their string parts.  IE `%2B -> +`.  [In version 12](https://github.com/elixir-lang/elixir/blob/v1.12/lib/elixir/lib/uri.ex#L415) if this function found a percent that didn't match a hex pattern, it threw an exception.  Since [version 13](https://github.com/elixir-lang/elixir/blob/75ac733bb5ac77c934388e04b1527d1d68952d40/lib/elixir/lib/uri.ex#L482) Elixir goes on its merry way and continues to decode the string when it finds a percent sign by a non hex character.

The intent of this plug was to find URIs that had bad percent encoded parts, and return a 404 before it hit the router.  I don't think this is best behavior.  We should let the router try to match it (bad `%` encodings will not match anything) and let the router handle the 404.  Elixir seems to agree this behavior should be handled outside of the decode function instead.  As there is no function I could find that still had similar functionality without checking more of the URI (I found suggestions running the whole URI into URL.new/1 which seemed like overkill).

.tool-versions
```diff
@@ -1,3 +1,3 @@
 nodejs 18.17.1
 erlang 24.3.4.14
```
anthonyshull, line 2: We could probably upgrade to Erlang 26.

anthonyshull (commented):
We should also make sure to check that all of the README information is accurate.

.tool-versions
kotva006: I am thinking that should be done later

lib/algolia/analytics.ex
```diff
@@ -54,15 +54,15 @@ defmodule Algolia.Analytics do
...
+      Logger.warning(
         "module=#{__MODULE__} Bad response from Algolia: #{inspect(response)} request body: #{body}"
       )
 
     {:error, response}
   end
 
   defp handle_click_response({:error, %HTTPoison.Error{} = response}, _) do
-    _ = Logger.warn("module=#{__MODULE__} Error connecting to Algolia: #{inspect(response)}")
+    _ = Logger.warning("module=#{__MODULE__} Error connecting to Algolia: #{inspect(response)}")
```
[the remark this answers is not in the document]
kotva006: Credo complains

mix.exs
```diff
@@ -74,47 +74,48 @@ defmodule DotCom.Mixfile do
       {:briefly, "~> 0.3"},
       {:bypass, "~> 1.0", [only: :test]},
       {:castore, "~> 0.1.11"},
-      {:con_cache, "~> 0.12.0"},
+      {:con_cache, "~> 1.0.0"},
```
anthonyshull, line 77: I thought the plan was to pin dependencies to a specific version. Could just take the version from the lock file and put it in here.

kotva006: We never decided on a solution. Just commented on the semvar behavior. I can lock these dependencies to one version if that is desired.

anthonyshull: Yeah, we should. To wit, `~> 1.0.0` won't upgrade to `1.1.0`.

mix.exs
```diff
@@ -99,22 +99,23 @@ defmodule DotCom.Mixfile do
       {:inflex, "~> 1.8.0"},
       {:jason, "~> 1.1"},
       {:laboratory, [github: "paulswartz/laboratory", ref: "cookie_opts"]},
-      {:logster, "~> 0.4.0"},
-      {:mail, "~> 0.2"},
+      {:logster, "~> 1.0"},
```
anthonyshull, line 102: We should make these explicit rather than implicit with the patch version number missing. `1.0.0` instead of `1.0`. Are we sure we can't upgrade to at least `1.1.1`?

mix.exs
```diff
@@ -125,15 +126,15 @@ defmodule DotCom.Mixfile do
        ]},
       {:recon, "~> 2.5.1", [only: :prod]},
       {:rstar, github: "armon/erl-rstar"},
-      {:sentry, "~> 7.0"},
+      {:sentry, "~> 10.0"},
```
anthonyshull, line 129: `10.1.0` is already out and this won't update to that because of the `~>`.

PR #1860, opened 2024-01-19
fix: some small things

thecristen:
#### Summary of changes

**Asana Ticket:** [Timetable | TypeError: Cannot read properties of undefined (reading 'stop_sequence')](https://app.asana.com/0/385363666817452/1205750830454329/f) and [Timetable | TypeError: undefined is not an object (evaluating 'e.forEach')](https://app.asana.com/0/385363666817452/1205750617538734/f)

Just helps these two errors from Sentry, along with fixing another problem I spotted on Splunk. With an extra commit to hopefully fix the problem devs have seen where SVG icons are not in the expected place.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -147,15 +147,10 @@ defmodule DotCom.Mixfile do
   end
 
   defp compile_assets(_) do
-    # starts the Phoenix framework mix phx.digest command, that takes content
-    # from assets/static and processes it into priv/static
-    print("(1/3) mix phx.digest")
-    Mix.Task.run("phx.digest", [])
-
     # builds the node script that lets us render some react components
     # server-side, compiling assets/react_app.js,
     # outputting react_renderer/dist/app.js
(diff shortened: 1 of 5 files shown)


mix.exs
```diff
@@ -147,23 +147,18 @@ defmodule DotCom.Mixfile do
...
+    print("(1/3) webpack --config webpack.config.react_app.js --env.production")
 
     {_, 0} =
       System.cmd("npm", ["run", "--prefix", "assets", "webpack:build:react"],
         stderr_to_stdout: true
       )
 
     # 3 - transpiles/builds our typescript/CSS/everything else for production
-    print("(3/3) webpack --config webpack.config.prod.js --env.production (long)")
+    print("2/3) webpack --config webpack.config.prod.js --env.production (long)")
```
kotva006, line 161: It looks like step `3/3` is missing

kotva006 (changes requested):
I think the mix file is missing a step

kotva006 (approved):

PR #1862, opened 2024-01-23
fix(alert): show suspensions as diversion

thecristen:
No ticket, was suggested by @shantigonzales to improve schedule page presentation of the Green Line suspensions happening this month.

| Before | After |
| --- | --- |
| green-b-before | green-b-after |


kotva006 (approved):
👍

PR #1863, opened 2024-01-24
Take an arbitrary path for delete cache

anthonyshull:
Deleting from the cache is failing for paths with > 2 component parts /one/two/three. This fix makes it so the path can be of any length.

### DIFF
--- a/lib/dotcom_web/controllers/cms_controller.ex
+++ b/lib/dotcom_web/controllers/cms_controller.ex
@@ -47,27 +47,18 @@ defmodule DotcomWeb.CMSController do
 
   @doc """
   Resets a cache key based on the URL params.
-  PATCH /cms/foo/bar will reset the cache key /cms/foo/bar.
-  This corresponds to the CMS page /foo/bar.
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/controllers/cms_controller.ex
```diff
@@ -50,24 +50,13 @@ defmodule DotcomWeb.CMSController do
   PATCH /cms/foo/bar will reset the cache key /cms/foo/bar.
```
amaisano, line 50: Docs here might need updating, too.

amaisano (approved):

thecristen (approved):
So we won't need the `"/cms/#{object}/#{id}"` sort of key anymore with this? Anyway, happy to give this a spin!

anthonyshull:
> So we won't need the `"/cms/#{object}/#{id}"` sort of key anymore with this? Anyway, happy to give this a spin!

Yeah, I replaced the two identical functions and tests with a single one that can take any number of path parts.

PR #1865, opened 2024-01-25
Proposed directory structure

anthonyshull:
https://app.asana.com/0/555089885850811/1206422830794349/f

This is my proposed directory structure for handling e2e tests, load tests, monitors, etc.

Setup:
```
%> npx playwright install chromium
%> npx playwright install-deps

%> export TARGET_URL=https://www.mbta.com
```

E2E tests:
```
%> npx playwright test all-scenarios
```

Load tests:
```
%> npx artillery run load_tests/all-scenarios.yml --target $TARGET_URL
```

Monitor:
```
%> node monitor/all-scenarios.js
```


deploy/monitor/splunk/user-seed.conf
```diff
@@ -0,0 +1,3 @@
+[user_info]
+USERNAME = admin
+PASSWORD = password
```
thecristen, line 3: this username and password works?? haha

package.json
```diff
@@ -1,5 +1,11 @@
 {
-  "name": "dotcom-scripts",
+  "name": "dotcom",
```
thecristen, line 2: am curious if this'll cause issue with `assets/package.json` having the same name. also, are you sure it won't work to define the dependencies in the other package.json file? I'd love to avoid two sets of `node_modules`...

thecristen (changes requested):
Cool! Proposed documentation changes:
- remove reference to now-deleted load testing doc in the `PULL_REQUEST_TEMPLATE` file
- I feel like the monitor could use some acknowledgement, in the "hi Dotcom newcomer, this project has a monitor, which is a distinct piece of infra deployed somewhere..." sense. Part of me feels like the monitor could be a standalone utility in some other repo, but I understand why it's useful to have alongside the tests... hrm.

package.json
anthonyshull: It's good to have two separate package declarations. That way, it is clear that the outer one is used for the entire project, and the assets one is used for the assets. I don't think the package's name being the same has any impact. We aren't publishing the package to a registry. The node_modules directories are ignored by docker and git so it doesn't really matter how many there are. Is there any problem with having two other than that there are two?

deploy/monitor/splunk/user-seed.conf
anthonyshull: This seeds the username and password into the UF. We don't care about them being secure because the UF is only being run on the instance we create.

anthonyshull:
> * Is `test-results/all-scenarios-All-scenarios-Plan-a-trip-chromium/trace.zip` something we should be gitignoring?
> * I think you'll need to update the other Docker-building workflow too since it relies on the default Dockerfile location - I think it's just the `file` argument for https://github.com/docker/build-push-action

Not sure where those results came from, removed and ignored. I'm pretty sure this .github/workflows/docker.yml handles the building of the image. Already changed it to point to the new file.

anthonyshull:
> Cool! Proposed documentation changes:
> 
> * remove reference to now-deleted load testing doc in the `PULL_REQUEST_TEMPLATE` file
> * I feel like the monitor could use some acknowledgement, in the "hi Dotcom newcomer, this project has a monitor, which is a distinct piece of infra deployed somewhere..." sense. Part of me feels like the monitor could be a standalone utility in some other repo, but I understand why it's useful to have alongside the tests... hrm.

Yeah, this PR isn't the final product. Just the proposed directory structure with the files showing they're working.

thecristen (changes requested):
The deploy workflow is separate and doesn't have anything to do with the docker build workflow. So thanks for updating the latter but the former needs adjusting too.

For the deploy workflow's image building step to work it needs to know where the Dockerfile is by adding the `file` argument as mentioned before.

See failed deploy attempt here: https://github.com/mbta/dotcom/actions/runs/7671042466/job/20908565463

anthonyshull:
> The deploy workflow is separate and doesn't have anything to do with the docker build workflow. So thanks for updating the latter but the former needs adjusting too.
> 
> For the deploy workflow's image building step to work it needs to know where the Dockerfile is by adding the `file` argument as mentioned before.
> 
> See failed deploy attempt here: https://github.com/mbta/dotcom/actions/runs/7671042466/job/20908565463

I added a symbolic link until I can figure out how to tell buildx where the Dockerfile is.

thecristen (approved):

PR #1868, opened 2024-01-26
Fix two issues with search

anthonyshull:
https://app.asana.com/0/555089885850811/1206080145852890/f

Fixes an issue with the search header bar and the search page.

For the search header bar, we weren't getting the referenced URL for a search node. So, it was just rendering the search node. It also wasn't filtering out search results that don't have a real URL.

For the search page, we weren't filtering out search results that don't have a real URL.

### DIFF
--- a/assets/js/algolia-results.js
+++ b/assets/js/algolia-results.js
@@ -199,9 +199,21 @@ export class AlgoliaResults {
   }
 
   render(results) {
+    // We don't want to render pages where the URL is reference to a node.
+    // That means that the result is a page that has been unpublished.
+    const filtered_pages_hits = results.pages.hits.filter(
+      result => !/node\//.test(result._content_url)
+    );
+    const filtered_pages = Object.assign({}, results.pages, {
+      hits: filtered_pages_hits
+    });
+    const filtered_results = Object.assign({}, results, {
+      pages: filtered_pages
+    });
+
     if (this._container) {
       this._container.innerHTML = this._groups
-        .map(group => this._renderGroup(results, group))
+        .map(group => this._renderGroup(filtered_results, group))
         .join("");
 
       Array.from(
@@ -209,7 +221,7 @@ export class AlgoliaResults {
       ).forEach(this.addResultClickHandler);
 
       this._groups.forEach(group => this._addShowMoreListener(group));
-      this._addLocationListeners(results);
+      this._addLocationListeners(filtered_results);
     }
   }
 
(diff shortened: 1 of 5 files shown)


assets/js/algolia-results.js
```diff
@@ -202,7 +202,7 @@ export class AlgoliaResults {
     // We don't want to render pages where the URL is reference to a node.
     // That means that the result is a page that has been unpublished.
```
amaisano, line 203: This line is not confirmed. Yet.

assets/js/algolia-results.js
```diff
@@ -199,17 +199,29 @@ export class AlgoliaResults {
   }
 
   render(results) {
+    // We don't want to render pages where the URL is reference to a node.
+    // That means that the result is a page that has been unpublished.
+    const filtered_pages_hits = results.pages.hits.filter(
+      result => !/node\//.test(result._search_result_url)
```
amaisano, line 205: So this assumption is false. There have been and will be `search_result` type drupal results where the value of the `_search_result_url` field will look like this and be 100% valid:

`"entity:node/5019"`

This is due to the potential way editors may (and should) be entering these URLs in the UI, where they may use a Drupal autocomplete to search for the node they wish to connect to this item, click on it, and store the system path (which is always what we want for internal links, due to relative URLs being ephemeral). This lets Drupal resolve it to the current canonical path.

amaisano (changes requested):

thecristen:
@anthonyshull Yeah great questions -- and I don't think we indicate that to the user, and I'm not sure we'd necessarily want to (since these point to functions that _are_ part of the T's services to the public, even though technically they're not hosted on the `mbta.com` website).

However there are also other ways that could be handled anyhow. In this case the content team could create a vanity link, for example `mbta.com/school-trip`, which redirects to the external site.. e.g. we could have `internal:/school-trip` be the search result URL. They already make vanity links for all sorts of things (usually for internal URLs that're too verbose for a Tweet or signage) though I don't know if there're any constraints around external links.

amaisano:
> though I don't know if there're any constraints around external links.

If the input does not start with `http` or `https`, then it will start with `internal:` or `entity:`.

There is nothing stopping editors from entering an internal URL as an external url, but that could be easily domain checked, unless it was hidden behind a non mbta.com redirect domain.

However, doing a quick DB audit of these destination URLs, they all follow the rule stated above.

amaisano:
@anthonyshull will there be breaking changes? Editors expect external destinations (beginning with http/s) to work for visitors. Such as the various payment programs which are funneled to subdomains and other orgs. These are valid and should continue working.

thecristen (approved):

anthonyshull:
> @anthonyshull will there be breaking changes? Editors expect external destinations (beginning with http/s) to work for visitors. Such as the various payment programs which are funneled to subdomains and other orgs. These are valid and should continue working.

It only changes the URL if internal: or entity: are found in the string.

PR #1874, opened 2024-02-01
Run the dotcom monitor application

anthonyshull:
https://app.asana.com/0/555089885850811/1206422830794353/f


deploy/monitor/Dockerfile
```diff
@@ -2,39 +2,12 @@ FROM node:21-bookworm
```
thecristen, line 2: Since this is so trimmed down now I wonder if we should just use the Playwright-provided image e.g. `FROM mcr.microsoft.com/playwright:v???-jammy`, I imagine then we wouldn't need to install the browsers and such

thecristen (approved):

monitor/worker.js
```diff
@@ -19,13 +21,12 @@ parentPort.on('message', async _ => {
...
-    const duration = end - start;
+    const duration = Math.floor(end - start);
 
     const metric = scenario.name.toLowerCase().replace(/ /g, '.');
 
-    client.gauge(metric, Math.floor(duration));
-
-    parentPort.postMessage(`Scenario: "${scenario.name}" took ${Math.floor(duration)}ms`);
+    client.gauge(metric, duration);
+    logger.info({metric, duration});
```
kotva006, line 29: Are these logs going anywhere?  Do we record the STD out of this process/image somewhere?

deploy/monitor/Dockerfile
anthonyshull: Ha, that's a good idea. I originally used that and then switched to be able to install the UF and AWS CLI. We can definitely switch back.

anthonyshull: Oh, I do remember one reason I wanted to move off of the official image is that we only need one browser. Their official image is just node bookworm with all of the browsers.

monitor/worker.js
anthonyshull: See the `log_configuration` is set to json where it is normally set to raw: https://github.com/mbta/devops/pull/1544/files

PR #1875, opened 2024-02-01
refactor(TripPlannerController): remove inputs 

thecristen:
## Summary of changes

**Asana Ticket:** eh, it's mostly cleanup before [Trip planner backend changes](https://app.asana.com/0/385363666817452/1206301319901542/f)

This PR removes the trip planner "best route" "fewest transfers" "least walking"radio  buttons, replacing the whole `<fieldset>` with a single checkbox to request accessible itineraries.

The major change this created was it removed the need to fetch itineraries twice -- before, if the rider was choosing to optimize by anything other than wheelchair accessible, we'd fetch both accessible and inaccessible itineraries, merging and deduplicating them to show in the results. With the single checkbox (set as checked by default, by the way), we just either show accessible itineraries (if checked), or all itineraries (if unchecked).

#### Better handling and testing of GraphQL response parsing and error handling. 
I had to read the [spec](http://spec.graphql.org/draft/#sec-Response-Format) a bit to get a grasp on the formats possible (and I still find it weird that 200 responses could still contain errors)... but I think I've covered the possible error cases there. 

I've also updated the ancient test fixture, and fixed a few bugs with our parser that were looking for certain bits of data in the OTP1 places.

#### Updated GraphQL query
Firstly I'm using the alternate format of specifying a request in the Absinthe client, by using a `{query, variables}` tuple. I've also un-JSON-stringified a few things, because it stopped being needed in the params, and the client gives us an already-JSON-decoded object.

The query itself is updated with a few extra parameters, I've left inline comments that hopefully explain the rationale. These were mostly from recommendations @EmmaSimon made in December documenting explainable properties. :)

#### Removed hidden checkbox inputs

[`Phoenix.HTML.Form.checkbox/3`](https://hexdocs.pm/phoenix_html/3.3.1/Phoenix.HTML.Form.html#checkbox/3) default behavior inserts a hidden checkbox with a `"false"` value that could be read on the backend on submit... we weren't utilizing it, and it was creating very unwieldy URL params:

```
&plan[modes][subway]=false
&plan[modes][subway]=true
&plan[modes][commuter_rail]=false
&plan[modes][commuter_rail]=true
&plan[modes][bus]=false
&plan[modes][bus]=true
&plan[modes][ferry]=false
&plan[modes][ferry]=true
```

So now we either have, for example the "commuter rail" checkbox,`plan[modes][commuter_rail]=true` if true or `` (nothing) if false

---

#### General checks


test/trip_plan/fixture/north_station_to_park_plaza.json
thecristen, line 1: This fixture was very old and from the OTP1 REST API. I've replaced it with a response from a similar request to OTP2.4's GraphQL API.

lib/trip_plan/api/open_trip_planner/parser.ex
```diff
@@ -14,82 +14,64 @@ defmodule TripPlan.Api.OpenTripPlanner.Parser do
...
-  def parse_ql(%{"data" => %{"plan" => nil}, "errors" => [head | _]}, _accessible?) do
-    _ =
-      Logger.warning(fn ->
-        "#{__MODULE__} trip_plan=error message=#{inspect(head["message"])}"
-      end)
-
-    {:error, :unknown}
-  end
+  @spec parse_ql(map()) :: {:ok, [Itinerary.t()]} | {:error, TripPlan.Api.error()}
+  # A field error is an error raised during the execution of a particular field which results in partial response data. This may occur due to an internal error during value resolution or failure to coerce the resulting value.
```
kotva006, line 18: nit-pick: Do we want to line break these comments?

kotva006 (approved):
Everything looks good overall

lib/trip_plan/api/open_trip_planner.ex
```diff
@@ -3,43 +3,66 @@ defmodule TripPlan.Api.OpenTripPlanner do
...
+
+             # Increased from 2.0 to reduce number of itineraries with significant walking
+             walkReluctance: 5.0
+
+             # Theoretically can be configured in the future for visitors using translation?
+             locale: "en"
+
+             # Proposed future use: prefer MBTA transit legs over Massport or others.
+             # Can't use this until we change the MBTA feed name (GraphQL can't do hyphens)
+             # preferred { agencies: [mbta-ma-us:1] }
```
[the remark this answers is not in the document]
thecristen: ahhhh thank you!

PR #1881, opened 2024-02-07
feat(Translations): Add translations analytics

kotva006:
#### Summary of changes
Added a listener to send an event to google analytics on language change.

**Asana Ticket:** [Log Translation Actions](https://app.asana.com/0/555089885850811/1206339704257865)

---

#### General checks


assets/js/translate-analytics.js
```diff
@@ -0,0 +1,42 @@
...
+  // Start by checking if the MutationObserver API is available
+  if (typeof MutationObserver === "function") {
+    // Tell the observer to monitor for changes to HTML attributes
+    const config = { attributes: true };
+    // Build the function to run when a change is observed
+    const callback = mutationList => {
+      // Loop through each observed change
+      for (let i = 0; i < mutationList.length; i += 1) {
+        // Only do something if the change was on an attribute
+        if (mutationList[i].type === "attributes") {
```
thecristen, line 18: Sorry I forgot to send you this earlier, [here's the other code that deals with mutations](https://github.com/mbta/dotcom/blob/master/assets/ts/ui/accordion.ts#L10)... so these can have a property called `oldValue`, that you can read if interested in the old value.

assets/js/translate-analytics.js
```diff
@@ -0,0 +1,42 @@
...
+          if (mutationList[i].attributeName === "lang") {
+            // Send an event to the dataLayer
+            const newLanguage = getNewLanguage();
```
thecristen, line 21: The mutations `target` property should have the `<html>` element in its latest state, so could potentially read the language attribute off of there instead of re-fetching the DOM element.

assets/js/translate-analytics.js
```diff
@@ -0,0 +1,42 @@
...
+            window.dataLayer = window.dataLayer || [];
+            window.dataLayer.push({
+              event: "translate"
+            });
+            if (newLanguage) {
```
thecristen, line 26: suggestion: should check if the new one is different from the old one?

assets/js/translate-analytics.js
```diff
@@ -0,0 +1,42 @@
...
+              });
+            }
+          }
+        }
+      }
+    };
+    // Create the actual observer
+    const observer = new MutationObserver(callback);
+    // Attach the observer to the <title> tag
+    observer.observe(document.getElementsByTagName("title")[0], config);
```
thecristen, line 38: question: I wonder if this ends up firing on Turbolinks-enhanced navigations (I think the title changes there?). I'm concerned it might get noisy

thecristen: does `<title>` ever get a `lang` attribute? Should this get a different callback then the `<html>` one?

kotva006: The only browser that updates the title is Edge.  Technically it adds an attribute to every element it tries to translate.  It unfortunately does not update the lang in the HTML when it does this.  From what I can tell there is no way to get the translated language from the browser.  So in Edge we can detect a langage change event, just not what it becomes.

On the other hand Safari and Chrome both update the lang attribute on the html element.  The helper function above is what that use case is for.

As for turbo links this shouldn't log any of that noise.  As we filter for specific attribute mutations (lang and _msttext) before logging the event.

assets/js/translate-analytics.js
kotva006: What we want is the new value of the lang attribute in the `html` element.  The helper code above gets the current one for us.

assets/js/translate-analytics.js
thecristen: I don't think you need to add the observer to the `<title>` tag.

assets/js/translate-analytics.js
```diff
@@ -0,0 +1,54 @@
...
+
+// adapted from: https://analytical42.com/2022/detect-track-translations-ga4/
+// This code does its best to detect if the user changes the language of the page.
+// Right now it fully supports logging from users changing the page language from the website itself regardless of the browser used
+// This struggles with detecting and reporting the new language when the user uses the browser to
+// change the language.
+// Chrome - Full support, we can detect when the user translates the page, and what language the page is translated to
+// Safari - Full support, we can detect when the user translates the page, and what language the page is translated to
+// Edge - Half support, we can detect whne the user translates the page, just not what language they changed it to
+// Firefox - No support, we cannot detect when the user translates the page
```
thecristen, line 11: It seems to work in Firefox for me (I can add a console.log to this and see the changes reflected)? Is it that it doesn't work in a specific scenario? I like this elaborate note here, but it needs to be accurate.

kotva006: Okay, I will update the comment.  I it appears that Firefox must have updated since this was written.  As initial testing Firefox did nothing to modify the `html` or `title`

kotva006:
Google Tag Manager is consuming these events.

thecristen (approved):

PR #1882, opened 2024-02-08
Upgrade all dependencies

anthonyshull:
https://app.asana.com/0/555089885850811/1206487082456630/f

Upgrade Erlang to 26.2.1
Upgrade Elixir to 1.16.1

I pinned every dependency and upgraded every one I could.
I also left a comment for every dependency I couldn't upgrade with information as to why.


thecristen (commented):
The Elixir linting task (which just runs `mix credo diff master -a`) seems to have an error now, might it need adjusting now that you've upgraded it?

mix.exs
```diff
@@ -65,72 +65,87 @@ defmodule DotCom.Mixfile do
...
-      {:benchfella, "~> 0.3", [only: :dev]},
-      {:bypass, "~> 1.0", [only: :test]},
-      {:castore, "~> 0.1.11"},
-      {:con_cache, "~> 0.12.0"},
+      {:absinthe_client, "0.1.0"},
+      {:benchfella, "0.3.5", [only: :dev]},
+      # latest version 2.1.0; cannot be upgraded because of errors when we use * in the url for tests
+      # rather than fixing tests and upgrading, we should convert the tests to use mox
+      {:bypass, "1.0.0", [only: :test]},
+      # latest version 1.0.5; cannot upgrade because of server_sent_event_stage expects castore < 1
```
thecristen, line 77: Some more detail for this note: technically we can change that (as [`:server_sent_event_stage`](https://github.com/mbta/server_sent_event_stage) is ours)!

Looking at other libraries that have more flexible requirements for `:castore`, I found that `:absinthe_client` and `:phoenix` list ">= 0.0.0" while "~> 0.1 or ~> 1.0" is used for `:elixir_make`, `:finch`, `:mint`, and `:redix`. So we could change `:server_sent_event_stage` to use that. 

However that might not be enough to help Dotcom since `:tesla` also expects "~> 0.1". That library is required by `:web_driver_client` which is required by `:wallaby`, which we use for parsing the DOM tree for one task.

anthonyshull:
> The Elixir linting task (which just runs `mix credo diff master -a`) seems to have an error now, might it need adjusting now that you've upgraded it?

I'll take a look. It works locally. It's complaining that it can't find master. Maybe a permissions issue? I'll sort it out.

mix.exs
anthonyshull: Yeah, I was going to note that we own that project in the follow-up ticket. Maybe we evaluate whether or not we still need it? There might be community libraries that do the same thing.

anthonyshull: I more meant...we can't upgrade it without more work. Technically, we could upgrade anything in the list given the right amount of effort.

anthonyshull:
> > The Elixir linting task (which just runs `mix credo diff master -a`) seems to have an error now, might it need adjusting now that you've upgraded it?
> 
> I'll take a look. It works locally. It's complaining that it can't find master. Maybe a permissions issue? I'll sort it out.

@thecristen I'm wondering if this is failing because master is using an older version of Credo. So when it is trying to compare it fails. I bet if we just merge this that error won't happen when we're comparing two branches with the same version of Credo. Not that they have to be the same version, but this check was introduced in 1.5.4. So, I bet it would cause that error any time you jump over that release threshold.

https://hexdocs.pm/credo/changelog.html#1-5-4

thecristen (approved):
:tada:

PR #1885, opened 2024-02-09
feat(GlobalNav): Focus on menu button when hitting escape in the subm…

kotva006:
…enu while focused.

#### Summary of changes
These changes allow us to focus on the menu button selected when the user presses escape in the sub menu.

**Asana Ticket:** [Tab Menu Focus](https://app.asana.com/0/555089885850811/1202299919599013)

---

#### General checks


assets/ts/app/global-navigation.ts
```diff
@@ -307,9 +310,29 @@ export function setup(rootElement: HTMLElement): void {
...
+    }
+  }
+
+  // Save the last focused menu toggle
+  var activeMenuElement: Element | null = null;
+  function saveActiveMenuElement(): void {
+    if (document.activeElement?.className.includes("m-menu--desktop__toggle")) {
+      activeMenuElement = document.activeElement;
+    } else if (
+      document.activeElement?.className.includes("custom-language-selector")
```
kotva006, line 325: This is kind of hacky, and I am not a huge fan.  I was trying to come up with a solution to clear the last active menu element when the user tab leaves the menu buttons.  Many of the sub menu options don't have classes so it was hard to tell where we were in the tree when they were tab selected.
The next item hit after leaving the menu buttons are the language translation button, so I decided to clear it then.
Open to other ideas.

thecristen (commented):
image


We have code that toggles the navbar buttons to `aria-expanded=true` or `false` when opening and closing the submenus. Have you considering working with that in some way?

kotva006:
> en opening and closing the subm

I was playing around with this a little bit and there is a weird user experience if we update the menu on only toggle open.
If you open the first menu item (Transit), close out, tab over to Fare, then press escape, it goes back to highlighting transit.
There might be more code to add to do nothing if there is none of the menus are expanded on escape press, but it seems like trying to solve the same problem in a different way.  I am thinking of just leaving it to save state on tab press.

thecristen:
@kotva006 This seems to work for me without UI quirks, without having to track `Tab` presses. Like you said, doing the same thing in a different way.

```typescript
  // Press Esc within open header should return focus to header
  header.addEventListener("keydown", e => {
    let activeNavButton: HTMLButtonElement | null | undefined;
    const activeNavSection = document.activeElement?.closest(
      "[data-nav='desktop-section']"
    );
    if (activeNavSection) {
      const openSectionId = activeNavSection.parentElement!.id;
      activeNavButton = document.querySelector<HTMLButtonElement>(
        `nav.m-menu--desktop [aria-controls=${openSectionId}]`
      );
      handleNativeEscapeKeyPress(e, () => {
        resetPage();
        if (activeNavButton) {
          activeNavButton.focus();
          // don't bubble up to the rootElement keydown listener
          e.stopPropagation();
        }
      });
    }
  });
```

thecristen (changes requested):
Some usability issues I found with this approach:

- If you close the open nav by clicking on the page, the focus moves to the menu button. This isn't desired behavior; no focus should be set in that case.
- Weird things happen when users behave unpredictably 😉 I find that if you tab a particular menu A open, but _then_ click one or more other menu B open, then close (using any method), the focus shows on the first opened menu A. Quite confusing.

thecristen (changes requested):
The weird behavior I noted is gone, but the original requirement appears to be no longer met (e.g. when I tab into a submenu and `Esc`, the focus didn't return to the menu button)

thecristen (approved):

PR #1887, opened 2024-02-13
Write Critical User Journeys

anthonyshull:
https://app.asana.com/0/555089885850811/1206498102526465/f

All 10 critical user journeys. I removed the expectation that the test runs within a threshold because it was pretty unstable. And, it doesn't make sense to have thresholds that are so high they can handle that volatility. Instead, we just attach an annotation to the test for how long it took to complete.

### DIFF
--- a/monitor/all-scenarios.js
+++ b/monitor/all-scenarios.js
@@ -5,8 +5,11 @@ const { Worker } = require('worker_threads');
 
 const filesPath = path.join(__dirname, '..', 'scenarios');
 
+const fileToName = (file) => file.replace(/-/g, '.').replace('.js', '').toLowerCase();
+
 const workers = fs.readdirSync(filesPath).map((file) => {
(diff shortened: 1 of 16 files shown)


monitor/all-scenarios.js
```diff
@@ -5,8 +5,11 @@ const { Worker } = require('worker_threads');
 
 const filesPath = path.join(__dirname, '..', 'scenarios');
 
+const fileToName = (file) => file.replace(/-/g, '.').replace('.js', '').toLowerCase();
```
kotva006, line 8: nit-pick: This could go in a helper file since it is reused in multiple files

kotva006 (approved):

monitor/all-scenarios.js
anthonyshull: Yeah, I thought about that, but I'm not sure where it would go. Let me think more about how we should structure all of this monitoring code.

PR #1889, opened 2024-02-13
Use elixir image because hex doesnt support otp 25 or 26

anthonyshull:
This hotfix uses the official Elixir image rather than one from Hex because Hex doesn't have any OTP 25 or 26 images.


thecristen (approved):

kotva006 (approved):

anthonyshull:
This was actually caused by an issue with Dockerhub not responding. The tag we were using exists and re-running the failing jobs now works.

PR #1894, opened 2024-02-20
Use docker compose for development

anthonyshull:
Pulling this branch will delete all of the contents of your .envrc file. That is because Docker can use a .env file to load environment variables. So, we use that instead. Before pulling this branch, copy your `.envrc` file to `/tmp`. Pull this branch and then copy `/tmp/.envrc` to `./.env`. The only difference between the two is that .env removes the `export ` from each line. Do that on the new file and it will work with Docker. `.env` is in `.gitignore` but `.envrc` no longer is so be *very* careful if you commit anything to this particular branch as I already accidentally committed sensitive information.

Check the README and make sure it accurately depicts how to use docker compose.


.envrc
```diff
@@ -0,0 +1 @@
+dotenv
```
kotva006, line 1: Can this file just be removed?

kotva006: Okay, I didn't understand this was a function/macro/command supported by direnv.  Based on the description of the pull request I thought this file was going away, and this was your attempt at clearing it out to be removed in the future.  Maybe add a comment about what the purpose of the `dotenv` is in this file?

thecristen (commented):
I can't run the dev application locally because the `redis-cluster-init` service keeps exiting for me, but the rest seems to be working well! I'd like to get it fully working before I approve though, will figure it out offline.

I also don't think it's necessary to switch from `.envrc` to `.env` at all! I managed to get it working by listing the values under the `environment` key and just pulling the values from the system e.g.

```
environment:
  - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
  - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
  - DRUPAL_ROOT=${DRUPAL_ROOT-https://live-mbta.pantheonsite.io}
  - MIX_ENV=dev
  - OPEN_TRIP_PLANNER_URL=${OPEN_TRIP_PLANNER_URL-http://otp2-local.mbtace.com}
  - REDIS_PORT=${REDIS_PORT}
  - V3_API_KEY=${V3_API_KEY}
  - V3_API_VERSION=${V3_API_VERSION}
  - V3_URL=${V3_URL-https://api-dev.mbtace.com}
```

And I even (last night) managed to figure out how to get the changes from #1940 to be compatible - I added this call to my `.envrc`, and with a tweak to the ExAws config to look for the environment variables, it all just worked ✨ 

```
# exports AWS_ACCESS_KEY_ID & AWS_SECRET_ACCESS_KEY
$(aws configure export-credentials --profile default --format env)
```

anthonyshull:
Yeah, we could always have manually listed every environment variable in the docker file, but then we have to maintain both. The idea was to have a single file to maintain that works whether or not you're working with Docker. If we pursue your path, adding or removing an environment variable means maintaining it in two places. Changing to `.env` means only having to maintain one file and it literally takes 5 seconds to search and replace `export ` with ``.

anthonyshull:
Also, btw, you can export the AWS credentials without the `export ` with:
```
$(aws configure export-credentials --format env-no-export)
```

.env.template
```diff
@@ -0,0 +1,38 @@
...
+# You can optionally set a Redis host and port.
+# The default host is 127.0.0.1
+# The default port is 6379
+# REDIS_HOST=
+# REDIS_PORT=
+
+# These credentials control access to resetting cache entries for the CMS.
+# You can set them to be whatever you want, but they'll need to match those on the Drupal side.
+# CMS_BASIC_AUTH_USERNAME=
+# CMS_BASIC_AUTH_PASSWORD=
```
thecristen, line 38: Can you update these to the newer names here

deploy/dev.yml
```diff
@@ -0,0 +1,107 @@
...
+            - ../:/app
+        networks:
+            redis_network:
+                ipv4_address: 10.0.0.2
+    monitor:
+        build:
+            context: ../
+            dockerfile: ./deploy/monitor/Dockerfile
+        env_file:
+            - ../.env
```
thecristen, line 28: I don't think the monitor uses any of the environment variables?

thecristen:
> Also, btw, you can export the AWS credentials without the `export ` with:
> 
> ```
> $(aws configure export-credentials --format env-no-export)
> ```

@anthonyshull Where would you want devs to invoke this? It doesn't appear to work in `.env` files.

Unless more useful elsewhere, might make sense to add `$(aws configure export-credentials --format env)` to the `.envrc` though!

.env.template
```diff
@@ -0,0 +1,38 @@
+# Example environment configuration file: copy to `.env` (this is git-ignored) and fill in any additional values you need.
+
+# ALGOLIA_APP_ID=
+# ALGOLIA_SEARCH_KEY=
+# ALGOLIA_WRITE_KEY=
+
+# AWS_ACCESS_KEY_ID=
+# AWS_SECRET_ACCESS_KEY=
+
```
thecristen, line 9: Not technically needed since it's being fetched via AWS CLI

deploy/dev.yml
```diff
@@ -0,0 +1,107 @@
...
+    dotcom:
+        build:
+            context: ../
+            dockerfile: ./deploy/dotcom/dev/Dockerfile
+        env_file:
+            - ../.env
+        depends_on:
+            - redis-cluster-init
+        environment:
+            - MIX_ENV=dev
```
thecristen, line 12: Since these values are elsewhere in this file anyway, I think it'd make sense to just add `REDIS_HOST=10.0.0.11` and `REDIS_PORT=6379` here

thecristen (approved):
Thanks for giving me more time with this. I think it'll work out. 🥳

deploy/dev.yml
anthonyshull: Yeah, all of the healthchecks do. E.g., https://github.com/mbta/dotcom/blob/fcca6a8aa967b21b6f7a457c8ab1b82c90ed0ad8/integration/health_checks/cms.js#L4

deploy/dev.yml
anthonyshull: The only issue there is that they're still needed for non-Docker development. So, it would split the environment variables based on your method of development. That was the idea with keeping them all in one file is that you only have to change them in one place no matter which dev method you use.

anthonyshull: Oh, also, the monitor needs those REDIS variables to run the healthcheck.

thecristen: Sounds good, thanks for clarifying!

.env.template
anthonyshull: What about some of these other vars? Can we remove any of them? Like, are we using Wiremock for anything? Some of these I've never set like 15-17.

thecristen: Yeah go ahead. Feel free to remove the Wiremock ones (we can remove the related scripts separately) and the Docker desktop password/username (which I don't recall having ever used...) and the Google API key (which shouldn't be needed at this point).

PR #1905, opened 2024-03-05
feat: run Critical User Journeys on deploy

thecristen:
No ticket, just run the test on deploy. Also updates the Playwright version to latest.

I've bundled in a few other CI-related changes.
- Attempt to fix the broken Link checker by using the correct Elixir version
- Update actions/cache version
- Add `paths-ignore` to the Docker build workflow
- Have the smoke tests **not** fail on certain known third-party uncaught exceptions.
- Fixed the "cleanup" step to successfully remove the dev-blue or dev-green label from other PRs


.github/workflows/use-deploy-ecs.yml
```diff
@@ -113,11 +113,54 @@ jobs:
...
+        node-version-file: .tool-versions
+        cache: npm
+        cache-dependency-path: package-lock.json
+    - run: npm install --ignore-scripts
+    - run: npx playwright install --with-deps chromium
+    - run: npx playwright test all-scenarios
+      env:
+        CI: true
+        HOME: /root
+        PLAYWRIGHT_TEST_BASE_URL: ${{ needs.deploy.outputs.deployment-url }}
```
anthonyshull, line 159: Pretty sure the only other user of this output is Cypress. So, we can change it to be HOST and not have to change the code to add another check. In general, we should have one config per each thing rather than per each consumer of a thing.

anthonyshull (approved):

PR #1906, opened 2024-03-06
Add memory limits on processes and setup log rotation

anthonyshull:
The scenarios script keeps crashing and failing to restart. Running it locally, it doesn't look like it gets anywhere near the memory limit of the task. But, we should set a memory limit anyway. When it reaches the limit, the process will just be restarted. I think the crashes might be related to log rotation. So, I set up some log rotation configuration.


thecristen (approved):
👍🏼

PR #1908, opened 2024-03-06
hotfix(TimetableController): don't use `Stops.Repo.by_route/3` to infer ordered stops

thecristen:
No ticket.

The Haverhill timetable was showing Reading in the wrong position for the inbound direction:

image

After a brief [discussion on Slack](https://mbta.slack.com/archives/C8THJ1NJY/p1709748275181689), we can acknowledge that the V3 API query behind `Stops.Repo.by_route/3` cannot be expected to return stops in a correct order.

As an alternate approach, this PR gets the list of stops from the canonical route patterns (which have an associated representative trip, which itself has the list of stops in order).

Since this controller no longer reads from `conn.assigns.all_stops`, and nothing else used it, and nothing else needs the plug which populated that value, we can also delete the `DotcomWeb.ScheduleController.AllStops` plug, so we do so here.

---

#### General checks


thecristen:
This is pretty random but I edited an unrelated smoke test here, because surfacing routes by searching for a stop name was disabled in Algolia. So now that test is aligned with expectations.

PR #1912, opened 2024-03-07
Close the context in Playwright to avoid crashes

anthonyshull:
We are seeing issues with Playwright where the browser becomes inaccessible. PM2 will continue to try to restart the process and fail. We don't know the root cause, but we should be closing the page and browser context before we close the browser. This *might* be the issue...causing some kind of memory leak. I've deployed this branch and we'll see if we have any issues.


deploy/monitor/Dockerfile
```diff
@@ -1,12 +1,15 @@
-FROM node:21-bookworm
+FROM mcr.microsoft.com/playwright:v1.42.1-jammy
```
thecristen, line 1: yay!

thecristen (approved):
Let's try it!

PR #1914, opened 2024-03-08
Flush caches by key

anthonyshull:
https://app.asana.com/0/555089885850811/1206042318802989/f

Adds the ability to flush the cache according to a given pattern (which could just be one key) by sending an HTTP request with Basic Auth.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -80,7 +80,7 @@ redis_config = [
 ]
 
 # This is used by PubSub, we only use the first node in the cluster
-config :dotcom, :redis, redis_config[:redis_cluster][:configuration_endpoints][:conn_opts]
+config :dotcom, :redis_config, redis_config[:redis_cluster][:configuration_endpoints][:conn_opts]
 
 # Set caches that use the Redis cluster
 config :dotcom, Dotcom.Cache.Multilevel,
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -95,15 +95,15 @@ config :dotcom, Dotcom.Cache.TripPlanFeedback.Cache, redis_config
 
 if config_env() == :test do
   config :dotcom, DotcomWeb.Router,
-    cms_basic_auth: [
+    basic_auth: [
       username: "username",
       password: "password"
     ]
 else
   config :dotcom, DotcomWeb.Router,
-    cms_basic_auth: [
-      username: System.get_env("CMS_BASIC_AUTH_USERNAME"),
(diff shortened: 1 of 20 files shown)


test/support/mocks.ex
```diff
@@ -0,0 +1,5 @@
+# This file houses definitions for defining Mox mocks.
+
+Mox.defmock(Redis.Mock, for: Dotcom.Redis.Behaviour)
+Mox.defmock(Redix.Mock, for: Dotcom.Redix.Behaviour)
+Mox.defmock(Redix.PubSub.Mock, for: Dotcom.Redix.PubSub.Behaviour)
```
kotva006, line 5: If the behaviours are only used for testing, should they be defined in a testing folder?

kotva006 (approved):
Minor Questions

config/runtime.exs
```diff
@@ -95,15 +95,15 @@ config :dotcom, Dotcom.Cache.TripPlanFeedback.Cache, redis_config
...
       password: "password"
     ]
 else
   config :dotcom, DotcomWeb.Router,
-    cms_basic_auth: [
-      username: System.get_env("CMS_BASIC_AUTH_USERNAME"),
-      password: System.get_env("CMS_BASIC_AUTH_PASSWORD")
+    basic_auth: [
+      username: System.get_env("BASIC_AUTH_USERNAME"),
+      password: System.get_env("BASIC_AUTH_PASSWORD")
```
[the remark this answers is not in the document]
anthonyshull: Yeah, this is why I don't like the idea of making environment variables "specific." Some things are generic. Like, SENTRY_ENVIRONMENT being used for things that have nothing to do with Sentry.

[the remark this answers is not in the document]
anthonyshull: SSO is a nightmare. It has taken the teams that did it multiple weeks to implement. And, we don't really have many needs for authentication. Also, we have really low-stakes privileges. Right now, the only thing you can do is signal that cache keys need to be removed.

test/support/mocks.ex
anthonyshull: Behaviours can be used for other purposes such as stubs or even custom implementations so it makes sense to keep them in lib. A good example would be a behaviour of the V3 API that we then implement ourselves and also mock.

anthonyshull: See this as an example: https://github.com/mbta/my_charlie/tree/main/lib/my_charlie/reduced_fares_applications/steps/identity_check/api

PR #1918, opened 2024-03-12
Add one key and set TTL to four hours

anthonyshull:
https://app.asana.com/0/555089885850811/1206607492620749/f

Increasing the TTL on the cms repo to four hours. If this works well, we can increase it.


lib/cms/repo.ex
```diff
@@ -104,7 +105,7 @@ defmodule CMS.Repo do
               cache: @cache,
               key: "/cms/whats-happening",
```
anthonyshull, line 106: This key was changed to `cms.repo|whats-happening` in another PR. Same as the one below. So, the one above is really just adding a key when we used a generated one before.

thecristen (approved):
Assuming #1914 is in place and operational.

PR #1919, opened 2024-03-14
Use dumb init

anthonyshull:
We need to actually call `dumb-init`. We can try redeploying the monitor every 12 hours until we can figure out how to run it somewhere where it won't crash. Maybe using dumb-init will solve the problem...probably not.


thecristen (approved):
🚀

PR #1920, opened 2024-03-15
fix(RealtimeSchedule): partial revert of #1915

thecristen:
More specifically, this is a partial revert of 8a50295235a40c7f8b7b9e77d72770a8af36cfaf

This takes `Dotcom.RealtimeSchedule` (used in Transit Near Me) back to the original request parameters for fetching predictions at a frequent interval. We're hoping it improves performance by (1) requesting less data from the V3 API and (2) having less data to parse through on our end.

### DIFF
--- a/lib/dotcom/realtime_schedule.ex
+++ b/lib/dotcom/realtime_schedule.ex
@@ -175,13 +175,13 @@ defmodule Dotcom.RealtimeSchedule do
       Task.async(fn ->
         next_two_predictions =
           [
-            route: route_pattern.route_id,
-            direction_id: route_pattern.direction_id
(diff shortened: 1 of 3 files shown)


anthonyshull:
We should deploy this and check the performance in whatever environment it is deployed to. Should see a decrease in predictions using `index=dotcom-prod-application "V3Api.get_json_response" | rex field=url "/(?<match>.*)/" | timechart useother=false span=1h sum(duration) by match`

anthonyshull (approved):

thecristen:
(avg & p95 duration for predictions) hahahahaha
image

PR #1921, opened 2024-03-15
feat(TimetableController): revised stop sorting

thecristen:
#### Summary of changes

**Asana Ticket:** [Fix incorrectly ordered timetable stops](https://app.asana.com/0/555089885850811/1206815695536344/f)

The primary change: Ensure stop sorting is consistent with each stop's `zone` number and schedule's `stop_sequence` values

So the simplest sorting that worked consitently across all commuter rail lines involved comparing the `stop_sequence` across all trips. Technically sorting initially by `zone` wasn't needed to show stops in a sensible order, but I found that adding it improved the legibility of a few timetables by placing certain termini stops more proximate to its adjacent stops.

This came with a slight caveat - for Newburyport/Rockport, having two branches which reach similar distances, the branched stops end up displaying interlined instead of clustering by branch. I'd tried a few different things which might help that (e.g. group by route pattern) but this proved more complicated and unreliable.

---

#### General checks

### DIFF
--- a/lib/dotcom_web/controllers/schedule/timetable_controller.ex
+++ b/lib/dotcom_web/controllers/schedule/timetable_controller.ex
@@ -147,8 +147,9 @@ defmodule DotcomWeb.ScheduleController.TimetableController do
 
       schedules ->
         schedules
-        |> Enum.reject(&Schedules.Schedule.no_times?/1)
-        |> Enum.filter(&(&1.route.line_id == route.line_id))
+        |> Enum.filter(
+          &(&1.route.line_id == route.line_id and not Schedules.Schedule.no_times?(&1))
+        )
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/controllers/schedule/timetable_controller.ex
```diff
@@ -148,6 +148,7 @@ defmodule DotcomWeb.ScheduleController.TimetableController do
       schedules ->
         schedules
         |> Enum.reject(&Schedules.Schedule.no_times?/1)
+        |> Enum.filter(&(&1.route.line_id == route.line_id))
```
thecristen, line 151: There was a bug that sometimes returned schedules for train's from another route (usually having overlapping stops with the requested route).
This matches on line_id instead of route ID so that we can match on shuttle routes that's serving this route.

anthonyshull: Depending on how long that list of schedules is, you could combine the `reject` and `filter` into a single private function. Otherwise, you're traversing the list twice.

anthonyshull (approved):

PR #1924, opened 2024-03-18
Use Mox instead of Bypass

anthonyshull:
https://app.asana.com/0/555089885850811/1206592383985338/f

This PR does a few things because they could only really be done in collaboration.

1. I added a tag of `:external` to all tests that call out to the V3API. I made it so that they are all excluded by default in the test_helper.
2. I added a library that will list out all failing tests in a nice way so they can be copied and re-run.
3. I renamed V3API to MBTA.Api because V3 is incredibly non-descriptive.
4. I created a behaviour for the MBTA.API and a matching mock. I also made a mock for HTTPoison, but I didn't have to create a behaviour because the library already has one.


config/test.exs
```diff
@@ -1,7 +1,11 @@
 import Config
 
 config :dotcom, :cache, Dotcom.Cache.TestCache
-config :dotcom, :redis, Redis.Mock
-config :dotcom, :redix, Redix.Mock
-config :dotcom, :redix_pub_sub, Redix.PubSub.Mock
+config :dotcom, :redis, Dotcom.Redis.Mock
+config :dotcom, :redix, Dotcom.Redix.Mock
+config :dotcom, :redix_pub_sub, Dotcom.Redix.PubSub.Mock
```
anthonyshull, line 6: Renaming these so they match the pattern for other config using `:dotcom` as the app.

config/test.exs
```diff
@@ -1,7 +1,11 @@
...
 config :dotcom, :trip_plan_feedback_cache, Dotcom.Cache.TestCache
+
+config :dotcom, :httpoison, HTTPoison.Mock
+
+config :dotcom, :mbta_api, MBTA.Api.Mock
```
anthonyshull, line 11: This is `MBTA.Api.Mock` instead of `Dotcom.MBTA...` because of the directory structure. We should move all non-web code under `dotcom` at some point. But, for now it's better to be consistent.

lib/dotcom_web/controllers/trip_plan/feedback.ex
```diff
@@ -7,7 +7,7 @@ defmodule DotcomWeb.TripPlan.Feedback do
   require Logger
 
   use DotcomWeb, :controller
-  use Nebulex.Caching
+  use Nebulex.Caching.Decorators
```
anthonyshull, line 10: Makes it more specific that we're only using the decorators in this file.

lib/dotcom_web/views/alert_view.ex
```diff
@@ -177,7 +177,7 @@ defmodule DotcomWeb.AlertView do
         Timex.format!(updated_at, "{M}/{D}/{YYYY}")
       end
 
-    time = Timex.format!(updated_at, "{h12}:{m} {AM} {Zabbr}")
+    time = Timex.format!(updated_at, "{h12}:{m} {AM} {Zabbr}") |> String.trim()
```
anthonyshull, line 180: This fixes that failing test where there was trailing whitespace after `AM`. I believe that happens when the timezone isn't present.

lib/mbta/api/facilities.ex
```diff
@@ -0,0 +1,17 @@
+defmodule MBTA.Api.Facilities do
+  @moduledoc """
+  Responsible for fetching Stop data from the V3 API.
+  """
+
+  @mbta_api Application.compile_env!(:dotcom, :mbta_api)
```
anthonyshull, line 6: You'll see this pattern throughout. These modules are really just intermediaries to the `MBTA.Api` so they determine whether to use the real one or a mock. This means that we ultimately always mock `MBTA.Api.Mock.get_json` since that's all they ever use.

test/alerts/parser_test.exs
```diff
@@ -58,15 +58,22 @@ defmodule Alerts.ParserTest do
                      }
                    ]),
                  active_period: [
-                   {~N[2016-06-06T14:48:48] |> Timex.to_datetime("Etc/GMT+4"),
-                    ~N[2016-06-06T19:53:51] |> Timex.to_datetime("Etc/GMT+4")}
+                   {~N[2016-06-06T14:48:48]
+                    |> Timex.to_datetime("Etc/GMT+4")
+                    |> Timex.Timezone.convert("America/New_York"),
```
anthonyshull, line 63: This fixes the failing test. The code does this timezone conversions, but the tests weren't changed to also do it.

test/cms/repo_test.exs
```diff
@@ -72,33 +72,35 @@ defmodule CMS.RepoTest do
     test "generates the correct key for /*" do
       path = "/foo"
 
-      assert Repo.generate(nil, nil, [path, %{}]) == "/cms" <> path
+      assert Repo.generate(nil, nil, [path, %{}]) == "cms.repo" <> String.replace(path, "/", "|")
```
anthonyshull, line 75: Just fixing failing tests.

test/dotcom_web/controllers/fare_controller_test.exs
```diff
@@ -3,22 +3,6 @@ defmodule DotcomWeb.FareControllerTest do
...
-          fare_path(conn, :show, :ferry, origin: "Boat-George", destination: "Boat-Logan")
-        )
-
-      response = html_response(conn, 200)
-
-      assert response =~ "Georges Island Fare"
-      assert response =~ "Child under 3"
-      assert response =~ "Family 4-pack"
-    end
-
```
anthonyshull, line 21: Hasn't been run in four years so I'm removing it.

test/dotcom_web/controllers/schedule/green_termini_api_test.exs
```diff
@@ -1,6 +1,6 @@
 defmodule DotcomWeb.ScheduleController.GreenTerminiApiTest do
-  use ExUnit.Case, async: true
```
anthonyshull, line 2: Don't need both this and the line below...

test/dotcom_web/controllers/trip_plan/feedback_test.exs
```diff
@@ -11,15 +11,17 @@ defmodule DotcomWeb.TripPlan.FeedbackTest do
       %{"things" => "are in here"}
     ]
   }
-  @expected_cache_key "dotcomweb.tripplan.feedback|local|123|2024-02-28T12:01:23|2"
+  @expected_cache_key "dotcom_web.trip_plan.feedback|local|123|2024-02-28T12:01:23|2"
```
anthonyshull, line 14: Fixes a failing test where the keys don't match.

test/mbta/api_test.exs
```diff
@@ -0,0 +1,111 @@
...
+  import Mox
+
+  setup :set_mox_global
+  setup :verify_on_exit!
+
+  describe "get_json/1" do
+    test "normal responses return a JsonApi struct" do
+      expect(HTTPoison.Mock, :get, fn _, _, _ ->
+        {:ok, %HTTPoison.Response{status_code: 200, body: ~s({"data": []})}}
+      end)
```
anthonyshull, line 15: An example of using Mox. This test will hit the underlying `HTTPoison.Mock`. All others will operate a level higher on the `MBTA.Api.Mock`.

test/stops/repo_test.exs
```diff
@@ -1,8 +1,6 @@
 defmodule Stops.RepoTest do
-  @moduledoc """
-  TODO: Mock API data in this whole file
-  """
```
anthonyshull, line 4: This is my favorite.

test/test_helper.exs
```diff
@@ -14,4 +14,9 @@ Mox.defmock(OpenTripPlannerClient.Mock, for: OpenTripPlannerClient.Behaviour)
 
 # Ensure the deps are all started
 Application.ensure_all_started(:dotcom)
+
+ExUnitSummary.start(:normal, %ExUnitSummary.Config{filter_results: :failed, print_delay: 100})
+ExUnit.configure(exclude: [external: true])
```
anthonyshull, line 19: Automatically excludes any test or module tagged with `:external`

test/mbta/stream_test.exs
```diff
@@ -0,0 +1,34 @@
...
+      types = [:add, :remove, :reset, :update]
+
+      known_events? =
+        [pid]
+        |> GenStage.stream()
+        |> Enum.take(4)
+        |> Enum.all?(fn %MBTA.Api.Stream.Event{event: type} -> Enum.member?(types, type) end)
+
+      assert known_events?
+    end
```
anthonyshull, line 32: I changed this test to *not* use Bypass, but didn't convert it to Mox because we'll have to figure out a way to mock a GenStage. Doesn't seem worth it to do just for this one test. Now that it uses external data, I tagged it as external.

test/predictions/repo_test.exs
```diff
@@ -1,24 +1,41 @@
 defmodule Predictions.RepoTest do
```
anthonyshull, line 1: This file needs a lot of work. I removed Bypass and started implementing Mox, but realized we need to come up with a strategy for testing repos because this is the only one that used Bypass. We probably need an entire suite of Factories and I didn't want to take on that chunk of work in this PR.

mix.exs
```diff
@@ -71,9 +71,6 @@ defmodule DotCom.Mixfile do
     [
       {:absinthe_client, "0.1.0"},
       {:benchfella, "0.3.5", [only: :dev]},
-      # latest version 2.1.0; cannot be upgraded because of errors when we use * in the url for tests
-      # rather than fixing tests and upgrading, we should convert the tests to use mox
-      {:bypass, "1.0.0", [only: :test]},
```
anthonyshull, line 76: What this has all been about...

coveralls.json
```diff
@@ -1,7 +1,7 @@
 {
   "coverage_options": {
-    "treat_no_relevant_lines_as_covered": true,
-    "minimum_coverage": 90
+    "minimum_coverage": 60,
```
anthonyshull, line 3: Set this to 60% since our exclusions count against coverage.

lib/cms/api/time_request.ex
amaisano, line 1: Why'd we lose the `@spec` here?

anthonyshull: We don't really use dialyzer and I'm trying to replace specs with detailed doc strings.

amaisano: OH man, we don't use dialyzer/dialyxer anymore? Where have i BEEN?

anthonyshull: BTW, this entire module is probably going to be rewritten. We can't upgrade HTTPoison without changing how we're using it. But, we want to unify on using a single HTTP Client. Right now we have multiple. So, we'll probably use something newer and more performant.

amaisano: Ok, please keep me and the CMS team in the loop when that change happens. Thanks for the heads up.

test/stops/repo_test.exs
thecristen: rofl it's a crowd pleaser!

image

thecristen (approved):
I appreciate your faithful translations of everything here, as well as the small enhancements and fixes throughout!

test/algolia/update_test.exs
```diff
@@ -1,95 +1,71 @@
...
-        Plug.Conn.send_resp(conn, 200, "ok")
+      expect(HTTPoison.Mock, :post, 2, fn url, _, _, _ ->
+        assert url == @url <> "/1/indexes/objects/batch"
+
+        {:ok, %HTTPoison.Response{status_code: 200}}
       end)
 
-      assert %{Algolia.MockObjects => result} =
-               Algolia.Update.update("http://localhost:#{bypass.port}")
+      response = Algolia.Update.update("http://localhost:9999")
```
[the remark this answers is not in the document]
anthonyshull: There is something in the Algolia code that only works with `localhost` or an actual Algolia URL.

PR #1926, opened 2024-03-19
Check for a prediction in the subway critical user journey

anthonyshull:


integration/scenarios/search-for-a-subway-line.js
```diff
@@ -7,16 +7,17 @@ exports.scenario = async ({ page, baseURL }) => {
   // But, the enter key does not load the page. So, we have to click the first result.
   await page
     .locator("div.search-wrapper input#input")
-    .pressSequentially("Green Line");
+    .pressSequentially("Blue Line");
```
thecristen, line 10: Let's change this to Red line since (a) you pointed out Blue line was working earlier today and (b) I think more eyes are on the Red line as opposed to the Blue line!

anthonyshull: For some reason the Red Line doesn't load consistently.

thecristen (approved):
I'd still rather test the Red Line, but it's better than nothing!

anthonyshull:
> I'd still rather test the Red Line, but it's better than nothing!

I think there is an issue with the rendering that it doesn't even load the schedule pane.

PR #1927, opened 2024-03-19
chore(DotcomWeb.Router): expose LiveDashboard behind basic auth

thecristen:


kotva006:
I see this pull request exposes the live view, but I don't see how it interacts with basic auth

thecristen:
@kotva006 `:basic_auth` was added as a plug in the pipeline. It should prompt you for a username and password if you try to visit the `/dashboard` route, did it not do that for you?

PR #1931, opened 2024-03-19
Match that a list of predictions isnt empty

anthonyshull:
We don't want to cache empty predictions. Sometimes, when a node comes online the API informs us that predictions haven't changed. It looks in the cache, sees that it is empty, and returns an error which then gets converted to an empty list.

### DIFF
--- a/lib/predictions/repo.ex
+++ b/lib/predictions/repo.ex
@@ -73,7 +73,7 @@ defmodule Predictions.Repo do
 
   @decorate cacheable(
               cache: @cache,
-              match: fn lst -> lst != [] end,
+              match: fn lst -> is_list(lst) && lst != [] end,
               on_error: :nothing,
               opts: [ttl: @ttl]
             )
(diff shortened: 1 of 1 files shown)


lib/predictions/repo.ex
```diff
@@ -71,7 +71,12 @@ defmodule Predictions.Repo do
     end
   end
 
-  @decorate cacheable(cache: @cache, on_error: :nothing, opts: [ttl: @ttl])
+  @decorate cacheable(
+              cache: @cache,
+              on_error: :nothing,
+              match: fn lst -> lst != [] end,
```
thecristen, line 76: Since it overrides the default matcher, wouldn't this mean this starts caching error tuples and such too?

anthonyshull: `fetch` only returns lists but, I can add a guard in there.

thecristen (approved):

PR #1940, opened 2024-03-22
feat: configure ExAws without env vars

thecristen:
In the interest of best security practice, we can remove reliance on hardcoded `AWS_ACCESS_KEY_ID` and
`AWS_SECRET_ACCESS_KEY` environment variables by configuring the `ExAws` dependency in this particular way! I also did some light updating to associated docs, which includes using AWS CLI.

Can the current devs confirm (via approval or comment on this PR) that this setup works on their machines? @kotva006 @anthonyshull @amaisano


anthonyshull:
My concern is with running this locally in Docker.

kotva006 (approved):
Works pretty slick

config/deps/deps.exs
```diff
@@ -1,5 +1,9 @@
 import Config
 
+config :ex_aws,
+  access_key_id: [{:awscli, "default", 30000}],
+  secret_access_key: [{:awscli, "default", 30000}]
```
anthonyshull, line 5: This assumes that the configuration being used is the default configuration. This won't work for people for whom these credentials are under another configuration name.

anthonyshull (changes requested):

config/deps/deps.exs
thecristen: Good call, I'll switch it to using the system's via (from `ExAws`):

> Alternatively, if you already have a profile name set in the AWS_PROFILE environment variable, you can use that with `{:awscli, :system, timeout}`

anthonyshull:
@thecristen pretty sure we don't need this now that we have this working both locally and in docker compose

thecristen:
Covered elsewhere.

PR #1943, opened 2024-03-25
When only given a key, we dont need to scan redis

anthonyshull:
It seems possible for L1 (ETS) and L2 (Redis) to get out of sync. When that happens, we can't flush the cache effectively if the key doesn't exist in Redis but does in ETS. We have to wait for the TTL. This PR makes it so that flushing a single key won't scan Redis for keys and will just send the command to delete it.


thecristen (approved):
💯 let's give it a spin!

PR #1946, opened 2024-03-27
fix(HoursOfOperation): Filter out schedules with same departure and a…

kotva006:
…rrival time

#### Summary of changes
The green line has a couple schedules that are single trips off of one line and on to another.
These cause the overall hours to be wrong.  This pull request filters the schedules from the results so we can correctly calculate the first and last trains of the line.

This is similar to the existing logic that was present on the front end before the hours card rewrite as seen here:
https://github.com/mbta/dotcom/commit/e57b9316bcb50c668f61772dcc8490883a5d58a0

**Asana Ticket:** [Green Line pages show inaccurate first/last train times](https://app.asana.com/0/555089885850811/1206891993416187)

---

#### General checks

### DIFF
--- a/lib/schedules/hours_of_operation.ex
+++ b/lib/schedules/hours_of_operation.ex
@@ -479,15 +479,17 @@ defmodule Schedules.HoursOfOperation do
     end)
   end
 
-  # This returns a single hours map for rapid transit routes
+  @doc """
+  This returns a single hours map for rapid transit routes
+  It ignores any departures that have the same start and end time, and are not the
+  terminus departure
+  """
   def departure_overall(data, headsigns, :rapid_transit) do
     departures = departure(data, headsigns, :rapid_transit)
 
-    # Remove single trip schedules (have the same first_departure and last_departure)
-    # from the overall calculation
     departures_filtered =
       Enum.filter(departures, fn x ->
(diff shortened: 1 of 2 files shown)


thecristen (changes requested):
Does bring a much more sensible display to the Green Line C! But I noticed a similar problem on the Green Line E, that surprisingly endures even with this change. 

image

Separately, I am noticing that the "last train" value seems to be a little off (can compare with the schedule finder's value for last train at the first stop - schedule finder usually shows a _later_ last train time than that displayed in the right corner). Going to be filing a separate ticket for that but it could be worth looking into if it helps uncover a different fix.

lib/schedules/hours_of_operation.ex
```diff
@@ -483,18 +483,21 @@ defmodule Schedules.HoursOfOperation do
   def departure_overall(data, headsigns, :rapid_transit) do
     departures = departure(data, headsigns, :rapid_transit)
 
-    if Enum.empty?(departures) do
+    # Remove single trip schedules (have the same first_departure and last_departure)
+    # from the overall calculation
```
anthonyshull, line 487: Should put any comments in a `@doc` block.

PR #1951, opened 2024-03-29
Chain service launch to reduce conflicts

anthonyshull:
Adds in dependencies between services so that multiple versions of dotcom don't have to compile the same things. This makes it slow and error prone. Instead, one version of the app launches and then a second one uses the work of the first one.


kotva006 (approved):
👍

PR #1955, opened 2024-04-02
Send images to sentry according to rate limit

anthonyshull:
When a monitor fails, we want to capture an image of what the "user" was seeing at the time. We'll send the images as attachments to Sentry. But, we are limited to 1GB per month. So, we want to rate limit how many we send. Currently, I limited it to 1 per hour. I also set the quality level of the image to 50% which makes them < 100kb. So, 1 * 24 * 30 * 100kb = 72MB.

If we can reduce the number of errors we're seeing in scenarios (always some kind of timeout), we can make this more sophisticated and rate limit by metric type instead of all metrics.


package.json
```diff
@@ -1,12 +1,13 @@
...
   "name": "dotcom",
   "dependencies": {
-    "artillery": "2.0.3",
-    "axios": "1.6.7",
+    "artillery": "2.0.8",
+    "axios": "1.6.8",
     "hot-shots": "10.0.0",
     "node-cron": "3.0.3",
-    "node-json-logger": "0.2.0",
+    "node-json-logger": "0.2.1",
```
anthonyshull, line 8: 📓 I just upgraded these while I was already in the file.

integration/tests/monitor/rate-limiter.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+const RateLimiter = require("../../monitor/rate-limiter");
+
+describe("lease()", () => {
+  it("rejects a lease within the window", () => {
+    const rateLimiter = new RateLimiter(1000, 1);
+
+    assert.equal(rateLimiter.lease(), true);
+    assert.equal(rateLimiter.lease(), false);
+  });
```
anthonyshull, line 12: 📓 This probably isn't the optimal way to test this. There was a lot of movement around testing between Node 18 (our version) and 21 (the latest version). We should come back and try to mock the timeout when we've upgraded.

integration/monitor/worker.js
```diff
@@ -26,8 +26,11 @@ parentPort.on("message", async (_) => {
 
   try {
     await scenario({ page, baseURL });
-  } catch(e) {
-    logger.error({ metric, error: e.message });
+  } catch(exception) {
+    const screenshot = await page.screenshot({quality: 50, type: "jpeg"});
```
anthonyshull, line 30: 📓 I set the quality to 50%, which resulted in an 80kb image.

integration/monitor/all-scenarios.js
```diff
@@ -1,12 +1,38 @@
...
 
 const { fileToMetricName } = require("../utils");
+const RateLimiter = require("./rate-limiter");
 
 const filesPath = path.join(__dirname, "..", "scenarios");
 
+/*
+ * Create a rate limiter that allows only one lease per hour.
+ */
+const rateLimiter = new RateLimiter(60 * 60 * 1000, 1);
```
anthonyshull, line 15: 📓 The limit is 1 image sent every hour.

integration/monitor/all-scenarios.js
```diff
@@ -1,12 +1,38 @@
...
+/*
+ * Initialize Sentry with the DSN and environment from the environment variables.
+ * Add a beforeSend callback that will only send events if the rate limiter allows it.
+ */
+Sentry.init({
+  dsn: process.env.SENTRY_DSN,
+  environment: process.env.SENTRY_ENVIRONMENT,
+  beforeSend(event) {
+    if (process.env.SENTRY_ENVIRONMENT == "prod") {
+      return rateLimiter.lease() ? event : null;
```
anthonyshull, line 26: 📓 Returning null in this callback does not send the event to Sentry.

https://docs.sentry.io/platforms/node/configuration/filtering/

integration/monitor/rate-limiter.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * A simple rate limiter that allows a limited number of leases in a window of time.
+ * @constructor
+ * @param {number} window - The window of time in milliseconds in which to count leases.
+ * @param {number} limit - The maximum number of leases in the window. Default 1.
+ */
+class RateLimiter {
```
anthonyshull, line 7: 📓 I looked at a few other rate limiters. But, they were designed more for rate limiting requests to a web app. One kept all the requests in a queue for later execution. That isn't what we want. We just want to discard if over the limit.

integration/monitor/all-scenarios.js
```diff
@@ -16,13 +42,30 @@ const workers = fs.readdirSync(filesPath).map((file) => {
...
 });
 
-cron.schedule("* * * * *", (_) => {
+/*
+ * A task runs every minute on the minute.
+ * Spread out the worker runs over the minute to avoid overwhelming the system.
+ * If we receive a message from a worker, it means that there was a failure.
+ * Capture the exception with Sentry and attach a screenshot to the event.
+ */
+cron.schedule("* 5-23 * * *", (_) => {
```
anthonyshull, line 51: 📓 Sneaking this in there to only run the monitor from 6am to midnight. A lot of the scenarios start failing when trains and buses stop running.

integration/monitor/all-scenarios.js
kotva006: Its a nit-pick, but the ternary (`?`) could be in the `if` above so this line doesn't need to also return `null` like the default return below

package.json
kotva006: These updates in the future should be done outside of unrelated changes.

kotva006 (approved):

PR #1958, opened 2024-04-03
Upgrades dependencies with focus on Credo

anthonyshull:
https://app.asana.com/0/555089885850811/1206615408159226/f


.github/workflows/tests.yml
```diff
@@ -77,6 +77,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: ./.github/actions/setup-all
+    - run: git fetch origin master:master
```
anthonyshull, line 80: 📓 This is what fixes the issue with credo not being able to find the master branch do run a diff.

kotva006: When we run `credo` locally does it find master successfully?

anthonyshull: Yep.

kotva006 (approved):

PR #1961, opened 2024-04-05
Allow for node connection and lower interval for faster startup

anthonyshull:
This will start our development Phoenix apps in a way that they can be communicated with while running.


kotva006 (approved):
👍

PR #1965, opened 2024-04-05
Add ignores and comments

anthonyshull:
This will get dependabot to stop bothering us about dependencies that we already know we can't upgrade. When we fix the dependency, we'll remove it from the ignore list.


kotva006 (approved):
😄

PR #1968, opened 2024-04-08
Charge a transfer fee for bus to subway transfers

anthonyshull:
Currently, for a subway -> bus transfer the fare for the first leg is $2.40 which covers the transfer. But, for bus -> subway the fare for the first leg is $1.70. So, we need to add the extra $0.70 transfer fee.


anthonyshull:
Waiting on verification from Claire that this solves more problems than it creates.

lib/dotcom_web/views/trip_plan_view.ex
```diff
@@ -1,3 +1,4 @@
+# credo:disable-for-this-file
```
anthonyshull, line 1: 📓 Disabling credo for this file because we're going to rewrite it...most likely moving this logic to a file in dotcom rather than dotcom_web since it has to do with domain logic rather than display.

PR #1977, opened 2024-04-12
refactor(StationAmenityCard): consolidate elevator/escalator logic

thecristen:
#### Summary of changes

**Asana Ticket:** [🐞 Alerts text does not reflect current status](https://app.asana.com/0/555089885850811/1206787777338665/f)

> [!Note]
> Viewing this fix relies on test alerts I've made available in the dev-green API

Ended up refactoring the mostly-identical `ElevatorsAmenityCard` and `EscalatorsAmenityCard` components into a single component that handles both.

The change that actually fixes the bug is passing `alerts.filter(isCurrentLifecycle).length` instead of `alerts.length` to the `availabilityMessage` function. No other changes were made further up the tree, because we still want other parts of that component to be able to show upcoming alerts too. Using `alerts.filter(isCurrentLifecycle)` also matches the logic used to render elevator/escalator alert status in tabular form within the modal.

---

#### General checks


assets/js/fixedsticky.js
```diff
@@ -7,11 +7,9 @@ export default function($) {
...
     $(".fixedsticky").fixedsticky();
   }
-  window.addEventListener(
-    "load",
-    () => window.nextTick(fixedsticky),
-    { passive: true }
-  );
+  window.addEventListener("load", () => window.nextTick(fixedsticky), {
+    passive: true
+  });
```
kotva006, line 12: It looks like nothing changed here other than whitespace.  Is this intentional?

assets/js/google-translate.js
```diff
@@ -27,8 +27,7 @@ const languages = [
   ["vi", "Vietnamese"]
 ];
 
-export default () =>
-  window.addEventListener("load", handleTranslation);
+export default () => window.addEventListener("load", handleTranslation);
```
kotva006, line 30: Same as above. Are these changes intnetional?

thecristen: My formatter did it! Seems to pass the suite though

kotva006: Okay, I bring it up because this can muddle the git history as no logic was changed here.  Its a small thing, but I wanted to bring it up

kotva006 (approved):
LGTM

PR #1980, opened 2024-04-12
adjust covered files and tagged tests

thecristen:
No ticket, just a minor improvement in coverage, mostly by replacing a few `@moduletag :external` with more targeted `@tag :external`. It also didn't seem to make sense to me to expect coverage of both behaviors and implementations so I omitted one, open to alternate suggestions around that.

### DIFF
--- a/coveralls.json
+++ b/coveralls.json
@@ -1,30 +1,35 @@
 {
   "coverage_options": {
-    "minimum_coverage": 60,
+    "minimum_coverage": 62,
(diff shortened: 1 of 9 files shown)


anthonyshull:
The idea was to tag modules with high % of external tests as external because the entire file will need to be considered when being rewritten. This just kind of replaces one tag with a lot of tags. But, not really important enough to disagree about.

anthonyshull (approved):

PR #1987, opened 2024-04-16
chore: improve error handling in `DotcomWeb.ScheduleController.HoursController` and `Schedules.HoursOfOperation`

thecristen:
No ticket, just noticing lots of errors in the backend in `DotcomWeb.ScheduleController.HoursController`, usually when the `route_id` does not correspond to a valid route. I wrote some logic that can account for this a bit more smoothly.

Locally, and a few times on dev, I also noticed errors in `Schedules.HoursOfOperation` when `departure(data, headsigns, :rapid_transit)` returns `:no_service`, so I wrote logic to accommodate this too.


kotva006 (approved):
LGTM

PR #1988, opened 2024-04-16
fix(ModuleNesting): Fix Credo Warnings about Module Nesting

kotva006:
#### Summary of changes
Fix Credo Warnings related to module nesting and alias.

**Asana Ticket:** [Credo Design.AliasUsage](https://app.asana.com/0/555089885850811/1207049863049668)

---

#### General checks

### DIFF
--- a/lib/dotcom/trip_plan/itinerary_row_list.ex
+++ b/lib/dotcom/trip_plan/itinerary_row_list.ex
@@ -4,6 +4,7 @@ defmodule Dotcom.TripPlan.ItineraryRowList do
 
   An optional to and from name can be passed in.
   """
+  alias Dotcom.TripPlan.Alerts, as: TripPlanAlerts
   alias Dotcom.TripPlan.ItineraryRow
   alias TripPlan.Itinerary
   alias Stops.Stop
@@ -62,7 +63,7 @@ defmodule Dotcom.TripPlan.ItineraryRowList do
   defp get_alerts(itinerary, deps) do
     itinerary.start
     |> deps.alerts_repo.()
-    |> Dotcom.TripPlan.Alerts.filter_for_itinerary(
+    |> TripPlanAlerts.filter_for_itinerary(
       itinerary,
       route_by_id: deps.route_mapper,
       trip_by_id: deps.trip_mapper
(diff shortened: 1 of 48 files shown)


amaisano (approved):
The one file I worked on looks fine - thanks!

anthonyshull (changes requested):
We shouldn't rename aliases with `as` as it will make the code less clear. In cases where we have naming conflicts, we should use the configuration options to result in the clearest code. You can alter the configuration so that certain names won't be flagged. `DateTime` is already in that list. https://hexdocs.pm/credo/Credo.Check.Design.AliasUsage.html#module-excluded_lastnames

lib/dotcom_web/controllers/bus_stop_change_controller.ex
```diff
@@ -8,6 +8,7 @@ defmodule DotcomWeb.BusStopChangeController do
 
   alias Alerts.{Alert, HistoricalAlert}
   alias Alerts.InformedEntity, as: IE
+  alias Alerts.Cache.BusStopChangeS3
```
anthonyshull, line 11: Should make this alpha with those above.

lib/facilities/repo.ex
```diff
@@ -3,11 +3,13 @@ defmodule Facilities.Repo do
   Repo to get facilitiy information.
   """
 
+  alias MBTA.Api.Facilities
```
anthonyshull, line 6: Alias should go below `use`. We want to order include directives based on the level of their effect. `require`, `use`, `import`, `alias`. Alias is last because it has no effect other than giving a shorter name.

lib/dotcom/trip_plan/itinerary_row_list.ex
```diff
@@ -4,6 +4,7 @@ defmodule Dotcom.TripPlan.ItineraryRowList do
 
   An optional to and from name can be passed in.
   """
+  alias Dotcom.TripPlan.Alerts, as: TripPlanAlerts
```
anthonyshull, line 7: This is just one example, but the same applies to all. `Alerts` should be added to the excluded lastnames list because `TripPlan.Alerts` is more clear than `TripPlanAlerts.`

anthonyshull (approved):

PR #1990, opened 2024-04-17
Update sentry configuration

anthonyshull:
https://app.asana.com/0/home/1204957179328166/1206719459037546

Related Devops PR: https://github.com/mbta/devops/pull/1713

### DIFF
--- a/.github/workflows/deploy-release.yml
+++ b/.github/workflows/deploy-release.yml
@@ -28,3 +28,14 @@ jobs:
         :t-intensifies: `prod` updated to release `${{ github.event.release.tag_name }}`
       failure-message: |
         release `${{ github.event.release.tag_name }}` failed to update `prod`
+
+  sentry:
+    needs: [release]
+    uses: getsentry/action-release@v1.7.0
+    env:
+      SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+      SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+      SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+    with:
+      environment: production
+      version: ${{ github.event.release.tag_name }}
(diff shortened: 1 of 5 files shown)


config/deps/sentry.exs
```diff
@@ -2,9 +2,8 @@ import Config
 
 # Centralize Error reporting
 config :sentry,
-  enable_source_code_context: false,
-  root_source_code_path: File.cwd!(),
+  enable_source_code_context: true,
```
anthonyshull, line 5: 📓 This is the only change here. Adding source code context.

kotva006 (approved):
Looks good to me

.github/workflows/deploy-release.yml
```diff
@@ -28,3 +28,14 @@ jobs:
...
+
+  sentry:
+    needs: [release]
+    uses: getsentry/action-release@v1.7.0
+    env:
+      SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+      SENTRY_ORG: mbtace
+      SENTRY_PROJECT: mbta-dotcom
+    with:
+      environment: production
```
thecristen, line 40: I believe we want `prod` rather than `production` here, to match the sentry config's `included_environments`.

.github/workflows/deploy-release.yml
```diff
@@ -28,3 +28,14 @@ jobs:
         :t-intensifies: `prod` updated to release `${{ github.event.release.tag_name }}`
       failure-message: |
         release `${{ github.event.release.tag_name }}` failed to update `prod`
+
```
thecristen, line 31: https://github.com/getsentry/action-release?tab=readme-ov-file#troubleshooting The action suggests doing `actions/checkout` first, it seems like it tries to compute a previous release.

.github/workflows/deploy-release.yml
thecristen: probably a typo from years ago 🥲

thecristen (approved):

PR #1992, opened 2024-04-17
feat: redirect bad schedules URLs

thecristen:
No ticket. I'm just tired of seeing tons of failed visits to `/schedules/:route_id/schedule` (e.g. the prior pattern for a schedule page, likely linked from many old pages across the Internet) constantly.

I wrote this plug, inspired by `Www.Redirector`, to redirect those to the appropriate place (`/schedules/:route_id`).

### DIFF
--- a/lib/dotcom_web/plugs/schedule_redirector.ex
+++ b/lib/dotcom_web/plugs/schedule_redirector.ex
@@ -1,0 +1,25 @@
+defmodule DotcomWeb.Plugs.ScheduleRedirector do
+  @moduledoc """
+  Catch-all for unsupported `/schedules/:route_id/*` URLs.
+  Redirects to the default `/schedules/:route_id` path.
+  """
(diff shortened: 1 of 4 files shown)


lib/dotcom_web/plugs/schedule_redirector.ex
```diff
@@ -0,0 +1,25 @@
+defmodule DotcomWeb.Plugs.ScheduleRedirector do
+  @moduledoc """
+  Catch-all for unsupported `/schedules/:route_id/*` URLs.
+  Redirects to the default `/schedules/:route_id` path.
+  """
+
+  @behaviour Plug
```
anthonyshull, line 7: 🌤️ I like to put the behaviours after the include statements so they're closer to the implementation definitions.

anthonyshull (approved):

PR #1993, opened 2024-04-17
Dont run credo on main

anthonyshull:


.github/workflows/tests.yml
thecristen, line 1: Any thoughts on skipping other jobs, or perhaps the entire suite on main?

anthonyshull: Yeah, that's a good point. I guess the question is what are we trying to get out of running these jobs on a push to main. The only thing I can think that might be useful is showing this branch is in good condition. If we turn off the ability to merge when jobs are failing then that wouldn't be a problem anymore.

thecristen: Yeah agreed. I find it kinda nice seeing the test results on main, but since all PRs have to go through it anyway it feels unnecessary.

thecristen (approved):

PR #1999, opened 2024-04-19
Rewrite MBTA Api using Req instead of HTTPoison

anthonyshull:
https://app.asana.com/0/555089885850811/1206825410497903/f

We want to instrument calls to the MBTA Api with telemetry. Req uses Finch which does that out of the box. It also does other things automatically like decompressing gzip. So, it makes sense to just switch to using Req.

Devops PR: https://github.com/mbta/devops/pull/1730


config/runtime.exs
```diff
@@ -107,10 +107,11 @@ else
...
 
-config :dotcom,
-  v3_api_base_url: System.get_env("V3_URL"),
-  v3_api_key: System.get_env("V3_API_KEY"),
-  v3_api_version: System.get_env("V3_API_VERSION", "2019-07-01")
+config :dotcom, :mbta_api,
+  base_url: System.get_env("MBTA_API_BASE_URL"),
+  enable_experimental_features: System.get_env("MBTA_API_ENABLE_EXPERIMENTAL_FEATURES"),
+  key: System.get_env("MBTA_API_KEY"),
+  version: System.get_env("MBTA_API_VERSION", "2019-07-01")
```
anthonyshull, line 114: 📓 I am renaming all of these ENV vars to be in line with what they do

.github/workflows/tests.yml
```diff
@@ -15,8 +15,8 @@ concurrency:
 
 env:
   MIX_ENV: test
-  V3_URL: ${{ secrets.V3_URL }}
-  V3_API_KEY: ${{ secrets.V3_API_KEY }}
+  MBTA_API_BASE_URLL: ${{ secrets.MBTA_API_BASE_URL }}
```
thecristen, line 18: 👀 typo!

kotva006 (approved):

PR #2001, opened 2024-04-22
fix(CMSRepo): Support cache key generation for map params

kotva006:
#### Summary of changes
Parse the params of a map to a string for cache key generation
https://app.asana.com/0/555089885850811/1206990758684890

---

#### General checks

### DIFF
--- a/lib/cms/repo.ex
+++ b/lib/cms/repo.ex
@@ -234,16 +234,20 @@ defmodule CMS.Repo do
   end
 
   defp params_to_string(params) when params == %{}, do: ""
-  defp params_to_string(nil), do: ""
   defp params_to_string(params) when is_binary(params), do: params
 
   defp params_to_string(params) when is_map(params) do
-    [head | tail] = Enum.map(params, fn {k, v} -> "#{k}=#{params_to_string(v)}" end)
+    [head | tail] =
+      params
+      |> Enum.reduce([], &CMS.API.HTTPClient.stringify_params/2)
+      |> Enum.map(fn {k, v} -> "#{k}=#{v}" end)
 
     ["?#{head}", "#{Enum.join(tail, "&")}"]
     |> Enum.reject(&(&1 == ""))
     |> Enum.join("&")
   end
+
+  defp params_to_string(_), do: ""
 
   @spec view_or_preview(String.t(), map) :: {:ok, map} | {:error, API.error()}
   defp view_or_preview(path, %{"preview" => _, "vid" => "latest"} = params) do
--- a/lib/cms/repo.ex
+++ b/lib/cms/repo.ex
@@ -237,14 +237,17 @@ defmodule CMS.Repo do
   defp params_to_string(params) when is_binary(params), do: params
 
   defp params_to_string(params) when is_map(params) do
-    [head | tail] =
-      params
-      |> Enum.reduce([], &CMS.API.HTTPClient.stringify_params/2)
-      |> Enum.map(fn {k, v} -> "#{k}=#{v}" end)
+    case params
+         |> Enum.reduce([], &CMS.API.HTTPClient.stringify_params/2)
+         |> Enum.map(fn {k, v} -> "#{k}=#{v}" end) do
+      [head | tail] ->
+        ["?#{head}", "#{Enum.join(tail, "&")}"]
+        |> Enum.reject(&(&1 == ""))
+        |> Enum.join("&")
 
-    ["?#{head}", "#{Enum.join(tail, "&")}"]
-    |> Enum.reject(&(&1 == ""))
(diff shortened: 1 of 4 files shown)


amaisano (approved):

test/cms/repo_test.exs
```diff
@@ -97,6 +97,16 @@ defmodule CMS.RepoTest do
                "cms.repo" <> String.replace(path, "/", "|") <> "?bam=bop&baz=qux"
     end
 
+    test "generates the correct key for a map" do
+      path = "/foo/bar"
+      params = %{"biz" => "bang", "data" => %{"some" => "map"}}
+
+      assert Repo.generate(nil, nil, [path, params]) ==
+               "cms.repo" <> String.replace(path, "/", "|") <> "?biz=bang&data=?some=map"
```
anthonyshull, line 105: 💥 This won't be parseable by the CMS. `?biz=bang&data=?some=map` will just take `?some=map` as a string when you get the params.

lib/cms/repo.ex
```diff
@@ -234,9 +234,11 @@ defmodule CMS.Repo do
   end
 
   defp params_to_string(params) when params == %{}, do: ""
+  defp params_to_string(nil), do: ""
```
anthonyshull, line 237: We should just make it `(_), do: ""` and put this after the string and map input types.

lib/cms/repo.ex
```diff
@@ -234,9 +234,11 @@ defmodule CMS.Repo do
...
+  defp params_to_string(params) when is_binary(params), do: params
 
   defp params_to_string(params) when is_map(params) do
-    [head | tail] = Enum.map(params, fn {k, v} -> "#{k}=#{v}" end)
+    [head | tail] = Enum.map(params, fn {k, v} -> "#{k}=#{params_to_string(v)}" end)
```
anthonyshull, line 241: We should not handle nested maps here as they don't really make sense. I would investigate what is calling this with a nested map. It is probably wrong.

anthonyshull (changes requested):

anthonyshull:
Look at https://github.com/mbta/dotcom/blob/abee6a61e4057d1675f492fb6bb3bcc3fd6e7797/lib/cms/api/http_client.ex#L48 because that is where it is converting nested maps to a query string. We could either make that public and use it to generate the cache key. But, I am also suspicious about whether or not a nested map should be passed in as params.

test/cms/repo_test.exs
anthonyshull: This is the structure that is being passed in: `?schedule_finder[direction_id]=1&schedule_finder[origin]=3939`

anthonyshull: Also, I can confirm that we pass those params on to the CMS and they are ignored.

amaisano: > Also, I can confirm that we pass those params on to the CMS and they are ignored.

There is a whitelist filter for CMS params somewhere in the dotcom code. Not sure how it interacts with this stuff but you may want to track it down.

kotva006: If the CMS ignores the params, then I am just going to reject the map

anthonyshull: There might be cases where we don't ignore the params. That's why I'm thinking just make the params to string function that the CMS code is using public and then use that to make the key. That way they're guaranteed to be the same.

anthonyshull (changes requested):
For the URL in the Sentry error, this code produces the following key:

"cms.repo|paragraphs|multi-column|bus-fares?schedule_finder=?direction_id=1&origin=3939"

It should match the actual URL params: `?schedule_finder[direction_id]=1&schedule_finder[origin]=3939`

anthonyshull:
I would go through and debug the entire process with the URL in the Sentry error because it looks like the params are being used for the cache but they are getting removed when a request is sent to the CMS. Our goal is to make sure that the cache key makes sense with the request that is actually being sent.

anthonyshull:
Let me know if you need more direction.

lib/cms/repo.ex
```diff
@@ -234,15 +234,24 @@ defmodule CMS.Repo do
...
+        ["?#{head}", "#{Enum.join(tail, "&")}"]
+        |> Enum.reject(&(&1 == ""))
+        |> Enum.join("&")
 
-    ["?#{head}", "#{Enum.join(tail, "&")}"]
-    |> Enum.reject(&(&1 == ""))
-    |> Enum.join("&")
+      _ ->
+        ""
+    end
```
anthonyshull, line 250: You could do this using pattern matching to support a list.

anthonyshull (approved):

PR #2027, opened 2024-05-02
refactor(RoutePatterns.Repo): use factories and mock for tests

thecristen:
No ticket, just a refactor that should help me test the timetable changes that's incoming.

This also removes what was left of our code relying on shape priority. As far as I can tell, we had already stopped parsing it. This field was removed in V3 API shapes in version 2020-05-01.


test/leaflet/map_data/polyline_test.exs
```diff
@@ -1,15 +1,22 @@
...
+  import Test.Support.Factory.RoutePattern
+
   alias Leaflet.MapData.Polyline
 
-  @route_patterns_repo_api Application.compile_env!(:dotcom, :route_patterns_repo_api)
+  setup do
+    route_pattern =
+      build(:route_pattern,
+        representative_trip_polyline:
+          "gfsaGvlaqL[Ek@Ck@A[?]BU?sAHg@DWDODKJMPM\\KNEQAOU`@c@`@EBC@[PeCn@g@HM@??[@cBG_BKmBIq@Eq@GyAGoAGcDQ??OAiDQoFYcCO_AEM?OA??yAIyAIeBK}AK_ACy@FiAPu@N??YFa@FSDaARw@\\UL[T??WT_B`Bm@r@e@n@Sb@_@|@gAxCq@hBIT]fACDGL??cAhBc@j@k@n@m@v@MNg@j@WX??q@v@oBxBiArA}@fAc@f@??STSTwAbBq@t@eAnAaAjA]b@qAxAMNuA~As@z@??MPsB~B{@jAgBbCgA~Ag@x@mA`C??IR[x@}@pCw@bCGP??y@vBGf@O\\}@tBSV??_AtAkAtAY`@mAzAMP??IJu@~@i@r@Y^cAtAa@h@w@fAe@n@KN??e@l@m@x@g@r@}@nAY\\w@hAg@p@OR??kBjCeB`CiA|AuAlBQV??Y^{AvBy@hAY`@w@fA??UZ]`@[\\k@b@u@^_ChAwBfAQH??uAr@{@b@uAv@yAz@a@`@??CBWXiApAaApAo@t@kAdBMX??a@~@[`AEPe@pAc@`BIZQr@??Sv@_@zAOfAAZAl@S`C]xECh@????]lEShCOjBMjB?H??UhCWfDOlBYxDMlAIf@Kd@s@lCg@zAEN??Uj@sAjDi@xAy@vBKRO`@??IT{@nBO^u@bBiAlCw@nBYh@??KRq@jBwA`DsBzE[b@KP??}@tAORc@l@o@~@uApBS`@IRMl@SnCEf@??AV}@`FQ`AOv@MhAEp@CvA?~AAl@???VCtBC|BC~CEbEAf@??Ad@[tDOpBAJaAPGdA"
```
anthonyshull, line 12: Does it need to be this long? Is there a simpler polyline?

anthonyshull (changes requested):
I think the factories can be minimized and organized to be reusable. You don't need to manually check the attributes.

See this as an example:

https://github.com/mbta/my_charlie/blob/115470dd086e2e0cd5f665ab06426de518447f21/test/support/factories/replacement_requests/factory.ex#L12

https://github.com/mbta/my_charlie/blob/115470dd086e2e0cd5f665ab06426de518447f21/lib/mix/tasks/seed/youth_pass_replacements.ex#L162

assets/ts/__v3api.d.ts
```diff
@@ -349,20 +349,19 @@ export interface Shape {
 }
 
 export interface RoutePattern {
-  typicality: 1 | 2 | 3 | 4;
-  time_desc: string | null;
-  route_id: string;
+  canonical: boolean;
```
anthonyshull, line 352: 👍 for alpha

assets/ts/schedule/components/ScheduleDirection.tsx
```diff
@@ -148,10 +148,6 @@ const ScheduleDirection = ({
         if (result.length === 0 || current.typicality < result[0].typicality)
           return [current];
         if (current.typicality === result[0].typicality) {
-          if (result[0].shape_priority < 0 && current.shape_priority > 0)
```
anthonyshull, line 151: I'm not sure why this is being removed in the context of this PR. A note in the PR would help.

config/config.exs
```diff
@@ -5,6 +5,7 @@ config :elixir, ansi_enabled: true
 config :dotcom, :httpoison, HTTPoison
 
 config :dotcom, :mbta_api_module, MBTA.Api
+config :dotcom, :repo_modules, route_patterns: RoutePatterns.Repo
```
anthonyshull, line 8: 👍

test/dotcom_web/controllers/stop_controller_test.exs
```diff
@@ -138,36 +136,23 @@ defmodule DotcomWeb.StopControllerTest do
...
-                         "headsign" => "Tree Hill",
-                         "name" => "Here Square - Tree Hill"
-                       }
-                       | _
-                     ]
-                   }
+      response =
+        get(conn, stop_path(conn, :grouped_route_patterns, "place-here")) |> json_response(200)
+
+      assert %{
```
anthonyshull, line 142: Is this hard-coded somewhere? This test is a little hard to figure out.

thecristen: I totally forgot about this and because the whole module's excluded it didn't fail the tests! I'll fix this.

assets/ts/schedule/components/ScheduleDirection.tsx
thecristen: Thanks, totally forgot to mention it. Priority was an attribute of V3 API shapes that's been removed in more recent versions, and as far as I could tell we had already stopped reading/parsing it in our own codebase. Added a note to the main PR text too.

test/route_patterns/repo_test.exs
```diff
@@ -1,50 +1,84 @@
...
+
+  alias RoutePatterns.RoutePattern
+
+  setup :verify_on_exit!
 
   describe "get" do
     test "returns a single route pattern" do
-      assert %RoutePattern{id: "111-5-0"} = Repo.get("111-5-0")
+      expect(MBTA.Api.Mock, :get_json, fn url, _ ->
+        assert url == "/route_patterns/111-5-0"
```
anthonyshull, line 14: Is there something magical about this route pattern number? If not, we should use a random one like the test below.

anthonyshull (changes requested):

PR #2029, opened 2024-05-03
fix(Schedules): Fix alerts not showing up for stops with inbound trains

kotva006:
#### Summary of changes
This fixes the start point of a shuttle disruption from showing predictions and showing the effect title instead

https://app.asana.com/0/home/1201123880594717/1206362494901588

---

#### General checks

### DIFF
--- a/lib/predictions/repo.ex
+++ b/lib/predictions/repo.ex
@@ -32,7 +32,7 @@ defmodule Predictions.Repo do
     opts
     |> add_all_optional_params()
     |> fetch()
-    |> filter_predictions(nil)
+    |> filter_predictions()
     |> load_from_other_repos
   end
 
@@ -56,6 +56,8 @@ defmodule Predictions.Repo do
 
   @spec filter_predictions([Parser.record()] | {:error, any}, DateTime.t() | nil) ::
           [Parser.record()] | {:error, any}
+  defp filter_predictions(predictions, min_time \\ nil)
+
   defp filter_predictions({:error, error}, _) do
     {:error, error}
   end
--- a/lib/predictions/repo.ex
+++ b/lib/predictions/repo.ex
@@ -24,7 +24,7 @@ defmodule Predictions.Repo do
     opts
     |> add_all_optional_params()
     |> cache_fetch()
-    |> filter_by_min_time(Keyword.get(opts, :min_time))
+    |> filter_predictions(Keyword.get(opts, :min_time))
     |> load_from_other_repos
   end
 
@@ -32,6 +32,7 @@ defmodule Predictions.Repo do
     opts
     |> add_all_optional_params()
     |> fetch()
+    |> filter_predictions(nil)
     |> load_from_other_repos
   end
 
@@ -51,6 +52,18 @@ defmodule Predictions.Repo do
       nil -> params
       value -> Keyword.put(params, key, value)
     end
+  end
+
+  @spec filter_predictions([Parser.record()] | {:error, any}, DateTime.t() | nil) ::
+          [Parser.record()] | {:error, any}
+  defp filter_predictions({:error, error}, _) do
+    {:error, error}
+  end
+
+  defp filter_predictions(predictions, min_time) do
+    Enum.filter(predictions, fn prediction ->
+      has_departure_time?(prediction) && after_min_time?(prediction, min_time)
+    end)
   end
 
   defp fetch(params) do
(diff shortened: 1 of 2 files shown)


test/predictions/repo_test.exs
```diff
@@ -58,6 +59,67 @@ defmodule Predictions.RepoTest do
       end
     end
 
+    test "filtes out predictions with no departure" do
```
anthonyshull, line 62: Typo

test/predictions/repo_test.exs
```diff
@@ -58,6 +59,67 @@ defmodule Predictions.RepoTest do
...
+                ],
+                "trip" => [],
+                "vehicle" => [],
+                "stop" => [
+                  %{id: "StopID"}
+                ]
+              }
+            }
+          ]
+        }
```
anthonyshull, line 110: Are all of these attributes necessary for exercising the test? Would it work just setting the departure_time? These two structs are exactly the same except for that one attribute, so there should be a shorter and clearer way to communicate that.

test/predictions/repo_test.exs
```diff
@@ -58,6 +59,67 @@ defmodule Predictions.RepoTest do
...
+      end)
+
+      with_mocks([
+        {Stops.Repo, [:passthrough], get_parent: fn _ -> %Stops.Stop{id: "Parent"} end},
+        {Routes.Repo, [:passthrough], get: fn _ -> Routes.MockRepoApi.get("Red") end}
+      ]) do
+        predictions = Repo.all(route: "39")
```
anthonyshull, line 117: Is this a magic string? What happens if we do `Repo.all(route: "38")`?

anthonyshull (changes requested):

test/predictions/repo_test.exs
kotva006: Yes, all these fields are required.  The parsing errors out if any of these fields are missing.  Ideally I'd like to mock away some of that, but it seemed outside the scope of this fix.

test/predictions/repo_test.exs
kotva006: Nope, its just a random number (I copied this call from the test above)

test/predictions/repo_test.exs
```diff
@@ -2,6 +2,7 @@ defmodule Predictions.RepoTest do
   use ExUnit.Case, async: false
   @moduletag :external
 
+  import Mock
```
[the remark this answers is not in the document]
kotva006: I am aware, please see my comment in slack

test/predictions/repo_test.exs
```diff
@@ -58,6 +59,67 @@ defmodule Predictions.RepoTest do
       end
     end
 
+    test "filtes out predictions with no departure" do
+      five_minutes_in_future = DateTime.add(Timex.now(), 5, :minute)
+
+      {:ok, five_minutes_in_future_string} =
+        Timex.format(five_minutes_in_future, "{ISO:Extended:Z}")
+
```
anthonyshull, line 67: Since you don't need the :ok and an error would just cause the test to fail you can do:

```
departure_time =
  Timex.now()
  |> DateTime.add(5, :minute)
  |> Timex.format!("{ISO:Extended:Z}")
```

test/predictions/repo_test.exs
```diff
@@ -58,6 +58,102 @@ defmodule Predictions.RepoTest do
...
+                ],
+                "trip" => [],
+                "vehicle" => [],
+                "stop" => [
+                  %{id: "StopID"}
+                ]
+              }
+            }
+          ]
+        }
```
anthonyshull, line 109: > These two structs are exactly the same except for that one attribute, so there should be a shorter and clearer way to communicate that.

Because these two structs are identical except for one attribute, you could define it once and then override that attribute for the second. It will make the test shorter. That also helps to communicate the difference between the two.

test/predictions/repo_test.exs
```diff
@@ -58,6 +58,102 @@ defmodule Predictions.RepoTest do
...
+      expect(MBTA.Api.Mock, :get_json, fn _, _ ->
+        %JsonApi{
+          data: [
+            %JsonApi.Item{
+              id: "Red",
+              attributes: %{
+                "type" => "1",
+                "long_name" => "Red Test Subway",
+                "direction_names" => ["North Test Name", "South Test Name"],
+                "direction_destinations" => ["North Test", "South Test"]
```
anthonyshull, line 122: These are all magic strings. As per the testing doc we all reviewed and agreed to, we should be using Faker for these (and all others in this file).

https://www.notion.so/mbta-downtown-crossing/The-Four-Legs-of-Quality-Programming-b35248799da84e02b93887361d5bf343

test/predictions/repo_test.exs
```diff
@@ -58,6 +58,102 @@ defmodule Predictions.RepoTest do
       end
     end
 
+    test "filtes out predictions with no departure" do
```
anthonyshull, line 61: This typo is still here.

lib/predictions/repo.ex
```diff
@@ -24,14 +24,15 @@ defmodule Predictions.Repo do
...
-    |> filter_by_min_time(Keyword.get(opts, :min_time))
+    |> filter_predictions(Keyword.get(opts, :min_time))
     |> load_from_other_repos
   end
 
   def all_no_cache(opts) when is_list(opts) and opts != [] do
     opts
     |> add_all_optional_params()
     |> fetch()
+    |> filter_predictions(nil)
```
anthonyshull, line 35: Using a default argument of nil rather than passing in nil makes both the function definition and it's application much clearer and cleaner.

anthonyshull (changes requested):

anthonyshull:
If we're going to have this test in there then we should remove the `external` moduletag and make sure the test runs. You can add an `external` tag to individual tests.

anthonyshull (approved):
🥂

PR #2034, opened 2024-05-06
feat(TimetableController): get stops from canonical route patterns

thecristen:
#### Summary of changes

**Asana Ticket:** [Spike: A forever Timetable fix](https://app.asana.com/0/555089885850811/1207150210540768/f) and [MatchError: no match of right hand side value: []](https://app.asana.com/0/555089885850811/1207207607989339/f)

- removes the unused `AllStops` plug
- fixes dialyzer (we were including a flag that'd been removed in Elixir 1.15)
- refactors fetching stops from the timetable

Instead of using `Stops.Repo.by_route/3` with the date specified to today, we get the stops via `RoutePatterns.Repo.by_route_id/2` with `canonical: true` set. Stops are retrieved from route patterns by using each route pattern's `representative_trip_id` to get all the stops for those trips: stops are in the correct order from this V3 API call. 

The resulting canonical stops list is reduced with the stops from the actual schedules being displayed on the timetable to create the final list. This can result in new stops added or canonical stops being removed from what's ultimately shown on the page.

Usually new stops are shuttle stops, and usually the shuttle stops can be matched with the corresponding canonical stations they serve. There are a few edge cases in which this isn't the case. In those cases we have to hardcode those stops' placement among the finalized list, as there's no programmatic way to determine where that shuttle stop is along the route otherwise: the schedule's `stop_sequence` might not be enough information depending on the schedule trip's stops, the shuttle stop itself doesn't have a zone... luckily these situations should be few and far between.

---

#### General checks


lib/dotcom/trip_plan/related_link.ex
```diff
@@ -14,7 +14,10 @@ defmodule Dotcom.TripPlan.RelatedLink do
             icon_name: nil
 
   @stops_repo Application.compile_env!(:dotcom, :repo_modules)[:stops]
-  @default_opts [route_by_id: &Routes.Repo.get/1, stop_by_id: {@stops_repo, :get_parent, 1}]
+  @default_opts [
+    route_by_id: &Routes.Repo.get/1,
+    stop_by_id: Function.capture(@stops_repo, :get_parent, 1)
```
kotva006, line 19: All of these default ops can be removed

lib/dotcom_web/views/schedule/timetable.ex
```diff
@@ -3,6 +3,7 @@ defmodule DotcomWeb.ScheduleView.Timetable do
   alias DotcomWeb.ViewHelpers, as: Helpers
   alias DotcomWeb.PartialView.SvgIconWithCircle
   alias Stops.Stop
+  alias Routes.Route
```
kotva006, line 6: Is this used?

test/dotcom_web/controllers/schedule/timetable_controller_test.exs
```diff
@@ -189,7 +287,7 @@ defmodule DotcomWeb.ScheduleController.TimetableControllerTest do
       assert platform_stop ==
                track_change_for_schedule(
                  schedule_1,
-                 MapSet.new(["stop-1-other-patform"]),
+                 ["stop-1-other-patform"],
```
kotva006, line 290: This should be random data

test/dotcom_web/views/schedule/timetable_view_test.exs
```diff
@@ -64,8 +64,8 @@ defmodule DotcomWeb.Schedule.TimetableViewTest do
       origin = destination = nil
       alerts = []
 
-      all_stops = [
-        %Stops.Stop{id: "stop", name: "Stop"}
+      header_stops = [
+        {%Stops.Stop{id: "stop", name: "Stop"}, 0}
```
kotva006, line 68: This could use the stop factory

lib/dotcom_web/views/schedule/timetable.ex
thecristen: uh yeah, by the types in the last function

test/dotcom_web/views/schedule/timetable_view_test.exs
thecristen: Not yet it can't (there actually isn't a factory for creating %Stop{} yet, just the JsonApi Item structured as a stop)! :P

lib/dotcom/trip_plan/related_link.ex
thecristen: I removed the stop one, but the routes repo isn't clean up yet, so I'd like to leave that in here and change the rest of this when we get to that point.

kotva006 (approved):

PR #2035, opened 2024-05-06
fix(force_ssl): Configure force_ssl in prod instead of runtime

kotva006:
#### Summary of changes
Move the `force_ssl` configuration to the `prod.exs` configuration file as it is not supported to be configured in the [`runtime.exs` file](https://hexdocs.pm/phoenix/using_ssl.html#force-ssl)

https://app.asana.com/0/555089885850811/1206388063560698

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -2,3 +2,13 @@ import Config
 
 config :dotcom, :cache, Dotcom.Cache.Multilevel
 config :dotcom, :trip_plan_feedback_cache, Dotcom.Cache.TripPlanFeedback.Cache
+
+unless System.get_env("PORT") do
+  # configured separately so that we can have the health check not require
+  # SSL
+  config :dotcom, :secure_pipeline,
+    force_ssl: [
+      host: nil,
+      rewrite_on: [:x_forwarded_proto]
+    ]
+end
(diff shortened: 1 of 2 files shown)


config/prod.exs
```diff
@@ -2,3 +2,13 @@ import Config
 
 config :dotcom, :cache, Dotcom.Cache.Multilevel
 config :dotcom, :trip_plan_feedback_cache, Dotcom.Cache.TripPlanFeedback.Cache
+
+unless System.get_env("PORT") do
+  # configured separately so that we can have the health check not require
```
kotva006, line 7: question: Do the health checks run as prod?

thecristen: This was a holdover from many moons ago added here to help Backstop testing that we don't do anymore. https://github.com/mbta/dotcom-legacy/commit/0d8b53a42016da6b5cec3506f79c5fcbb170b095 We can probably revert that to how it was before

anthonyshull:
There is also config for this in `config/deps/phoenix.exs` that sets this for the test environment. We should move that to the `config/test.exs` file so they both follow the same pattern.

thecristen (approved):

PR #2036, opened 2024-05-07
test(Predictions): Added the mox for predictions

kotva006:
No ticket.  Adding some mox for the predictions, and updating tests that use the repo functions


config/test.exs
```diff
@@ -6,7 +6,10 @@ config :dotcom, :httpoison, HTTPoison.Mock
 
 config :dotcom, :cms_api_module, CMS.Api.Static
 config :dotcom, :mbta_api_module, MBTA.Api.Mock
-config :dotcom, :repo_modules, route_patterns: RoutePatterns.Repo.Mock
+
+config :dotcom, :repo_modules,
+  route_patterns: RoutePatterns.Repo.Mock,
+  predictions: Predictions.Repo.Mock
```
anthonyshull, line 12: Should alphabetize this list and the one above because they are going to be quite long.

lib/dotcom/realtime_schedule.ex
```diff
@@ -25,10 +24,12 @@ defmodule Dotcom.RealtimeSchedule do
 
   @predicted_schedules_per_stop 2
 
+  @predictions_repo Application.compile_env!(:dotcom, :repo_modules)[:predictions]
+
   @default_opts [
     stops_fn: &StopsRepo.get/1,
     routes_fn: &RoutesRepo.by_stop_with_route_pattern/1,
-    predictions_fn: &PredictionsRepo.all_no_cache/1,
+    predictions_fn: Function.capture(@predictions_repo, :all_no_cache, 1),
```
anthonyshull, line 32: We should just remove the dependency injection. It's pretty simple to just replace all instances of `predictions_fn.(params)` with a call to `@predictions_repo.all_no_cache(params)`.

test/dotcom_web/controllers/schedule/predictions_test.exs
```diff
@@ -25,50 +16,63 @@ defmodule DotcomWeb.ScheduleController.PredictionsTest do
...
-        stop: %Stops.Stop{id: "origin"},
-        trip: 1234,
-        departing?: false
-      }
+      expect(Predictions.Repo.Mock, :all, fn _ ->
+        [
+          %Predictions.Prediction{
+            time: ~N[2017-01-01T00:00:00],
+            stop: %Stops.Stop{id: "origin"},
+            trip: 1234,
```
anthonyshull, line 63: Should use Faker for all of these unless they mean something.

test/dotcom_web/controllers/schedule/predictions_test.exs
```diff
@@ -2,20 +2,11 @@ defmodule DotcomWeb.ScheduleController.PredictionsTest do
   use DotcomWeb.ConnCase, async: true
 
   import DotcomWeb.ScheduleController.Predictions
+  import Mox
```
anthonyshull, line 5: I like to make sure there are empty lines between the different kinds of includes.

anthonyshull (changes requested):
Looks a lot cleaner, but there are still a lot of magic strings.

test/support/mocks.ex
```diff
@@ -13,3 +13,4 @@ Mox.defmock(CMS.Api.Mock, for: CMS.Api.Behaviour)
 Mox.defmock(MBTA.Api.Mock, for: MBTA.Api.Behaviour)
 Mox.defmock(RoutePatterns.Repo.Mock, for: RoutePatterns.Repo.Behaviour)
 Mox.defmock(OpenTripPlannerClient.Mock, for: OpenTripPlannerClient.Behaviour)
+Mox.defmock(Predictions.Repo.Mock, for: Predictions.Repo.Behaviour)
```
anthonyshull, line 16: Should probably separate out the `Repo.Mock`s and make sure they're alphabetized. I would label them too.

test/dotcom_web/controllers/schedule/trip_info_test.exs
```diff
@@ -578,7 +589,8 @@ defmodule DotcomWeb.ScheduleController.TripInfoTest do
 
   describe "test that wollaston station is properly inserted when expected" do
     test "Does not add Wollaston to non Red line routes", %{conn: conn} do
-      init = init(trip_fn: &trip_fn/2, vehicle_fn: &vehicle_fn/1, prediction_fn: &prediction_fn/1)
+      expect(Predictions.Repo.Mock, :all, &prediction_fn/1)
```
anthonyshull, line 592: Why are we still referencing `prediction_fn`? Should be able to remove that.

test/dotcom_web/controllers/schedule/predictions_test.exs
anthonyshull: Would also be super simple to turn this into a factory.

test/dotcom_web/controllers/schedule/predictions_test.exs
```diff
@@ -25,50 +16,63 @@ defmodule DotcomWeb.ScheduleController.PredictionsTest do
...
+      end)
 
       conn =
         conn
         |> assign(:origin, %Stops.Stop{id: "origin"})
         |> assign(:destination, nil)
         |> assign(:route, %{id: "4"})
         |> assign(:direction_id, "0")
-        |> call(predictions_fn: fn _ -> [prediction] end)
+        |> call(init())
```
anthonyshull, line 75: `init()` will just return an empty list. And `call/1` will default to an empty list. In fact `call/2` just ignores the options. So, I'm pretty sure you can just remove it.

test/dotcom_web/controllers/schedule/trip_info_test.exs
kotva006: `prediction_fn/1` is a helper defined in this file. It seemed fine to keep it used here instead of redefining it for every test

anthonyshull: We shouldn't need it. It looks like it is being used as a kind of factory.

test/dotcom/transit_near_me_test.exs
```diff
@@ -689,19 +696,19 @@ defmodule Dotcom.TransitNearMeTest do
...
-      track: "2",
-      trip: @trip2
-    }
+    @prediction2 build(:prediction, %{
+                   departing?: true,
+                   direction_id: @direction_id,
+                   id: Faker.Internet.slug(),
+                   route: @route,
+                   schedule_relationship: nil,
+                   status: nil,
```
anthonyshull, line 705: See the below comment for the factory.

test/support/factory/prediction.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Test.Support.Factory.Prediction do
+  @moduledoc """
+  Generated fake data for %RoutePattern{}
+  """
+  use ExMachina
+
+  alias Predictions.Prediction
+
+  def prediction_factory do
+    %Prediction{}
```
anthonyshull, line 10: In the above tests, you are building a prediction and passing in every single value. I'm fairly certain that not setting anything here will set them all to struct default which is probably nil. So, passing in values set to nil won't do anything. You should set reasonable defaults here and then you only have to override ones that are necessary for the tests. Makes the tests shorter and clearer.

anthonyshull (changes requested):

anthonyshull (approved):

PR #2039, opened 2024-05-08
Adds telemetry metrics to dotcom

anthonyshull:
https://app.asana.com/0/555089885850811/1206719459037550/f

These are all of the metrics we now have access to.
Screenshot Capture - 2024-05-08 - 12-36-27


lib/telemetry/helper.ex
```diff
@@ -0,0 +1,61 @@
...
+  Stop the tracer and therefore stop listening for telemetry events.
+  """
+  def stop do
+    :dbg.stop()
+  end
+
+  defp handler(name, metadata, measure, filter) do
+    unless filtered?(name, filter) do
+      # credo:disable-for-next-line Credo.Check.Warning.IoInspect
+      IO.inspect(%{name: name, metadata: metadata, measure: measure})
```
kotva006, line 50: Why is this an `IO.inspect`?

lib/telemetry/helper.ex
```diff
@@ -0,0 +1,61 @@
+defmodule Telemetry.Helper do
+  @moduledoc """
+  Helper functions for seeing all telemetry events.
+
+  Only for use in development.
+  """
+
+  require Logger
```
kotva006, line 8: Logger is required here, but I don't see it being used.  Should it be used over IO.inspect

mix.lock
```diff
@@ -100,6 +100,7 @@
...
+  "telemetry_metrics_splunk": {:hex, :telemetry_metrics_splunk, "0.0.2-alpha", "f0a25d582b4315b5461840a55c77f2b0ad625c8afceaec56ce0b198f6c87f741", [:mix], [{:jason, "1.4.1", [hex: :jason, repo: "hexpm", optional: false]}, {:recase, "0.7.0", [hex: :recase, repo: "hexpm", optional: false]}, {:telemetry, "1.2.1", [hex: :telemetry, repo: "hexpm", optional: false]}, {:telemetry_metrics, "1.0.0", [hex: :telemetry_metrics, repo: "hexpm", optional: false]}], "hexpm", "c6ea80ee987a19df3c5d3bb87a5b94c3acf4a8d4df778e0f3ede2987426eba8f"},
   "telemetry_metrics_statsd": {:hex, :telemetry_metrics_statsd, "0.7.0", "92732fae63db31ef2508df6faee7d81401883e33f2976715a82f296a33a45cee", [:mix], [{:nimble_options, "~> 0.4 or ~> 1.0", [hex: :nimble_options, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}, {:telemetry_metrics, "~> 0.6", [hex: :telemetry_metrics, repo: "hexpm", optional: false]}], "hexpm", "797e34a856376dfd4e96347da0f747fcff4e0cadf6e6f0f989598f563cad05ff"},
```
kotva006, line 104: I see this got removed from our `mix.exs` file but is still present here.  It should be removed from the lock file as well

kotva006 (changes requested):
A couple questions

lib/telemetry/helper.ex
anthonyshull: This is made to be used in `iex` and `IO.inspect` is cleaner to read than a log. Also, with a log you have to choose a log level. So, if we chose `Logger.info` and the app is set to `:notice` then you wouldn't see the log, etc. etc.

mix.lock
anthonyshull: It was `telemetry_metrics_statsd` that got removed and `telemetry_metrics_splunk` got added. Looks like our lock file was very behind. So there are some other changes in there now.

PR #2054, opened 2024-05-14
Make sure trip isnt nil before accessing it

anthonyshull:
https://app.asana.com/0/555089885850811/1207286739672613/f


kotva006 (approved):
Looks like this will also fix the Asana bug

PR #2055, opened 2024-05-14
fix(Credo): Remove negated check in if-else

kotva006:
https://app.asana.com/0/555089885850811/1207286684779740


thecristen (changes requested):
This is confusing. I think [the guidance](https://hexdocs.pm/credo/Credo.Check.Refactor.NegatedConditionsWithElse.html) would suggest handling this the other way around:

```elixir
if Enum.member?(days_to_avoid, check_date) do
   (lots of stuff)
else
   check_date
end
```

thecristen (approved):

PR #2061, opened 2024-05-16
hotfix(TimetableController): schedules can add new stops

thecristen:
No ticket, I noticed this weekend's shuttle on the Newburyport/Rockport line was missing the stops from the local bus shuttle. And realized I had no mechanism for allowing the day's scheduled trips to add new stops to the timetable stops. So this PR does that (along with adding the positions for the two new stops, at Wonderland and Lynn).


thecristen:
**Current prod** - shows all the shuttle stops, they're in the wrong order though
image

**Current staging** (with the incoming timetable stop ordering fix) - stops are in the right order, but it doesn't include the new stops being served by those shuttles
image

**This PR** (using dev API) - includes newly served shuttle stops _and_ in the correct order:
image

kotva006 (approved):

PR #2063, opened 2024-05-16
feat(Search): Remove search bar from schedules and mode

kotva006:
Removes some low usage search bars and cleans up unused search code

https://app.asana.com/0/home/1201123880594717/1203059756247787


anthonyshull (changes requested):
Looks like the smoke tests and critical user journeys are failing. You'll probably have to edit them to accommodate the changes to the search bar.

PR #2069, opened 2024-05-21
refactor(Schedules): Consolidate schedules react trees into one tree

kotva006:
Consolidate React Trees on the schedule page.  Cleans up some old code as well.

https://app.asana.com/0/385363666817452/1204375294893168


assets/ts/schedule/schedule-loader.tsx
```diff
@@ -59,101 +29,156 @@ const updateURL = (origin: SelectedOrigin, direction?: DirectionId): void => {
...
-          />
-        </Provider>,
-        scheduleFinderRoot
-      );
-    }
-  }
 };
 
-const renderDirectionAndMap = (
+export const getDirectionAndMap = (
```
thecristen, line 70: as far as I can tell this function isn't called outside this file, so no need to export

thecristen (commented):
✨ It's looking real promising, works well! I'm a little surprised you decided to keep `<ScheduleLoader />` around and intact -- is it even a useful abstraction anymore when we could also call the underlying components directly?

assets/ts/schedule/components/SchedulePage.tsx
```diff
@@ -0,0 +1,446 @@
...
+    pdfs,
+    hours,
+    route,
+    schedule_note: scheduleNote,
+    route_patterns: routePatternsByDirection,
+    today,
+    services,
+    stops
+  } = schedulePageData;
+  const { modalOpen, modalMode, selectedOrigin } = state;
```
thecristen, line 225: **question**: Would it be better to do `const state = useSelector((state: StoreProps) => state);` here instead of passing it in as a prop? Also applies to the `<ScheduleFinder />` component itself.

assets/ts/schedule/components/schedule-finder/upcoming-departures/UpcomingDepartures.tsx
```diff
@@ -298,8 +341,19 @@ export const upcomingDeparturesTable = (
 };
 
 export const UpcomingDepartures = ({
-  state
+  routeId,
+  selectedOrigin,
+  selectedDirection,
+  today
 }: Props): ReactElement<HTMLElement> | null => {
+  const [state, updateData] = useProvider(fetchData, [
```
thecristen, line 349: nice move!

assets/ts/schedule/components/line-diagram/LineDiagram.tsx
```diff
@@ -58,23 +57,40 @@ const LineDiagram = ({
...
+      type: "CHANGE_ORIGIN",
+      newStoreValues: {
+        selectedOrigin: origin
+      }
+    });
+    // reopen modal depending on choice:
+    dispatch({
+      type: "OPEN_MODAL",
+      newStoreValues: {
+        modalMode: origin ? "schedule" : "origin"
```
thecristen, line 81: **suggestion**: I _think_ if you added a test, or adjusted a current test, to use the "origin" case here, that'll bump up your coverage to 95%. so close!

assets/ts/schedule/components/SchedulePage.tsx
```diff
@@ -0,0 +1,446 @@
...
+    let newDirection: DirectionId | undefined;
+    const newOrigin: SelectedOrigin | undefined =
+      query["schedule_finder[origin]"];
+
+    // modify the store values in case URL has parameters:
+    const queryDirectionId =
+      query["schedule_finder[direction_id]"] ||
+      query["schedule_direction[direction_id]"];
+    if (queryDirectionId !== undefined) {
+      newDirection = queryDirectionId === "0" ? 0 : 1;
```
thecristen, line 350: **suggestion**: I believe `queryDirectionId.toString()` will work in this case? will also let us have one fewer code branch to test. :)

thecristen (changes requested):
👍🏼👍🏼👍🏼

just a few mostly minor comments, and I think we should update some logic to not evaluate `false` when the selected direction is `0`.

assets/ts/schedule/store/ScheduleStore.tsx
```diff
@@ -82,14 +82,6 @@ export const createScheduleStore = (directionId: DirectionId): Store =>
 // initial 'dummy' value (will be overwritten when schedule-loader loads)
 export const store: Store = createScheduleStore(0);
 
-export const getCurrentState = (): StoreProps => store.getState();
-
 export const storeHandler = (newValues: Action): void => {
```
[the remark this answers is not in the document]
kotva006: There is a [skipped test](https://github.com/mbta/dotcom/pull/2069/files#diff-c15f27f0168f7723d579f8fe8593f808f3084c6c970f28ffd64861df22a52ca2R632) that I wasn't able to rewrite that still imports these functions.
That is the only place that still uses these functions

assets/ts/schedule/components/line-diagram/LineDiagram.tsx
kotva006: I am wondering why it is under 95% as it was meeting the coverage requirement when I was running it locally

assets/ts/schedule/components/SchedulePage.tsx
kotva006: newDirectionId is only allowed to be the number 0 or 1 so this stops users from entering 76, and also converts the string to s number

assets/ts/schedule/components/line-diagram/LineDiagram.tsx
kotva006: Screenshot 2024-06-06 at 9 16 57 AM
I changed zero tests, and it isn't complaining about coverage locally

assets/ts/schedule/store/ScheduleStore.tsx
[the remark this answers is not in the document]
thecristen: The link doesn't go to the specific spot and I can't find the skipped test you mention. But if it's as you describe it seems like we can remove the test (and these function)!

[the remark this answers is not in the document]
kotva006: The test has been removed

thecristen (approved):
🎉

PR #2072, opened 2024-05-21
refactor(TransitNearMeControllerTest): update mocks

thecristen:
Quick followup to https://github.com/mbta/dotcom/commit/aa9f9915f4ecaf4e5f7e220a54dc8fc9ce7b5ddd

- Removes the dependency injection from `TransitNearMeController`
- Refactors test to reflect changes

The rest of this test suite could use some more work... but I have reviews to do. ;)

### DIFF
--- a/test/dotcom_web/controllers/transit_near_me_controller_test.exs
+++ b/test/dotcom_web/controllers/transit_near_me_controller_test.exs
@@ -1,12 +1,13 @@
 defmodule DotcomWeb.TransitNearMeControllerTest do
-  use DotcomWeb.ConnCase
-  @moduletag :external
+  use DotcomWeb.ConnCase, async: true
 
   alias LocationService.Address
   alias Leaflet.{MapData, MapData.Marker}
-  alias Dotcom.TransitNearMe
-  alias DotcomWeb.TransitNearMeController, as: TNMController
+  alias DotcomWeb.TransitNearMeController
   alias Stops.Stop
+
+  import Mox
+  import Test.Support.Factory.MbtaApi
 
   @orange_line %{
     id: "Orange",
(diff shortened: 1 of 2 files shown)


test/dotcom_web/controllers/transit_near_me_controller_test.exs
```diff
@@ -161,42 +99,77 @@ defmodule DotcomWeb.TransitNearMeControllerTest do
 
   describe "with valid location params" do
     test "assigns stops with routes", %{conn: conn} do
-      params = %{"address" => %{"latitude" => "valid", "longitude" => "valid"}}
+      stops = build_list(3, :stop_item)
 
-      conn = get(conn, transit_near_me_path(conn, :index, params))
+      expect(LocationService.Mock, :reverse_geocode, fn latitude, longitude ->
```
anthonyshull, line 104: I *think* you can just `@latitude` and `@longitude` here. If not, you *might* be able to pin them.

test/dotcom_web/controllers/transit_near_me_controller_test.exs
```diff
@@ -161,42 +99,77 @@ defmodule DotcomWeb.TransitNearMeControllerTest do
...
+        assert latitude == @latitude
+        assert longitude == @longitude
+        {:ok, [@located_address]}
+      end)
 
-      assert conn.status == 200
+      expect(MBTA.Api.Mock, :get_json, 3, fn "/stops/", opts ->
```
anthonyshull, line 110: Is there a reason we're mocking the `MBTA.Api` (which is higher up) than the `Stops.Repo` here?

anthonyshull (changes requested):

test/dotcom_web/controllers/transit_near_me_controller_test.exs
thecristen: ....huh, that _is_ weird.

thecristen: It's populated via `TransitNearMe.build/2`, which calls `Stops.Nearby.nearby_with_varying_radius_by_mode/1`, which calls `MBTA.Api.Stops.all()` directly. Bananas!

anthonyshull: Might be making TNM slower than it needs to be.

anthonyshull (approved):

PR #2074, opened 2024-05-22
actually use the apple-touch-icon file

thecristen:
Just trying to see something....


thecristen:
Yay, this works! 

If I use Safari to add the homepage as a bookmark, I can see the dev-green one (left) has the desired image, as opposed to the current image (right, from bookmarking dev)

image

kotva006 (approved):
Good catch :P

PR #2076, opened 2024-05-24
feat(TimetableController): advanced override placement

thecristen:
Needed to accommodate next weekend's Needham Line's rail replacement bus which has a stop at Newton Highlands, at the very start of the route. The current shuttle stop placement logic was inadequate for this, as it assumed an existing stop _before_ the inserted stop. This PR enhances the logic by allowing specification of whether the shuttle stop is before or after a reference stop.

Made a minor correction for Anderson/Woburn while here. Other checks seem to come out ok.

### DIFF
--- a/lib/dotcom_web/controllers/schedule/timetable_controller.ex
+++ b/lib/dotcom_web/controllers/schedule/timetable_controller.ex
@@ -358,15 +358,23 @@ defmodule DotcomWeb.ScheduleController.TimetableController do
 
   # Some shuttle stops must be placed manually within the existing stops. Each
   # key is a scheduled stop ID from a shuttle, and the corresponding value is
-  # the name of the canonical rail stop it should be placed adjacent to.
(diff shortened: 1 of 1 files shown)


lib/dotcom_web/controllers/schedule/timetable_controller.ex
```diff
@@ -477,15 +485,27 @@ defmodule DotcomWeb.ScheduleController.TimetableController do
...
+          case before_or_after do
+            :after -> index + 1
+            :before -> index
+          end
+        else
+          case before_or_after do
+            :after -> index
+            :before -> index + 1
+          end
+        end
```
anthonyshull, line 503: This is going to get flagged as too much nesting.

anthonyshull (changes requested):

anthonyshull (approved):

PR #2081, opened 2024-05-30
Setup OTP Evaluation

anthonyshull:
This [livebook](https://livebook.dev) allows you to add a new scenario to a yaml file with a name, origin, and destination. Evaluating the livebook will emit a yaml file with itineraries where each step is spelled out in an easy-to-read way. This will allow us to add or alter test cases and evaluate them easily.

An example of the input for two trips:
```yaml
- name: Parallel subway & commuter rail
  from: 186 Canal St, Boston, MA 02114
  to: Commercial St &, Pleasant St, Malden, MA 02148
- name: Walking-only itineraries
  from: 700 Atlantic Ave, Boston, MA 02110
  to: 630 Washington St, Boston, MA 02111
```

An example of the output for those two trips:

```yaml
- from: 186 Canal St, Boston, MA 02114
  itineraries:
    - - WALK FROM 186 Canal St, Boston, MA, 02114, USA TO North Station
      - TAKE Orange Line SUBWAY FROM North Station TO Malden Center
      - WALK FROM Malden Center TO Commercial St & Pleasant St, Malden, MA, 02148, USA
    - - WALK FROM 186 Canal St, Boston, MA, 02114, USA TO North Station
      - TAKE Haverhill Line RAIL FROM North Station TO Malden Center
      - WALK FROM Malden Center TO Commercial St & Pleasant St, Malden, MA, 02148, USA
  name: Parallel subway & commuter rail
  to: Commercial St &, Pleasant St, Malden, MA 02148
- from: 700 Atlantic Ave, Boston, MA 02110
  itineraries:
    - WALKING_BETTER_THAN_TRANSIT
  name: Walking-only itineraries
  to: 630 Washington St, Boston, MA 02111
```


thecristen (commented):
I think the structure's great, but I wonder how we'd go about making _assertions_ based on this setup. Maybe we can use this example from Notion as an assertion test case?

Origin: 2401 Mass Ave, Cambridge, MA 02140
Destination: Park Street station
- Assert Red Line from Davis to Park Street appears as an option, and is tagged as most direct.
- Assert 77 bus from stop 2270 or 2296 to Red Line appears as an option, and is tagged as least walking.

(notably this also requires applying and saving the itinerary tags here. we can probably get away with running the same tags for every request).

From there it looks like we'd want to aggregate the pass/fails too.

anthonyshull:
> I think the structure's great, but I wonder how we'd go about making _assertions_ based on this setup. Maybe we can use this example from Notion as an assertion test case?
> 
> Origin: 2401 Mass Ave, Cambridge, MA 02140 Destination: Park Street station
> 
> * Assert Red Line from Davis to Park Street appears as an option, and is tagged as most direct.
> * Assert 77 bus from stop 2270 or 2296 to Red Line appears as an option, and is tagged as least walking.
> 
> (notably this also requires applying and saving the itinerary tags here. we can probably get away with running the same tags for every request).
> 
> From there it looks like we'd want to aggregate the pass/fails too.

Looks like the Notion doc had some details added after I already started/finished this work. We can add labels; afaict, that is the only thing needed to check any of the cases in the doc. These aren't really tests. Since it's just for evaluation, I made it flexible enough that it can be easily changed over time.

anthonyshull:
I added a `dates.yml` file and the plans will be run for any datetimes in that file.

livebooks/evaluate-open-trip-planner.livemd
```diff
@@ -6,13 +6,15 @@
...
+  def attach_itineraries_to_plan(%{"from" => from, "to" => to} = plan, datetime) do
     itineraries =
-      plan(from, to)
+      plan(from, to, depart_at: datetime)
       |> Kernel.elem(1)
       |> itineraries_to_strings()
 
-    Map.put(plan, "itineraries", itineraries)
+    plan
+    |> Map.put("datetime", Timex.format!(datetime, "%FT%H:%M", :strftime))
```
thecristen, line 16: thoughtful to format it even more readable here

thecristen (approved):

PR #2082, opened 2024-05-31
Use the ferry-map vanity url

anthonyshull:
I confirmed this works on Chrome, Vivaldi, and Safari. If someone can check on Firefox, I think we're good.


thecristen (approved):
Yeah, it works in Firefox. Still amazed it works as a `src` value!

PR #2083, opened 2024-05-31
fix(Credo): Reduce function nesting

kotva006:
https://app.asana.com/0/home/1201123880594717/1207286684779746

Refactor to reduce function nesting

### DIFF
--- a/lib/dotcom_web/controllers/schedule_controller.ex
+++ b/lib/dotcom_web/controllers/schedule_controller.ex
@@ -85,18 +85,18 @@ defmodule DotcomWeb.ScheduleController do
     redirect(conn, external: "https://capeflyer.com")
   end
 
-  defp time_fn(nil), do: "nil"
+  defp format_time(nil), do: "nil"
 
-  defp time_fn(time) do
+  defp format_time(time) do
     case Timex.format(time, "{h24}:{m}") do
       {:ok, time} -> time
       {:error, err} -> err
     end
   end
 
-  defp map_trip_to_time(%Schedule{time: t, trip: trip}) do
+  defp trip_and_time_tuple(%Schedule{time: time, trip: trip}) do
     trip_id = if is_nil(trip), do: "nil", else: Map.get(trip, :id)
-    time = time_fn(t)
+    time = format_time(time)
     {trip_id, time}
   end
 
@@ -116,13 +116,13 @@ defmodule DotcomWeb.ScheduleController do
       log_entries =
         Enum.map(
           out_schedules,
-          &map_trip_to_time/1
+          &trip_and_time_tuple/1
         )
         |> Enum.take(-5)
 
       _ =
         Logger.info(
-          "module=#{__MODULE__} fun=future_departures stop=#{params["stop_id"]} removed=#{inspect(log_entries)} time=#{time_fn(now)}"
+          "module=#{__MODULE__} fun=future_departures stop=#{params["stop_id"]} removed=#{inspect(log_entries)} time=#{format_time(now)}"
         )
     end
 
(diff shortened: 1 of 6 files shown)


lib/dotcom_web/controllers/trip_plan_controller.ex
```diff
@@ -573,6 +573,18 @@ defmodule DotcomWeb.TripPlanController do
     end)
   end
 
+  defp chunk_subway_legs({leg, _idx}) do
+    highest_fare =
+      leg
+      |> Fares.get_fare_by_type(:highest_one_way_fare)
```
anthonyshull, line 579: Not sure why it didn't, but Credo should be flagging the use of pipe here. We can just call the function since there is only one pipe.

lib/dotcom_web/views/page_content_view.ex
```diff
@@ -68,15 +68,19 @@ defmodule DotcomWeb.CMS.PageView do
     Enum.any?(paragraphs, &right_rail_check(&1))
   end
 
+  defp teasers?(paragraph) do
+    if Map.has_key?(paragraph, :teasers) do
+      if Enum.empty?(paragraph.teasers), do: false, else: true
+    else
+      true
+    end
+  end
```
anthonyshull, line 77: Can use pattern matching to remove that first level of if/else.

lib/mix/tasks/header_footer.ex
```diff
@@ -60,6 +60,18 @@ if Mix.env() in [:dev, :test] do
       make_zip()
     end
 
+    defp click_lang(dropdown, lang_code) do
+      if lang_code != "en" do
```
anthonyshull, line 64: Can use pattern matching to remove this if.

lib/mix/tasks/header_footer.ex
```diff
@@ -133,6 +135,21 @@ if Mix.env() in [:dev, :test] do
       {:ok, _files} = File.rm_rf("export")
     end
 
+    defp files_list_reducer(filename, path, base_path, acc) do
+      filename_path = Path.join(path, filename)
+
+      if File.dir?(filename_path) do
+        acc ++ create_files_list(File.ls!(filename_path), filename_path, base_path)
+      else
+        filenm =
```
anthonyshull, line 144: Variables aren't immutable. We can reuse the word `filename`.

lib/mix/tasks/header_footer.ex
```diff
@@ -133,6 +135,21 @@ if Mix.env() in [:dev, :test] do
       {:ok, _files} = File.rm_rf("export")
     end
 
+    defp files_list_reducer(filename, path, base_path, acc) do
```
anthonyshull, line 138: This function is doing a lot. It's hard to read. It can be broken out further to be more clear.

anthonyshull (changes requested):

lib/dotcom_web/controllers/schedule_controller.ex
```diff
@@ -85,6 +85,21 @@ defmodule DotcomWeb.ScheduleController do
...
   end
 
+  defp time_fn(nil), do: "nil"
+
+  defp time_fn(time) do
+    case Timex.format(time, "{h24}:{m}") do
+      {:ok, time} -> time
+      {:error, err} -> err
+    end
+  end
```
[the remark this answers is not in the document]
kotva006: Future helpful feedback could include possible names of the functions, or why the name is considered bad and what to avoid in the future

lib/dotcom_web/controllers/schedule_controller.ex
```diff
@@ -85,6 +85,21 @@ defmodule DotcomWeb.ScheduleController do
...
+
+  defp map_trip_to_time(%Schedule{time: t, trip: trip}) do
```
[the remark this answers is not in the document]
kotva006: Future helpful feedback could include possible names of the functions, or why the name is considered bad and what to avoid in the future

anthonyshull (approved):
👍

PR #2084, opened 2024-05-31
Show "No real-time data" when no predictions

anthonyshull:
https://app.asana.com/0/555089885850811/1207376931845116/f

The best way I found to test this with no predictions is to comment out the `PredictionsSupervisor` in `application.ex`.


thecristen (approved):
What a surgical change! Looks good in both spots.

PR #2086, opened 2024-06-03
feat(MBTA.Api): guard checking type and length

thecristen:
#### Summary of changes

No ticket, I noticed `Stops.Repo.get("")` returned a stop result and realized something needs fixing. For certain functions where we construct a URL, a value of `""` as our ID value risks producing an undesired result by, say, requesting `/stops/` instead of `/stops/123`. I wrote a guard function and applied it to the handful of places this would impact our usage of the V3 API.

---

#### General checks

### DIFF
--- a/test/dotcom_web/controllers/trip_plan_controller_test.exs
+++ b/test/dotcom_web/controllers/trip_plan_controller_test.exs
@@ -813,7 +813,7 @@ defmodule DotcomWeb.TripPlanControllerTest do
     end
 
     test "doesn't set custom_route? flag for regular routes", %{itineraries: itineraries} do
-      expect(MBTA.Api.Mock, :get_json, fn "/routes/" <> _, _ ->
+      stub(MBTA.Api.Mock, :get_json, fn "/routes/" <> _, _ ->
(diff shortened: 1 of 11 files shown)


test/dotcom_web/controllers/trip_plan_controller_test.exs
```diff
@@ -813,7 +813,7 @@ defmodule DotcomWeb.TripPlanControllerTest do
     end
 
     test "doesn't set custom_route? flag for regular routes", %{itineraries: itineraries} do
-      expect(MBTA.Api.Mock, :get_json, fn "/routes/" <> _, _ ->
+      stub(MBTA.Api.Mock, :get_json, fn "/routes/" <> _, _ ->
```
thecristen, line 816: changed because may or may not get called depending on the itinerary (needs improved factory to create better base itineraries)

anthonyshull (changes requested):
This is going to cause a crash every time we don't send in a valid id. We should have catchall functions that return some kind of failure API response.

anthonyshull (approved):

PR #2091, opened 2024-06-06
Change Trip Planner Form Inputs

anthonyshull:
https://app.asana.com/0/555089885850811/1207426907582376/f
https://app.asana.com/0/555089885850811/1207495504741834/f


lib/dotcom_web/templates/trip_plan/_options.html.eex
```diff
@@ -1,14 +1,17 @@
+<h6>When</h6>
+<%=
+  render "_departure.html", Map.merge(assigns, %{
+    title: format_plan_type_for_title(@query),
+    prefix: "trip-plan-departure",
+    datetime: datetime_from_query(@query)
+  })
+%>
+<h6>Modes</h6>
```
thecristen, line 9: Why use `<h6>` here? I'm not sure that headings make more sense here semantically than fieldset labels, but given the structure of this page they'd be `<h2>` if anything. [Heading levels shouldn't skip levels](https://dequeuniversity.com/rules/axe/4.8/heading-order).

lib/dotcom_web/templates/trip_plan/_options.html.eex
```diff
@@ -1,12 +1,12 @@
...
+<label>When</label>
 <%=
   render "_departure.html", Map.merge(assigns, %{
     title: format_plan_type_for_title(@query),
     prefix: "trip-plan-departure",
     datetime: datetime_from_query(@query)
   })
 %>
-<h6>Modes</h6>
+<label>Modes</label>
```
thecristen, line 9: **question**: Do `<label>` elements make sense here? This label has no association made with another element when used here like this.

anthonyshull: No elements really make sense there. But, the way things are rendered doesn't build a cohesive form. So, for now, unless we want to rewrite the whole thing, we have to choose some element to go there. I'm open to whatever.

thecristen (approved):

PR #2099, opened 2024-06-11
predictions working

anthonyshull:
https://app.asana.com/0/555089885850811/1207481893303699/f

### DIFF
--- a/assets/ts/helpers/departureInfo.tsx
+++ b/assets/ts/helpers/departureInfo.tsx
@@ -158,45 +158,34 @@ const departuresListFromInfos = (
   omitCancelledAndSkipped = false,
   WrapperEl: typeof DefaultWrapper = DefaultWrapper
 ): React.ReactElement[] => {
-  // optional cutoff time, before which we won't show schedules.
-  // just used with subway for now.
-  const predictionTimeCutoff = chain(departureInfos)
-    .filter(d => d.routeMode === SUBWAY)
-    .maxBy("prediction.time")
-    .value()?.prediction!.time;
-
-  const routeId = departureInfos[0]?.route?.id;
-  const tripId = departureInfos[0]?.trip?.id;
-
-  if (isSubway && !predictionTimeCutoff) {
-    return [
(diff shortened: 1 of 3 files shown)


assets/ts/helpers/departureInfo.tsx
```diff
@@ -158,45 +158,34 @@ const departuresListFromInfos = (
...
-    .value()?.prediction!.time;
+  const predictions = chain(departureInfos)
+    .reject(
+      departure =>
+        omitCancelledAndSkipped &&
+        (!!departure.isCancelled || !!departure.isSkipped)
+    )
+    .reject(
+      departure => isSubway && typeof departure.prediction === "undefined"
+    )
```
kotva006, line 169: `reject` should only be called once.  Also for readability the conditions could be put behind a function (ex. `isCanceledSkippedAndNoPrediction`)

kotva006 (changes requested):
One optimization and linting is failing and needs to be fixed

assets/ts/helpers/departureInfo.tsx
anthonyshull: I actually kept it as a separate call to reject because the other one is already loaded and this rejection has nothing to do with that rejection. And, creating a function wouldn't make it much easier to read since it passes in a variable from a closure like the one above.

kotva006: It's more efficient to call a single reject, and falls inline with our code standards in other places of the code:
https://app.asana.com/0/search?q=credo%20reject&sort=last_modified&child=1207115993391023

anthonyshull: I understand it is more efficient to only traverse the list once. It's more communicative to do it twice in this situation because the two really have nothing to do with each other and the first one is already overloaded. Furthermore, we don't enforce Credo Elixir checks on Typescript just like we don't do vice-versa. Learning how to be a good programmer is about nuance and understanding why you are or aren't doing something. But, I'll make the change as this code is going in the dustbin so I'm not very concerned about its longevity.

kotva006 (approved):
👍

PR #2100, opened 2024-06-12
update(Elixir/OTP): Updating elixir/otp to version 17/27 and depndencies

kotva006:
https://app.asana.com/0/555089885850811/1207269824078369


anthonyshull:
You need to also update the dev docker images. We use a very old version of html_sanitize_ex which requires mochiweb and it looks like mochiweb won't compile with the latest elixir. I would use this as an opportunity to remove html_santize_ex.

PR #2065, opened 2024-05-20
refactor: create `LocationService` behaviour

thecristen:
No ticket, just refactoring and a bunch of test rewriting.

Since `LocationService` uses `ExAws`, and `ExAws` comes with its own behaviour module, I went ahead and setup mocking for that as well, and used that to rewrite the `LocationService` tests.

PR #2033, opened 2024-05-06
refactor(Stops.Repo): create a behaviour

thecristen:
Similar idea to #2027 

- added `Stops.Repo.Behaviour` & the configuration needed to swap out `Stops.Repo` for `Stops.Repo.Mock`
- updated everything else accordingly

The `Dotcom.GreenLine.Supervisor` test was a bit harder to wrangle with and it's probably still terrible, but it's mocked better now and has some coverage!

Also figured out that the `stop_count` property on `TripInfo` wasn't used for anything, so I removed it so that I didn't need to adjust yet another test.

(extra update:)

- Removed unused part of `Stops.Repo.stop_features/2`
- Removed duplicated address field in `Stops.Api` call
- Removed unused `Stops.Repo.stop_exists_on_route?/3` function
- Updated most but not every test module which relied on `Stops.Repo`. I just got tired, and the remaining modules are very large with problematic tests that check for actual data :(

PR #2067, opened 2024-05-20
Show diversion diagrams in diversion alerts

anthonyshull:
https://app.asana.com/0/555089885850811/1207209404321117/f

PR #2108, opened 2024-06-24
feat(Routes.Repo): use a behaviour

thecristen:
This is similar in goal to  
- #2033 
- #2065

Just this time for routes.

Highlights:
- removes `Routes.MockRepoApi`, which is a nice way of hardcoding data but inconsistent with our current approach.
- adds nice factory for `%Routes.Route{}` structs


anthonyshull (changes requested):
In this PR, I changed the structure of factories so that they mirror the entities they create. It's unfortunate this PR has just lingered here for weeks on end. But, we should still do that part of the work so we have consistency.

https://github.com/mbta/dotcom/pull/2067/files

anthonyshull (approved):

PR #2111, opened 2024-06-24
feat(Routes.Route): add external_agency_name

thecristen:
This is going to replace `custom_route?`. There's additional handling in the trip planner controller that will be added, but this maintains the current state of affairs while populating `external_agency_name` with a meaningful name.


lib/dotcom_web/controllers/trip_plan_controller.ex
```diff
@@ -520,7 +520,8 @@ defmodule DotcomWeb.TripPlanController do
       type: type
     } = Enum.find(legs, &(Leg.route_id(&1) == {:ok, id}))
 
-    custom_route = %Route{
+    %Route{
+      external_agency_name: if(type == "2", do: "Massport"),
```
thecristen, line 524: Confusingly, `type` doesn't come from the route type, it comes from an OTP field with the agency ID. That parsing's getting overhauled, anyway.

lib/dotcom_web/controllers/trip_plan_controller.ex
```diff
@@ -529,25 +530,6 @@ defmodule DotcomWeb.TripPlanController do
...
-          | type: "Massport-" <> type
-        }
-
-      # Handle MBTA busses not present via api
-      {"1", "BUS"} ->
-        %Route{
-          custom_route
-          | type: 3
-        }
-
```
thecristen, line 547: This never made sense and isn't possible to hit now, so I'm just removing it.

anthonyshull (approved):

PR #2112, opened 2024-06-24
refactor(Routes.Route): remove deprecated custom_route? value

thecristen:
Followup to:
- #2111 

This still doesn't fix the Trip Planner's handling of Massport routes, that needs some more refactoring that will follow next.


kotva006 (approved):

kotva006:
There are some tests, and linting that needs to be cleaned up

PR #2119, opened 2024-07-03
chore(Fares): remove obsolete logic

thecristen:
While cleaning up code from my (ongoing) trip planner refactoring, I realized there was some fare-calculating logic involved in `Fares.fare_for_stops/4` that is no longer relevant

- The Foxboro pilot logic is from 2019, I'm pretty sure there's no pilot effort involved anymore...
- The combo zones that were introduced in 2020 (PR #601) are no longer relevant, as there are no more `1A-2` or such zones.

This made the `trip` argument no longer needed, so I got to remove it from way too many function calls.

In the meantime I also removed the `:external` tag from a bunch of related tests to stop skipping them, and added the necessary mocks to make those tests work.

### DIFF
--- a/test/fares/fares_test.exs
+++ b/test/fares/fares_test.exs
@@ -1,7 +1,12 @@
 defmodule FaresTest do
   use ExUnit.Case, async: true
   doctest Fares
-  alias Routes.Route
+
+  import Mox
+
+  alias Test.Support.Factories.{Routes.Route, Stops.Stop}
+
+  setup :verify_on_exit!
 
   describe "calculate_commuter_rail/2" do
     test "when the origin is zone 6, finds the zone 6 fares" do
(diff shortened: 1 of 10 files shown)


test/fares/fares_test.exs
```diff
@@ -1,7 +1,12 @@
 defmodule FaresTest do
   use ExUnit.Case, async: true
   doctest Fares
-  alias Routes.Route
+
+  import Mox
+
+  alias Test.Support.Factories.{Routes.Route, Stops.Stop}
```
bklebe, line 7: suggestion (non-blocking): I try not to nitpick style in general but I've come to believe in the utility of this Credo check: https://hexdocs.pm/credo/Credo.Check.Consistency.MultiAliasImportRequireUse.html because splitting up the aliases instead of grouping them by root (as here) makes it easier to grep for module usage in the codebase. Thoughts?

bklebe (approved):

PR #2120, opened 2024-07-05
When we have predictions show those over alerts

anthonyshull:
https://app.asana.com/0/555089885850811/1207562164283201/f

This will work in most cases, but there is one case where it will present incorrect information. If there are one or fewer predictions *and* there is an alert then it will say there is no service. I can see this happening at the end of service for the day. But, it is still better than it saying there is no service for the entire day.

Screenshot Capture - 2024-07-05 - 09-10-42

### DIFF
--- a/assets/ts/stop/components/DeparturesWithBadge.tsx
+++ b/assets/ts/stop/components/DeparturesWithBadge.tsx
@@ -48,6 +48,8 @@ const DeparturesWithBadge = ({
   if (!(priorityBadge || secondaryBadge)) return <>{children}</>;
   const displayBadge = (priorityBadge || secondaryBadge)!;
 
+  // if the priority badge is present and there is only one departure, show the badge
+  // this is because we check for predictions and return a single time list if there are none
   return (
     <>
       {priorityBadge && timeListLength === 1 ? (
--- a/assets/ts/stop/components/DeparturesWithBadge.tsx
+++ b/assets/ts/stop/components/DeparturesWithBadge.tsx
@@ -29,24 +29,15 @@ const toInformativeAlertBadge = (alerts: Alert[]): JSX.Element | undefined => {
   return undefined;
 };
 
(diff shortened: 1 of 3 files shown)


assets/ts/stop/components/DeparturesWithBadge.tsx
```diff
@@ -57,16 +48,23 @@ const DeparturesWithBadge = ({
   if (!(priorityBadge || secondaryBadge)) return <>{children}</>;
   const displayBadge = (priorityBadge || secondaryBadge)!;
 
+  // if the priority badge is present and there is only one departure, show the badge
+  // this is because we check for predictions and return a single time list if there are none
```
thecristen, line 52: A single time list is also always returned if there are commuter rail predictions (as it's designed to only display one time at time). 

So. I'm not sure that tying _this_ PR's logic to "time list happens to have length of 1" makes all that much sense.

assets/ts/stop/components/DeparturesWithBadge.tsx
```diff
@@ -57,16 +48,23 @@ const DeparturesWithBadge = ({
...
   return (
     <>
-      {priorityBadge ? (
-        <div className="font-helvetica-neue fs-14" style={{ float: "right" }}>
-          See alternatives
+      {priorityBadge && timeListLength === 1 ? (
```
thecristen, line 55: There's some logic involved in the generation of `priorityBadge` that attempts to weight the number of predictions available against the particular type of alerts – have you looked at `isSuppressiveAlert` to see whether modifying that would better fit our use case?

anthonyshull:
@thecristen can you spell out better logic to implement? There are a lot of assumptions and implicit things encoded in this code and, frankly, I'm not sure what fixes/breaks anything.

anthonyshull:
Also, since I'm picking up the ticket to add direction id to the alerts, I'm not sure if it is even worth merging this PR since it will just be reversed in the next few days.

PR #2125, opened 2024-07-10
Ags/add direction to station closure

anthonyshull:
https://app.asana.com/0/555089885850811/1207746584777124/f

Note, I have limited this to station closures at subways. It only works with one direction. I'm not sure if there are subway stops with two directions. But, the API only supports one direction on alert informed entities.

Screenshot Capture - 2024-07-09 - 12-15-54
Screenshot Capture - 2024-07-09 - 12-18-01
Screenshot Capture - 2024-07-09 - 12-18-12
Screenshot Capture - 2024-07-09 - 12-16-27
Screenshot Capture - 2024-07-09 - 12-17-04
Screenshot Capture - 2024-07-09 - 12-17-24

### DIFF
--- a/lib/alerts/alert.ex
+++ b/lib/alerts/alert.ex
@@ -149,10 +149,10 @@ defmodule Alerts.Alert do
 
   defp set_direction_ids(%__MODULE__{} = alert), do: alert
 
-  defp set_direction_id(%InformedEntity{} = entity) do
+  defp set_direction_id(%InformedEntity{direction_id: nil, route_type: 1} = entity) do
     stop = Stops.Repo.get(entity.stop)
 
-    if entity.direction_id == nil && stop.child_ids == [] do
+    if stop.child_ids == [] do
       direction_id =
         stop.id
         |> RoutePatterns.Repo.by_stop_id()
@@ -165,6 +165,8 @@ defmodule Alerts.Alert do
       entity
     end
   end
+
+  defp set_direction_id(entity), do: entity
 
   @spec set_priority(map) :: map
   defp set_priority(%__MODULE__{} = alert) do
--- a/lib/alerts/alert.ex
+++ b/lib/alerts/alert.ex
@@ -157,8 +157,8 @@ defmodule Alerts.Alert do
         stop.id
         |> RoutePatterns.Repo.by_stop_id()
         |> Enum.filter(&(&1.route_id == entity.route))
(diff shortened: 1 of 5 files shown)


lib/alerts/alert.ex
```diff
@@ -137,9 +137,37 @@ defmodule Alerts.Alert do
...
+
+    Map.put(alert, :informed_entity, informed_entities)
+  end
+
+  defp set_direction_ids(%__MODULE__{} = alert), do: alert
+
+  defp set_direction_id(%InformedEntity{direction_id: nil, route_type: 1} = entity) do
+    stop = Stops.Repo.get(entity.stop)
+
+    if stop.child_ids == [] do
```
thecristen, line 155: `stop.child?` will work here too, reads a bit more cleanly IMO

lib/alerts/alert.ex
```diff
@@ -137,9 +137,37 @@ defmodule Alerts.Alert do
...
+      direction_id =
+        stop.id
+        |> RoutePatterns.Repo.by_stop_id()
```
thecristen, line 158: let's get the repo module via `Application.compile_env!(:dotcom, :repo_modules)[:route_patterns]`

lib/alerts/alert.ex
```diff
@@ -137,9 +137,37 @@ defmodule Alerts.Alert do
...
+  defp set_direction_ids(%__MODULE__{effect: :station_closure} = alert) do
+    informed_entities = Enum.map(alert.informed_entity, &set_direction_id/1)
+
+    Map.put(alert, :informed_entity, informed_entities)
+  end
+
+  defp set_direction_ids(%__MODULE__{} = alert), do: alert
+
+  defp set_direction_id(%InformedEntity{direction_id: nil, route_type: 1} = entity) do
+    stop = Stops.Repo.get(entity.stop)
```
thecristen, line 153:
```suggestion
    stop = @stops_repo.get(entity.stop)
```

thecristen (changes requested):
let's use our configured repo modules here. you might need to fix some tests/add some mock expectations as a result.

lib/alerts/alert.ex
thecristen: might need to watch for if this returns `nil`, the following logic will break..

thecristen (approved):

PR #2127, opened 2024-07-11
Filter location autocomplete results by USA and appropriate states

anthonyshull:
Screenshot Capture - 2024-07-11 - 13-27-15

### DIFF
--- a/lib/location_service/location_service.ex
+++ b/lib/location_service/location_service.ex
@@ -13,14 +13,18 @@ defmodule LocationService do
   @cache Application.compile_env!(:dotcom, :cache)
   @ttl :timer.hours(24)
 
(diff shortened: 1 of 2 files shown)


lib/location_service/location_service.ex
```diff
@@ -13,14 +13,18 @@ defmodule LocationService do
   @cache Application.compile_env!(:dotcom, :cache)
   @ttl :timer.hours(24)
 
+  @filter ~r/,\s(MA|NH|RI),\s/
```
thecristen, line 16: I'm so glad you included the commas so that we don't filter out places/streets which happen to have these characters! :D

thecristen (approved):

PR #2128, opened 2024-07-12
Remove predictions that don't have a departure time

anthonyshull:
https://app.asana.com/0/555089885850811/1207779834092092/f

Note, I am rewriting tests, but we want to get this out before this weekend's Red Line issues.


thecristen (approved):
:D glad you're finding the factory helpers useful too

PR #2129, opened 2024-07-16
Rewrite predictions pubsub tests

anthonyshull:
No ticket, this is just the test rewrite to go along with *not* showing predictions with no departure time.

Please reference: https://www.notion.so/mbta-downtown-crossing/Mocking-External-Dependencies-625665802fb84c8fa6e6edfa030d273f


lib/predictions/pub_sub.ex
```diff
@@ -31,14 +34,13 @@ defmodule Predictions.PredictionsPubSub do
...
-  @spec subscribe(String.t(), GenServer.server()) :: [Prediction.t()] | {:error, term()}
-  def subscribe(topic, server \\ __MODULE__) do
+  @impl Behaviour
+  def subscribe(topic) do
     case StreamTopic.new(topic) do
       %StreamTopic{} = stream_topic ->
         :ok = StreamTopic.start_streams(stream_topic)
 
-        {registry_key, predictions} = GenServer.call(server, {:subscribe, stream_topic})
+        {registry_key, predictions} = GenServer.call(__MODULE__, {:subscribe, stream_topic})
```
bklebe, line 43: question: why not leave `__MODULE__` as a default argument to this function? Then we can inject different dependencies to test if we'd like.

lib/predictions/pub_sub.ex
```diff
@@ -54,9 +56,9 @@ defmodule Predictions.PredictionsPubSub do
   # Server
 
   @impl GenServer
-  def init(opts) do
-    subscribe_fn = Keyword.get(opts, :subscribe_fn, &Phoenix.PubSub.subscribe/2)
-    subscribe_fn.(Predictions.PubSub, "predictions")
+  def init(_) do
+    Phoenix.PubSub.subscribe(@predictions_phoenix_pub_sub, "predictions")
```
bklebe, line 60: question: why are we making this compile-time instead of a runtime option? It's conventional for callbacks like `init` to actually configure the process they're starting dynamically from the options being passed in.

lib/predictions/store.ex
```diff
@@ -5,55 +5,46 @@ defmodule Predictions.Store do
...
-    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @doc "Deletes predictions associated with the input fetch keys, e.g. clear([route: 'Red', direction: 1])"
+  @impl Behaviour
+  def clear(keys) do
+    GenServer.cast(__MODULE__, {:remove, Enum.map(fetch(keys), & &1.id)})
   end
```
bklebe, line 27: suggestion: I would also suggest making `__MODULE__` an argument to this function.

lib/predictions/store/behaviour.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Predictions.Store.Behaviour do
```
bklebe, line 1: question: could we come up with a different name for this behaviour? I find it confusing to read code with `@behaviour Behaviour` everywhere, it makes it seem like this is some kind of generic library code for behaviours but it seems like it's actually a Predictions Store Behaviour.

lib/predictions/supervisor.ex
```diff
@@ -8,19 +8,21 @@ defmodule Predictions.Supervisor do
...
+
   def start_link(_) do
     Supervisor.start_link(__MODULE__, [])
   end
 
   @impl Supervisor
   def init(_) do
     children = [
-      {Phoenix.PubSub, [name: Predictions.PubSub]},
+      {Phoenix.PubSub, [name: @predictions_phoenix_pub_sub]},
```
bklebe, line 20: same suggestion re: dynamic vs static configuration

lib/predictions/pub_sub.ex
anthonyshull: Because we're using Mox and determining the implementation at compile time.

lib/predictions/supervisor.ex
anthonyshull: We are moving away from DI and toward using Mox. That's 50% of the purpose of this PR.

lib/predictions/store/behaviour.ex
anthonyshull: I actually think it is pretty straightforward. `Predictions.Store` implements the `Predictions.Store.Behaviour`. Makes it so that any time I want to find the custom behaviour being implemented, I know right where to look.

lib/predictions/pub_sub.ex
anthonyshull: Because we are moving away from DI and towards using Mox. We don't need to inject variables in dynamically. That is, it will *never* be called with a second argument.

lib/predictions/store.ex
anthonyshull: Again, we aren't doing DI. We will never need to message a different GenServer.

anthonyshull:
The tests around starting streams seem flaky. I'm going to write a helper that does assertions on a retry basis so we don't have to try to sleep for the 'right' amount of time.

bklebe (approved):

PR #2131, opened 2024-07-21
Upgrade dependencies 2024 07

anthonyshull:
https://app.asana.com/0/555089885850811/1207649895037277/f

I chose to upgrade Timex this month. There were changes to the handling of daylight savings time.
https://github.com/bitwalker/timex/blob/main/CHANGELOG.md#fixed-8


lib/util/util.ex
```diff
@@ -188,10 +188,10 @@ defmodule Util do
 
   @spec handle_ambiguous_time(Timex.AmbiguousDateTime.t() | DateTime.t() | {:error, any}) ::
           DateTime.t() | {:error, any}
-  defp handle_ambiguous_time(%Timex.AmbiguousDateTime{before: before}) do
+  defp handle_ambiguous_time(%Timex.AmbiguousDateTime{after: aft}) do
```
anthonyshull, line 191: 📓 Timex suggests you use after unless you have a reason to use before. `after` is a reserved word hence the use of `aft`.

mix.exs
```diff
@@ -83,8 +83,8 @@ defmodule DotCom.Mixfile do
       {:ex_aws, "2.5.4"},
       {:ex_aws_s3, "2.5.3"},
       {:ex_aws_ses, "2.4.1"},
-      {:ex_doc, "0.34.0", only: :dev},
-      {:ex_machina, "2.7.0"},
+      {:ex_doc, "0.34.2", only: :dev},
+      {:ex_machina, "2.8.0", only: [:dev, :test], override: true},
```
anthonyshull, line 87: 📓 We don't need ex_machina in production.

test/dotcom_web/views/cms_view_test.exs
```diff
@@ -135,7 +135,7 @@ defmodule DotcomWeb.CMSViewTest do
         )
 
       # could also be November 6th, 1:00 AM (test daylight savings)
-      expected = "November 5, 2016 1 AM - November 6, 2016 2 AM"
+      expected = "November 5, 2016 1 AM - November 6, 2016 1 AM"
```
anthonyshull, line 138: 📓 Ambiguous daylight savings times were wrong in previous versions of Timex. We probably shouldn't test this in our code because it just tests the underlying library. But, they're already there so we can just correct them.

test/dotcom_web/views/schedule/timetable_view_test.exs
```diff
@@ -145,7 +145,11 @@ defmodule DotcomWeb.Schedule.TimetableViewTest do
       header_schedules = [%Schedules.Schedule{trip: trip}]
 
       trip_schedules = %{
-        {"Test-Trip-ID", "Test-Stop-ID"} => %Schedules.Schedule{trip: trip, stop: original_stop}
+        {"Test-Trip-ID", "Test-Stop-ID"} => %Schedules.Schedule{
+          stop: original_stop,
+          time: Timex.now(),
```
anthonyshull, line 150: 📓 Not sure how this was passing previously, but `time` is referenced in a template and so needs to be set.

test/feedback/mailer_test.exs
```diff
@@ -253,13 +253,13 @@ defmodule Feedback.MailerTest do
     end
 
     test "converts 'America/New_York' date to UTC accounting for beginning of daylight saving time" do
-      dt = Util.convert_using_timezone(~N[2020-03-08T02:00:00], "America/New_York")
-      assert Mailer.formatted_utc_timestamp(dt) == "03/08/2020 06:00"
+      dt = Util.convert_using_timezone(~N[2020-03-08T03:00:00], "America/New_York")
```
anthonyshull, line 256: 📓 These and all others below are just changes to the ambiguous daylight savings time changes in Timex.

bklebe (approved):

PR #2138, opened 2024-08-05
feat: add accessibility test with report

thecristen:
#### Summary of changes

**Asana Ticket:** [Dev | First pass at accessibility process](https://app.asana.com/0/555089885850811/1207726969014242/f)


thecristen:
Actually, now that all the smoke tests were moved from Cypress to Playwright, I might co-opt _that_ test for this purpose.

PR #2144, opened 2024-08-09
Fix route id access path and create fallback pattern match to prevent…

anthonyshull:
https://app.asana.com/0/555089885850811/1207993324656149/f

Logs when we can't pattern match. It also solves cases where we have data that doesn't fit the format because it will just default to a description of "unknown."


bklebe (approved):
Did you figure out why this broke in the first place? Did the itinerary format change without us noticing it or something?

anthonyshull:
> Did you figure out why this broke in the first place? Did the itinerary format change without us noticing it or something?

The underlying data changed from using a key `route_id` to two keys `route` => `id`.

PR #2145, opened 2024-08-12
This is a copy of the alerts/diversions task with fewer changes

anthonyshull:
Copying over the changes from https://github.com/mbta/dotcom/pull/2067 that are still relevant.

https://app.asana.com/0/555089885850811/1207209404321117/f

Visit https://dev-green.mbtace.com/diversions/alternative-travel-options-during-july-red-line-closure and you'll see an alert in the middle of the page. If you edit that alert to add an image URL, it will appear in the alert.


assets/js/alert-item.js
```diff
@@ -10,6 +10,10 @@ export const addAlertItemEventHandlers = () => {
     [...document.querySelectorAll(`.${ITEM_SELECTOR}`)].forEach(alertItem => {
       alertItem.addEventListener("click", handleAlertItemClick);
       alertItem.addEventListener("keydown", handleAlertItemKeyPress);
+
+      if (document.querySelector(".diversions-template") && alertItem.querySelector("img")) {
```
anthonyshull, line 14: 📓 On the diversions page, we want to open any alerts that have an image.

lib/alerts/informed_entity.ex
```diff
@@ -29,8 +29,6 @@ defmodule Alerts.InformedEntity do
           | :using_escalator
           | :using_wheelchair
 
-  alias __MODULE__, as: IE
```
anthonyshull, line 32: 📓 I just remove this `IE` shortcut wherever I find it. I don't know why it is so hard to type `InformedEntity` and I have seen `InformedEntitySet` also referred to as `IE` so it's better to disambiguate.

lib/alerts/informed_entity.ex
```diff
@@ -72,21 +70,24 @@ defmodule Alerts.InformedEntity do
...
 
   def mapsets_match?(%MapSet{} = a, %MapSet{} = b), do: has_intersect?(a, b)
 
   defp has_intersect?(a, b), do: Enum.any?(a, &(&1 in b))
 
   defp do_match?(f, s) do
-    @fields
+    __MODULE__.__struct__()
+    |> Map.keys()
+    |> List.delete(:__struct__)
```
anthonyshull, line 90: 📓 Instead of duplicating the list of fields and hard-coding them, I get them from the struct.

lib/alerts/parser.ex
```diff
@@ -10,17 +10,20 @@ defmodule Alerts.Parser do
...
-        effect: effect(attributes),
+        banner: description(attributes["banner"]),
         cause: cause(attributes["cause"]),
-        severity: severity(attributes["severity"]),
+        created_at: parse_time(attributes["created_at"]),
+        description: description(attributes["description"]),
+        effect: effect(attributes),
+        header: attributes["header"],
+        image: attributes["image"],
+        image_alternative_text: attributes["image_alternative_text"],
```
anthonyshull, line 21: 📓 `created_at`, `image`, and `image_alternative_text` are the new alert fields we are using in this feature. Everything else is just alphabetical re-arrangement.

lib/cms/page/diversions.ex
```diff
@@ -0,0 +1,64 @@
+defmodule CMS.Page.Diversions do
```
anthonyshull, line 1: 📓 This is just a copy of the default page type, but we need one to match.

lib/dotcom_web/controllers/cms_controller.ex
```diff
@@ -137,6 +138,17 @@ defmodule DotcomWeb.CMSController do
     end
   end
 
+  defp render_page(conn, %Page.Diversions{} = page) do
+    conn
+    |> assign(
+      :alerts,
+      Alerts.Repo.diversions_by_route_ids(page.related_transit, conn.assigns.date_time)
```
anthonyshull, line 145: 📓 If we're rendering a diversion page then we get diversions for any route that the page is tagged with. That's how we know that this diversion page is for the Red Line, for example.

test/dotcom/trip_plan/location_test.exs
```diff
@@ -4,7 +4,7 @@ defmodule Dotcom.TripPlan.LocationTest do
   alias TripPlan.NamedPosition
 
   import Mox
-  import Test.Support.Factories.LocationService
+  import Test.Support.Factories.LocationService.LocationService
```
anthonyshull, line 7: 📓 I renamed this to match the directory structure like all other factories.

lib/cms/helpers.ex
```diff
@@ -69,6 +69,22 @@ defmodule CMS.Helpers do
...
+    |> Enum.map(&Map.get(&1, "name"))
+    |> Enum.filter(&(!is_nil(&1)))
+  end
+
+  @spec parse_related_transit(map) :: list(String.t())
+  def parse_related_transit(%{} = data) do
+    data
+    |> Map.get("field_related_transit", [])
+    |> Enum.map(&Map.get(&1, "name"))
+    |> Enum.filter(&(!is_nil(&1)))
```
bklebe, line 85: suggestion: use `Enum.reject/2` instead of `Enum.filter/2`.

This code may also be made clearer using the `Access` module for nested traversal instead of mapping `Map.get`:

```suggestion
    data
    |> Map.get("field_related_transit", [])
    |> get_in([Access.all(), "name"])
    |> Enum.reject(&is_nil(&1))
```

it's possible to combine the first `Map.get` in the call to `get_in` and use `Access` for that too, but it would require checking if the result was nil before calling `reject`/`filter` because it doesn't allow us to specify a default value of `[]` as we do here so it's probably a wash.

bklebe (changes requested):

bklebe (approved):

PR #2151, opened 2024-08-19
Fix date sorting using timex

anthonyshull:
Comparison operators don't work with date times.

### DIFF
--- a/lib/alerts/repo.ex
+++ b/lib/alerts/repo.ex
@@ -39,7 +39,10 @@ defmodule Alerts.Repo do
     |> by_route_ids(now)
     |> Enum.filter(&Alert.diversion?/1)
     |> Enum.sort(fn a, b ->
-      a.active_period |> List.first() |> elem(0) < b.active_period |> List.first() |> elem(0)
+      first = a.active_period |> List.first() |> elem(0)
+      second = b.active_period |> List.first() |> elem(0)
+
+      Timex.before?(first, second)
     end)
   end
 


lib/alerts/repo.ex
```diff
@@ -39,7 +39,10 @@ defmodule Alerts.Repo do
     |> by_route_ids(now)
     |> Enum.filter(&Alert.diversion?/1)
     |> Enum.sort(fn a, b ->
-      a.active_period |> List.first() |> elem(0) < b.active_period |> List.first() |> elem(0)
+      first = a.active_period |> List.first() |> elem(0)
+      second = b.active_period |> List.first() |> elem(0)
+
+      Timex.before?(first, second)
```
bklebe, line 45: suggestion(non-blocking):
```suggestion
      first = a.active_period |> List.first() |> elem(0)
      second = b.active_period |> List.first() |> elem(0)

      :lt == DateTime.compare(first, second)
```

bklebe: https://hexdocs.pm/elixir/1.17.2/DateTime.html#compare/2

anthonyshull: Why not use `Timex`?

bklebe: I figure since `DateTime` offers an equivalent function here we might as well use the native functionality. (sorry my first suggestion used `compare` before I saw DateTime also has `before`)

PR #2153, opened 2024-08-20
Rearrange struct keys and update schedule

anthonyshull:
https://app.asana.com/0/555089885850811/1205890385667025/f

### DIFF
--- a/assets/ts/schedule/components/__schedule.d.ts
+++ b/assets/ts/schedule/components/__schedule.d.ts
@@ -120,12 +120,12 @@ export interface SimpleStopMap {
 }
 
 export interface ScheduleNote {
-  peak_service: string;
+  alternate_text: string | null;
+  exceptions: ServiceException[];
   offpeak_service: string | null;
+  peak_service: string;
   saturday_service: string;
   sunday_service: string;
-  exceptions: ServiceException[];
-  alternate_text: string | null;
 }
 
 export interface ServiceException {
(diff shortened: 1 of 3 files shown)


thecristen (changes requested):
lol at nothing reading those exceptions.

can you please make the same update to the TypeScript type at `assets/ts/schedule/components/__schedule.d.ts`?

anthonyshull:
> lol at nothing reading those exceptions.
> 
> can you please make the same update to the TypeScript type at `assets/ts/schedule/components/__schedule.d.ts`?

Oh, we hard-code these things in two places? Why not?

anthonyshull:
@thecristen I'm not seeing a hardcoding of data in the TS file. Do you just mean the key order?

assets/ts/schedule/components/__schedule.d.ts
```diff
@@ -120,12 +120,12 @@ export interface SimpleStopMap {
 }
 
 export interface ScheduleNote {
-  peak_service: string;
+  alternate_text: string | null;
+  exceptions: ServiceException[];
```
thecristen, line 124: Oh yeah, good catch on the order. 

And for some reason I thought you'd removed this `exceptions` from the Elixir type, though now I see I was mistaken. You _could_ remove them, since you deleted that Elixir test which was as far as I can tell the only thing that needed that value. Or you could put the test back in.

thecristen (approved):

assets/ts/schedule/components/__schedule.d.ts
anthonyshull: I left it because I assume it served some purpose before and that purpose might resurface at some point.

PR #2154, opened 2024-08-20
refactor(ViewHelpers): add svg error state

thecristen:
#### Summary of changes

**Asana Ticket:** [ArgumentError: unknown SVG icon-logan-express-DV.svg](https://app.asana.com/0/555089885850811/1208102975899686/f)

I decided to tackle this by having the unknown SVG return an error tuple instead of raising an `ArgumentError`.

Also adds bus icons as a fallback for the Logan Express and Massport cases.


lib/dotcom_web/views/helpers.ex
```diff
@@ -115,13 +115,26 @@ defmodule DotcomWeb.ViewHelpers do
...
       when is_binary(name) do
     route_number = String.slice(name, 0..1)
-    svg("icon-massport-#{route_number}.svg")
+
+    case svg("icon-massport-#{route_number}.svg") do
+      {:error, {:unknown_svg, _}} ->
+        svg("icon-mode-bus-default.svg")
+
+      icon ->
+        icon
```
anthonyshull, line 124: We could remove both of these case statements by just making the `svg/1` function return the bus mode default. Otherwise, you get a strange situation where calling it with an unknown string returns an error and then calls itself.

anthonyshull (changes requested):

lib/dotcom_web/views/helpers.ex
```diff
@@ -53,7 +53,13 @@ defmodule DotcomWeb.ViewHelpers do
   def subway_lines, do: @subway_lines
 
   def svg(unknown) do
-    raise ArgumentError, message: "unknown SVG #{unknown}"
+    cond do
+      unknown =~ ~r/massport|logan-express/ ->
+        svg("icon-mode-bus-default.svg")
+
+      true ->
+        raise ArgumentError, message: "unknown SVG #{unknown}"
```
anthonyshull, line 61: I can't think of a time we would want to raise an error instead of returning *some* icon. Worst case should send an "error" to Sentry saying that an SVG isn't found and just return the default.

anthonyshull: https://hexdocs.pm/sentry/Sentry.html#capture_message/2

thecristen: Yeah. It's not always gonna merit a bus representation! But it's a fine enough fallback. But what if the file for `svg("icon-mode-bus-default.svg")` mysteriously disappears from `/priv`? But am probably overthinking this point...

thecristen: If we want an even safer default I think the only option left would be to return a `""` or something that'll render but not notably impact the HTML.

anthonyshull (approved):
Yeah, I think we could talk about having some kind of "default" icon that is just blank with a defined height and width.

thecristen:
Heh I didn't notice you'd already approved it. 👍🏼 I was being annoying and really wanting _two_ fallbacks - a bus for when we know we're trying to show a Logan Express or Massport route, and the "true" fallback for when all other options have been exhausted. So I've added both (which could look like the two icons at the end of this screenshot).

image

PR #2158, opened 2024-08-22
feat(DotcomWeb.Live.TripPlanner): selecting and mapping locations

thecristen:
**Asana Ticket:** some wacky combination of [Generic LiveView-compatible Map component](https://app.asana.com/0/555089885850811/1207828528645192/f)  [`Phoenix.Component` for selecting locations](https://app.asana.com/0/555089885850811/1207828528645180/f) and [Trip planner form `Phoenix.LiveComponent`](https://app.asana.com/0/555089885850811/1207828528645185/f)

### Creating `<.TripPlanner.input_form />` for reuse 

- Based on `TripPlanner.InputForm` schema from #2159 
- Renders each location field this way:
```elixir
<.algolia_autocomplete config_type="trip-planner">
# a list of hidden inputs to capture each location's latitude, longitude, name, and stop ID
</.algolia_autocomplete>
```
The JS in the `AlgoliaAutocomplete` hook handles synchronizing these hidden inputs to the user's interactions with the AutocompleteJS piece.
- The form actually works from other non-LiveView pages if initialized minimally as `<.TripPlanner.input_form params={%{}} id="id" />` – submitting will navigate you to `/preview/trip-planner` with the fields filled in and map initialized accordingly ⭐ 
  - This required making the form a function component as opposed to a `LiveComponent`. Acknowledging here that this choice makes communicating form changes a little clunkier. In this case we can pass a LiveView process ID to this input form, and that PID will receive notification when the form data has been revalidated. This seems to work OK.
- Reiterating: URL params will populate the form. This uses the same format as the current trip planner (e.g. `plan[from][latitude]=foo` and such)
- Displays successfully submitted data
- Displays form validation errors
- I gave up on crafting some nice reusable LiveView-enabled map, and instead focused on making a Trip Planner specific map.
  - LiveView handles getting the AutocompleteJS selection and telling the map about this new data
  - The map's JS hook listens to the event and adds/moves/removes the corresponding map marker.

### Adds a new configuration to `<.algolia_autocomplete />`

- Adds the `"trip-planner"` config_type which specifies an AutocompleteJS configuration which:
  - Firstly specifies settings the same as the current trip planner: geolocation, popular locations, find up to 5 locations from AWS, find GTFS stops
  - Adds function for setting the initial state of the displayed input field based on the loaded form values (which are in turn initialized by the form params)
  - Adds an onSelect handler to each source, which will set the displayed input field text to the selected item, and push the selected data to LiveView
  - Adds an onReset handler to communicate to LiveView
- Adjusts the geolocation component with an additional prop, `onLocationFound`, to do something custom with the browser location (in this case... you guessed it... communicate to LiveView).
- Allows adding the supplemental form fields via a newly-added [slot](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#module-the-default-slot). 

#### Other tweaks

- Refactors the geolocation component with improved React code, because it was quite hard to test 🙈 
- CSS: Fix the location text being right-aligned
- JS: corrections to the location and popular sources, so that they're able to render templates without `<a>`
- CSS: Refactor the `_autocomplete-theme.scss` for cleaner output, and support adjusted style for error state
- Adds test coverage to much of the AutocompleteJS-related code


anthonyshull (changes requested):
There seems to be some weird behavior around the state of the inputs. Two major things. One, when I enter in 'North Station' and 'South Station' and plan a trip *then* refresh the page, the input values are replaced with coordinates. Two, there is inconsistency with intermediate error states. If I go to a fresh trip planner page and enter 'North Station' and then remove it, it doesn't show the error. If I plan a trip for 'North Station' to 'South Station' and then remove either one I do get the error.

I would suggest we don't show the intermediate errors and only show them on form submission. My guess as to why this is behaving this way is because you're checking for the dirtiness of the form and it will be dirty after submission.

anthonyshull (approved):

PR #2170, opened 2024-09-09
feat: use `MbtaMetro.Components`

thecristen:
- installs `:mbta_metro`, removes `:heroicons` since they're already loaded in Metro
- adjusts our Tailwind configuration:
  - removes `tw-` prefixing.
  - adds the `important` key to help reduce conflicts. 
  - The one breaking conflict I found was with the `collapse` utility, which is used in all the accordions via a Bootstrap plugin. So I disabled that [family of utilities](https://tailwindcss.com/docs/visibility), and all seems well.
  - formatted that file
- manually removed a bunch of utility class conflicts by prefixing _ours_
- removed CoreComponents, as Metro will replace those

Used the `<.error>` component in the preview trip planner, it works (styles + icon)

image


mix.lock
```diff
@@ -1,5 +1,6 @@
 %{
   "absinthe_client": {:hex, :absinthe_client, "0.1.1", "1e778d587a27b85ecc35e4a5fedc64c85d9fdfd05395745c7af5345564dff54e", [:mix], [{:castore, ">= 0.0.0", [hex: :castore, repo: "hexpm", optional: false]}, {:req, "~> 0.4", [hex: :req, repo: "hexpm", optional: false]}, {:slipstream, "~> 1.0", [hex: :slipstream, repo: "hexpm", optional: false]}], "hexpm", "e75a28c5bb647f485e9c03bbc3a47e7783742794bd4c10f3307a495a9e7273b6"},
+  "bandit": {:hex, :bandit, "1.5.7", "6856b1e1df4f2b0cb3df1377eab7891bec2da6a7fd69dc78594ad3e152363a50", [:mix], [{:hpax, "~> 1.0.0", [hex: :hpax, repo: "hexpm", optional: false]}, {:plug, "~> 1.14", [hex: :plug, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}, {:thousand_island, "~> 1.0", [hex: :thousand_island, repo: "hexpm", optional: false]}, {:websock, "~> 0.5", [hex: :websock, repo: "hexpm", optional: false]}], "hexpm", "f2dd92ae87d2cbea2fa9aa1652db157b6cba6c405cb44d4f6dd87abba41371cd"},
```
anthonyshull, line 3: Can you try cleaning your deps and re-pulling mbta_metro? We shouldn't need these deps in dotcom. I *think* I fixed it upstream.

lib/dotcom_web.ex
```diff
@@ -134,6 +134,8 @@ defmodule DotcomWeb do
     quote do
       # Use all HTML functionality (forms, tags, etc)
       use Phoenix.HTML
+      # Use all MBTA Metro components
+      use MbtaMetro
```
anthonyshull, line 138: Can we make this alphabetical with the `use` above?

assets/tailwind.config.js
```diff
@@ -1,80 +1,103 @@
 // See the Tailwind configuration guide for advanced usage
```
anthonyshull, line 1: Would be a good opportunity to copy some of the config from MbtaMetro that pulls in preline and default colors.

https://github.com/anthonyshull/mbta_metro/blob/main/assets/tailwind.config.js

anthonyshull (changes requested):
Some small changes and we should install Preline since it's in the Mbta Metro installation docs ;)

PR #2173, opened 2024-09-18
refactor(TripPlannerForm): use LiveComponent

thecristen:
**Asana Ticket:** [LiveView-compatible trip planner form inputs](https://app.asana.com/0/555089885850811/1207999988422557/f)

This is primarily a refactor of the trip planner form, from using a functional component to a `Phoenix.LiveComponent`.

Highlights:

- Improvements to validation: checks whether the field is touched before rendering any errors. Also should eagerly revalidate (e.g. as soon as you fill in the field the error will disappear, as opposed to leaving the error on screen until resubmitting the form)
- Datetime selection
  - The datepicker toggles visibility when choosing between now/arrive/depart
  - The form will always set the chosen datetime to the current date and time if "now" is selected, otherwise it'll validate whether the datetime is in the past
- Mode selection
  - Verifies whether at least one mode is present
  - Adjusts the accordion text based on selected modes

Other changes:

- Uses several components from `MbtaMetro`, so this PR installs and sets this up as well.
  - Adjusts our Tailwind configuration:
    - Follows the guidance from `MbtaMetro`
    - Adds two class names to the `blocklist` since they conflict with our existing Bootstrap styles in terrible ways
    - Adds `important: true`
  - Thanks to the prior two bulletpoints: our CSS now imports the **base** Tailwind CSS
  - Removes our `CoreComponents`
- Sets up Phoenix verified routes so I can use `~p` in a test


assets/tailwind.config.js
```diff
@@ -1,80 +1,66 @@
...
+  },
+  blocklist: ["container", "collapse"],
+  content: [
+    ...content,
+    "./js/**/*.js",
+    "./ts/**/*.ts",
+    "./ts/**/*.tsx",
+    "../lib/dotcom_web.ex",
+    "../lib/dotcom_web/**/*.*ex",
+    "../deps/mbta_metro/lib/mbta_metro/**/*.ex"
```
anthonyshull, line 26: You shouldn't need this.

assets/css/app.scss
```diff
@@ -1,4 +1,5 @@
 // TailwindCSS
+@import 'tailwindcss/base';
 @import 'tailwindcss/components';
 @import 'tailwindcss/utilities';
```
anthonyshull, line 4: You actually don't need. You can include the mbta metro stylesheet that has these as well as the mapgl styles.

https://github.com/anthonyshull/mbta_metro?tab=readme-ov-file#install-mbta-metro

assets/tailwind.config.js
thecristen: This PR is outdated so this line works for this PR, which predates the /component and /live directories in Metro 😓

thecristen: Which I guess means I need to downgrade the npm lib? Umm hmm

thecristen: Ha nope, the Elixir version of Metro being used in this PR predates mbta_metro 0.0.1. I'll be sure to remove this line when I update it in a later PR!

assets/css/app.scss
thecristen: Since this PR is pretty old I'll tackle including the mbta metro stylesheet in a later PR. At least this PR gets us one step closer to Tailwind-style-compatibility. 👍🏼

anthonyshull (approved):

PR #2176, opened 2024-10-01
Adds Mbta Metro to dotcom

anthonyshull:


thecristen (changes requested):
I had to re-introduce this into the Dotcom Tailwind config but otherwise it's quite workable as-is (except for the one note below)!

```js
corePlugins: {
  preflight: false
},
blocklist: ["container", "collapse"],
```


The other thing that needs to change is this CSS from Metro:

```css
// default.css
@layer utilities {
  * {
    font-family: "InterNormal", system-ui, sans-serif;
  }
}
```
Why? It takes precedence over any elements that's depending on inheriting the font-family from their parent element. Hilariously it also mangles the MBTA logo  (which has a `<text>` with `font-family="Helvetica-Bold, Helvetica"` set):

image

That CSS _might_ work if it's in the `base` layer instead of in `utilities`. I question why you even need this though, as opposed to just let consumers (and these components) add a `font-inter` class to get the desired result (this feels like the more "tailwind" way at least).

thecristen:
Sorry for the belated comment but I also just noticed that importing the JS hooks doesn't work (I'd started trying to use the datepicker). 😞 

image

anthonyshull:
> Sorry for the belated comment but I also just noticed that importing the JS hooks doesn't work (I'd started trying to use the datepicker). 😞
> 
> image

I can tackle this today. I'm sure it's something small.

PR #2182, opened 2024-10-02
fix: Don't overwrite dep directories in `docker-compose`'ed containers

joshlarson:
#### Summary of changes

By mounting `..` into `/app`, we were overwriting the `deps/` directory in the containers with whatever was in the repository when running `docker-compose`. That means that the results from the build steps in the dockerfiles were being overwritten. A particularly annoying effect is that if you run `docker-compose -f deploy/dev.yml up` on a fresh clone, then even though the build step does fetch and build dependencies, `/app/deps` is missing, and the app doesn't start.

This fixes that by explicitly re-mounting the image's `/app/deps` folder into the container, regardless of what's in `..:/app`

---

#### General checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ For details on environment configuration, including optional variables, see
 The easiest way to develop MBTA dotcom is to use Docker Compose.
 
 ```
-docker compose -f deploy/dev.yml up -d
+docker compose -f deploy/dev.yml up -d --build
 ```
 
 This will set up Redis in cluster mode and run two versions of Dotcom with nginx load balancing requests between them.
--- a/deploy/dev.yml
+++ b/deploy/dev.yml
@@ -50,6 +50,7 @@ services:
             timeout: 10s
         volumes:
             - ../:/app
+            - /app/_build
             - /app/deps
             - /app/assets/node_modules
         networks:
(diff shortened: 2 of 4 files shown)


thecristen (approved):
works on a fresh install! thanks so much

anthonyshull:
Can we make sure it works when upgrading a dependency and changing config?

joshlarson:
@anthonyshull changing config works the same as before - a simple `docker compose restart dotcom-1 dotcom-2` will work (I have validated this)

I think updating a dependency will require `docker compose down; docker compose up --build`. Any deps in particular that I can/should test this out with?

If we would prefer `docker compose restart` to also pick up new deps, we can add `mix deps.get` to the `CMD` step. We still gain by keeping it earlier in the Dockerfile as well, because if the image's deps are up-to-date, then `mix deps.get` in the `CMD` will just breeze through the cache.

anthonyshull:
> I think updating a dependency will require `docker compose down; docker compose up --build`. Any deps in particular that I can/should test this out with?

Any dep should be fine; we want to make sure it picks up the change.

joshlarson:
`down` and then `up` picks up dependency changes.

I have a few other ideas for how to speed this up more, but those can be follow-up tasks.

joshlarson:
@thecristen in case it affects your approval, I added [another commit](https://github.com/mbta/dotcom/pull/2182/commits/fff1f56076a5f44366dfbd0a58b18c756bdf3ebd)

thecristen:
Sounds good to me @joshlarson ! 👍🏼

PR #2190, opened 2024-10-07
fix: Restore cache calls in LocationService, fixing some warnings

joshlarson:
This removes these warnings, which show up any time we run any test

```ex
     warning: module attribute @ttl was set but never used
     │
  10 │   @ttl :timer.hours(24)
     │   ~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/location_service/location_service.ex:10: LocationService (module)

     warning: module attribute @cache was set but never used
     │
   9 │   @cache Application.compile_env!(:dotcom, :cache)
     │   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/location_service/location_service.ex:9: LocationService (module)

```

### DIFF
--- a/lib/location_service/location_service.ex
+++ b/lib/location_service/location_service.ex
@@ -6,6 +6,8 @@ defmodule LocationService do
   use Nebulex.Caching.Decorators
 
   @aws_client Application.compile_env!(:dotcom, :aws_client)
+  @cache Application.compile_env!(:dotcom, :cache)
+  @ttl :timer.hours(24)
 
   @base_options %{
     "FilterCountries" => ["USA"],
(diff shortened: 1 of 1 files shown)


thecristen (changes requested):
😓 This was my sloppiness, I'd changed the format of the responses here and commented these out to disable the caching here, but had always meant to reenable these once the TTL window had passed (thus old responses no longer cached).

@joshlarson Can you please put these back and un-comment the `@decorate cacheable` calls? That'll remove the warnings by resuming usage of those attributes 😇

joshlarson:
> Can you please put these back and un-comment the @decorate cacheable calls?

@thecristen Finally did ☝️! I think this should be ready for re-review now!

thecristen (approved):

PR #2191, opened 2024-10-08
Make sure Rancher Desktop works

anthonyshull:
We want to make sure that we can run the Docker Compose setup w/out using Docker Desktop in case the bootjacks come for us in the night. Copying over the package lock in the monitor image was failing because it was requiring some Mac-only deps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,40 @@ For details on environment configuration, including optional variables, see
 
 The easiest way to develop MBTA dotcom is to use Docker Compose.
 
+### Installing Rancher Desktop
+
+We have tested this setup using [Rancher Desktop](https://rancherdesktop.io) on Mac.
+
+Install Rancher Desktop and then download the [docker-credential-osxkeychain](https://github.com/docker/docker-credential-helpers/releases) binary.
+Add it to your `$PATH`.
+
+Install docker and docker compose:
+
+```
+brew install docker docker-compose
+```
+
+Make sure your `~/.docker/config.json` looks like the following:
+
+```json
+{
+  "auths": {},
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -123,6 +123,42 @@ For details on environment configuration, including optional variables, see
 
 The easiest way to develop MBTA dotcom is to use Docker Compose.
 
+### Installing Rancher Desktop
+
+We have tested this setup using [Rancher Desktop](https://rancherdesktop.io) on Mac.
```
joshlarson, line 128: The implication here is that we've tested and confirmed that it works, but I wasn't able to get Rancher working on my machine. I would either remove this sentence, or replace it with something along the lines of:

```suggestion
[Rancher Desktop](https://rancherdesktop.io) is a free and open source container runtime that backs Docker containers and provides a Docker-for-Desktop-like frontend for viewing and managing containers and images.

(*Note: Rancher is backed by Lima, which has been known to have problems with Mac M2 processors. We currently don't know of any reliable workarounds.*)
```

README.md
```diff
@@ -123,6 +123,42 @@ For details on environment configuration, including optional variables, see
...
+  "credsStore": "osxkeychain",
+  "currentContext": "rancher-desktop"
+}
+```
+
+Login to Docker:
+
+```
+docker login
+```
```
joshlarson, line 156: How come the `docker login` step is necessary?

anthonyshull: I guess it's not necessary for Dotcom since we use a different registry. But, it is part of setting up your local Docker if you're going to be using it for anything else.

PR #2194, opened 2024-10-11
feat(LiveComponents.TripPlannerForm): fixes for mode selector + updated MbtaMetro & overhauled typography

thecristen:
#### Summary of changes

**Asana Ticket:** [TP UI | Inputs | Polished accessibility selector](https://app.asana.com/0/555089885850811/1208227271108372/f) and [TP UI | Inputs | Polished mode selector](https://app.asana.com/0/555089885850811/1208217344262978/f)

Technically not 100% polished yet since these don't yet use the official design system tokens for colors and spacing. These are pretty close.

Sorry for the omnibus PR! 😅 But the recent updates to `MbtaMetro` included [CSS I wanted to incorporate here](https://github.com/anthonyshull/mbta_metro/blob/main/assets/css/default.css), including:
- Updated font files (which I've downloaded here, replacing our current font files). This should fix the current issue on dev (where it can't find the Inter font and thus falls back to Helvetica)
- Updated header styles. They're a 99.9% match of what's already on the site, with much simpler CSS.
These let me start phasing out a whole bunch of our homegrown utility files... which accounts for most of this PR diff. This phaseout is not even close to complete but is enough for now. I've checked many pages to ensure I'm not introducing style regressions! 
- Our tailwind config now uses the `@tailwindcss/forms` plugin in full, including its style resets. This required some edits to our custom `<select>` styling to avoid showing multiple carets.
- Our tailwind config now uses `important: true` as the forms plugin is kinda buggy otherwise, since it inserts its reset CSS in a way that overwrites our own styles. 🙄 
- Adds extra configuration to LiveSocket, to avoid `<details>` open state changing inside a LiveView.


assets/js/app.js
```diff
@@ -58,7 +58,20 @@ let csrfToken = document
 
 let liveSocket = new LiveSocket("/live", Socket, {
   params: { _csrf_token: csrfToken },
-  hooks: { ...Hooks, ...DotcomHooks }
+  hooks: { ...Hooks, ...DotcomHooks },
+  dom: {
+    onBeforeElUpdated(from, to) {
+      // Clone open state of <details>
```
joshlarson, line 64: Just to refresh my memory: The point of this is because the `<.accordion />` component doesn't remember its open/closed state if its summary changes? Can we add a comment to that effect (I know there's a comment, but I mean something that explains more clearly what the consequences of not doing this are).

For instance,

```suggestion
      // By default the <.accordion /> component from metro will not remember its
      // open/closed state if the summary changes. This creates some surprising
      // behavior in the mode-selector on the trip planner, where each time you
      // change your selected mode, the menu closes.
      //
      // This code forces the `<details>` tag (what the accordion uses under the
      // hood) to remember its open/closed state to avoid that bug.
```

thecristen: Yep! It's not so much specific to Metro's accordion but it is to the HTML `<details>` open/closed state.

PR #2195, opened 2024-10-11
Polished date picker implementation

anthonyshull:
It looks like there are some slight conflicts between the CSS that Cristen did in her polish PR and the changes I made in mine. I figure we can work those out after the big pieces are done being built. Another outstanding task is getting the default date of the date picker to be "now" in East Coast time. This is only a problem for people not on the East Coast, but something we should fix. We can do that in a separate task since it will necessitate a new release of Metro.


lib/dotcom/trip_plan/input_form.ex
```diff
@@ -25,7 +24,7 @@ defmodule Dotcom.TripPlan.InputForm do
   typed_embedded_schema do
     embeds_one(:from, __MODULE__.Location)
     embeds_one(:to, __MODULE__.Location)
-    field(:datetime_type, Ecto.Enum, values: @time_types)
+    field(:datetime_type, :string)
```
anthonyshull, line 27: 📓 I changed this for now because we need to have a way to navigate between these atoms and the string we're given back by the form. We can figure this out later.

lib/dotcom_web/components/live_components/trip_planner_form.ex
```diff
@@ -31,7 +31,7 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerForm do
      assign(socket, %{
        form: form,
        location_keys: InputForm.Location.fields(),
-       show_datepicker: input_value(form, :datetime_type) != :now
+       show_datepicker: false
```
anthonyshull, line 34: 📓 This is always false for now until we implement some way to load a plan.

assets/css/_trip-plan-form.scss
```diff
@@ -220,6 +220,12 @@
   }
 }
 
+#date-picker-calendar {
+  .form-control[readonly] {
+    background-color: $body-bg;
```
anthonyshull, line 225: 📓 Bootstrap was setting this to grey so we need this override.

lib/dotcom/trip_plan/input_form.ex
thecristen: the atom/string juggling was definitely my least favorite part about trying to handle mode changes too...

thecristen (approved):
Works well! Some additional polish notes:
- There's a bit too much space between the timepicker and the buttons, and I think it's due to some margins added to all lists via Dotcom's CSS. If you add `mb-0` to the input group's `<ul>` tag that'll help.
- The timepicker input should have the same border width as the buttons, right now it's a lot thicker
- The calendar icon position is a little wonky, also ought to be blue rather than low-contrast grey

test/dotcom_web/live/trip_planner_test.exs
```diff
@@ -25,22 +25,6 @@ defmodule DotcomWeb.Live.TripPlannerTest do
...
-        |> render_click()
-
-      assert html =~ date_input
-
-      html =
-        view
-        |> element("input[value=now]")
-        |> render_click()
-
-      refute html =~ date_input
```
anthonyshull, line 43: 📓 I'm removing these until this stabilizes and we can test the entire form.

test/dotcom_web/live/trip_planner_test.exs
```diff
@@ -78,12 +62,6 @@ defmodule DotcomWeb.Live.TripPlannerTest do
     end
 
     test "shows errors on form submit", %{view: view} do
-      html =
-        view
-        |> element("form")
-        |> render_submit()
-
-      assert html =~ "Please add a destination."
```
anthonyshull, line 86: 📓 Same here. I think we'll want to figure out our final error display approach before writing these tests.

PR #2196, opened 2024-10-11
refactor(Live.TripPlanner): show form submission while waiting for trips

thecristen:
I wasn't quite satisfied with the current experience of waiting for results. This change simplies the form code a bit, and lets us show the submitted form details immediately (black text), while showing a spinner until the results are retrieved (status of this is shown below the black text).

image

It can also show the error code.

image

The formatting is also adjusted to be closer to the desired output from the Figma mock, though I couldn't figure out how to format dates with suffixes (e.g. October 11**th** instead of October 11)..

### DIFF
--- a/lib/dotcom_web/live/trip_planner.ex
+++ b/lib/dotcom_web/live/trip_planner.ex
@@ -11,8 +11,7 @@ defmodule DotcomWeb.Live.TripPlanner do
   alias Dotcom.TripPlan.{InputForm.Modes, ItineraryGroups}
 
   import DotcomWeb.Components.TripPlanner.ItineraryGroup, only: [itinerary_group: 1]
-  import MbtaMetro.Components.Feedback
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/live/trip_planner.ex
```diff
@@ -11,6 +11,8 @@ defmodule DotcomWeb.Live.TripPlanner do
   alias Dotcom.TripPlan.{InputForm.Modes, ItineraryGroups}
 
   import DotcomWeb.Components.TripPlanner.ItineraryGroup, only: [itinerary_group: 1]
+  import MbtaMetro.Components.Feedback
+  import MbtaMetro.Components.Spinner
```
anthonyshull, line 15: Can combine these `MbtaMetro.Components.{Feedback, Spinner}`.

anthonyshull (approved):

PR #2199, opened 2024-10-22
fix: Failing "publish docs" build

joshlarson:
The `Publish docs` action has been failing consistently for a few weeks with this message:

```
** (Mix) You're trying to run :dotcom on Elixir v1.16.1 but it has declared in its mix.exs file it supports only Elixir ~> 1.17
Error: Process completed with exit code 1.
```

This should fix that.

Example: https://github.com/mbta/dotcom/actions/runs/11371895399/job/31634982030


thecristen (approved):
🎉

PR #2203, opened 2024-10-23
Show the ad for the mobile app to everyone not on an android

anthonyshull:
https://app.asana.com/0/555089885850811/1208568379309256/f

I didn't put it on the schedule finder because it is a modal, not a page. It is on the schedule page below the schedule finder.


assets/js/mobile-app-banner.js
```diff
@@ -1,28 +1,45 @@
...
+const isNot = url => {
+  return isNotAndroid() && isNotCommuterRail(url);
+};
+
 // Show the mobile app banner
 export default function mobileAppBanner() {
-  if (isIPhone() && (isAlertsPage() || isSchedulePage() || isStopPage())) {
+  const url = window.location.href;
+
+  if (is(url) && isNot(url)) {
```
thecristen, line 42: lolololol at this naming

thecristen (approved):
fine!

assets/js/mobile-app-banner.js
anthonyshull: I thought about `ises` and `isNots`...but the former probably isn't a word.

assets/js/mobile-app-banner.js
```diff
@@ -11,7 +11,7 @@ const isNotCommuterRail = url => {
 };
 
 const isSchedulePage = url => {
-  return /\/schedules\/[\w.-]+\/line/.test(url);
+  return isNotCommuterRail(url) && /\/schedules\/[\w.-]+\/line/.test(url);
 };
```
thecristen: I don't see what is not commuter rail condition has anything to do with whether it's a schedule page, did we mix something up here

thecristen (dismissed):
I like the revised approach, just a question affecting the readability of one of your functions here

assets/js/mobile-app-banner.js
anthonyshull: Yeah, it's a little weird because they want it to be on schedule pages that aren't commuter rail schedules.

anthonyshull: So, yeah, there wouldn't have been overlap between the two. I thought it was better to communicate that it shouldn't be commuter rail even though they won't overlap. But, I can remove it.

mix.exs
```diff
@@ -103,7 +103,7 @@ defmodule DotCom.Mixfile do
       {:jason, "1.4.4", override: true},
       {:logster, "1.1.1"},
       {:mail, "0.3.1"},
-      {:mbta_metro, "0.0.49"},
+      {:mbta_metro, "0.0.50"},
```
joshlarson, line 106: Was upgrading metro necessary as part of this PR?

assets/js/mobile-app-banner.js
```diff
@@ -1,28 +1,44 @@
...
-  const url = window.location.href;
-
-  return /\/alerts\/[\w.-]+/.test(url);
+const isNotAndroid = () => {
+  return !/Android/.test(navigator.userAgent);
 };
 
-const isIPhone = () => {
-  return /iPhone/.test(navigator.userAgent);
+const isNotCommuterRail = url => {
```
joshlarson, line 5: I think we also need to filter out ferry pages. The banner is showing up, for instance, on http://localhost:4001/schedules/Boat-F4/alerts

Screenshot 2024-10-25 at 2 42 25 PM

joshlarson (changes requested):
1. https://github.com/mbta/dotcom/pull/2203/files#r1817198719 <-- I think we need to suppress the banner on ferry pages too, per the A/C's of the ticket.
2. I also think that the logic in `mobile-app-banner.js` is getting complicated enough, especially with all the regexes, that it would be worth adding tests. This can be a follow-up, but unless that logic gets a lot simpler (for instance, I don't really understand why we're not just showing it on every page), I'd hesitate to iterate on it anymore without tests.

mix.exs
anthonyshull: No, but we might as well.

joshlarson: I won't block approval on this, but in general I don't like lumping library upgrades together with unrelated features. It runs the risk of creating hard-to-revert surprises, especially when the library is in alpha, as metro is.

joshlarson (approved):

PR #2204, opened 2024-10-23
feat: Add a nicer summary view for trip planner output

joshlarson:
# Before
Screenshot 2024-10-23 at 2 19 41 PM

# After
Screenshot 2024-10-23 at 2 20 25 PM

---

**Parent Ticket:** [Trip Planner Preview | Output summary view](https://app.asana.com/0/555089885850811/1208278332362878/f)
**Subtask Ticket:** [Add mode icons (not worrying about the smushed icons for now) to the summary view](https://app.asana.com/0/0/1208594821721719/f)

---

#### General checks

---

## Assorted Implementation Notes

- This does not fully solve the parent ticket - there are some elements missing, and the icons won't always be right. I'm treating those as follow-up tasks.
- Smushed icons are a follow-up ticket. Showing icons side-by-side will be part of this ticket, but also a follow-up PR. This will require changing the `ItineraryGroups` algorithm to group the legs within itineraries as well as grouping similar itineraries themselves.
- I wasn't able to use the Metro badges for this because they rendered incorrectly. Happy to fix that now if it's easy, but otherwise I don't want to block the functionality here on fixing the library, especially since the icons I had already built do do the job correctly.
Screenshot 2024-10-23 at 5 46 50 PM


lib/dotcom_web/components/trip_planner/itinerary_group.ex
```diff
@@ -14,12 +16,31 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryGroup do
   """
   def itinerary_group(assigns) do
     ~H"""
-    <div class="mb-3 p-2 border border-2 border-slate-200">
-      <% [first | rest] = @group %>
-      <div class="text-slate-800 font-bold">Group with <%= Enum.count(@group) %> options</div>
+    <% [first | rest] = @group %>
+    <% [first_leg | rest_of_legs] = first.legs %>
+    <div class="border border-gray-lighter border-solid m-4 p-4">
+      <div class="flex flex-row mb-3">
```
joshlarson, line 22: I'm on the fence about this being a whole big thing. It feels like a lot is going on in this one HEEx blob, but I'm not totally sure about the optimal way to slice and dice it.

Happy to either slice it up as part of this PR, or refactor as a follow-up.

lib/dotcom_web/components/trip_planner/itinerary_group.ex
```diff
@@ -33,20 +54,59 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryGroup do
...
       </.accordion>
     </div>
     """
   end
 
+  attr :class, :string, default: ""
+  attr :mode, :any, required: true
+  attr :leg, :any, required: true
+
+  defp leg_icon(%{mode: %PersonalDetail{}, leg: %{duration: duration}} = assigns) do
```
joshlarson, line 67: A few thoughts on this `<.leg_icon />` component:

- I'm not convinced that this is the right place for this, but since this is the only consumer of this component for now, that's where this landed.
- I'm partly thinking that this does belong in the same file as (or at least close to) _some part_ of the blob mentioned above, which is part of why I haven't moved it yet.
- I'm not even a little bit attached to the name `.leg_icon`. It's what came to me as I was building this, and couldn't immediately think of a better name. Open to ideas!

lib/dotcom_web/components/svg.ex
```diff
@@ -0,0 +1,38 @@
+defmodule DotcomWeb.Components.Svg do
+  use Phoenix.Component
+
+  attr :rest, :global
+
+  def logan_express_bb_icon(assigns) do
```
joshlarson, line 6: Open to other ways of doing SVG's, but this way was a clean way I could think of that allowed me to specify properties on the SVG element itself, which I _think_ was lacking in the other `svg` functions I found. If there's another prior art for SVG's that would work, and that we haven't implicitly deprecated let me know!

lib/dotcom_web/components/mode_icons.ex
```diff
@@ -0,0 +1,112 @@
...
+
+  def mode_icon(%{route_name: "BB", type: :logan_express} = assigns) do
+    ~H"""
+    <.logan_express_bb_icon class={"h-6 w-6 #{@class}"} />
+    """
+  end
+
+  def mode_icon(assigns) do
+    ~H"""
+    <.subway_icon class={"bg-silver-500 #{@class}"}>?</.subway_icon>
```
thecristen, line 100: `?` :sob:
I like the idea of a fallback icon but I don't think it should be labeled as a "subway" icon.

lib/dotcom_web/components/svg.ex
thecristen: You'll need different `fill` colors in here for the other Logan Express routes. But I think you can make this a generic Logan Express icon function, and pattern match on the route name to then decide the right fill colors.

lib/dotcom_web/components/trip_planner/itinerary_group.ex
thecristen: I think the name's good and the placement makes sense!

thecristen (commented):
Feel free to do some refactoring as it makes sense for you! Left a few comments.

anthonyshull:
I don't think any of those badge or icon components are necessary.

assets/tailwind.config.js
```diff
@@ -29,8 +29,13 @@ module.exports = {
   theme: {
     extend: {
       colors: {
-        ...colors
-      }
+        ...colors,
+        "gray-lighter": "#b0b5c0",
+        "gray-light": "#788093",
```
anthonyshull, line 34: These colors should be in the format `hue-shade`.

lib/dotcom_web/components/badges.ex
```diff
@@ -0,0 +1,36 @@
+defmodule DotcomWeb.Components.Badges do
+  use Phoenix.Component
+
+  attr :class, :string, default: ""
+  slot :inner_block
+
+  def circle_icon(assigns) do
```
anthonyshull, line 7: I don't get why we need this.

lib/dotcom_web/components/badges.ex
```diff
@@ -0,0 +1,36 @@
...
+    ]}>
+      <%= render_slot(@inner_block) %>
+    </span>
+    """
+  end
+
+  attr :class, :string, default: ""
+  slot :inner_block
+
+  def rect_icon(assigns) do
```
anthonyshull, line 29: Or this.

lib/dotcom_web/components/mode_icons.ex
```diff
@@ -0,0 +1,112 @@
...
+    ~H"""
+    <.subway_icon class={"bg-orange-500 #{@class}"}>OL</.subway_icon>
+    """
+  end
+
+  def mode_icon(%{type: :red_line} = assigns) do
+    ~H"""
+    <.subway_icon class={"bg-red-500 #{@class}"}>RL</.subway_icon>
+    """
+  end
```
anthonyshull, line 84: These could all use metro and be done programmatically.

```
def mode_icon(%{type: type} = assigns) do
  color = type |> Atom.to_string() |> String.split("_") |> List.first()
  
  ~H"""
  <.badge color={color} type="circle"><%= type_to_string(type) %></.badge>
  """
end
```

lib/dotcom_web/components/mode_icons.ex
```diff
@@ -0,0 +1,112 @@
...
+  def mode_icon(assigns) do
+    ~H"""
+    <.subway_icon class={"bg-silver-500 #{@class}"}>?</.subway_icon>
+    """
+  end
+
+  attr :class, :string, default: ""
+  slot :inner_block
+
+  defp subway_icon(assigns) do
```
anthonyshull, line 107: Don't think this is necessary.

lib/dotcom_web/components/svg.ex
```diff
@@ -0,0 +1,38 @@
...
+  def walk_icon(assigns) do
+    ~H"""
+    <svg role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" {@rest}>
+      <g fill="#1C1E23" fill-rule="nonzero" transform="matrix(-1 0 0 1 14 0)">
+        <circle cx="4.747" cy="1.318" r="1.318" />
+        <path d="M.602 7.086c.366-.064 1.833-.318 1.915-.345.093-.031 0 0 .03-.047.032-.047 1.311-2.745 1.436-2.886.125-.14.468-.499.951-.655.484-.156.967.094.967.094S8.304 4.62 8.46 4.697c.157.079.297.36.297.36l1.121 2.108a.678.678 0 1 1-1.191.579L7.641 5.766 6.985 5.4c.047.14.648 3.065.648 3.065l.608 3.463 2.132 2.568a.783.783 0 1 1-1.17 1.025l-2.545-2.993c-.21-.25-.265-.375-.296-.538-.031-.164-.383-2.278-.383-2.278l-2.012 2.013v3.548h-.004a.788.788 0 0 1-1.567 0h-.004v-.068-.012-.016c.004-.54.02-3.233 0-3.842-.024-.679.218-.827.273-.89.055-.062 1.934-2.058 1.934-2.058l-.413-2.004s-.538 1.052-.593 1.17c-.055.116-.226.327-.405.366-.18.039-2.418.491-2.418.491l-.001-.004c-.033.004-.066.01-.1.01a.666.666 0 1 1-.066-1.33z" />
+      </g>
+    </svg>
+    """
+  end
```
anthonyshull, line 37: I would put these svgs into files and then use something like: https://github.com/anthonyshull/mbta_metro/blob/main/lib/mbta_metro/components/icon.ex to compile them when needed. I still think they would best go in a separate library.

lib/dotcom_web/components/trip_planner/itinerary_group.ex
```diff
@@ -3,6 +3,8 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryGroup do
   A component to render an itinerary group.
   """
 
+  import DotcomWeb.Components.{ModeIcons, Svg}
+  alias Dotcom.TripPlan.{TransitDetail, PersonalDetail}
   use DotcomWeb, :component
```
anthonyshull, line 8: We try to order things `use`, `require`, `import`, `alias`. Ordered by impact. Also, an empty line in between the different include types and everything alphabetized. Really helps when includes become large.

```
use DotcomWeb, :component

import DotcomWeb.Components.{ModeIcons, Svg}

alias Dotcom.TripPlan.{PersonalDetail, TransitDetail}
```

anthonyshull:
Can definitely use metro now to put thos SVGs into a directory and turn them into metro icons. Then put that icon into a circle badge.

lib/dotcom_web/components/svg.ex
anthonyshull: If we do: https://github.com/mbta/rider-design-system/pull/1/files we can use https://github.com/anthonyshull/mbta_metro/pull/10 and then just set the fill like other metro icons.

lib/dotcom_web/components/badges.ex
joshlarson: Historical context:

I had actually built this before you added the badges to metro, so the path of least resistance was to leave them. I did try to migrate to using metro, but ran into a bug where they came out looking wrong. I actually did leave a screenshot in the PR description. 

I'm willing to spend a small amount of time trying to get metro to work with this, but depending on how long that takes, it might make more sense to do that as a follow-up PR, rather than blocking this PR on it.

lib/dotcom_web/components/mode_icons.ex
joshlarson: I could see some amount of DRY benefit to this suggestion, but that would also make this code pretty opaque.

Plus, we still need a giant case statement or pattern-match for `type_to_string`.

Plus, it doesn't look like the team is aligned on this suggestion.

lib/dotcom_web/components/trip_planner/itinerary_group.ex
joshlarson: Ahh - thanks for this! I remember a conversation somewhere else (slack?) where we were talking this through - this is a helpful example of that.

assets/tailwind.config.js
joshlarson: What does this mean?

I couldn't find those hex codes elsewhere in dotcom or in metro, and I named them so that the names in Figma and the names in our code would line up.

lib/dotcom_web/components/mode_icons.ex
anthonyshull: You don't need a case statement or large pattern match.

```
def type_to_string(type) do
  type
  |> Atom.to_string()
  |> String.split("_")
  |> Enum.map(&String.at(&1, 0))
  |> Enum.map(&String.upcase/1)
  |> letters_to_initials()
end

def letters_to_initials([a, b, c]), do: c

def letters_to_initials(letters), do: Enum.join(letters)
```

anthonyshull: This just programmatically expresses what you're doing manually in the above examples.

lib/dotcom_web/components/badges.ex
anthonyshull: If you have problems integrating metro, let me know and I can probably figure it out quickly. Merging a PR with stopgaps that we just have to change later doesn't really save us anything.

assets/tailwind.config.js
anthonyshull: All other colors take the form of `hue-shade`. E.g., `blue-500`, `red-300`.

https://github.com/anthonyshull/mbta_metro/blob/main/assets/js/styles/colors.js

lib/dotcom_web/components/badges.ex
anthonyshull:
```
<.badge type="square" color="yellow" class="text-yellow-800 rounded-sm">66</.badge>
```
Fix the override for `p` in bootstrap's `_reboot.css`.

Screenshot Capture - 2024-10-29 - 15-40-45

thecristen: Beware! That might very well have consequence on other pages across Dotcom using `<p>`

assets/tailwind.config.js
thecristen: The names in Figma are that way because that's what Dotcom uses currently. 

These are actually captured as Sass variables ([link](https://github.com/mbta/dotcom/blob/603367ef55285c1ed8ad131f0d601c3e46b993d7/assets/css/_colors.scss#L20-L24)) but did not yet have corresponding utility classes. This adds them! I think they're useful to have until we derive everything from the design system.

lib/dotcom_web/components/badges.ex
anthonyshull:
```
<.badge type="square" color="yellow" class="mbta-metro-badge text-yellow-800 rounded-sm">66</.badge>
```

```
.mbta-metro-badge p {
  margin: 0;
}
```

anthonyshull: Could also change it to a `span` inside the badge.

lib/dotcom_web/components/svg.ex
joshlarson: I think I accomplished these suggestions (minus `rider-design-system` - that can be a follow-up) in [this commit](https://github.com/mbta/dotcom/pull/2204/commits/8f83729d395a773560d08eba9bab46865d8d67d2) - lemme know what you think!

assets/tailwind.config.js
```diff
@@ -29,8 +29,18 @@ module.exports = {
...
+        "navy": {
+          500: "#104c8f"
+	},
+        "logan-express-bb": "#f16823",
```
joshlarson, line 38: I'm open to suggestions about how to name these colors. I do get the sense that the jury's still out regarding how MBTA-brand-specific colors should be named, but it seems to me like there's a place for some very specific hard-coded brand colors, like "The Logan Express Braintree color".

Aaaaand actually, I'm kind of going back to my original thought that the MBTA blue line blue (or subway blue) or whatever should also get its own very explicit and obvious name (`mbta-subway-blue`, for instance), rather than relying on `blue-500` being the subway-brand color.

lib/dotcom_web/components/mode_icons.ex
thecristen: This program would work for 7 of the 10 highlighted clauses (there's a forgotten extra one, `:mattapan_line`, which fits this use case _but_ doesn't follow this logic, as we'd want an M instead of ML, there's no red in the name...). 

But I actually find the above examples easier to understand at a glance than this programmatic expression. 

I do agree it could stand to be a bit more concise. I think a guard expression would achieve this more easily _and_ more clearly, e.g.

```elixir
def mode_icon(%{type: type} = assigns) when type in [:red_line, :mattapan_line, :blue_line, :orange_line, ..all the greens] do
   color = tiny function that can map the color to these types
   text =  tiny function that can map the text to these types, much like cited above
  ~H"""
    <.whatever_icon class={"#{color} #{@class}"}>{text}</.thing>
  """
end
```

anthonyshull: Yeah, I can't edit a highlight.

anthonyshull: I just want to be more clear because I think people are confused. You use pattern matching to create a function for each unique case you'll encounter. We have 'bus', 'massport_shuttle', 'mattapan_line', then generic atom with three words and generic atom with two words where initials and color are in the atom.

```
def mode_icon(%{type: :bus} = assigns) do
  ~H"""
  <.badge type="square" color="yellow" class="text-yellow-800 rounded-sm">
    <%= @route_name %>
  </.badge>
  """
end

def mode_icon(%{type: :massport_shuttle} = assigns) do
 ...
end

def mode_icon(%{type: :mattapan_line} = assigns) do
 ...
end 

# other cases that don't fit the general pattern

def mode_icon(%{type: type} = assigns) do
  ~H"""
  <.badge color={type_to_color(type)} type="circle"><%= type_to_initials(type) %></.badge>
  """
end

defp type_to_color(type) do
  type
  |> Atom.to_string()
  |> String.split("_")
  |> List.first()
end

defp type_to_initials(type) do
  type
  |> Atom.to_string()
  |> String.split("_")
  |> Enum.map(&String.at(&1, 0))
  |> Enum.map(&String.upcase/1)
  |> letters_to_initials()
end

defp letters_to_initials([a, b, c]), do: c

defp letters_to_initials(letters), do: Enum.join(letters)
```

anthonyshull: It should take care of all cases with less, more expressive code. It eliminates the need for the other various components, as it just uses the ones already in the metro. I was hoping to communicate this without having to type out the entire module 🤷

lib/dotcom_web/components/mode_icons.ex
```diff
@@ -0,0 +1,154 @@
...
+        "WO" -> "fill-logan-express-wo"
+        _ -> ""
+      end
+
+    assigns = assign(assigns, :fill_color_class, fill_color_class)
+
+    ~H"""
+    <.icon class={"w-6 #{@fill_color_class} #{@class}"} type="icon-svg" name="icon-logan-express" />
+    """
+  end
```
anthonyshull, line 118:
```
// This could and probably should live where routes are defined
@logan_express_routes ~w[BB BT DV WO]

def logan_express_routes, do: @logan_express_routes

// Then in this module

@logan_express_routes Routes.logan_express_routes()

def mode_icon(%{route_name: route_name, type: :logan_express} = assigns) when route_name in @logan_express_routes do
  ~H"""
  <.icon class={"w-6 fill-logan-express-#{String.downcase(@route_name}"} ... />  
  """
end
```

Makes it much easier to test because you can get a list of routes.

Screenshot Capture - 2024-11-06 - 16-53-29

joshlarson: I like that, especially because then there's one (maybe two) places that need to be updated if we ever get another Logan Express route.

The one thing that's throwing me off is that tailwind seems to only reliably recognize `fill-logan-express-#{String.downcase(@route_name}` as a valid utility class when `fill-logan-express-bb` appears explicitly in the code - the string interpolation thing sometimes worked for me and sometimes didn't (I ran into this problem when I tried string-interpolating the subway line colors into `bg-#{color}-500` as well). Not knowing 100% about how tailwind works under the hood, I guessed that it was scanning the code for usages of the various utility classes, and only generating the ones that it thought were needed, which left out ones that were programmatically generated, but maybe that's not actually how it works?

Do you know how tailwind does this? And/or how to convince tailwind to just generate the utility classes we need?

lib/dotcom_web/components/trip_planner/itinerary_group.ex
joshlarson: Heh - side benefit - I noticed two import statements that I was able to combine.

lib/dotcom_web/components/mode_icons.ex
anthonyshull: https://github.com/mbta/mbta_metro/blob/main/assets/js/styles/safelist.js

lib/dotcom_web/components/components.ex
```diff
@@ -71,4 +76,92 @@ defmodule DotcomWeb.Components do
...
+    <.icon type="icon-svg" name={"icon-massport-#{@route_number}"} class="w-6 h-6" />
+    """
+  end
+
+  defp route_svg(%{route: %Route{external_agency_name: "Logan Express", name: name}} = assigns)
+       when name in ["BB", "BT", "DV", "FH", "WO"] do
+    ~H"""
+    <.icon type="icon-svg" name={"icon-logan-express-#{@route.name}"} class="w-6 h-6" />
+    """
+  end
```
thecristen, line 150: These match the logic from `DotcomWeb.ViewHelpers.line_icon/2`

assets/tailwind.config.js
```diff
@@ -29,8 +29,18 @@ module.exports = {
...
+        "logan-express-bt": "#06c",
+        "logan-express-dv": "#704c9f",
+        "logan-express-fh": "#e81d2d",
+        "logan-express-wo": "#00954c",
```
anthonyshull, line 42: Can you remove these since you just generated an icon for each color?

lib/dotcom_web/components/components.ex
```diff
@@ -71,4 +81,145 @@ defmodule DotcomWeb.Components do
...
+
+  ```elixir
+  <.route_symbol route={%Routes.Route{id: "Red"}} />
+  <.route_symbol route={%Routes.Route{type: 3}} />
+  ```
+  """
+  def route_symbol(assigns) do
+    ~H"""
+    <span class="leading-[0]">
+      <%= if @route.type == 3 and is_nil(@route.external_agency_name) do %>
```
anthonyshull, line 103: 💡 This could be done with pattern matching.

lib/dotcom_web/components/components.ex
```diff
@@ -71,4 +81,145 @@ defmodule DotcomWeb.Components do
...
+    """
+  end
+
+  defp bus_pill(assigns) do
+    assigns =
+      assigns
+      |> assign_new(:class, fn -> "" end)
+      |> assign_new(:color_class, fn ->
+        cond do
+          Routes.Route.silver_line?(assigns.route) ->
```
anthonyshull, line 119: 💡 This could be done with pattern matching.

lib/routes/route.ex
```diff
@@ -85,9 +85,16 @@ defmodule Routes.Route do
   @type subway_lines_type :: :orange_line | :red_line | :green_line | :blue_line | :mattapan_line
   @type branch_name :: String.t() | nil
 
+  # Icons we know we have SVGs for, modify if new icons are added/removed
+  # These names are equivalent to the route names from the Massport GTFS
+  @logan_express_icon_names ["BB", "BT", "DV", "FH", "WO"]
+  @massport_icon_names ["11", "22", "33", "44", "55", "66", "77", "88", "99"]
```
anthonyshull, line 91: 💡 There is probably a better name for these than `*_icon_names` as they have to represent something more than the fact that we happen to have icons for them. `@logan_express_routes` and `@massport_routes` is probably more domain-driven.

lib/routes/route.ex
```diff
@@ -85,9 +85,16 @@ defmodule Routes.Route do
...
   @silver_line ~w(741 742 743 746 749 751)
   @silver_line_set MapSet.new(@silver_line)
 
+  def logan_express_icon_names, do: @logan_express_icon_names
+  def massport_icon_names, do: @massport_icon_names
```
anthonyshull, line 96: 💡 Same as above, I think we would want to be able to get all of the logan express routes regardless of their intended use.

lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -0,0 +1,145 @@
...
+
+  defp unique_leg_to_tuple(%Leg{mode: %PersonalDetail{}} = leg) do
+    {:WALK, leg.from.name, leg.to.name}
+  end
+
+  defp unique_leg_to_tuple(%Leg{mode: %{route: route}} = leg) do
+    {Routes.Route.type_atom(route.type), leg.from.name, leg.to.name}
+  end
+
+  defp group_departures({_hash, grouped_itineraries}) do
```
anthonyshull, line 44: 💡 This function is doing more than one thing so should probably be split into two.

lib/dotcom_web/components/gtfs.ex
```diff
@@ -0,0 +1,188 @@
...
+  def route_symbol(
+        %{
+          route: %Route{
+            description: description,
+            external_agency_name: nil,
+            type: 3
+          }
+        } = assigns
+      ) do
+    if description in [:rail_replacement_bus, :rapid_transit] do
```
anthonyshull, line 60: 💡 Could be done with pattern matching.

lib/dotcom_web.ex
```diff
@@ -128,8 +128,9 @@ defmodule DotcomWeb do
       use Gettext, backend: DotcomWeb.Gettext
       use MbtaMetro
 
+      import DotcomWeb.{Components, Components.Gtfs, ErrorHelpers}
```
joshlarson, line 131: Nit: I think this wants to replace the line below it.

lib/dotcom_web.ex
```diff
@@ -128,8 +128,9 @@ defmodule DotcomWeb do
...
       import DotcomWeb.{Components, ErrorHelpers}
-      import Phoenix.{HTML, LiveView.Helpers}
+      import Phoenix.{HTML, LiveView.Helpers, View}
```
joshlarson, line 133: Is this intended to be part of this PR?

lib/routes/route.ex
joshlarson: I see that these were moved from elsewhere and otherwise unchanged, but if it's not too much effort, I do agree with this comment - these aren't just the icon names - it's a comprehensive list of the routes.

IMO this is non-blocking - happy to see this fixed either in this PR or in a follow-up.

lib/dotcom_web/components/gtfs.ex
```diff
@@ -0,0 +1,188 @@
+defmodule DotcomWeb.Components.Gtfs do
```
joshlarson, line 1: As long as this yields the correct icons, then this comment is non-blocking, but...

If I'm reading this correctly, doesn't this re-implement a lot of `&Routes.Route.icon_atom/1`?

It strikes me as though most of the logic in this file might be easier to follow if it passed the route into `icon_atom`, and then used the result to actually render the icon. I _think_ that would make the pattern-matching easier.


---

Unrelated, but another comment at this same code-location: I don't think I like the name `Gtfs` for this module. While routes (and route icons/symbols) are related to GTFS, I think that they are fundamentally different concepts. GTFS is the data format used to transit route data, while routes are their own thing, whether communicated through GTFS, the V3 API, or carrier pigeon. 

Also, related to our chat earlier today, while I do sort of get why Phoenix crams all of its "core" components into one `core_components.ex` file, I don't think it's necessary to do that for the rest of our stuff. If it's easier to manage by having a separate file per component, or a separate file per class-of-component, or whatever, then I'm all for separating things out. All that to say - I don't think we need these components to live under a wide umbrella for right now.

What about calling this module `RouteSymbols` or `RouteIcons`?

lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -0,0 +1,145 @@
+defmodule Dotcom.TripPlan.ItineraryGroups do
```
joshlarson, line 1: Ooooh thanks for the file-rename! It's been bugging me that the file name and the module name were mis-matched!

lib/dotcom_web.ex
thecristen: it came along when i merged main 🫠

lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -0,0 +1,145 @@
...
+  defp unique_leg_to_tuple(%Leg{mode: %PersonalDetail{}} = leg) do
+    {:WALK, leg.from.name, leg.to.name}
+  end
+
+  defp unique_leg_to_tuple(%Leg{mode: %{route: route}} = leg) do
+    {Routes.Route.type_atom(route.type), leg.from.name, leg.to.name}
+  end
+
+  defp group_departures({_hash, grouped_itineraries}) do
+    summarized_legs =
```
joshlarson, line 45: This set of chained functions goes through a few different layers of abstraction, which makes it a little harder to follow. I wonder if it would make sense to extract everything before `remove_short_intermediate_walks()`, so that the whole thing reads like:

```elixir
summarized_legs =
  grouped_itineraries
  |> group_legs()
  |> aggregate_cost_and_walk_distance() # or maybe aggregate_leg_stats()
  |> remove_short_intermediate_walks()
```

Not saying that that's exactly the right way to group it, and I'm not sold on my naming choices, but that way, we keep a consistent level of abstraction, which lets it read more like a high-level description of what's happening.

lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -0,0 +1,145 @@
...
+      grouped_itineraries
+      |> Enum.map(& &1.legs)
+      |> Enum.zip_with(&Function.identity/1)
```
joshlarson, line 48:
```suggestion
      |> Enum.zip()
```

I might be misunderstanding the [Enum.zip/1 docs](https://hexdocs.pm/elixir/1.12/Enum.html#zip/1), but I think this would do the same thing?

lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -0,0 +1,145 @@
...
+
+    %{itineraries: ItineraryTag.sort_tagged(grouped_itineraries), summary: summary}
+  end
+
+  defp remove_short_intermediate_walks(summarized_legs) do
+    summarized_legs
+    |> Enum.with_index()
+    |> Enum.reject(fn {leg, index} ->
+      index > 0 && index < Kernel.length(summarized_legs) - 1 &&
+        (leg.routes == [] && leg.walk_minutes < 5)
```
joshlarson, line 78: Nit (non-blocking):
```suggestion
        (leg.routes == [] && leg.walk_minutes < @short_walk_threshold)
```

So that 5 minutes isn't a magic value hidden away in the middle, but displayed prominently at the top.

I could even see a justification for making it an env variable, but I think that's overkill.

lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -0,0 +1,145 @@
...
+    end)
+    |> Enum.map(&elem(&1, 0))
```
joshlarson, line 80: Nit (non-blocking):

```suggestion
    |> Enum.map(fn {thing, _} -> thing end)
```

Toooooootally a stylistic thing on my end, but I prefer the explicitness of `{thing, _}` over calling `elem`.

joshlarson: Uhhhh I trust your judgment to come up with a better name than `thing`! 😅

lib/dotcom_web/components/gtfs.ex
joshlarson: I agree with this, although per my comment about about possibly using `&Routes.Route.icon_atom/1`, I think the optimal pattern-matching might not be a one-to-one match with the code that's here.

lib/dotcom_web/components/trip_planner/itinerary_detail.ex
```diff
@@ -0,0 +1,41 @@
...
+  """
+  use DotcomWeb, :component
+
+  import DotcomWeb.Components.TripPlanner.Leg
+
+  alias Dotcom.TripPlan.PersonalDetail
+
+  def itinerary_detail(assigns) do
+    assigns =
+      assign(assigns, :accordion_id, Ecto.UUID.generate())
```
joshlarson, line 13: Is the `accordion_id` assign used?

thecristen: whoops, not anymore!

lib/dotcom_web/components/trip_planner/itinerary_group.ex
thecristen: I did change this slightly but not much. :D I still think it's okay here for now.

lib/routes/route.ex
thecristen: FWIW they're not comprehensive. Comprehensive to me would involve having the Massport GTFS routes.txt involved. These lists _are_ noting the ones we have SVG icons created for and do indeed come from a prior change.

lib/dotcom_web/components/route_symbols.ex
```diff
@@ -0,0 +1,167 @@
...
+  use Phoenix.Component
+
+  import MbtaMetro.Components.Icon
+
+  alias Routes.Route
+
+  @logan_express_icon_names Route.logan_express_icon_names()
+  @massport_icon_names Route.massport_icon_names()
+
+  defguardp is_external_route?(assigns) when not is_nil(assigns.route.external_agency_name)
```
joshlarson, line 15: Whoa - I didn't know you could do this!

lib/dotcom/trip_plan/itinerary_groups.ex
thecristen: Huh, interesting - almost the same!

```elixir
iex(1)> [[1, 2, 3],[:a, :b, :c]] |> Enum.zip()
[{1, :a}, {2, :b}, {3, :c}]
iex(2)> [[1, 2, 3],[:a, :b, :c]] |> Enum.zip_with(&Function.identity/1)
[[1, :a], [2, :b], [3, :c]]
```

I think I'll keep it the way it is since it saves me the step of converting the tuples to enmuerables.

priv/static/icon-svg/icon-logan-express.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 8.5 8.5">
```
joshlarson, line 1: I personally prefer the approach implied by adding this icon (the approach where we use this icon and just render it in different colors depending on the route), but it looks like the actual implementation isn't doing it that way.

Based on that, do we still need to include this icon?

thecristen: 😅 good catch, we do not need to include it in that case

lib/dotcom_web/components/trip_planner/itinerary_group.ex
```diff
@@ -2,47 +2,146 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryGroup do
...
+      @class
+    ]}>
+      <.icon name="person-walking" class="h-4 w-4" />
+      <span><%= @walk_minutes %> min</span>
+    </span>
+    """
+  end
+
+  # Group of commuter rail routes are summarized to one symbol.
+  defp leg_icon(%{routes: [%Routes.Route{type: 2} | _]} = assigns) do
```
joshlarson, line 92: Non-blocking, partly because I think my ideal outcome here might be a fairly large restructuring of how routes and modes are represented in dotcom (which is _definitely_ too much scope for this PR), but I feel like there are a lot of places where I have to memorize what `type: 2`/`type: 4`/etc mean in order to understand our code.

lib/dotcom_web/components/trip_planner/itinerary_group.ex
```diff
@@ -2,47 +2,146 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryGroup do
...
+       ) do
+    slashed? = type == 3 && is_nil(agency)
+
+    assigns =
+      assigns
+      |> assign(:slashed?, slashed?)
+      |> assign(
+        :grouped_classes,
+        if(slashed?,
+          do: "[&:not(:first-child)]:rounded-l-none [&:not(:last-child)]:rounded-r-none !px-3",
```
joshlarson, line 116: Screenshot 2024-11-15 at 5 06 22 PM
Screenshot 2024-11-15 at 5 15 06 PM


Something is going wonky with respect to the spacing. It looks like for whatever reason, the wrapping `div` thinks that each bus pill is a bit larger than it is (hence the fact that the weird empty space to the right is bigger when there are three routes versus when there are two).

What's especially weird is that it looks like this is only happening in Firefox. It looks great in Chrome.

thecristen (approved):

PR #2206, opened 2024-10-29
fix: Have vanity links point to dotcom, rather than CDN

joshlarson:
This should fix this problem 👇 on https://www.mbta.com/schedules/subway.

Screenshot 2024-10-29 at 10 46 48 AM

### DIFF
--- a/lib/dotcom/map_helpers.ex
+++ b/lib/dotcom/map_helpers.ex
@@ -55,14 +55,14 @@ defmodule Dotcom.MapHelpers do
 
   @spec image(atom) :: String.t()
   def image(:subway) do
-    static_url(DotcomWeb.Endpoint, "/subway-map-image")
+    "/subway-map-image"
   end
 
   def image(:commuter_rail) do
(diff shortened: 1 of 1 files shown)


lib/dotcom/map_helpers.ex
```diff
@@ -55,14 +55,14 @@ defmodule Dotcom.MapHelpers do
 
   @spec image(atom) :: String.t()
   def image(:subway) do
-    static_url(DotcomWeb.Endpoint, "/subway-map-image")
+    "/subway-map-image"
```
joshlarson, line 58: I didn't use the `~p` sigils because those come in through `DotcomWeb`'s `view_helpers` macro, which generated conflicting function names with something else that's imported here.

I figured it wasn't a huge deal because `~p` sigils' main benefit is to let you know if you're using a path that isn't defined in `Router`, but we [match on all paths anyway](https://github.com/mbta/dotcom/blob/c892bd12d44cab84b7696d58729e7c64f5015cff/lib/dotcom_web/router.ex#L306).

thecristen (approved):

PR #2212, opened 2024-11-04
Switch away from using npm for metro

anthonyshull:
We don't have an organizational npm account so it will probably be easier and faster in the short-term to not publish metro's JS via npm. That means dotcom has to make some changes to accommodate.


.github/workflows/tests.yml
```diff
@@ -101,6 +101,7 @@ jobs:
             priv
             react_renderer/dist/app.js
           key: ci-application-cache-${{ github.sha }}
+      - run: npm install --prefix assets -S -install-links deps/mbta_metro/priv/
```
anthonyshull, line 104: 📓 This is necessary to install mbta_metro from a path, but getting its deps and putting them in `node_modules`.

deploy/dotcom/prod/Dockerfile
```diff
@@ -37,8 +37,14 @@ COPY --from=elixir-builder /root/deps /root/deps
...
 ADD ./assets /root/assets
 
+RUN npm i -g husky@latest
+
 WORKDIR /root/assets
+
+RUN npm install --package-lock-only
 RUN npm ci --ignore-scripts
+RUN npm install -S -install-links ../deps/mbta_metro/priv/
+
```
anthonyshull, line 47: 📓 This just does all three of those things in the build process.

thecristen: cool. we should probably create an alias to group these commands at some point

thecristen (approved):
No blockers here!

PR #2216, opened 2024-11-07
feat(Plugs.SecureHeaders): add secure headers to errors and static files

thecristen:
#### Summary of changes

**Asana Ticket:** [CSP | Web Application Security](https://app.asana.com/0/555089885850811/1208252819442819/f) / [CSP | Web Application Headers](https://app.asana.com/0/555089885850811/1208252819442827/f)

While we were pretty good about returning most secure headers generally, they were totally absent for error pages and static files. This fixes that.


lib/dotcom_web/plugs/secure_headers.ex
```diff
@@ -0,0 +1,63 @@
...
+      "img-src 'self' cdn.mbta.com px.ads.linkedin.com www.linkedin.com www.facebook.com *.google.com *.googleapis.com *.gstatic.com *.s3.amazonaws.com data: i.ytimg.com www.googletagmanager.com *.arcgis.com",
+    style: "style-src 'self' 'unsafe-inline' www.gstatic.com cdn.jsdelivr.net",
+    script:
+      "script-src 'self' 'unsafe-eval' 'unsafe-inline' insitez.blob.core.windows.net snap.licdn.com connect.facebook.net www.instagram.com www.google-analytics.com *.google.com www.gstatic.com www.googletagmanager.com *.googleapis.com data.mbta.com *.arcgis.com",
+    font: "font-src 'self'",
+    connect:
+      "connect-src 'self' *.googleapis.com analytics.google.com www.google-analytics.com www.google.com px.ads.linkedin.com stats.g.doubleclick.net *.arcgis.com *.s3.amazonaws.com",
+    frame:
+      "frame-src 'self' data.mbta.com www.youtube.com www.google.com cdn.knightlab.com livestream.com www.instagram.com *.arcgis.com",
+    worker: "worker-src blob: ;"
```
anthonyshull, line 21: Just a suggestion, but since we're cleaning this up, is there a better way to make this manageable? Thinking something like...
```
@base_img_directives ~w[
  img-src
  'self'
  ...
]
```

We can put all of the different types into these all in alphabetical order. Then:

```
@base_csp_directives [
  img: @base_image_directives,
  ...
]
```

config/runtime.exs
```diff
@@ -144,6 +144,23 @@ end
...
+    end
+
+  config :dotcom,
+         :content_security_policy_definition,
+         DotcomWeb.Plug.SecureHeaders.build_csp_directives(
+           img: "#{System.get_env("STATIC_HOST", "")} #{System.get_env("CMS_API_BASE_URL", "")}",
+           style: "#{System.get_env("STATIC_HOST", "")}",
+           script: "#{System.get_env("STATIC_HOST", "")}",
+           font: "#{System.get_env("STATIC_HOST", "")}",
+           connect: "wss://#{host} #{sentry_dsn_host || ""}"
```
anthonyshull, line 161: This would probably be easier if you make @base_img_directives a list and then export it via base_image_directives/0 and then make a function that just takes a keyword list of lists of strings:

[
  img: base_image_directives() ++ [System.get_env("STATIC_HOST", ""), System.get_env("CMS_API_BASE_URL", "")],
  ...
]

lib/dotcom_web/plugs/secure_headers.ex
anthonyshull: Would be nice to alphabetize too since these lists are long and have a lot of overlap...multiple references to google.

thecristen: There's lots of overlap between the directives and somehow lots of differences too, i'll give it a crack!

thecristen: didn't do much to account for overlaps since they're so inconsistent, but just alphabetizing and formatting with newlines gave this a big readability win!

anthonyshull (approved):

PR #2217, opened 2024-11-07
update package-lock to reflect metro deps

thecristen:
I ran `npm --prefix assets install` and got some updates in the package-lock.json! Does this seem legit?


anthonyshull (approved):
This seems right. I think this is a one time issue to do with changing the way that metro was included. I don't think this got flagged as necessary because the automated builds do this.

PR #2218, opened 2024-11-08
Take trip planner query params into account in the new trip planner

anthonyshull:
The time selection is a little trickier because of the change and hide/show behavior. But, we decided to just set it to 'now' since links generated in the past won't work anyway. All other selections are reflected. You can test this by planning a trip with selections and then just switching to `/preview/trip-planner`.


thecristen (changes requested):
Appreciate how smoothly this works! I have some ideas how we can extend this to handle the `/trip-planner/to/Special+Location` use cases too. ;)
Let's add a cursory test for the new module.

thecristen (approved):

PR #2223, opened 2024-11-14
chore: add svg img roles, improve title tags

thecristen:
#### Summary of changes

**Asana Ticket:** No ticket, but noticed some `<title>` tags weren't a nice as they _could_ be while using a few of these SVG files for the `<.route_symbol />` component. So, this PR:

- Re-cases the `<title>` tag content to sentence/title case.
- Verifies the presence of `role="img"` on all the SVGs, in alignment with 'Pattern 5' in this guidance: [Accessible SVGs: Perfect Patterns For Screen Reader Users — Smashing Magazine](https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/#best-in-show-1)

Additional optimizations because these are all inlined:
- they SHOULD NOT have xml preambles. apparently those belong at the start of a document and there shouldn't be more than one of these in a given page (https://stackoverflow.com/questions/38169475/svg-in-html5-when-is-xml-declaration-xml-version-1-0-encoding-utf-8#comment63771563_38172170)
- xmlns/version/similar attributes on the SVG are not required. so goodbye!


priv/static/icon-svg/403.svg
```diff
@@ -1,6 +1,5 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<svg viewBox="0 0 759 740" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
joshlarson, line 2: I confess that this is an area of non-expertise for me, but both the Github diff viewer and Chrome (which my computer has decided is the default tool for viewing SVG files) are telling me that something is invalid about a bunch of these icons.

When I open the icon itself in Chrome, I get this:
```
error on line 122 at column 34: Namespace prefix xlink for href on use is not defined
```

Adding the `xmlns` and `xmlns:xlink` back in fixes that.

It renders just fine in-context in both Firefox and Chrome though 🤯

thecristen: well that's silly lol. well i may as well put them back then...

priv/static/icon-svg/green-line-b.svg
```diff
@@ -1,8 +1,5 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="24px" height="24px" role="img" version="1.1" viewBox="0 0 24 24">
-    <title>
-        green line B
-    </title>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="24px" height="24px" role="img" viewBox="0 0 24 24">
```
joshlarson, line 1: Screenshot 2024-11-18 at 10 14 06 AM

Definitely out of scope for this PR, but I would be pretty surprised if we're actually using all of icons in this folder. Maybe at some point it would be worth going through and identifying which icons we still need to keep?

thecristen: Yep. Added to the dev discussion list!

joshlarson (approved):
Oh oops - I didn't realize I hadn't approved this already. Looks good!

PR #2228, opened 2024-11-18
storybook: Add storybook and stories for route symbols

joshlarson:
Can be accessed by going to http://localhost:4001/storybook, and using the same basic auth credentials that you use to access the admin page.


lib/dotcom_web/router.ex
```diff
@@ -314,6 +316,16 @@ defmodule DotcomWeb.Router do
     get("/about_the_mbta/news_events", Redirector, to: "/news")
   end
 
+  scope "/" do
+    storybook_assets()
+  end
+
+  scope "/", DotcomWeb do
+    # pipe_through([:secure, :browser, :browser_live, :basic_auth])
```
thecristen, line 324: Did you mean to use this line?

config/config.exs
```diff
@@ -37,6 +37,26 @@ config :sentry,
 
 config :mbta_metro, custom_icons: ["#{File.cwd!()}/priv/static/icon-svg/*"]
 
+config :esbuild,
+  version: "0.17.11",
+  default: [
+    # args: ~w(js/app.js js/storybook.js --bundle --target=es2017 --outdir=../priv/static/assets),
```
thecristen, line 43: Did you want to delete this?

lib/dotcom_web/router.ex
thecristen: In any case just wanted to flag that this is _not_ behind basic auth!

joshlarson:
Closing this because I don't feel like dealing with the deployment-CSS issues for right now.

Will probably either:
- Reopen this with a better strategy for dealing with deployment-CSS, OR
- Open a different one with a different approach (likely a home-rolled storybook simulacrum that just uses our regular CSS assets by default)

PR #2229, opened 2024-11-18
fix(PredictionsChannel): filter out past predictions

thecristen:
#### Summary of changes

**Asana Ticket:** [Fix: Extraneous <1 minute predictions appearing on Stop Pages](https://app.asana.com/0/555089885850811/1208505215613687/f)

Some facts:
- because the backend endpoint that gives the stop page its schedules is cached, the frontend component refreshes every 15s, filtering out the past schedules. 
  - these get merged with the predictions. but plenty of predictions have no corresponding schedule. 
  - and apparently there's no frontend mechanism to filter out stale predictions
- in the backend, past predictions get culled out of our in-memory storage every 5 minutes (gotta keep the ETS table size low)
- the streaming connection we have to the V3 API tells us when to add/remove/update/reset our predictions

Filtering the predictions here in `DotcomWeb.PredictionsChannel`, right before we publish the predictions to stop pages, seems to work ok. My guess is that this logic wasn't included previously because we thought (a) we'd get timely "remove" events from the streaming predictions or (b) something else in our logic would filter these out before making it to the UI.

### DIFF
--- a/test/dotcom_web/channels/predictions_channel_test.exs
+++ b/test/dotcom_web/channels/predictions_channel_test.exs
@@ -86,6 +86,31 @@ defmodule DotcomWeb.PredictionsChannelTest do
       # Verify
       assert_push("data", %{predictions: ^predictions})
     end
+
+    test "filters out past predictions", context do
+      # Setup
+      now = Timex.now() |> Timex.shift(seconds: 1)
+      past = Timex.shift(now, seconds: -15)
+      future = Timex.shift(now, seconds: 15)
+
+      predictions =
+        [past, now, future]
+        |> Enum.map(&Prediction.build(:canonical_prediction, %{departure_time: &1}))
+
+      expect(@predictions_pub_sub, :subscribe, fn _ ->
+        predictions
+      end)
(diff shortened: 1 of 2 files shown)


test/dotcom_web/channels/predictions_channel_test.exs
```diff
@@ -86,6 +86,31 @@ defmodule DotcomWeb.PredictionsChannelTest do
...
+      expect(@predictions_pub_sub, :subscribe, fn _ ->
+        predictions
+      end)
+
+      {:ok, _, socket} = subscribe_and_join(context.socket, PredictionsChannel, context.channel)
+
+      # Exercise
+      PredictionsChannel.handle_info({:new_predictions, predictions}, socket)
+
+      [_past_prediction | expected_predictions] = predictions
```
joshlarson, line 109:
```suggestion
      expected_predictions = [now, future]
```

Not a huge deal, but I'm more in favor of being super explicit about exactly what we expect, rather than having to reason about what's in the `predictions` array.

PR #2230, opened 2024-11-19
chore: add integrity attribute to external links

thecristen:
Another try at #2215 but with infrastructure changes to the CDN for dev-blue and dev-green enabled. Let's test it!


thecristen:
replaced by #2236

PR #2232, opened 2024-11-21
feat: Toggle between summary and details view in itinerary panel

joshlarson:
https://github.com/user-attachments/assets/a3551943-cac9-4cf2-9590-56bd80954239

#### Summary of changes

**Asana Ticket:** [Trip Planner Preview | Toggle between list and details views](https://app.asana.com/0/555089885850811/1208788271277295/f)

### DIFF
--- a/lib/dotcom_web/components/live_components/trip_planner_results_section.ex
+++ b/lib/dotcom_web/components/live_components/trip_planner_results_section.ex
@@ -62,6 +62,7 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
     <.itinerary_group
       :for={{result, index} <- Enum.with_index(@results)}
       index={index}
+      details_click_event="show_itinerary_details"
       target={@target}
       {result}
     />
--- a/lib/dotcom_web/components/live_components/trip_planner_results_section.ex
+++ b/lib/dotcom_web/components/live_components/trip_planner_results_section.ex
@@ -41,23 +41,7 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
     """
   end
 
-  defp itinerary_panel(%{results: results, details_index: details_index} = assigns) do
-    case details_index do
-      nil ->
-        ~H"""
-        <.itinerary_panel_with_all_results results={@results} target={@target} />
-        """
-
-      _ ->
-        assigns = assign(assigns, :result, Enum.at(results, details_index))
-
-        ~H"""
-        <.itinerary_panel_with_specific_result result={@result} target={@target} />
-        """
-    end
-  end
-
-  defp itinerary_panel_with_all_results(assigns) do
+  defp itinerary_panel(%{details_index: nil} = assigns) do
     ~H"""
     <.itinerary_group
       :for={{result, index} <- Enum.with_index(@results)}
@@ -69,8 +53,9 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
     """
   end
 
-  defp itinerary_panel_with_specific_result(%{result: result} = assigns) do
-    assigns = assign(assigns, :itineraries, Map.get(result, :itineraries))
+  defp itinerary_panel(%{results: results, details_index: details_index} = assigns) do
+    assigns =
+      assign(assigns, :itineraries, results |> Enum.at(details_index) |> Map.get(:itineraries))
 
     ~H"""
     <div class="mt-30">
--- a/lib/dotcom_web/components/live_components/trip_planner_results_section.ex
+++ b/lib/dotcom_web/components/live_components/trip_planner_results_section.ex
@@ -9,8 +9,9 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
   import DotcomWeb.Components.TripPlanner.ItineraryDetail
   import DotcomWeb.Components.TripPlanner.ItineraryGroup, only: [itinerary_group: 1]
 
-  def mount(_params, _session, socket) do
-    {:ok, socket}
+  @impl true
+  def mount(socket) do
+    {:ok, socket |> assign(:expanded_itinerary_index, nil)}
   end
 
   @impl true
@@ -22,7 +23,11 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
       </div>
       <.async_result :let={results} assign={@results}>
         <div :if={results} class="w-full p-4">
-          <.itinerary_panel results={results} details_index={@itinerary_details_index} />
+          <.itinerary_panel
+            results={results}
+            details_index={@expanded_itinerary_index}
+            target={@myself}
+          />
         </div>
       </.async_result>
       <.live_component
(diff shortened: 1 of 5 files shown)


lib/dotcom_web/live/trip_planner.ex
```diff
@@ -88,7 +90,65 @@ defmodule DotcomWeb.Live.TripPlanner do
     """
   end
 
+  defp itinerary_panel(%{details_index: details_index} = assigns) do
```
joshlarson, line 93: I think I want to pull `itinerary_panel` out into a separate live component, and have _it_ handle the summary versus details toggling, rather than having that logic handled in the top-level `TripPlanner` live.

The one complicating factor is the map, since on mobile the map gets rendered in the middle of the details panel. That complicating factor is why I left it like this in this PR, but I sort of do want to figure out at least some strategy for this before merging.

lib/dotcom_web/components/trip_planner/itinerary_group.ex
```diff
@@ -60,13 +58,15 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryGroup do
           Similar trips depart at <%= Enum.map(@summary.next_starts, &format_datetime_short/1)
           |> Enum.join(", ") %>
         </div>
-        <button class="btn-link font-semibold underline" phx-click={JS.toggle(to: "##{@group_id}")}>
+        <button
+          class="btn-link font-semibold underline"
+          phx-click="show_itinerary_details"
```
joshlarson, line 63: In react-land, this would be a callback that would be explicitly passed into this component - `onDetailsClick` or something like that. I feel like we need _some_ analogous strategy for that here, because I really don't like having `"show_itinerary_details"` being hard-coded in both this file and `trip_planner.ex`.

Do you have any suggestions? I can also experiment before merging, maybe by passing in a `phx-details-click` `assign` or something like that?

lib/dotcom_web/live/trip_planner.ex
```diff
@@ -88,7 +90,65 @@ defmodule DotcomWeb.Live.TripPlanner do
...
+        </p>
+      </button>
+      <.itinerary_detail :for={itinerary <- @itineraries} itinerary={itinerary} />
+    </div>
+    """
+  end
+
   @impl true
+  def handle_event("show_itinerary_details" = event, %{"index" => index_str} = params, socket) do
+    dbg(event)
```
joshlarson, line 140: Oops

lib/dotcom_web/components/trip_planner/itinerary_group.ex
thecristen: ha, I was going to link to this exact doc! can totally do that.

joshlarson: I didn't end up passing in a callback because `itinerary_group` isn't a live component (pretty sure only live components can do anything meaningful with callbacks?). I did pass `details_click_event` and `target` in as assigns, which sort of accomplishes the same goal.

lib/dotcom_web/live/trip_planner.ex
joshlarson: I wound up pulling out `TripPlannerResultsSection` instead, which includes the map. I think this is a passable solution for now, and might wind up being actually good? Interested to know what you think.

lib/dotcom_web/components/trip_planner/itinerary_group.ex
```diff
@@ -4,18 +4,24 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryGroup do
   """
   use DotcomWeb, :component
 
-  import DotcomWeb.Components.TripPlanner.ItineraryDetail
-
   attr(:summary, :map, doc: "ItineraryGroups.summary()", required: true)
   attr(:itineraries, :list, doc: "List of %Dotcom.TripPlan.Itinerary{}", required: true)
 
+  attr(:index, :integer,
```
thecristen, line 10: It took me a while to understand what this was for... ultimately I'd favor a more descriptive attribute name, maybe `:group_index`?

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
```diff
@@ -0,0 +1,99 @@
...
+  def handle_event("show_itinerary_details", %{"index" => index_str}, socket) do
+    {index, ""} = Integer.parse(index_str)
+
+    {:noreply, socket |> assign(:expanded_itinerary_index, index)}
+  end
+
+  @impl true
+  def handle_event("show_itinerary_summary", _params, socket) do
+    {:noreply, socket |> assign(:expanded_itinerary_index, nil)}
+  end
```
thecristen, line 98: nit: I wonder if these could be more explicitly related? They're setting and unsetting the same property... what if they use the same event and pass different params?

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
```diff
@@ -0,0 +1,99 @@
...
+        <.itinerary_panel_with_all_results results={@results} target={@target} />
+        """
+
+      _ ->
+        assigns = assign(assigns, :result, Enum.at(results, details_index))
+
+        ~H"""
+        <.itinerary_panel_with_specific_result result={@result} target={@target} />
+        """
+    end
```
thecristen, line 57: nit: I think this would read better without the intermediate `itinerary_panel_` components 🤔 
that and using pattern matching here, e.g.

```
defp itinerary_panel(%{details_index: nil} = assigns) do
  ~H"""
    <.itinerary_group
      :for={{result, index} <- Enum.with_index(@results)}
      index={index}
      details_click_event="show_itinerary_details"
      target={@target}
      {result}
    />
  """
end

defp itinerary_panel(assigns) do
  ~H"""
    <div class="mt-30">
      <button type="button" phx-click="show_itinerary_summary" phx-target={@target} class="btn-link">
        <p class="flex flex-row items-center">
          <.icon class="fill-brand-primary h-4 mr-2" name="chevron-left" />
          <span class="font-medium">View All Options</span>
        </p>
      </button>
      <.itinerary_detail :for={itinerary <- @results.itineraries} itinerary={itinerary} />
    </div>
  """
end
```

thecristen (approved):
Just a few minor comments!

lib/dotcom_web/components/trip_planner/itinerary_group.ex
joshlarson: Thinking for a moment... the only reason an itinerary group cares about its `index` is because it does _the right thing_ in response to clicking on `Details`. So maybe the thing to do is have setting `index` correctly be handled in a callback instead of implicitly through this param.

I think that would require converting `itinerary_group` to a live component though, so if we think that's a good idea, my vote would be to do that as a follow-up PR.

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
joshlarson: D'oh! Yes, that would be better!

lib/dotcom_web/components/trip_planner/itinerary_group.ex
thecristen: I think that would work fine (and agree that might need a live component), but it might be overkill? 

An alternate thought: the summary and itineraries here are populated by the output of `Dotcom.TripPlan.ItineraryGroups.from_itineraries/1`, which outputs a list of `%{itineraries: itineraries, summary: summary}` maps. Why not have this function compute the indexes and have the map include that `index` as well?

A more meta thought - we're using index to really mean "unique identifier". The past iteration made up a unique ID (`assign(assigns, :group_id, "group-#{:erlang.phash2(assigns.itineraries)}")`). We might want to reconsider using the index value for that purpose (what if we want to let end users re-sort the list later, for example).

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
joshlarson: [Like this comment](https://github.com/mbta/dotcom/pull/2232#discussion_r1857335348), I think this will also be taken care of by making this into a legit callback, which would require a live-component refactor, so I want to punt on it for now.

I agree with this though - this could be clearer.

PR #2237, opened 2024-11-26
feat(trip_planner): Use <.itinerary_summary /> in the details panel

joshlarson:
Extract 👇 to a re-usable component called `<.itinerary_summary />` and re-use it in the details view.

Screenshot 2024-11-26 at 2 07 25 PM


thecristen (approved):
✨

PR #2239, opened 2024-11-27
Revert "chore: convert root template to .heex (#2233)"

joshlarson:
This reverts commit e1e3dc1de91d1786ac8f1f60e2e8ae5d6f0bc0f0 and PR #2233. 

There are some infra changes that need to take effect before we can deploy this, so we'd rather have this wait and get merged once we can confidently release it to production.


thecristen (approved):
Yup that's the one

PR #2240, opened 2024-11-29
feat: Arrange map and plan results correctly on desktop and mobile

joshlarson:
# Mobile

## Before
Screenshot 2024-11-29 at 12 51 00 PM

Screenshot 2024-11-29 at 12 51 09 PM

## After
Screenshot 2024-11-29 at 12 47 43 PM
Screenshot 2024-11-29 at 12 48 03 PM

# Desktop

Looks pretty much the same

---

**Asana Ticket:** [Trip Planner Preview | Scaffold out the responsive mobile/desktop layout](https://app.asana.com/0/555089885850811/1208788271277297/f)

### DIFF
--- a/assets/css/_events.scss
+++ b/assets/css/_events.scss
@@ -181,10 +181,6 @@ $mobile-control-height: 73px;
   cursor: pointer;
 }
 
-.hidden {
-  display: none !important;
-}
-
 .m-event {
   padding: 16px;
   @include media-breakpoint-down(xs) {
--- a/lib/dotcom_web/components/live_components/trip_planner_results_section.ex
+++ b/lib/dotcom_web/components/live_components/trip_planner_results_section.ex
@@ -16,27 +16,61 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
 
   @impl true
   def render(assigns) do
+    grid_column_layout = "md:grid-cols-[1fr_1fr]"
+
+    grid_column_layout =
+      case assigns.results.result do
+        nil -> "md:grid-cols-[1fr]"
+        _ -> "md:grid-cols-[1fr_1fr]"
+      end
+
+    assigns = assign(assigns, :grid_column_layout, grid_column_layout)
+
     ~H"""
-    <section class="flex w-full border border-solid border-slate-400">
-      <div :if={@error} class="w-full p-4 text-rose-400">
-        <%= inspect(@error) %>
-      </div>
-      <.async_result :let={results} assign={@results}>
-        <div :if={results} class="w-full p-4">
-          <.itinerary_panel
-            results={results}
-            details_index={@expanded_itinerary_index}
-            target={@myself}
-          />
+    <div>
+      <section class={"flex flex-col md:grid #{@grid_column_layout} md:grid-rows-[min-content_1fr] w-full border border-solid border-slate-400"}>
+        <div :if={@error} class="w-full p-4 text-rose-400">
+          <%= inspect(@error) %>
         </div>
-      </.async_result>
-      <.map
-        map_config={@map_config}
-        from={@from}
-        to={@to}
-        hide_on_mobile={@expanded_itinerary_index == nil}
-      />
-    </section>
+        <.async_result :let={results} assign={@results}>
+          <div
+            :if={results && @expanded_itinerary_index}
+            class="row-start-1 col-start-1 h-min w-full p-4"
+          >
+            <button
+              type="button"
+              phx-click="set_expanded_itinerary_index"
+              phx-value-index="nil"
+              phx-target={@myself}
+              class="btn-link"
+            >
+              <span class="flex flex-row items-center">
+                <.icon class="fill-brand-primary h-4 mr-2" name="chevron-left" />
+                <span class="font-medium">View All Options</span>
+              </span>
+            </button>
+          </div>
+        </.async_result>
+
+        <.map
+          map_config={@map_config}
+          from={@from}
+          to={@to}
+          hide_on_mobile={@expanded_itinerary_index == nil}
+          class="row-span-2"
+        />
+
+        <.async_result :let={results} assign={@results}>
+          <div :if={results} class="w-full p-4 row-start-2 col-start-1">
+            <.itinerary_panel
+              results={results}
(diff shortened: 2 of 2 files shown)


assets/css/_events.scss
```diff
@@ -181,10 +181,6 @@ $mobile-control-height: 73px;
   cursor: pointer;
 }
 
-.hidden {
-  display: none !important;
```
joshlarson, line 185: This feels like it could be a big risky change, but also this utility was overriding the tailwind utility that does the same thing (except that the tailwind utility doesn't have `!important` anymore, since we removed that [here](https://github.com/mbta/dotcom/commit/3c6456cac09cd3359d0287de1bc3eeed07a26329#diff-52a43e64271620d979d25d597733b7930e6f3771ab30e71ec459bff2398e43dbR19)).

Interested in your thoughts on getting rid of this, and whether there's another way to get tailwind to conditionally give us `display: none` in the right settings.

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
```diff
@@ -16,28 +16,61 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
...
-      </div>
-      <.async_result :let={results} assign={@results}>
-        <div :if={results} class="w-full p-4">
-          <.itinerary_panel
-            results={results}
-            details_index={@expanded_itinerary_index}
-            target={@myself}
-          />
+    <div>
+      <section class={"flex flex-col md:grid #{@grid_column_layout} md:grid-rows-[min-content_1fr] w-full border border-solid border-slate-400"}>
```
joshlarson, line 31: I could also see a world in which this was a `grid` on both desktop and mobile, just with a different template, and different `row-start`/`row-span`/etc for the grid entries, but this felt easier as I was building it. This way, lexically the pieces are arranged according to the mobile design, and then on desktop, they get placed onto the grid instead, and their lexical ordering is ignored.

Open to feedback/ideas on this.

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
```diff
@@ -16,28 +16,61 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
...
-      <.live_component
-        module={MbtaMetro.Live.Map}
-        id="trip-planner-map"
-        class="h-96 w-full relative overflow-none"
-        config={@map_config}
-        pins={[@from, @to]}
-      />
-    </section>
+        <.async_result :let={results} assign={@results}>
+          <div
```
joshlarson, line 36: This added some extra spacing between `View all options` and the content below it. 

I'm currently leaning towards leaving it as-is for this PR, and fixing it, as well as a bunch of other minor visual issues as a follow-up.

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
```diff
@@ -16,27 +16,56 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
...
+      case assigns.results.result do
+        nil -> "md:grid-cols-[1fr]"
+        _ -> "md:grid-cols-[1fr_1fr]"
+      end
+
+    assigns = assign(assigns, :grid_column_layout, grid_column_layout)
+
     ~H"""
-    <section class="flex w-full border border-solid border-slate-400">
+    <section class={"flex flex-col md:grid #{@grid_column_layout} md:grid-rows-[min-content_1fr] w-full border border-solid border-slate-400"}>
```
joshlarson, line 28: I could also see a world in which this was a grid on both desktop and mobile, just with a different template, and different row-start/row-span/etc for the grid entries, but this felt easier as I was building it. This way, lexically the pieces are arranged according to the mobile design, and then on desktop, they get placed onto the grid instead, and their lexical ordering is ignored.

Open to feedback/ideas on this.

(This was originally [this comment](https://github.com/mbta/dotcom/pull/2240#discussion_r1863831461), but got outdated when I changed the spacing, and is still relevant)

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
```diff
@@ -16,27 +16,56 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
...
+
+      <.map
+        map_config={@map_config}
+        from={@from}
+        to={@to}
+        hide_on_mobile={@expanded_itinerary_index == nil}
+        class="row-span-2"
+      />
+
+      <.async_result :let={results} assign={@results}>
```
joshlarson, line 60: Something about how I re-arranged the DOM elements added some extra spacing between `View all options` and the content below it.

I'm currently leaning towards leaving it as-is for this PR, and fixing it, as well as a bunch of other minor visual issues as a follow-up.

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
```diff
@@ -16,27 +16,56 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerResultsSection do
...
+            </span>
+          </button>
+        </div>
+      </.async_result>
+
+      <.map
+        map_config={@map_config}
+        from={@from}
+        to={@to}
+        hide_on_mobile={@expanded_itinerary_index == nil}
```
anthonyshull, line 56: Instead of wrapping the map in another component and passing in this attribute, you could use the same ternary trick from above:

```elixir
~H"""
  <.live_component
    module={MbtaMetro.Live.Map}
    id="trip-planner-map"
    class={[
      "h-64 md:h-96 w-full relative overflow-none",
      @expanded_itinerary_index == nil && "hidden md:block"
    ]}
    config={@map_config}
    pins={[@from, @to]}
  />
  """
```

anthonyshull (approved):
Something we could look out for in the future: if we're wrapping a component in another component just to set the value of an attribute, that attribute can be set in the component directly.

lib/dotcom_web/components/live_components/trip_planner_results_section.ex
joshlarson: Oh yeah, that is cleaner.

PR #2241, opened 2024-11-29
refactor: Inline <.itinerary_group /> component into TripPlannerResultsSection

joshlarson:
No Asana ticket.

I think this addresses [this concern](https://github.com/mbta/dotcom/pull/2232/files#r1857250141) raised in https://github.com/mbta/dotcom/pull/2232, by bringing both of the `phx-click="set_expanded_itinerary_index"`'es into the same file where the resulting `expanded_itinerary_index` is actually used.


anthonyshull (approved):
Only thing is we should have a fallback for `itinerary_panel` that logs if it doesn't get pattern matched and returns some default.

joshlarson:
@anthonyshull how does https://github.com/mbta/dotcom/pull/2241/commits/bb429341f985fcdba999b51892635c696d4ebb5e look? IMO it doesn't look great, but also that scenario _should_ never happen, so I don't think we need to make it look too pretty O_o

(Also, `Sentry.capture_message` for logging, or do we have a better way?)

Screenshot 2024-12-02 at 11 41 53 AM

Screenshot 2024-12-02 at 11 41 58 AM

anthonyshull:
> @anthonyshull how does [bb42934](https://github.com/mbta/dotcom/commit/bb429341f985fcdba999b51892635c696d4ebb5e) look? IMO it doesn't look great, but also that scenario _should_ never happen, so I don't think we need to make it look too pretty O_o
> 

Yeah, that looks good.

> (Also, `Sentry.capture_message` for logging, or do we have a better way?)

Yeah, that's what we've been using. I think you can use the `extra` option instead of converting the assigns to a string:
https://hexdocs.pm/sentry/Sentry.html#capture_message/2-options

PR #2243, opened 2024-12-03
feat(TripPlanner): display backend error message

thecristen:
#### Summary of changes

**Asana Ticket:** [Trip Planner returns error messages](https://app.asana.com/0/555089885850811/1208651714256934/f)

Using the recent update from the [OpenTripPlannerClient](https://github.com/mbta/open_trip_planner_client/pull/1) (which is now migrated to the MBTA Github org!), we can now show more detailed error messages in the trip planner.  (from the Asana ticket, this is the section "New errors for OTP input errors")

Technically the "Unable to plan your trip" error is located in the wrong spot – it should be where the itineraries display. We can move that in a followup PR after @joshlarson refactors `TripPlannerResultsSection` from a LiveComponent into a function component as previously discussed.

- Also fixes showing error messages _within_ the form, as these were broken in both Metro and our own backend 😅 

image
image
image
image


test/dotcom_web/components_test.exs
```diff
@@ -30,4 +32,27 @@ defmodule DotcomWeb.ComponentsTest do
...
+        rendered_to_string(~H"""
+        <.error_container title={assigns.title}>
+          <%= assigns.content %>
+        </.error_container>
+        """)
+
+      assert str =~ assigns.title
+      assert str =~ assigns.content
+
+      assert rendered_to_string(~H"""
```
joshlarson, line 53: Nit (non-blocking): I think the assertion that would make this section justify its existence would be to somehow assert that there is no title.

joshlarson (approved):

PR #2246, opened 2024-12-05
Add walking legs to the trip plan details

anthonyshull:


lib/dotcom_web/components/trip_planner/itinerary_detail.ex
```diff
@@ -77,15 +78,19 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryDetail do
...
+        <% else %>
+          <.leg
+            start_time={leg.start}
+            end_time={leg.stop}
+            from={leg.from}
+            to={leg.to}
+            mode={leg.mode}
+            realtime={leg.realtime}
+            realtime_state={leg.realtime_state}
+          />
```
anthonyshull, line 92: 📓 We should probably rename this to TransitLeg and not destructure the leg.

lib/dotcom_web/components/trip_planner/walking_leg.ex
```diff
@@ -0,0 +1,54 @@
...
+    - `duration` (integer) in minutes
+    - `mode` with a list of `steps` (list of `Step` structs)
+    - `polyline` (string) to use as the accordion ID
+  """
+
+  attr :leg, :map, required: true
+
+  def walking_leg(assigns) do
+    ~H"""
+    <hr class="my-2 border-t-solid border-slate-200" />
```
thecristen, line 28: heh it's different because you used `border-slate-300` in the accordion! anyways, it strikes me as kinda funky that some of the bordering is already baked into the component but this other one we needed to add here. hopefully that'll look better in a followup.

lib/dotcom_web/components/live_components/trip_planner_form.ex
```diff
@@ -129,11 +129,6 @@ defmodule DotcomWeb.Components.LiveComponents.TripPlannerForm do
                   </.inputs_for>
                 </div>
               </:content>
-              <:extra :if={used_input?(f[:modes])}>
```
thecristen, line 132: Can we undo this removal? This removes the "please select a mode" message if all modes are left unchecked, and this change doesn't seem to have anything to do with the rest of this PR anyway.

lib/dotcom_web/components/trip_planner/itinerary_detail.ex
thecristen: did you look inside `DotcomWeb.Components.TripPlanner.Leg.leg/1`? it's basically a wrapper that shows a transit leg or walking leg. Might want to either (a) move this to there or (b) simplify `leg/1` internal logic to remove handling of the walking directions, then rename that to `transit_leg`.

anthonyshull: That's what I meant by this comment. I don't think it is in the scope of this ticket to refactor the previous work. But, we should do it because the two different legs have completely different rendering methods. And, they should both take a `leg` instead of that destructuring.

lib/dotcom_web/components/trip_planner/walking_leg.ex
anthonyshull: Well, this line isn't a part of the accordion. It's a separator from the rest of the steps. So, it shouldn't be part of the accordion.

lib/dotcom_web/components/live_components/trip_planner_form.ex
thecristen: Ah, gotcha.

image

As I just described: "Please select at least one mode of transit". But this could just as easily be included as a `<div>` after the accordion instead of using the prior `extra` attribute.

lib/dotcom_web/components/trip_planner/itinerary_detail.ex
thecristen: Yeah, I agree. I just think since this change makes part of the existing `leg/1` logic [unreachable](https://github.com/mbta/dotcom/blob/929128f3dbd6f6411f792b3bab4cee846d16e72f/lib/dotcom_web/components/trip_planner/leg.ex#L55-L58) already, it feels a little odd not to make that refactor here. Just my 2c.

thecristen (approved):

PR #2248, opened 2024-12-05
fix: Make Service.serves_date?/2 use added_stops as a full override

joshlarson:
This accommodates a very specific type of GTFS service, which looks like

      %Services.Service{
        added_dates: ["2024-12-05"],
        added_dates_notes: %{"2024-12-05" => nil},
        description: "Weekday schedule",
        end_date: ~D[2024-12-04],
        name: "Weekday",
        id: "LRV42024-hlb44011-Weekday-01",
        removed_dates: [],
        removed_dates_notes: %{},
        start_date: ~D[2024-11-27],
        type: :weekday,
        typicality: :typical_service,
        valid_days: [1, 2, 3, 5],
        rating_start_date: ~D[2024-08-25],
        rating_end_date: nil,
        rating_description: "Fall"
      }

The edge case is when the `valid_days` doesn't include the day of the week for the `added_date(s)`. When that happens, the added dates should still be considered served, but the previous version was only handling that correctly if `valid_days` was `[]`.

# Before

Screenshot 2024-12-05 at 5 29 39 PM

# After

Screenshot 2024-12-05 at 5 29 43 PM

### DIFF
--- a/lib/services/service.ex
+++ b/lib/services/service.ex
@@ -188,9 +188,9 @@ defmodule Services.Service do
     removed_dates = parse_listed_dates(removed_dates)
 
     (dates
-     |> Enum.reject(fn date -> Enum.member?(removed_dates, date) end)
-     |> Enum.reject(fn date -> Timex.weekday(date) not in valid_days end)
-     |> Enum.map(&Timex.to_date/1)
+     |> Stream.reject(fn date -> Enum.member?(removed_dates, date) end)
+     |> Stream.reject(fn date -> Timex.weekday(date) not in valid_days end)
+     |> Stream.map(&Timex.to_date/1)
      |> Enum.uniq()) ++ parse_listed_dates(added_dates)
   end
 
--- a/lib/services/service.ex
+++ b/lib/services/service.ex
@@ -184,6 +184,7 @@ defmodule Services.Service do
         |> Interval.new()
         |> Enum.map(& &1)
       end
+      |> Enum.map(&Timex.to_date/1)
 
     removed_dates = parse_listed_dates(removed_dates)
 
--- a/lib/services/service.ex
+++ b/lib/services/service.ex
@@ -194,7 +194,6 @@ defmodule Services.Service do
       dates
       |> Stream.reject(fn date -> Enum.member?(removed_dates, date) end)
       |> Stream.reject(fn date -> Timex.weekday(date) not in valid_days end)
-      |> Stream.map(&Timex.to_date/1)
       |> Enum.uniq()
 
     explicitly_added_dates ++ valid_dates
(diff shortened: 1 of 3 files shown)


lib/services/service.ex
```diff
@@ -187,15 +187,11 @@ defmodule Services.Service do
...
-      # if valid_days is an empty array, the service's dates are likely those in
-      # added_dates, so we can ignore evaluating the day of the week here
-      Enum.member?(removed_dates, date) ||
-        if valid_days != [], do: Timex.weekday(date) not in valid_days
-    end)
-    |> Enum.map(&Timex.to_date/1)
-    |> Enum.uniq()
+    (dates
+     |> Enum.reject(fn date -> Enum.member?(removed_dates, date) end)
+     |> Enum.reject(fn date -> Timex.weekday(date) not in valid_days end)
```
anthonyshull, line 192: Credo is going to complain about the double reject. Probably easiest to make it a private function that does both checks.

lib/services/service.ex
```diff
@@ -187,15 +187,11 @@ defmodule Services.Service do
...
+     |> Enum.map(&Timex.to_date/1)
+     |> Enum.uniq()) ++ parse_listed_dates(added_dates)
```
anthonyshull, line 194: This syntax is really funky. Would be pretty easy to break it out a little

lib/services/service.ex
joshlarson: Ah yep - I saw that - I used `Stream` and Credo stopped complaining 🙊

lib/services/service.ex
joshlarson: Yeah I didn't love it either :\

Will push up a change

lib/services/service.ex
```diff
@@ -184,18 +184,20 @@ defmodule Services.Service do
...
-    end)
-    |> Enum.map(&Timex.to_date/1)
-    |> Enum.uniq()
+    explicitly_added_dates = parse_listed_dates(added_dates)
+
+    valid_dates =
+      dates
+      |> Stream.reject(fn date -> Enum.member?(removed_dates, date) end)
+      |> Stream.reject(fn date -> Timex.weekday(date) not in valid_days end)
+      |> Stream.map(&Timex.to_date/1)
```
thecristen, line 197: You don't need this line anymore

lib/services/service.ex
```diff
@@ -184,18 +184,20 @@ defmodule Services.Service do
...
+      |> Enum.uniq()
+
+    explicitly_added_dates ++ valid_dates
```
thecristen, line 200: I wonder if you ought to do the `Enum.uniq()` after this step?

thecristen (approved):

PR #2250, opened 2024-12-06
feat: Suppress "Depart at" buttons when there's only one itinerary

joshlarson:
# Before

Screenshot 2024-12-06 at 2 59 28 PM

# After
Screenshot 2024-12-06 at 2 58 53 PM

**Asana Ticket:** [Trip Planner Preview | Suppress "Depart At" buttons if only one itinerary is surfaced in an itinerary group](https://app.asana.com/0/555089885850811/1208923455785213/f)

### DIFF
--- a/lib/dotcom_web/components/trip_planner/itinerary_detail.ex
+++ b/lib/dotcom_web/components/trip_planner/itinerary_detail.ex
@@ -36,14 +36,9 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryDetail do
   attr :selected_itinerary_detail_index, :integer
   attr :target, :string
 
-  defp depart_at_buttons(%{itineraries: [_]} = assigns) do
-    ~H"""
-    """
-  end
-
   defp depart_at_buttons(assigns) do
     ~H"""
-    <div>
+    <div :if={Enum.count(@itineraries) > 1}>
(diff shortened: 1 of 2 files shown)


lib/dotcom_web/components/trip_planner/itinerary_detail.ex
```diff
@@ -20,6 +20,28 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryDetail do
...
+  attr :target, :string
+
+  defp depart_at_buttons(%{itineraries: [_]} = assigns) do
+    ~H"""
+    """
+  end
+
+  defp depart_at_buttons(assigns) do
     ~H"""
     <div>
```
thecristen, line 46:
```suggestion
    <div :if={Enum.count(@itineraries) < 2}>
```
😅 I feel like adding the `if` directive to the existing markup would suffice (and be a bit more readable)? Unless I'm missing something?

joshlarson: Nope! I just got into pattern-matching brain. The `:if` directive is easier to read in this context!

https://github.com/mbta/dotcom/pull/2250/commits/09081ebf157e86d49d93b7a5d8de077d379e24ef

PR #2251, opened 2024-12-06
WIP: feat(TripPlanner): transit leg and itinerary locations

thecristen:
#### Summary of changes

**Asana Ticket:** [Trip Planner Preview | Details: Transit segment](https://app.asana.com/0/555089885850811/1208788271277316/f) and [Trip Planner Preview | Details: Location Item](https://app.asana.com/0/555089885850811/1208788271277302/f)

Somewhat rough around the edges and still missing some acceptance criteria, but wanted to get a chance for review before leaving for a couple days. Notable items remaining:

- [ ] Missing the trip headsign - we can add this to the query in `OpenTripPlannerClient`
- [ ] It should only show _one_ location in between two legs, not both the end of one leg + beginning of next leg. I tried to write some logic to look at those two legs and consolidate them while preferring to display the transit icon over the generic location pin... but it was terrible and didn't quite work. I'm thinking it might be better just to adjust `OpenTripPlannerClient` to give us the stop's `locationType` .. then we could read that value to decide which icon to render.
- [ ] The text needs some improvement for commuter rail train rides - in the AC it's technically "TBD" but IMO it should at least show the train number.
- [x] The MBTA stop names are supposed to be clickable links to their respective stop pages
- [x] I haven't yet tested this against really long location names (it should wrap a certain way)

image


joshlarson:
>  I'm thinking it might be better just to adjust OpenTripPlannerClient to give us the stop's locationType .. then we could read that value to decide which icon to render.

Could we do something even dumber than that, and just have some logic at time-of-rendering that tells a walking segment not to render the locations at either end unless it's the beginning or end of an itinerary?

joshlarson (changes requested):
This looks so great right now!

I have a thought about `<.place />`. What if, instead of wrapping each `<.leg />` in two places (which guarantees that all places will get doubled up, as it currently is), we wrapped _only_ transit legs with `<.place />`'s and left walking legs place-less. We would then need to add a location-dot-icon place at the beginning and the end of the whole itinerary.

(If we do ☝️then it probably makes sense to have a `<.transit_leg />` include its places and leave them out of the main `itinerary_detail` function, but that's kind of a minor thing.)

In order to get that fully 100% correct, I think we'd need to add a data transformation layer to solve (at least) two corner-cases:
- Transferring without walking, which can happen if you need to transfer at a bus stop that serves multiple routes, or if you need to switch green line trains in the Central Subway. (This would involve having a "waiting leg", which hasn't been fully designed yet)
- If the itinerary starts or ends exactly at a station or stop, then we don't want to double up the locations (like in the screenshot below)
Screenshot 2024-12-09 at 5 02 48 PM

IMO ☝️ these corner cases and the data transformation layer can happen as fast-follow-up PR's, rather than getting lumped into this one. If we get rid of the guaranteed duplication of `<.place />`'s, then I'm good to merge this.


---

TL;DR - there are a bunch of corner cases that we need to address before releasing, but can happen as follow-up PR's. For this PR, I'm good to merge once we get rid of the duplicated `<.place />`'s?

lib/dotcom_web/components/trip_planner/itinerary_detail.ex
```diff
@@ -63,35 +64,26 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryDetail do
...
-        :all_routes,
-        assigns.itinerary.legs
-        |> Enum.reject(&match?(%PersonalDetail{}, &1.mode))
-        |> Enum.map(& &1.mode.route)
-      )
+      assigns
+      |> assign(:start_place, List.first(assigns.itinerary.legs).from)
+      |> assign(:start_time, List.first(assigns.itinerary.legs).start)
+      |> assign(:end_place, List.last(assigns.itinerary.legs).to)
+      |> assign(:end_time, List.last(assigns.itinerary.legs).stop)
```
joshlarson, line 71: This whole blob is kind of gross, but @thecristen and I talked offline, and we're going to remove it in a fast-follow PR, so that this one doesn't get bloated.

joshlarson: Side note, at least the variable name lengths decrement by one letter per line, so they form a nice shape.

joshlarson (approved):

PR #2252, opened 2024-12-09
refactor: Convert TripPlannerResultsSection to a function component

joshlarson:
Follow up to [this comment](https://github.com/mbta/dotcom/pull/2238#pullrequestreview-2472824039), which suggested biasing much more strongly towards function components, and [this PR](https://github.com/mbta/dotcom/pull/2242), which consolidated the `TripPlannerResultsSection` state into a single variable.

User-facing-wise, this should be a pure refactor, so no visible changes.

No ticket.

### DIFF
--- a/lib/dotcom_web/live/trip_planner.ex
+++ b/lib/dotcom_web/live/trip_planner.ex
@@ -8,8 +8,9 @@ defmodule DotcomWeb.Live.TripPlanner do
   use DotcomWeb, :live_view
 
   import MbtaMetro.Components.{Feedback, Spinner}
+  import DotcomWeb.Components.TripPlanner.TripPlannerResultsSection
 
-  alias DotcomWeb.Components.LiveComponents.{TripPlannerForm, TripPlannerResultsSection}
+  alias DotcomWeb.Components.LiveComponents.{TripPlannerForm}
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/live/trip_planner.ex
```diff
@@ -71,19 +73,53 @@ defmodule DotcomWeb.Live.TripPlanner do
...
         from={@from}
         to={@to}
+        itinerary_selection={@itinerary_selection}
       />
     </div>
     """
   end
 
+  @impl true
+  def handle_event("set_itinerary_group_index", %{"index" => index_str}, socket) do
```
joshlarson, line 89: One of the reasons I originally made everything live components was because of these event names; it felt a bit odd to have the event names spread around the way they are - the handler is here, but the thing that generates the event is tucked away in `trip_planner_results_section` (or for the `set_itinerary_index` event, tucked away in `itinerary_detail`).

Is that okay? Do we want to pass the event names around as `attr`s? Some other solution?

joshlarson: Realistically, I could see the answer to this being a follow-up, rather than trying to "fix" it in this PR, but I am interested in what our overarching vision for this type of question is.

anthonyshull (approved):

PR #2257, opened 2024-12-10
cleanup: We only need to alias PersonalDetail once

joshlarson:
No ticket. Just happened to notice that we were _really_ making sure that we imported the `PersonalDetail` module.

### DIFF
--- a/test/fares/fares_test.exs
+++ b/test/fares/fares_test.exs
@@ -4,7 +4,7 @@ defmodule FaresTest do
 
   import Mox
 
-  alias Dotcom.TripPlan.{NamedPosition, Leg, PersonalDetail, TransitDetail}
+  alias Dotcom.TripPlan.{Leg, NamedPosition, PersonalDetail, TransitDetail}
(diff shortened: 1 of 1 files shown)


test/fares/fares_test.exs
```diff
@@ -4,7 +4,7 @@ defmodule FaresTest do
 
   import Mox
 
-  alias Dotcom.TripPlan.{NamedPosition, Leg, PersonalDetail, PersonalDetail, TransitDetail}
+  alias Dotcom.TripPlan.{NamedPosition, Leg, PersonalDetail, TransitDetail}
```
anthonyshull, line 7: Can you alphabetize these? Makes it easier to see if there are duplicates.

anthonyshull (changes requested):
See comment.

test/fares/fares_test.exs
joshlarson: [Done](https://github.com/mbta/dotcom/pull/2257/commits/0bcdf4b52b5d09e99b9b7a58d8a385506239d4ad)

anthonyshull (approved):

PR #2258, opened 2024-12-10
Itinerary detail map

anthonyshull:
Screenshot Capture - 2024-12-10 - 10-26-44

### DIFF
--- a/lib/dotcom/trip_plan/map.ex
+++ b/lib/dotcom/trip_plan/map.ex
@@ -53,7 +53,7 @@ defmodule Dotcom.TripPlan.Map do
 
   This just gets the longitude and latitude from each marker.
   """
-  def get_points(map) do
+  def get_points(itinerary_map) do
     itinerary_map
     |> Map.get(:markers, [])
     |> Enum.map(&marker_to_point/1)
@@ -78,7 +78,7 @@ defmodule Dotcom.TripPlan.Map do
 
   defp polyline_to_line(polyline) do
     %{
-      color: Map.get(polyline, :color, "black"),
+      color: Map.get(polyline, :color) |> tailwind_color(),
       coordinates: Map.get(polyline, :positions, []) |> invert_coordinates(),
       width: Map.get(polyline, :weight, 4)
     }
@@ -197,4 +197,13 @@ defmodule Dotcom.TripPlan.Map do
   def z_index(%{current: idx, start: idx}), do: 100
   def z_index(%{current: idx, end: idx}), do: 100
   def z_index(%{}), do: 0
+
+  defp tailwind_color("#003DA5"), do: "blue-500"
+  defp tailwind_color("#80276C"), do: "purple-500"
+  defp tailwind_color("#00843D"), do: "green-500"
+  defp tailwind_color("#ED8B00"), do: "orange-500"
+  defp tailwind_color("#DA291C"), do: "red-500"
+  defp tailwind_color("#7C878E"), do: "silver-500"
+
+  defp tailwind_color(_), do: "black"
 end
--- a/lib/dotcom/trip_plan/map.ex
+++ b/lib/dotcom/trip_plan/map.ex
@@ -2,6 +2,7 @@ defmodule Dotcom.TripPlan.Map do
   @moduledoc """
   Handles generating the maps displayed within the TripPlan Controller
   """
+
   alias Dotcom.TripPlan.{Leg, NamedPosition, TransitDetail}
   alias Leaflet.{MapData, MapData.Marker}
   alias Leaflet.MapData.Polyline, as: LeafletPolyline
@@ -15,7 +16,6 @@ defmodule Dotcom.TripPlan.Map do
     |> MapData.new(14)
   end
 
-  # Maps for results
   @doc """
   Returns the static map data and source URL
   Accepts a function that will return either a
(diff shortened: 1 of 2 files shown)


lib/dotcom/trip_plan/map.ex
```diff
@@ -150,4 +197,13 @@ defmodule Dotcom.TripPlan.Map do
   def z_index(%{current: idx, start: idx}), do: 100
   def z_index(%{current: idx, end: idx}), do: 100
   def z_index(%{}), do: 0
+
+  defp tailwind_color("#003DA5"), do: "blue-500"
+  defp tailwind_color("#80276C"), do: "purple-500"
+  defp tailwind_color("#00843D"), do: "green-500"
+  defp tailwind_color("#ED8B00"), do: "orange-500"
+  defp tailwind_color("#DA291C"), do: "red-500"
+  defp tailwind_color("#7C878E"), do: "silver-500"
```
anthonyshull: 📓 For now, we have to use the Metro colors because it will only accept those. We can change Metro to pass in an arbitrary hex code.

lib/dotcom_web/components/trip_planner/trip_planner_results_section.ex
```diff
@@ -39,19 +41,41 @@ defmodule DotcomWeb.Components.TripPlanner.TripPlannerResultsSection do
...
+            id="trip-planner-map"
+            class={[
+              "h-64 md:h-96 w-full",
+              "relative overflow-none row-span-2",
+              @itinerary_selection == :summary && "hidden md:block"
+            ]}
+            config={@map_config}
+            pins={[@from, @to]}
+          />
+        <% end %>
```
anthonyshull, line 76: 📓 None of this will be here after we change the state management. It must be embedded like this because the results component includes the map.

lib/dotcom/trip_plan/map.ex
```diff
@@ -37,6 +37,53 @@ defmodule Dotcom.TripPlan.Map do
...
+
+  defp invert_coordinates(coordinates) do
+    coordinates
+    |> Enum.map(&invert_coordinate/1)
+    |> Enum.reject(&is_nil/1)
+  end
+
+  defp invert_coordinate([a, b]), do: [b, a]
+
+  defp invert_coordinate(_), do: nil
```
joshlarson, line 71: Is this necessary?

lib/dotcom/trip_plan/map.ex
```diff
@@ -150,4 +197,13 @@ defmodule Dotcom.TripPlan.Map do
   def z_index(%{current: idx, start: idx}), do: 100
   def z_index(%{current: idx, end: idx}), do: 100
   def z_index(%{}), do: 0
+
+  defp tailwind_color("#003DA5"), do: "blue-500"
```
joshlarson, line 201: I'm torn about this. Most of the other places where we translate things from GTFS to color, we do it based on the route, but here we're pulling the color out of the polyline directly.

If it turns out that we can get this to work, should we use the colors directly in otther contexts too?

lib/dotcom_web/components/trip_planner/trip_planner_results_section.ex
```diff
@@ -39,19 +41,43 @@ defmodule DotcomWeb.Components.TripPlanner.TripPlannerResultsSection do
...
-        config={@map_config}
-        pins={[@from, @to]}
-      />
-
+      <.async_result :let={results} assign={@results}>
+        <%= if results && @itinerary_selection != :summary do %>
+          <% {:detail,
+           %{itinerary_group_index: itinerary_group_index, itinerary_index: itinerary_index}} =
+            @itinerary_selection %>
+          <% itinerary =
```
joshlarson, line 49: You can move a bunch of this logic out of the HEEx template itself and into the body of `&trip_planner_results_section/1` right before the HEEx template by doing something like:

```elixir
def trip_planner_results_section(
      %{itinerary_selection: itinerary_selection, results: itinerary_results} = assigns
    ) do
  assigns =
    if itinerary_results.result && itinerary_selection != :summary do
      {:detail,
       %{itinerary_group_index: itinerary_group_index, itinerary_index: itinerary_index}} =
        itinerary_selection

      itinerary =
        Enum.at(itinerary_results.result, itinerary_group_index).itineraries
        |> Enum.at(itinerary_index)

      itinerary_map = TripPlan.Map.itinerary_map(itinerary)
      lines = TripPlan.Map.get_lines(itinerary_map)
      points = TripPlan.Map.get_points(itinerary_map)

      assign(assigns, %{lines: lines, points: points})
    else
      assign(assigns, %{lines: [], points: []})
    end

  ~H"""
  <!-- ... -->
  """
end
```

(I'm not attached to ☝️ code exactly as-is - I think it could be cleaned up a bit, although I would buy the argument that polishing this chunk of code isn't worth the effort)

That lets you have `lines` and `points` be `[]` unless there's an itinerary powering real data for them, which then lets you just unconditionally render the map, with or without the itinerary data:

```heex
<.live_component
  module={MbtaMetro.Live.Map}
  id="trip-planner-map"
  class={[
    "h-64 md:h-96 w-full",
    "relative overflow-none row-span-2",
    @itinerary_selection == :summary && "hidden md:block"
  ]}
  config={@map_config}
  lines={@lines}
  pins={[@from, @to]}
  points={@points}
/>
```

...and then get rid of _all_ of the extra logic that's embedded inside the HEEx blob.

That also has the advantage of fixing [this bug](https://github.com/mbta/dotcom/pull/2258#issuecomment-2532661926), by bringing the map back outside of the `<.async_result />`.

joshlarson (changes requested):
Looks pretty great! 

I made a few comments, but the big one that's holding up approval for me is [the map-flickering bug](https://github.com/mbta/dotcom/pull/2258#issuecomment-2532661926), which I think would be reasonably-elegantly solved by [this change](https://github.com/mbta/dotcom/pull/2258/files#r1878698178), but as long as it's fixed, I'm not too worried about how.

lib/dotcom/trip_plan/map.ex
thecristen: The color that's in the polyline came from the associated route, it's just parsed roundaboutly because legacy....

lib/dotcom/trip_plan/map.ex
```diff
@@ -37,6 +37,53 @@ defmodule Dotcom.TripPlan.Map do
...
+    |> Enum.reject(&is_nil/1)
+  end
+
+  defp invert_coordinates(coordinates) do
+    coordinates
+    |> Enum.map(&invert_coordinate/1)
+    |> Enum.reject(&is_nil/1)
+  end
+
+  defp invert_coordinate([a, b]), do: [b, a]
```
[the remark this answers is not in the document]
anthonyshull: This is right before it is being passed into maplibre gl.

lib/dotcom/trip_plan/map.ex
anthonyshull: If we ever get a coordinate that isn't a tuple it will crash the process otherwise.

joshlarson (approved):

PR #2260, opened 2024-12-10
refactor(TransitDetail): use trip struct

thecristen:
Changing the `%TransitDetail{}` spec to have a trip rather than just a trip_id.
This'll set us up for displaying the trip headsign and trip name directly from OTP.

---

Made some more updates. This PR now also updates the preview Trip Planner so we can see headsigns instead of trip IDs. It also improves the transit leg's text to better match the acceptance criteria in [Trip Planner Preview | Details: Transit segment](https://app.asana.com/0/555089885850811/1208788271277316/f)

image
image


lib/dotcom/trip_plan/related_link.ex
```diff
@@ -104,10 +104,10 @@ defmodule Dotcom.TripPlan.RelatedLink do
     new("Massport schedules", url, :massport_shuttle)
   end
 
-  defp route_link(route, trip_id, itinerary) do
+  defp route_link(route, _, itinerary) do
     icon_name = Route.icon_atom(route)
     date = Date.to_iso8601(itinerary.start)
-    url = schedule_path(DotcomWeb.Endpoint, :show, route, date: date, trip: trip_id)
+    url = schedule_path(DotcomWeb.Endpoint, :show, route, date: date)
```
thecristen, line 110: This was just a legacy thing, as far as I can tell putting a trip in the query parameters doesn't render anything different on a schedule page nowadays...

lib/dotcom_web/components/trip_planner/itinerary_detail.ex
```diff
@@ -19,6 +19,7 @@ defmodule DotcomWeb.Components.TripPlanner.ItineraryDetail do
       ) do
     assigns =
       assign(assigns, :selected_itinerary, Enum.at(itineraries, selected_itinerary_detail_index))
+      |> assign_new(:target, fn -> nil end)
```
thecristen, line 22: A test broke without this 🤷🏼‍♀️

thecristen: https://github.com/mbta/dotcom/blob/acb1cbad75bd8e35d46f1a69642125ae480519e3/lib/dotcom_web/components/trip_planner/itinerary_detail.ex#L28 it's still here :/

joshlarson: Based on 👇 you should be able to get rid of this now:
- https://github.com/mbta/dotcom/pull/2261

thecristen: Done!

lib/dotcom_web/components/trip_planner/trip_planner_results_section.ex
```diff
@@ -72,26 +72,27 @@ defmodule DotcomWeb.Components.TripPlanner.TripPlannerResultsSection do
               }}
          } = assigns
        ) do
-    %{itineraries: itineraries, summary: summary} = results |> Enum.at(itinerary_group_index)
+    with %{itineraries: itineraries, summary: summary} <-
+           results |> Enum.at(itinerary_group_index) do
```
joshlarson, line 76: Is this change necessary? 

I only ask because Anthony is currently working on a refactor that will overhaul (possibly remove or rename?) this file, and this could generate git conflicts.

test/dotcom_web/live/trip_planner_test.exs
```diff
@@ -226,13 +226,13 @@ defmodule DotcomWeb.Live.TripPlannerTest do
 
       view |> element("button", "Details") |> render_click()
 
-      assert render_async(view) =~ trip_id_1
-      refute render_async(view) =~ trip_id_2
+      assert render_async(view) =~ trip_headsign_1
```
joshlarson, line 229: The tests here have become pretty flaky. (Wild guess) about one in four times I run them, one of these headsign assertions fails.

I'm not 100% certain, but I'm guessing that sometimes the factoried-up route is "external", which makes `<.transit_leg />` look for the route long-name instead of the trip headsign.

I think the solution here is to force the factoried-up route in this test to not be external.

It would also be good to have explicit coverage of the use case where the route is an external agency. Currently undecided whether that should be approval-blocking or not.

At some point, I'd like us as a team to revisit the explicit versus implicit data in tests question, because I think that's causing flakiness elsewhere as well.

lib/dotcom_web/components/trip_planner/trip_planner_results_section.ex
thecristen: I put it there because I ran into an error where this returned `nil` and failed the pattern match. hopefully the refactor will clearly obviate this!

test/dotcom_web/live/trip_planner_test.exs
thecristen: Great catch, your assessment sounds likely correct to me. Maybe this would be easier to pull off after your refactor, if we could maybe stub out _segments_ in tests instead of piecing togeher ever more intricate itineraries... hmmm.
In the meantime I'll hack internal routes into these tests.

joshlarson (changes requested):
Mostly looks good, but the [test flakiness](https://github.com/mbta/dotcom/pull/2260#discussion_r1880437938) is a tad too much for me!

I think once that's resolved, I'll be ready to ✅

test/dotcom_web/live/trip_planner_test.exs
joshlarson: 🤔 I think I'd vote in favor of hacking internal routes into the tests and keeping it that way, since the segment abstraction that I came up with is still very much an implementation detail (and new enough that it's not particularly stable).

joshlarson: The new version of the tests are still flaky (actually flakier than they were before - now we're at 50%). This time I think it's because you removed the `start_time` logic, so now `trip_headsign_1` sometimes happens before `trip_headsign_2`, but sometimes they happen in the other order, which means that half the time, the test is expecting the trips to be swapped.

I think the easiest solution would be to restore the overall test structure back to what it was before, replace `update_trip_id()` with `update_headsign()`, and just hard-code something in `base_itinerary` that forces it to not be an external agency.

test/dotcom_web/live/trip_planner_test.exs
```diff
@@ -20,19 +21,31 @@ defmodule DotcomWeb.Live.TripPlannerTest do
...
+      })
+
+    base_itinerary = Factory.build(:itinerary, legs: [base_leg])
+
+    Enum.reduce(headsigns, [], fn headsign, itineraries ->
+      leg = update_in(base_leg, [:trip, :trip_headsign], fn _ -> headsign end)
+      itinerary = update_in(base_itinerary, [:legs], fn _ -> [leg] end)
+      [itinerary | itineraries]
+    end)
+    |> Enum.reverse()
```
joshlarson, line 48:
```suggestion
    Enum.map(headsigns, fn headsign ->
      leg = update_in(base_leg, [:trip, :trip_headsign], fn _ -> headsign end)
      %{base_itinerary | legs: [leg]}
    end)
```

Per [this comment](https://github.com/mbta/dotcom/pull/2260/files#r1882440446), I'm not sure we need this logic at all, but if we do, then I think this version is equivalent?

test/dotcom_web/live/trip_planner_test.exs
thecristen: dang, yeah I forgot the test expects the list to be ordered. 
I made a change where the test itineraries now have sequential start datetimes, maybe that'll be enough?

joshlarson: Ahh there's one more issue (I'll push up a fix for this, since you're out for the weekend).

I was so confused by this failure:

```elixir
assert render_async(view) =~ trip_headsign_1 # <-- passes
refute render_async(view) =~ trip_headsign_2 # <-- fails
```

There's basically no way that the code should allow both headsigns to appear on the screen at the same time - if something was wrong with the ordering, then I would have expected the `assert` statement to fail.

Turns out:

We're using `Faker.App.name()` as the headsign. But we're also using it elsewhere in factories that get rendered on the screen, so if we end up with something like:
```elixir
trip_headsign_2 = Faker.App.name() # returns "Zathon"

# Elsewhere, deep in some factory...
list_item = Faker.App.name() # also happens to returns "Zathon"
```

Then `list_item` will get rendered on the screen as `Zathon`, and `refute render_async(view) =~ trip_headsign_2` will fail.

I'm pushing up a fix to have the headsigns be `"Headsign:#{Faker.App.name()}"` instead. I tested that out by running the tests 20 times, and they passed every time, so I think that completely resolves it.

joshlarson (approved):

PR #2262, opened 2024-12-11
refactor: Convert itinerary legs to segments for easier rendering

joshlarson:
Pure refactor.

No ticket.

The point of this refactor is that the data returned from OpenTripPlanner is subtly different from what we want to render on the screen. Even in a simple case, when OTP returns `[walking, transit, walking]`, what we actually want to draw on the screen is `[location, walking, transit, walking, location]`*, and there are some corner-cases that make this even more complicated.

*I'm making the assumption that drawing `transit` on the screen actually means drawing the transit leg's start location and end location, each with the right icon for that transit leg, on either side of the transit leg details itself. This works out, since all transit legs are bracketed by locations indicating their stops or stations.

---

This'll enable follow-ups to address two use cases:

- When OTP returns `[transit, walking, transit]` or something like that, where the first or last leg of an itinerary is transit, then we want to draw `[transit, walking, transit]` - in other words, we don't want to show `location` right next to transit. [Ticket for this](https://app.asana.com/0/555089885850811/1208939733320494/f), which includes a screenshot of what it looks like now.
- When OTP returns something like `[walking, transit, transit, walking]`, that is, when there's a transfer that involves no walking, then we want to add in a "waiting leg", so the end result should be `[location, walking, transit, waiting, transit, walking, location]`. [Ticket for this](https://app.asana.com/0/home/1205718271156544/1208939733320483) (same deal - has a screenshot).

### DIFF
--- a/lib/dotcom/trip_plan/input_form.ex
+++ b/lib/dotcom/trip_plan/input_form.ex
@@ -49,6 +49,7 @@ defmodule Dotcom.TripPlan.InputForm do
       date: PlanParams.to_date_param(datetime),
       time: PlanParams.to_time_param(datetime),
       transportModes: __MODULE__.Modes.selected_mode_keys(modes) |> PlanParams.to_modes_param(),
+      numItineraries: 40,
       wheelchair: wheelchair
     }
     |> PlanParams.new()
--- a/lib/dotcom/trip_plan/input_form.ex
+++ b/lib/dotcom/trip_plan/input_form.ex
@@ -49,7 +49,6 @@ defmodule Dotcom.TripPlan.InputForm do
       date: PlanParams.to_date_param(datetime),
       time: PlanParams.to_time_param(datetime),
       transportModes: __MODULE__.Modes.selected_mode_keys(modes) |> PlanParams.to_modes_param(),
-      numItineraries: 40,
       wheelchair: wheelchair
     }
     |> PlanParams.new()
--- a/lib/dotcom_web/components/trip_planner/leg_to_segment_helper.ex
+++ b/lib/dotcom_web/components/trip_planner/leg_to_segment_helper.ex
@@ -1,0 +1,15 @@
+defmodule DotcomWeb.Components.TripPlanner.LegToSegmentHelper do
+  alias Dotcom.TripPlan.{PersonalDetail, TransitDetail}
+
+  def legs_to_segments(legs) do
+    Enum.map(legs, &to_segment/1)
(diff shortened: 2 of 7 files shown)


lib/dotcom_web/components/trip_planner/leg_to_segment_helper.ex
```diff
@@ -0,0 +1,37 @@
+defmodule DotcomWeb.Components.TripPlanner.LegToSegmentHelper do
```
joshlarson, line 1: I'm not sold on either this name or this location for this file, but I couldn't think of a better one off the top of my head.

thecristen: Since it's not technically a rendered component, you can move this to the `Dotcom.TripPlan.` namespace. I don't mind the current name.

thecristen (changes requested):
this is so elegant 🤩 

I think you should move the helper module. And as the linter says you can add a `@moduledoc` to it too. And don't forget the formatting!

Otherwise I think this looks great.

lib/dotcom_web/components/trip_planner/leg_to_segment_helper.ex
joshlarson: 💥 https://github.com/mbta/dotcom/pull/2262/commits/5caf9b799796fbd1980a5ae5b5aa4b4d11c5b1de 💥

joshlarson:
@thecristen I think I addressed everything - want to take another look?

lib/dotcom/trip_plan/input_form.ex
```diff
@@ -49,6 +49,7 @@ defmodule Dotcom.TripPlan.InputForm do
       date: PlanParams.to_date_param(datetime),
       time: PlanParams.to_time_param(datetime),
       transportModes: __MODULE__.Modes.selected_mode_keys(modes) |> PlanParams.to_modes_param(),
+      numItineraries: 40,
```
thecristen, line 52: I kinda wanna let this slide, but 40 seems kinda excessive?

thecristen (approved):

lib/dotcom/trip_plan/input_form.ex
joshlarson: Oh oops - didn't mean to include this change. I've been using that to get more itineraries so that _the one I'm looking for_ is more likely to actually show up.

PR #2263, opened 2024-12-11
refactor(ServiceOptGroup): simplified text

thecristen:
#### Summary of changes

**Asana Ticket:** [Schedule Finder schedule picker updates](https://app.asana.com/0/555089885850811/1208959344219840/f)

Refactoring to prefer showing the service's description directly instead of describing individual values from `valid_days`.
- Bonus: formats the service text for "extra service" in the same way we do for "holiday service" (which fixes the NYE service, and should also be fine for any other extra service we'll have)
- Bonus: fix the sorting of holiday schedules


joshlarson (approved):
From a code perspective, this looks good to me, and I'd be shocked if this caused any problems.

(Also this is vastly better than what it's replacing!)

That said, we should make sure that Plans & Schedules knows that we're using the `description` field in this way so that they're not surprised later.

PR #2266, opened 2024-12-11
Improve state management

anthonyshull:
https://app.asana.com/0/555089885850811/1208975073802630/f


lib/dotcom/trip_plan/input_form.ex
```diff
@@ -244,7 +229,9 @@ defmodule Dotcom.TripPlan.InputForm do
       |> selected_modes()
     end
 
-    def selected_modes([]), do: "No transit modes selected"
+    def selected_modes(%{RAIL: true, SUBWAY: true, BUS: true, FERRY: true}), do: "All modes"
+
+    def selected_modes(%{}), do: "Walking directions only"
```
joshlarson, line 234:
```suggestion
```

This doesn't do what it looks like it does. The `"Walking directions only"` chunk would match any map passed in as a parameter, and I'm reasonably sure that the only reason it doesn't short-circuit and catch all inputs is because it never matches because of `def selected_modes(%__MODULE__{} = modes)` a few lines up, and is actually dead code.

lib/dotcom_web/live/trip_planner.ex
```diff
@@ -8,189 +8,354 @@ defmodule DotcomWeb.Live.TripPlanner do
...
-        itinerary_selection={@itinerary_selection}
+        module={MbtaMetro.Live.Map}
+        id="trip-planner-map"
+        class="h-64 md:h-96 w-full"
+        config={@map.config}
+        lines={@map.lines}
+        pins={@map.pins}
+        points={@map.points}
       />
+      <.results results={@results} />
```
joshlarson, line 79: The way that this is structured at the top level is (as far as I can tell) a regression, since now the map-plus-results-panel layout is wrong on both desktop and mobile. The map is supposed to get rendered to the side of the results panel on desktop, and on mobile, it's only supposed to be visible when viewing a specific itinerary (at which point it's rendered in between bits of the results panel).

I had originally introduced `TripPlannerResultsSection` (yeah, not the most elegant name 😇) in order to accomplish ☝️, and I'm struggling to see how I can avoid just re-doing that work.

thecristen (approved):

PR #2273, opened 2024-12-16
fix: Make the "Service every 15 minutes or better" text black

joshlarson:
## Before
Screenshot 2024-12-16 at 3 53 19 PM

Screenshot 2024-12-16 at 3 53 34 PM

## After

Screenshot 2024-12-16 at 3 52 31 PM


thecristen (approved):
🐈‍⬛

PR #2278, opened 2024-12-19
feat(TripPlanner): Show more itineraries

joshlarson:
Screenshot 2024-12-19 at 2 58 11 PM

#### Summary of changes

**Before**
- We retrieved 5 itineraries from OTP, then grouped them, and returned them however they were grouped.
- This sometimes led to all of the itineraries being shown as one group, or two groups, and sometimes left out certain itineraries.
  - For instance, Logan Airport to North Station can easily be done with the Back Bay Logan Express shuttle, followed by an OL transfer, but if that wasn't in the next 5 trips, then it wouldn't show up.

**After**
- We retrieve 100 itineraries from OTP, group those, and then within each group, surface the next five itineraries (so as not to overwhelm).

**Asana Ticket:** [Increase number of itineraries returned from 5 ](https://app.asana.com/0/555089885850811/1208975267675615/f)

### DIFF
--- a/test/dotcom/trip_plan/itinerary_groups_test.exs
+++ b/test/dotcom/trip_plan/itinerary_groups_test.exs
@@ -34,6 +34,21 @@ defmodule Dotcom.TripPlan.ItineraryGroupsTest do
       assert Kernel.length(grouped_itineraries) == 1
     end
 
+    test "only includes the first five itineraries in a group", %{stops: [a, b, c]} do
+      # SETUP
+      bus_a_b_leg = TripPlanner.build(:bus_leg, from: a, to: b)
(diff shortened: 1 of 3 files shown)


thecristen (approved):

test/dotcom/trip_plan/itinerary_groups_test.exs
```diff
@@ -34,6 +34,21 @@ defmodule Dotcom.TripPlan.ItineraryGroupsTest do
       assert Kernel.length(grouped_itineraries) == 1
     end
 
+    test "only includes the first five itineraries in a group", %{stops: [a, b, c]} do
```
anthonyshull, line 37: I like how this new test sticks with the pattern of other tests in the file.

PR #2291, opened 2024-12-23
WIP - Make map and results summary sticky

joshlarson:
**Asana Ticket:** [Map should stick to the top when scrolling on desktop layout](https://app.asana.com/0/555089885850811/1209037564056797/f)


thecristen:
I think this is good as a first pass, merging

PR #2292, opened 2024-12-24
deps: update mbta_metro to minor version 1

thecristen:
#### Summary of changes

Updates [mbta/mbta_metro](https://www.github.com/mbta/mbta_metro) to the version which references [mbta/rider-design-system](https://www.github.com/mbta/rider-design-system) which references the design tokens created in Figma by the design system team.

While individual components are still being finalized, this gives us a mechanism for pushing updates to them via the design system. However, the button component from Metro is already updated with this PR as well.

**Asana Ticket:** ["Get trip suggestions" button should match dotcom primary button styling](https://app.asana.com/0/555089885850811/1209037564056785/f)

The updated [`<.button>` component definition](https://github.com/mbta/mbta_metro/blob/main/lib/mbta_metro/components/button.ex) with styling via [this Tailwind plugin definition](https://github.com/mbta/mbta_metro/blob/main/assets/js/plugins/button.js).


assets/css/app.scss
```diff
@@ -11,6 +11,8 @@
   --tw-content: '' !important;
 }
 
+@import "../node_modules/@mbta/rider-design-system/dist/variables.light.css" (prefers-color-scheme: light);
+@import "../node_modules/@mbta/rider-design-system/dist/variables.dark.css" (prefers-color-scheme: dark);
```
thecristen, line 15: because the button styles (and in the future, other component styles) are defined using these CSS variables, we get light and dark mode versions this easily!

thecristen:
It works!!

PR #2300, opened 2025-01-03
feat(autocomplete): on mobile for trip planner, use detached experience

thecristen:
#### Summary of changes

**Asana Ticket:** [Cannot scroll through location results on mobile (menu disappears when input loses focus)](https://app.asana.com/0/555089885850811/1209037564056799/f)

Trying out this library's [full-screen search experience on mobile](https://www.algolia.com/doc/ui-libraries/autocomplete/core-concepts/detached-mode/), enabled only for the trip planner for now.

### DIFF
--- a/assets/tailwind.config.js
+++ b/assets/tailwind.config.js
@@ -39,7 +39,7 @@ module.exports = {
     preflight: false
   },
   blocklist: ["container", "collapse"],
-  important: ".body-wrapper",
+  important: "body",
   content: [
     ...content,
     "./js/**/*.js",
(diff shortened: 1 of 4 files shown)


assets/tailwind.config.js
```diff
@@ -39,7 +39,7 @@ module.exports = {
     preflight: false
   },
   blocklist: ["container", "collapse"],
-  important: ".body-wrapper",
+  important: "body",
```
joshlarson, line 42: Didn't we get rid of this field? Why do we need this?

thecristen: It's in the elaborated commit message hehe 
image

joshlarson (approved):

PR #2301, opened 2025-01-03
feat: hide CR timetable in the presence of special alerts

paulswartz:
#### Summary of changes

**Asana Ticket:** [CR disruption timetable improvements](https://app.asana.com/0/116562505129567/1208217281866205/f)

This adds special-case handling for two types of alerts where we want to hide the timetable from view. One case is expected to have a PDF as the URL for the alert, and the other does not. In the PDF case, we hide the timetable and provide the link to the PDF. In the other, we hide the timetable and say "PDF upcoming."

This also changes the rendering of the alert to drop the special-case text, and to say "View PDF schedule" if the PDF link is available.

---

#### General checks

* [x] **Are interactive elements accessible?** This includes at minimum having [relevant keyboard interactions](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#keyboard-interaction) and [visible focus](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#visible-focus), but can also include verification with screen reader testing.
* [x] **Other accessibility checks** such as sufficient [color constrast](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#color-contrast), or whether the layout holds up at 200% [zoom level](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#visible-focus).

* [x] **Have we load-tested** any new pages or internal API endpoints that will receive significant traffic? [See load testing docs](./../load_tests/README.md)

### DIFF
--- a/lib/dotcom_web/templates/schedule/_timetable.html.eex
+++ b/lib/dotcom_web/templates/schedule/_timetable.html.eex
@@ -1,4 +1,4 @@
-<%= DotcomWeb.AlertView.group(alerts: @alerts, route: @route, date_time: @date_time, priority_filter: :high) %>
+<%= DotcomWeb.AlertView.group(alerts: @alerts, route: @route, date_time: @date_time, priority_filter: :high, always_show: List.wrap(@blocking_alert)) %>
 <%= render "_trip_view_filters.html", assigns %>
 
 <div class="calendar-covered m-timetable">
--- a/lib/dotcom_web/templates/schedule/_timetable.html.eex
+++ b/lib/dotcom_web/templates/schedule/_timetable.html.eex
@@ -3,9 +3,10 @@
 
 <div class="calendar-covered m-timetable">
   <%= content_tag(:div, "", [class: "calendar-cover", hidden: !@show_date_select?]) %>
-  <%= if Enum.empty?(@header_schedules) do %>
+  <%= cond do %>
+  <% Enum.empty?(@header_schedules) -> %>
     <%= render "_empty.html", date: @date, direction: Routes.Route.direction_name(@route, @direction_id), conn: @conn, error: assigns[:schedule_error] %>
-  <% else %>
+  <% true -> %>
     <% # only show scroll controllers if we have 2 or more schedules
         should_show_scroll_controls? = match?([_, _ | _], @header_schedules)
     %>
--- a/lib/dotcom_web/templates/schedule/_timetable.html.eex
+++ b/lib/dotcom_web/templates/schedule/_timetable.html.eex
@@ -4,6 +4,8 @@
 <div class="calendar-covered m-timetable">
   <%= content_tag(:div, "", [class: "calendar-cover", hidden: !@show_date_select?]) %>
   <%= cond do %>
+  <% @blocking_alert != nil -> %>
+      <%= render "_timetable_blocked.html", formatted_date: @formatted_date, alert: @blocking_alert %>
   <% Enum.empty?(@header_schedules) -> %>
     <%= render "_empty.html", date: @date, direction: Routes.Route.direction_name(@route, @direction_id), conn: @conn, error: assigns[:schedule_error] %>
   <% true -> %>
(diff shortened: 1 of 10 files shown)


lib/dotcom_web/templates/schedule/_timetable.html.eex
```diff
@@ -1,11 +1,14 @@
-<%= DotcomWeb.AlertView.group(alerts: @alerts, route: @route, date_time: @date_time, priority_filter: :high) %>
+<%= DotcomWeb.AlertView.group(alerts: @alerts, route: @route, date_time: @date_time, date: @date, priority_filter: ~w[high blocking]a) %>
```
thecristen, line 1: **suggestion**: This and the new code in in `DotcomWeb.AlertView.group/1`  took me a while to wrap my head around since `:blocking` isn't an actual [priority level](https://github.com/mbta/dotcom/blob/main/lib/alerts/priority.ex#L11). 

I think this would be easier to understand as a separate argument, maybe `show_blocking?: true`, and have `DotcomWeb.AlertView.group/1` read it in a new, separate `filter_by_blocking` (or somesuch) private function that handles the matching on blocking alerts (instead of combining it with the existing `filter_by_priority/2`).

test/dotcom/timetable_blocking_test.exs
```diff
@@ -0,0 +1,58 @@
+defmodule Dotcom.TimetableBlockingTest do
+  @moduledoc false
+  use ExUnit.Case, async: true
+  alias Dotcom.TimetableBlocking, as: TB
```
thecristen, line 4: **nit**: though we haven't yet enforced it via [this credo check](https://hexdocs.pm/credo/Credo.Check.Readability.AliasAs.html), we'd like to avoid renaming modules like this; can you please remove the `as:`?

test/dotcom_web/controllers/schedule/timetable_controller_test.exs
```diff
@@ -333,4 +333,61 @@ defmodule DotcomWeb.ScheduleController.TimetableControllerTest do
...
+        |> assign(:route, %Routes.Route{id: "CR-route"})
+
+      {:ok, %{conn: conn}}
+    end
+
+    test "assigns @blocking_alert if there's a specially tagged alert", %{conn: conn} do
+      alert =
+        Alerts.Alert.new(
+          header: "No timetable. Schedule will be available soon on the MBTA website.",
+          informed_entity: [%Alerts.InformedEntity{route: "CR-route"}],
```
thecristen, line 351: **nit**: we can use `conn.assigns.route.id` here to more explicitly tie the entity to the expected value

test/dotcom_web/views/alert_view_test.exs
```diff
@@ -185,6 +185,74 @@ defmodule DotcomWeb.AlertViewTest do
...
+          alerts: [alert],
+          route: @route,
+          stop?: false,
+          show_empty?: false,
+          date_time: DateTime.utc_now()
+        )
+
+      text = safe_to_string(response)
+
+      refute text =~ "View PDF Timetable on the MBTA website."
```
thecristen, line 207: Should this be refuting it's equal to `Dotcom.TimetableBlocking.pdf_available_text()` directly?

lib/dotcom_web/templates/schedule/_timetable.html.eex
paulswartz: Good idea. I ended up going with an `always_show:` option which takes a list of alerts we should always show, even if they were otherwise filtered out.

paulswartz:
@thecristen I made the changes as `fixup!` commits; I'll clean up the history after approval.

lib/dotcom_web/views/alert_view.ex
```diff
@@ -49,17 +49,8 @@ defmodule DotcomWeb.AlertView do
...
-    alert_ids = Enum.map(alerts, & &1.id)
-    ashmont_shuttle_alert_ids = ["519314", "529291"]
-
-    if Enum.all?(ashmont_shuttle_alert_ids, &Enum.member?(alert_ids, &1)) do
-      # remove the second one
-      Enum.reject(alerts, &(&1.id == "529291"))
-    else
-      alerts
-    end
+    Enum.uniq_by(alerts, & &1.id)
```
thecristen, line 53: Thanks!!

thecristen (approved):
Love it!! Thanks so much 👍🏼

PR #2303, opened 2025-01-06
Swap in new trip planner with a monitor update

anthonyshull:
https://app.asana.com/0/555089885850811/1209078853452837/f

I also did https://app.asana.com/0/555089885850811/1209078853452832/f as that needs to be shipped with this change.

Screenshot Capture - 2025-01-07 - 10-14-55


thecristen (commented):
It works! 

I notice you deleted the old trip plan controller but stopped short at removing corresponding trip plan view or the associated templates, how come?

PR #2312, opened 2025-01-07
small changes and restart at midnight

anthonyshull:
https://app.asana.com/0/555089885850811/1208924776345287/f

I made some small efficiency changes to the monitor and made it so that every night at midnight the all-scenarios script restarts. I couldn't find any correlation between CPU usage or memory usage and degradation of performance. So, I'm hoping simply restarting the script every night works. We already don't run it during off-service hours so no data will be missed.


joshlarson (approved):
Let's give it a try!

PR #2314, opened 2025-01-08
fix: Have the map show the selected itinerary, rather than a random member of the group

joshlarson:
## Before
Screenshot 2025-01-08 at 3 16 23 PM
(Note that the selected itinerary uses the Orange Line, but the line on the map is the Green Line color)

## After
Screenshot 2025-01-08 at 3 16 40 PM

---

No Ticket.

### DIFF
--- a/lib/dotcom_web/live/trip_planner.ex
+++ b/lib/dotcom_web/live/trip_planner.ex
@@ -204,9 +204,25 @@ defmodule DotcomWeb.Live.TripPlanner do
   def handle_event("select_itinerary", %{"index" => index}, socket) do
     index = String.to_integer(index)
 
+    new_map = %{
+      lines:
+        itinerary_groups_to_lines(
+          socket.assigns.results.itinerary_groups,
+          socket.assigns.results.itinerary_group_selection,
+          index
+        ),
+      points:
+        itinerary_groups_to_points(
+          socket.assigns.results.itinerary_groups,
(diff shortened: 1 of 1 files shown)


lib/dotcom_web/live/trip_planner.ex
```diff
@@ -204,9 +204,25 @@ defmodule DotcomWeb.Live.TripPlanner do
   def handle_event("select_itinerary", %{"index" => index}, socket) do
     index = String.to_integer(index)
 
+    new_map = %{
```
joshlarson, line 207: There is a fundamental issue here, which is that `lines` and `points` have to be re-computed at various different points, and forgetting to do that somewhere introduces bugs like this one. (It also creates a bunch of duplicative code for re-calculating `lines` and `points` any time they ought to change)

I think the Right™ solution is to calculate `lines` and `points` (and a few other things, like `pins`) at the top of `render`, based on the result set and user interactions, and not try to have this one-to-one mapping between components-on-screen and top-level state fields.

I'd rather actually tackle that refactor as a follow-up, since fixing this bug in this way was pretty easy, and because trying to refactor the state right now risks conflicting with other ongoing trip planner work.

thecristen: ⭐ Thanks for this! I agree with your take on what an improvement on this would be.

thecristen (approved):
🤩

PR #2231, opened 2024-11-19
feat: Add swap-to-and-from button to Trip Planner Form

joshlarson:
#### Summary of changes

**Asana Ticket:** [Add switch origin/destination button](https://app.asana.com/0/1205718271156548/1208603342838393/f)

---

#### General checks
* [ ] **Are the changes organized into self-contained commits with descriptive and well-formatted commit messages?** This is a good practice that can facilitate easier reviews.
* [ ] **Testing.** Do the changes include relevant passing updates to tests? This includes updating screenshots. Preferably tests are run locally to verify that there are no test failures created by these changes, before opening a PR.
* [ ] **Tech debt.** Have you checked for tech debt you can address in the area you're working in? This can be a good time to address small issues, or create Asana tickets for larger issues.

#### New UI, or substantial UI changes
* [ ] **Cross-browser compatibility** is less of an issue now that we're no longer supporting IE, but changes still need to work as expected in Safari, Chrome, and Firefox.
* [ ] **Are interactive elements accessible?** This includes at minimum having [relevant keyboard interactions](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#keyboard-interaction) and [visible focus](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#visible-focus), but can also include verification with screen reader testing.
* [ ] **Other accessibility checks** such as sufficient [color constrast](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#color-contrast), or whether the layout holds up at 200% [zoom level](https://thinkdobecreate.com/articles/4-required-tests-before-shipping-new-features/#visible-focus).

#### New endpoints, or non-trivial changes to current endpoints
* [ ] **Have we load-tested** any new pages or internal API endpoints that will receive significant traffic? [See load testing docs](./../load_tests/README.md)
* [ ] **If this change involves routes,** does it work correctly with pertinent "unusual" routes such as the combined Green Line, Silver Line, Foxboro commuter rail, and single-direction bus routes like the 170?

PR #2317, opened 2025-01-09
feat(Live.TripPlanner): Add switch origin/destination button

thecristen:
#### Summary of changes

**Asana Ticket:** [Add switch origin/destination button](https://app.asana.com/0/555089885850811/1208603342838393/f)

Building off of Josh's earlier work in #2231 and elsewhere, implements switching the locations on the form, thus switching the displayed markers on the map, and rerunning the trip planner with new results.

This also includes some cleanup on the `AlgoliaAutocomplete` Phoenix JS hook. Now instead of triggering 4 "change" events for each location input change, we more elegantly fire off _one_ (1) event for the LiveView to handle (just another `"input_form_change"` event, transmitting data selected from the location dropdown).

### DIFF
--- a/lib/dotcom_web/components/trip_planner/input_form.ex
+++ b/lib/dotcom_web/components/trip_planner/input_form.ex
@@ -25,35 +25,24 @@ defmodule DotcomWeb.Components.TripPlanner.InputForm do
     <section class={["rounded px-xl py-lg lg:px-2xl lg:py-xl bg-charcoal-90", @class]}>
       <.form
         :let={f}
-        class="md:grid md:grid-cols-2 gap-x-8 gap-y-2"
+        class="flex flex-col md:grid md:grid-cols-[1fr_max-content_1fr] md:gap-x-lg gap-y-sm pt-md"
         id="trip-planner-input-form"
         for={@changeset}
         method="get"
         phx-change="input_form_change"
         phx-submit="input_form_submit"
       >
-        <fieldset :for={field <- [:from, :to]} id={"trip-planner-locations-#{field}"} class="mb-sm">
-          <legend class="text-charcoal-40 m-0 py-sm">{Phoenix.Naming.humanize(field)}</legend>
-          <.algolia_autocomplete
-            config_type="trip-planner"
-            placeholder={"Enter #{if(field == :from, do: "an origin", else: "a destination")} location"}
-            id={"trip-planner-input-form--#{field}"}
+        <.location_search_box name="trip-planner-input-form--from" field={f[:from]} />
+        <div class="self-end rotate-90 -mt-xs -mb-md md:mt-0 md:mb-0 md:rotate-0">
+          <button
+            type="button"
+            class="px-xs md:py-3 bg-transparent fill-brand-primary hover:fill-black"
           >
-            <.inputs_for :let={location_f} field={f[field]} skip_hidden={true}>
-              <input
-                :for={subfield <- InputForm.Location.fields()}
-                type="hidden"
-                class="location-input"
-                id={location_f[subfield].id}
-                value={location_f[subfield].value}
-                name={location_f[subfield].name}
-              />
-            </.inputs_for>
-            <.error_container :for={{msg, _} <- f[field].errors}>
-              {msg}
-            </.error_container>
-          </.algolia_autocomplete>
-        </fieldset>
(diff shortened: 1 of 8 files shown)


lib/dotcom_web/live/trip_planner.ex
```diff
@@ -368,7 +409,9 @@ defmodule DotcomWeb.Live.TripPlanner do
   # on "Leave at" or "Arrive by", the actual value of "datetime"
   # doesn't always appear in params. When that happens, we want to set
   # "datetime" to a reasonable default.
-  defp add_datetime_if_needed(%{"datetime_type" => "now"} = params), do: params
+  defp add_datetime_if_needed(%{"datetime_type" => "now"} = params),
+    do: params |> Map.put("datetime", nearest_5_minutes())
```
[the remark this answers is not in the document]
joshlarson: I think the behavior we want in this case would be `now()` rather than `nearest_5_minutes()`.

lib/dotcom_web/components/trip_planner/input_form.ex
```diff
@@ -108,6 +104,32 @@ defmodule DotcomWeb.Components.TripPlanner.InputForm do
...
+        <.inputs_for :let={location_f} field={@field} skip_hidden={true}>
+          <input
+            :for={subfield <- @location_keys}
+            type="hidden"
+            id={location_f[subfield].id}
+            value={location_f[subfield].value}
+            name={location_f[subfield].name}
+          />
+        </.inputs_for>
+        <.feedback :for={{msg, _} <- @field.errors} :if={used_input?(@field)} kind={:error}>
```
joshlarson, line 125: This change is creating some spurious errors (as soon as you enter an origin, it complains that there's no destination, and vice versa), but it's also conflicting with `main`. I wonder if resolving the conflicts with `main` would fix this.

(I think resolving the conflicts with `main` would actually just main using exactly this version of this file, but with the new "Enter [an origin|a destination] location" placeholders, since the big change to `main` was re-adding the error feedback component, which you also added here.)

thecristen: I think resolving the conflict fixed it?!? Feel free to try again :)

lib/dotcom_web/components/trip_planner/input_form.ex
```diff
@@ -111,6 +104,32 @@ defmodule DotcomWeb.Components.TripPlanner.InputForm do
...
 
+  defp location_search_box(assigns) do
+    assigns =
+      assigns
+      |> assign(:location_keys, InputForm.Location.fields())
+
+    ~H"""
+    <fieldset class="mb-sm -mt-md" id={"#{@name}-wrapper"}>
+      <legend class="text-charcoal-40 m-0 py-sm">{Phoenix.Naming.humanize(@field.field)}</legend>
+      <.algolia_autocomplete config_type="trip-planner" placeholder="Enter a location" id={@name}>
```
joshlarson, line 115:
```suggestion
      <.algolia_autocomplete config_type="trip-planner" placeholder={@placeholder} id={@name}>
```

We lost [this change](https://github.com/mbta/dotcom/pull/2300/files#diff-6c63a22d4645df0238c7ebc4e7754b74177b5b8ff240cfde7212fadb5a74dadfR39), which is helpful for the detached search box experience.

lib/dotcom_web/components/trip_planner/input_form.ex
```diff
@@ -25,35 +25,28 @@ defmodule DotcomWeb.Components.TripPlanner.InputForm do
...
-            <.error_container :for={{msg, _} <- f[field].errors}>
-              {msg}
-            </.error_container>
-          </.algolia_autocomplete>
-        </fieldset>
+            <span class="sr-only">Swap origin and destination locations</span>
+            <.icon class="h-6 w-6 rotate-90 md:rotate-0" name="right-left" />
+          </button>
+        </div>
+        <.location_search_box name="trip-planner-input-form--to" field={f[:to]} />
```
joshlarson, line 49:
```suggestion
        <.location_search_box name="trip-planner-input-form--to" field={f[:to]} placeholder="Enter a destination location" />
```

joshlarson (changes requested):
Love this!

I'm ready to approve once the placeholder text is back.

thecristen:
Yikes, thanks for catching!
image

PR #2319, opened 2025-01-09
fix: Decrease min zoom on the map to allow riders to zoom out further

joshlarson:
A `minZoom` of 12 doesn't allow riders to zoom out very far at all, which means that trips over a certain distance won't fit on the screen. 

(Lower zoom corresponds to a bigger map - a zoom level of 1 is the entire planet)

## Before
Screenshot 2025-01-09 at 12 06 04 PM

## After
Screenshot 2025-01-09 at 12 04 05 PM

---

No ticket.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -44,7 +44,7 @@ end
 config :mbta_metro, :map, %{
   center: [-71.0589, 42.3601],
   maxZoom: 18,
-  minZoom: 9,
+  minZoom: 8,
   style: %{
     "version" => 8,
     "sources" => %{
--- a/config/config.exs
+++ b/config/config.exs
@@ -44,7 +44,7 @@ end
 config :mbta_metro, :map, %{
   center: [-71.0589, 42.3601],
   maxZoom: 18,
-  minZoom: 12,
+  minZoom: 9,
   style: %{
     "version" => 8,
     "sources" => %{


config/config.exs
```diff
@@ -44,7 +44,7 @@ end
 config :mbta_metro, :map, %{
   center: [-71.0589, 42.3601],
   maxZoom: 18,
-  minZoom: 12,
+  minZoom: 9,
```
thecristen, line 47:
```suggestion
  minZoom: 8,
```

I feel like we could go further here, thinking about even longer trips on the smallest screen. Though I'm not sure it's even worth enforcing a minZoom at all...

thecristen (changes requested):

config/config.exs
joshlarson: Yeah good point. Logan to Providence still doesn't fit with a minZoom of 9.

I agree about not enforcing a minZoom. I wonder what happens if we just drop that param...

joshlarson: Screenshot 2025-01-09 at 2 36 15 PM

Eeeeeehhhhhh, this isn't the best thing in the world.

joshlarson: Screenshot 2025-01-09 at 2 38 10 PM

Zoom level 7

Screenshot 2025-01-09 at 2 39 29 PM

thecristen (approved):

PR #2321, opened 2025-01-09
fix(TripPlanner): Restore A and B on desktop in search boxes

joshlarson:
## Before

Screenshot 2025-01-09 at 2 16 33 PM

## After

Screenshot 2025-01-09 at 2 15 49 PM

---
No ticket.


thecristen (approved):
🚀 woohoo thank you!!

PR #2322, opened 2025-01-09
fix: change metro map config tile URL for prod

thecristen:
#### Summary of changes

**Asana Ticket:** [Map tiles should load on production environment](https://app.asana.com/0/555089885850811/1209117947602740/f)

Possible fix! I'm not sure we'll really know til we try it in prod.


joshlarson (approved):
🚀

PR #2325, opened 2025-01-13
Change the tab order of itineraries and the map

anthonyshull:
https://app.asana.com/0/555089885850811/1209110864370903/f

I just changed the order of the elements in the DOM.

### DIFF
--- a/lib/dotcom_web/live/trip_planner.ex
+++ b/lib/dotcom_web/live/trip_planner.ex
@@ -76,7 +76,7 @@ defmodule DotcomWeb.Live.TripPlanner do
       <div>
         <.results_summary class="mt-2 mb-6" changeset={@input_form.changeset} results={@results} />
         <div class={[
-          "flex flex-col gap-4 md:flex-row md:gap-7"
+          "flex flex-col-reverse gap-4 md:flex-row md:gap-7"
         ]}>
           <.results
             :if={Enum.count(@results.itinerary_groups) > 0 || @results.loading?}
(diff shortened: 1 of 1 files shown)


joshlarson (changes requested):
This almost works, but makes the order wrong on mobile. I think replacing `flex-col` with `flex-col-reverse` on the parent `div` would do the trick.

Related: there's an `md:order-last` on the map component, which I think you could remove now, since it's now DOM-ordered-last 😅

lib/dotcom_web/live/trip_planner.ex
```diff
@@ -76,27 +76,25 @@ defmodule DotcomWeb.Live.TripPlanner do
...
+            results={@results}
+          />
           <.live_component
             module={MbtaMetro.Live.Map}
             id="trip-planner-map"
             class={[
-              "md:order-last md:sticky md:top-4",
-              "h-64 md:h-[32rem] w-full",
-              @results.itinerary_group_selection == nil && "hidden md:block",
-              @results.itinerary_group_selection != nil && "block"
```
joshlarson, line 88: I'm pretty sure we still want this logic, since we only want the map to show up if you're looking at a specific itinerary. (Unless that part of the design changed?)

joshlarson (changes requested):
Unless I missed something, [we still don't want the map to render on mobile unless an itinerary is selected](https://github.com/mbta/dotcom/pull/2325/files#r1913505188)

joshlarson (approved):
🗺️

PR #2326, opened 2025-01-13
only submit if valid

anthonyshull:
No ticket. We noticed the map was only taking up part of the page. This is due to the fact that we were calling submit changeset on mount and it was setting loading to true. Now we only try to load itineraries if the changeset is valid.

### DIFF
--- a/lib/dotcom_web/live/trip_planner.ex
+++ b/lib/dotcom_web/live/trip_planner.ex
@@ -426,16 +426,24 @@ defmodule DotcomWeb.Live.TripPlanner do
   #
   # - Update the input form state with the new changeset
   # - Update the map state with the new pins
-  # - Set the results state to loading
-  # - Start an async operation to get the itinerary groups
+  # - If the changeset is valid:
+  #   - Set the results state to loading
+  #   - Start an async operation to get the itinerary groups
   defp submit_changeset(socket, changeset) do
     new_changeset = Map.put(changeset, :action, :submit)
 
-    socket
-    |> assign(:input_form, Map.put(@state.input_form, :changeset, new_changeset))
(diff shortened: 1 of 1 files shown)


lib/dotcom_web/live/trip_planner.ex
```diff
@@ -426,16 +426,24 @@ defmodule DotcomWeb.Live.TripPlanner do
...
+      socket
+      |> assign(:input_form, Map.put(@state.input_form, :changeset, new_changeset))
+      |> assign(:map, Map.put(@state.map, :pins, input_form_to_pins(new_changeset)))
+
+    if new_changeset.valid? do
+      new_socket
+      |> assign(:results, Map.put(@state.results, :loading?, true))
+      |> start_async("get_itinerary_groups", fn -> get_itinerary_groups(new_changeset) end)
+    else
+      new_socket
```
joshlarson, line 445: I think this needs to explicitly clear the results if they're already set. Something like 👇 maybe?
```suggestion
      new_socket
      |> assign(:results, @state.results)
```

joshlarson (changes requested):
We also need to [clear the results](https://github.com/mbta/dotcom/pull/2326/files#r1913514370) if the input changes to something invalid.

lib/dotcom_web/live/trip_planner.ex
anthonyshull: > I think this needs to explicitly clear the results if they're already set. Something like 👇 maybe?

That will make the existing results disappear even if the new form submission isn't valid. Is that what we want?

joshlarson: Haha great question! I _think_ the answer is yes, but let's ask Chase.

joshlarson (approved):

PR #2329, opened 2025-01-14
refactor(TripPlan.ItineraryGroups): improve handling for "arrive by" selection

thecristen:
#### Summary of changes

**Asana Ticket:** [TP | Show appropriate itineraries for the "Arrive by" selection](https://app.asana.com/0/555089885850811/1209129102394304/f) and [TP | Alternate time selection should be reflected in the itinerary summary](https://app.asana.com/0/555089885850811/1209129102394297/f)

The main crux is that, when the user selects "arrive by" instead of the default or "leave at" time, many of our code's assumptions turned out to be false... so this PR makes a slew of changes around that. 

- when limiting the number of itineraries for each group, we don't want the first N itineraries if using "arrive by" -- we want the last
- move (and rename) the 'depart at buttons' to same level as itinerary_summary and itinerary_detail. this one just felt more legible to me
- fix which button from the group is the initially active one (as it will be the last button, not first, for "arrive by")
- fix the displayed summary so that it reflects the chosen (via button click) itinerary from the group, instead of always reflecting the representative itinerary
- adjust the display text in various ways, to better describe 'arrive by' results
- adjust `ItineraryGroups.from_itineraries/2` to take 5 groups *after* all the sorting
- add more docs all around


lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -8,48 +8,34 @@ defmodule Dotcom.TripPlan.ItineraryGroups do
...
-  def from_itineraries(itineraries) do
+  @max_per_group 4
+
+  @doc """
+  From a large list of itineraries, collect them into 5 groups of at most
+  #{@max_per_group} itineraries each, sorting the groups in favor of tagged
+  groups first
+  """
+  @spec from_itineraries([Itinerary.t()], boolean()) :: [ItineraryGroup.t()]
+  def from_itineraries(itineraries, take_from_end? \\ false) do
```
joshlarson, line 23: Nit: The interface for this function would be clearer if `take_from_end?` were passed in as a keyword arg or an options map, rather than a plain old boolean.

That's shown off pretty clearly by how this function is invoked in the test 👇 
```elixir
sorted_itineraries
|> ItineraryGroups.from_itineraries(true)
```

What `true` means in this context is pretty tough to tease out. It would be a lot clearer if the function were invoked like

```elixir
sorted_itineraries
|> ItineraryGroups.from_itineraries(:take_from_end)
```

Or

```elixir
sorted_itineraries
|> ItineraryGroups.from_itineraries(take_from_end: true)
```

Or

```elixir
sorted_itineraries
|> ItineraryGroups.from_itineraries(
     %{take_from_end: true}
   )
```

test/dotcom/trip_plan/itinerary_group_test.exs
```diff
@@ -0,0 +1,88 @@
...
+        itineraries: itineraries,
+        representative_index: representative_index,
+        representative_time: start_or_stop
+      }
+
+      # EXERCISE
+      text = ItineraryGroup.options_text(group)
+
+      # VERIFY
+      if Enum.count(itineraries) == 2 do
```
joshlarson, line 43: This is actually four different tests, but the way it's written, ExUnit will just non-deterministically pick one to run and ignore the other three*.

I'd urge splitting this up and getting rid of the `if` statements. As a bonus, the assertions could then assert on the more complete `"Similar trips depart at"`/`"Similar trip arrives at"`/etc strings, which is both a slightly stronger assertion and (at least per my sensibilities) easier to tell what the assertion is.

---

*Best case scenario, this test runs each of its embedded test cases with some non-zero probability, and I think that is the case here (1/3 odds each for the plural cases and 1/6 odds each for the singular cases), but I've seen this kind of test construction where some test cases are actually executed with probability zero, and I wouldn't guarantee that that's not the case here.

joshlarson: Oh one additional point here is that if the use cases don't reliably run every time `mix test` is run, then it's a lot harder to refactor. My typical refactor workflow is to do a little refactor, run the tests, and make sure everything passes. If `mix test` succeeds, then I assume I didn't break anything, and move on.

If a use case is only executed 17% of the time, then I'll advance on average 6 steps past when I made a breaking change before `mix test` will tell me that I did something wrong, which makes it a lot harder to figure out where I went astray.

test/dotcom/trip_plan/itinerary_group_test.exs
```diff
@@ -0,0 +1,88 @@
...
+
+      stop_times =
+        %ItineraryGroup{
+          itineraries: itineraries,
+          representative_time: :stop
+        }
+        |> ItineraryGroup.all_times()
+
+      # VERIFY
+      assert start_times != stop_times
```
joshlarson, line 85: Nit (non-blocking): It doesn't look like this test provides any guarantee that `start_times` and `stop_times` will be different.

This is non-blocking, because having all four start times line up with all four stop times is low-enough probability that I doubt it'll ever happen, but we did run into [some issues](https://github.com/mbta/dotcom/pull/2260/files#r1882985986) where a test implicitly required that Faker was returning unique data, which was leading to some very hard-to-debug flaky tests.

lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -69,22 +55,43 @@ defmodule Dotcom.TripPlan.ItineraryGroups do
     grouped_itineraries
   end
 
-  defp to_summarized_group(grouped_itineraries) do
-    %{
-      itineraries: ItineraryTag.sort_tagged(grouped_itineraries),
-      summary: summary(grouped_itineraries)
-    }
+  defp limit_itinerary_count(itineraries, take_from_end?) do
```
joshlarson, line 58: Not sure how this suggestion will play with [this other comment above](https://github.com/mbta/dotcom/pull/2329/files#r1919161002), but this could be accomplished with pattern matching

```elixir
defp limit_itinerary_count(itineraries, true), do: Enum.take(itineraries, -@max_per_group)
defp limit_itinerary_count(itineraries, false), do: Enum.take(itineraries, @max_per_group)
```

Or similarly
```elixir
defp limit_itinerary_count(itineraries, %{take_from_end: true}), do: Enum.take(itineraries, -@max_per_group)
defp limit_itinerary_count(itineraries, _), do: Enum.take(itineraries, @max_per_group)
```

lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -69,22 +55,43 @@ defmodule Dotcom.TripPlan.ItineraryGroups do
...
 
-  defp limit_itinerary_count(itineraries) do
-    Enum.take(itineraries, @max_per_group)
+  # Summarize a group of chronologically-sorted itineraries, noting the
+  # - representative_index: Which itinerary should represent the group? It will
+  #   be either the first or final one.
+  # - representative_time: An itinerary departs at the :start time and arrives
+  #   by the :stop time. This denotes which of those is relevant to the group.
+  defp to_group(grouped_itineraries, take_from_end?) do
+    representative_index = if(take_from_end?, do: Enum.count(grouped_itineraries) - 1, else: 0)
```
joshlarson, line 72: **Praise**: Oooh I like this. I had in my mind that the decision of which button should start out highlighted would happen in the LiveView, but this is a better place for it.

joshlarson (changes requested):
This all looks great! And the UX is spot-on!

I think the only comment that I made that's blocking my approval is [this one about the tests](https://github.com/mbta/dotcom/pull/2329#discussion_r1919169432). I'd be interested in your thoughts on the others, but I would be happy for those to either be follow-ups or disputed.

lib/dotcom/trip_plan/itinerary_groups.ex
thecristen: My first (unpublished) iteration did that, but turns out it was way easier to handle here

test/dotcom/trip_plan/itinerary_group_test.exs
thecristen: Yeah, too low probability IMO

lib/dotcom/trip_plan/itinerary_groups.ex
thecristen: Ended up using keyword here, mostly because we often use that approach elsewhere in Dotcom, thanks!

test/dotcom/trip_plan/itinerary_group_test.exs
thecristen: You did the math! Okay, I've separated out these tests.

lib/dotcom/trip_plan/itinerary_group.ex
```diff
@@ -0,0 +1,73 @@
...
+        %__MODULE__{
+          representative_index: representative_index,
+          representative_time: start_or_stop
+        } =
+          group
+      ) do
+    {_, other_times} =
+      group
+      |> all_times()
+      |> List.pop_at(representative_index)
```
joshlarson, line 50: Nit (non-blocking): I think this is just the same as `List.delete_at`

```suggestion
    other_times =
      group
      |> all_times()
      |> List.delete_at(representative_index)
```

lib/dotcom/trip_plan/itinerary_group.ex
```diff
@@ -0,0 +1,73 @@
...
+
+    formatted_times =
+      other_times
+      |> Enum.map_join(", ", &Timex.format!(&1, "%-I:%M", :strftime))
+
+    "Similar #{phrase} #{formatted_times}"
+  end
+
+  defp options_phrase(:stop, 1), do: "trip arrives by"
+  defp options_phrase(:stop, _), do: "trips arrive by"
```
joshlarson, line 62: Nit (non-blocking): "Arrive by" is a constraint, not a description, but (if I'm reading the code right), these strings are descriptive rather than constraining. 

What I mean is that even if the option that a user selected is "Arrive by", it still feels more natural when describing a trip to say "Similar trips arrive at 3:23, 3:53", rather than "Similar trips arrive by..."

```suggestion
  defp options_phrase(:stop, 1), do: "trip arrives at"
  defp options_phrase(:stop, _), do: "trips arrive at"
```

lib/dotcom/trip_plan/itinerary_groups.ex
```diff
@@ -8,48 +8,34 @@ defmodule Dotcom.TripPlan.ItineraryGroups do
...
+      %ItineraryGroup{summary: %{tag: tag}} ->
         Enum.find_index(ItineraryTag.tag_priority_order(), &(&1 == tag))
-
-      _ ->
-        -1
     end)
+    |> Enum.take(5)
   end
 
+  def max_per_group, do: @max_per_group
```
joshlarson, line 37: Nit (non-blocking): This is bugging me a bit - the only reason it exists is to expose this value to the tests, but I feel like the tests can afford to be explicit about the fact that the itinerary-count-limit is 4.

lib/dotcom/trip_plan/itinerary_group.ex
```diff
@@ -0,0 +1,73 @@
...
+          summarized_legs: [summarized_leg()],
+          tag: String.t(),
+          total_cost: non_neg_integer(),
+          walk_distance: float()
+        }
+
+  @type t :: %__MODULE__{
+          itineraries: [Itinerary.t()],
+          representative_index: non_neg_integer(),
+          representative_time: :start | :stop,
```
joshlarson, line 29: Nit (non-blocking): The name `representative_time` makes me think that the value stored here will be a time, but it's actually an atom indicating where to look for an itinerary's time. Maybe `representative_time_key`?

lib/dotcom_web/components/trip_planner/results.ex
```diff
@@ -43,32 +45,27 @@ defmodule DotcomWeb.Components.TripPlanner.Results do
...
   end
 
+  # When an itinerary group is selected, show a list of group summaries, each
+  # optionally displaying a tag and description of alternate times
   defp itinerary_panel(%{results: %{itinerary_group_selection: nil}} = assigns) do
     ~H"""
     <div class="flex flex-col gap-4">
       <div
-        :for={{%{summary: summary}, index} <- Enum.with_index(@results.itinerary_groups)}
+        :for={{group, index} <- Enum.with_index(@results.itinerary_groups)}
```
joshlarson, line 54: **Praise** This way of doing this IS a decent bit easier to read. I think I got a bit pattern-happy when I built it the other way.

lib/dotcom/trip_plan/itinerary_group.ex
```diff
@@ -0,0 +1,73 @@
...
+  @type summarized_leg :: %{
+          routes: [Routes.Route.t()],
+          walk_minutes: non_neg_integer()
+        }
+
+  @type summary :: %{
+          accessible?: boolean() | nil,
+          duration: non_neg_integer(),
+          start: DateTime.t(),
+          stop: DateTime.t(),
```
joshlarson, line 19: Nit (non-blocking): The names `start` and `stop` don't convey that these are times. And `stop` more frequently refers to a location where a bus allows passengers to embark and disembark. I would advocate for `start_time` and `end_time`.

(Note: This is only non-blocking because I know that the names aren't being introduced in this PR, and were like this already, and that fixing them could lead to PR bloat.)

joshlarson (approved):
I made a bunch of non-blocking comments, but this looks good! I turned off auto-merge so that Github won't merge this out from under your nose, just in case you do want to address some or all of these comments, but by all means merge and address later (or not at all) if that makes more sense.


Most are code-cosmetic, but [one is rider-facing](https://github.com/mbta/dotcom/pull/2329/files#r1924393556).

PR #2327, opened 2025-01-13
feat: Add module that retrieves system-status relevant alerts

joshlarson:
#### Summary of changes

This adds a `Dotcom.SystemStatus` module, as well as `Dotcom.SystemStatus.Alerts`, which combined will retrieve a list of all alerts that are either active now, or will become active later today (or whatever day is passed in as `now`).

This also adds a live view that I was using to test this out. I see that live view morphing into one that previews the system status widget as we get closer to done on this project.

Some questions that I'd be interested in noodling on:
- Is `Dotcom.SystemStatus.Alerts` the right place for that filtering logic, or would it make more sense for some or all of that to live in [the `Alerts` namespace](https://github.com/mbta/dotcom/tree/ff3a464e68853a65872f63decac0c63565a9c4b0/lib/alerts)?
- Did I miss any corner cases where an alert should or shouldn't show up?
- I'm open to any and all renaming suggestions.
- ...other?

---

**Asana Ticket:** [System Status | Add ability to get all future alerts for the day](https://app.asana.com/0/555089885850811/1209091199210564/f)

---

Blocks:
- https://github.com/mbta/dotcom/pull/2330

PR #2330, opened 2025-01-14
feat: Create a data structure to populate the system status widget

joshlarson:
#### Summary of changes

**Asana Ticket:** [System Status Flowchart | Implement the bulk of the flowchart](https://app.asana.com/0/555089885850811/1209229382061161/f)
**Parent Ticket:** [System Status | Add module that implements the system-status flowchart](https://app.asana.com/0/555089885850811/1209091199210566/f)

---

Depends on:
- https://github.com/mbta/dotcom/pull/2327

### DIFF
--- a/lib/dotcom/system_status.ex
+++ b/lib/dotcom/system_status.ex
@@ -36,4 +36,11 @@ defmodule Dotcom.SystemStatus do
     |> SystemStatus.Alerts.for_day(datetime)
     |> SystemStatus.Alerts.filter_relevant()
   end
+
+  @doc """
+  Returns a map indicating the subway status for each of the subway lines.
+  """
+  def subway_status() do
+    subway_alerts_for_today() |> SystemStatus.Subway.subway_status(Timex.now())
+  end
 end
--- a/lib/dotcom/system_status.ex
+++ b/lib/dotcom/system_status.ex
@@ -1,13 +1,8 @@
 defmodule Dotcom.SystemStatus do
   @moduledoc """
-  The system status feature is a widget that's intended to appear on the
-  homepage, as well as eventually several other places throughout the site.
-  The widget will show statuses for each of the subway lines (Red, Orange,
-  Green, Blue), which might look like "Normal Service", or information
-  about current or upcoming alerts.
-
-  This module is responsible for reporting data in a format that can easily
-  be plugged into a component on the frontend.
+  This module exists to provide a simple view into the status of the
+  subway system, for each line, showing whether its running normally,
+  or whether there are alerts that impact service.
   """
 
   alias Dotcom.SystemStatus
--- a/lib/dotcom/system_status/groups.ex
+++ b/lib/dotcom/system_status/groups.ex
@@ -7,7 +7,7 @@ defmodule Dotcom.SystemStatus.Groups do
 
   @lines ["Blue", "Orange", "Red", "Green"]
   @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
-  @routes ["Blue", "Orange", "Red"] ++ @green_line_branches
+  @routes ["Blue", "Orange", "Red", "Mattapan"] ++ @green_line_branches
 
   def groups(alerts, now) do
     grouped_alerts = Map.new(@routes, &{&1, alerts_for_line(alerts, &1)})
@@ -26,6 +26,7 @@ defmodule Dotcom.SystemStatus.Groups do
       %{route_id: route_id, sub_routes: [], statuses: statuses}
     end)
     |> combine_green_line_branches()
+    |> combine_mattapan_with_red_line()
     |> sort_routes_and_sub_routes()
   end
 
@@ -168,18 +169,53 @@ defmodule Dotcom.SystemStatus.Groups do
     end)
   end
 
+  defp combine_mattapan_with_red_line(statuses_by_route) do
+    {mattapan_entries, other_entries} =
+      statuses_by_route
+      |> Enum.split_with(fn %{route_id: route_id} -> route_id == "Mattapan" end)
+
+    new_mattapan_entries =
+      case mattapan_entries do
+        [%{statuses: [%{description: "Normal Service"}]}] ->
+          []
+
+        _ ->
+          mattapan_entries
+          |> Enum.map(fn %{statuses: statuses} ->
+            %{route_id: "Red", sub_routes: ["Mattapan"], statuses: statuses}
+          end)
+      end
+
+    other_entries ++ new_mattapan_entries
+  end
+
   defp sort_routes_and_sub_routes(entries) do
     line_indexes = @lines |> Enum.with_index() |> Map.new()
 
     entries
-    |> Enum.sort_by(fn %{route_id: route_id, statuses: [%{description: description} | _]} ->
+    |> Enum.sort_by(fn %{
+                         route_id: route_id,
+                         sub_routes: sub_routes,
+                         statuses: [%{description: description} | _]
+                       } ->
       description_sort_order =
         case description do
           "Normal Service" -> 1
           _ -> 0
         end
 
-      {line_indexes |> Map.get(route_id), description_sort_order}
+      base_route_sort_order =
+        case sub_routes do
+          [] -> 0
+          _ -> 1
+        end
+
+      {
+        line_indexes |> Map.get(route_id),
+        base_route_sort_order,
+        description_sort_order,
+        sub_routes
+      }
     end)
   end
 end
--- a/lib/dotcom/system_status/groups.ex
+++ b/lib/dotcom/system_status/groups.ex
@@ -1,4 +1,4 @@
-defmodule Dotcom.SystemStatus.Grouping do
+defmodule Dotcom.SystemStatus.Groups do
   @moduledoc """
   A module that groups alerts into statuses for the system status widget.
   """
@@ -9,7 +9,7 @@ defmodule Dotcom.SystemStatus.Grouping do
   @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
   @routes ["Blue", "Orange", "Red"] ++ @green_line_branches
 
-  def grouping(alerts, now) do
+  def groups(alerts, now) do
     grouped_alerts = Map.new(@routes, &{&1, alerts_for_line(alerts, &1)})
 
     @routes
--- a/lib/dotcom/system_status/groups.ex
+++ b/lib/dotcom/system_status/groups.ex
@@ -5,6 +5,8 @@ defmodule Dotcom.SystemStatus.Groups do
 
   alias Alerts.Alert
 
+  # &groups/2 returns an ordered data structure, sorted in the order
+  # given by `@lines`.
   @lines ["Blue", "Orange", "Red", "Green"]
   @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
   @routes ["Blue", "Mattapan", "Orange", "Red"] ++ @green_line_branches
@@ -30,6 +32,9 @@ defmodule Dotcom.SystemStatus.Groups do
     |> sort_routes_and_sub_routes()
   end
 
+  # Given `alerts` and `line_id`, filters out only the alerts
+  # applicable to the given line, using the alert's "informed
+  # entities".
   defp alerts_for_line(alerts, line_id) do
     alerts
     |> Enum.filter(fn %Alert{informed_entity: informed_entity} ->
@@ -41,10 +46,19 @@ defmodule Dotcom.SystemStatus.Groups do
     end)
   end
 
+  # Maps a list of alerts to a list of statuses, where a status is a
+  # simple structure with a route, a description, and a few additional
+  # fields that determine how it will render in the frontend.
+  defp alerts_to_statuses(alerts, time)
+
+  # If there are no alerts, then we want a single status indicating
+  # "Normal Service".
   defp alerts_to_statuses([], _time) do
     [%{description: "Normal Service", time: nil}]
   end
 
+  # If there are alerts, then create a starting list of statuses that
+  # maps one-to-one with the alerts provided.
   defp alerts_to_statuses(alerts, time) do
     alerts
     |> Enum.map(fn alert ->
@@ -52,6 +66,11 @@ defmodule Dotcom.SystemStatus.Groups do
     end)
   end
 
+  # Translates an alert to a status:
+  #  - The effect is humanized into a description for the status.
+  #  - If the alert's already active, `time` is set to `nil`.
+  #  - If the alert is in the future, `time` is set to the alert's
+  #    start time
   defp alert_to_status(alert, time) do
     description =
       case alert.effect do
@@ -66,19 +85,6 @@ defmodule Dotcom.SystemStatus.Groups do
     %{description: description, time: time}
   end
 
-  defp stringify_times(statuses) do
-    statuses
-    |> Enum.map(fn status ->
-      case status do
-        %{time: nil} ->
-          status
-
-        %{time: time} ->
-          %{status | time: Timex.format!(time, "%-I:%M%p", :strftime) |> String.downcase()}
-      end
-    end)
-  end
-
   # - If the active period is in the future, returns its start_time.
   # - If the active period indicates that the alert is currently active, returns nil.
   # - Raises an error if the alert is completely in the past.
@@ -90,30 +96,9 @@ defmodule Dotcom.SystemStatus.Groups do
     end
   end
 
-  defp sort_statuses(statuses) do
-    statuses
-    |> Enum.sort_by(fn %{time: time, description: description} -> {time, description} end)
-  end
-
-  defp maybe_add_now_text(statuses) do
-    if any_future_statuses?(statuses) do
-      statuses |> Enum.map(&add_now_text/1)
-    else
-      statuses
-    end
-  end
-
-  defp any_future_statuses?(statuses) do
-    statuses
-    |> Enum.any?(fn
-      %{time: nil} -> false
-      %{} -> true
-    end)
-  end
-
-  defp add_now_text(%{time: nil} = status), do: %{status | time: "Now"}
-  defp add_now_text(status), do: status
-
+  # Combines statuses that have the same active time and description
+  # into a single pluralized status (e.g. "Station Closures" instead
+  # of "Station Closure").
   defp consolidate_duplicate_descriptions(statuses) do
     statuses
     |> Enum.group_by(fn %{time: time, description: description} -> {time, description} end)
@@ -123,6 +108,9 @@ defmodule Dotcom.SystemStatus.Groups do
     end)
   end
 
+  # Replaces the description of a status with its plural form, for use
+  # in &consolidate_duplicate_descriptions/2, when multiple statuses
+  # have the same time and effect.
   defp pluralize_description(%{description: description} = status) do
     new_description =
       case description do
@@ -134,6 +122,85 @@ defmodule Dotcom.SystemStatus.Groups do
     %{status | description: new_description}
   end
 
+  # Sorts the given list of statuses first by time, then by
+  # description, so that earlier statuses show up before later ones,
+  # and then to keep statuses in a stable order.
+  #
+  # This takes advantage of the fact that `nil` is sorted before
+  # anything else, which allows it to automatically sort active
+  # statuses before future ones.
+  # 
+  # This should be called before &stringify_times/1, otherwise times
+  # will get sorted lexically instead of temporally (e.g. 10:00pm will
+  # get sorted ahead of 9:00pm).
+  defp sort_statuses(statuses) do
+    statuses
+    |> Enum.sort_by(fn %{time: time, description: description} -> {time, description} end)
+  end
+
+  # Transforms the time in each status given into a human-readable
+  # string or nil.
+  defp stringify_times(statuses) do
+    statuses
+    |> Enum.map(fn status ->
+      %{
+        status
+        | time: stringify_time(status.time)
+      }
+    end)
+  end
+
+  # Returns a human-readable version of the time given, formatted like
+  # "8:30pm", for example. Leaves nil unchanged.
+  defp stringify_time(nil), do: nil
+  defp stringify_time(time), do: Timex.format!(time, "%-I:%M%p", :strftime) |> String.downcase()
+
+  # The `time` attribute of a status governs how a status should be
+  # rendered. An alert that's currently active should typically only
+  # show its description, while an alert in the future should show the
+  # stringified time as well. If there is both an active status and a
+  # future status, then active statuses should show "Now" instead of
+  # not showing a time.
+  #
+  # We accomplish this by setting the `time` attribute of a status to
+  # `nil` if we don't want it to be rendered, and to some string if we
+  # want that string rendered.
+  #
+  # This function checks whether there are any future statuses, and if
+  # so, replaces any `nil` times with "Now".
+  defp maybe_add_now_text(statuses) do
+    if any_future_statuses?(statuses) do
+      statuses |> Enum.map(&add_now_text/1)
+    else
+      statuses
+    end
+  end
+
+  # Checks the list of statuses to see if any of them have a non-nil
+  # `time` field, which would indicate that at least one status is for
+  # the future, rather than currently active.
+  defp any_future_statuses?(statuses) do
+    statuses
+    |> Enum.any?(fn
+      %{time: nil} -> false
+      %{} -> true
+    end)
+  end
+
+  # If the time for a status is `nil`, replaces it with the string "Now".
+  # Leaves the status unchanged if it has a non-nil `time` attribute.
+  defp add_now_text(%{time: nil} = status), do: %{status | time: "Now"}
+  defp add_now_text(status), do: status
+
+  # This is a special-purpose function to accommodate the fact that
+  # the Green line is actually composed of four different routes under
+  # the hood, but we want to display Green line entries grouped
+  # together when that makes sense.
+  #
+  # This function takes the entries corresponding to the four
+  # different Green line branches, combines ones that are the same,
+  # and uses the specific branch name (e.g. "Green-B") as a
+  # `sub_route` instead of a `route_id`.
   defp combine_green_line_branches(statuses_by_route) do
     {green_line_entries, other_entries} =
       statuses_by_route
@@ -148,6 +215,16 @@ defmodule Dotcom.SystemStatus.Groups do
     other_entries ++ consolidated_green_line_entries
   end
 
+  # Used by &combine_green_line_branches/1, takes the result of
+  # grouping Green line entries and returns one or more entries with
+  # the `route_id` set to "Green", and the `sub_routes` set to
+  # the specific branch route ID's (e.g. "Green-B").
+  defp convert_branches_to_sub_routes(entries)
+
+  # If there's only one entry, that means that all of the Green line
+  # branches have the same entries. Since we only use `sub_routes`
+  # when some entries are different from others, we don't need
+  # `sub_routes` in this case, so we leave it empty.
   defp convert_branches_to_sub_routes([{statuses, _}]) do
     [
       %{
@@ -158,6 +235,11 @@ defmodule Dotcom.SystemStatus.Groups do
     ]
   end
 
+  # If there are multiple entries, that means that not all Green line
+  # branches have the same entries, which means that we do need to
+  # distinguish them using `sub_routes`. This function returns a list
+  # of the entries given, with the `route_id` set to "Green", and the
+  # `sub_routes` set to all of the applicable Green line branch ID's.
   defp convert_branches_to_sub_routes(entries) do
     entries
     |> Enum.map(fn {statuses, routes} ->
@@ -169,6 +251,18 @@ defmodule Dotcom.SystemStatus.Groups do
     end)
   end
 
+  # This is a special-purpose function to accomodate the fact that
+  # most riders consider the Mattapan trolley an extension of the Red
+  # line, so we group Mattapan entries under the Red line if there
+  # are any.
+  #
+  # If there are no Mattapan entries, then we drop the whole Mattapan
+  # entry, leaving riders to infer that "Normal Service" on the Red
+  # line includes the Mattapan trolley.
+  #
+  # If there are Mattapan entries, then this transforms them into
+  # entries under the "Red" `route_id`, and uses "Mattapan" as the
+  # `sub_route`.
   defp combine_mattapan_with_red_line(statuses_by_route) do
     {mattapan_entries, other_entries} =
       statuses_by_route
@@ -189,6 +283,16 @@ defmodule Dotcom.SystemStatus.Groups do
     other_entries ++ new_mattapan_entries
   end
 
+  # Sorts entries by the following criteria:
+  #  - The subway lines should be sorted in the order given by @lines.
+  #  - Entries with no sub-routes should come before entries with
+  #    sub-routes (this mainly serves to sort Red line entries before
+  #    Mattapan ones).
+  #  - "Normal Service" should come before other descriptions (this
+  #    applies mostly to Green line entries where some branches might
+  #    be normal, and others not).
+  #  - Sub-routes should be sorted lexically, so all else equal, Green-B
+  #    should be sorted before Green-C, for instance.
   defp sort_routes_and_sub_routes(entries) do
     line_indexes = @lines |> Enum.with_index() |> Map.new()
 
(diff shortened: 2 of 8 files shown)


lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,221 @@
+defmodule Dotcom.SystemStatus.Groups do
+  @moduledoc """
+  A module that groups alerts into statuses for the system status widget.
+  """
+
+  alias Alerts.Alert
+
+  @lines ["Blue", "Orange", "Red", "Green"]
```
anthonyshull, line 8: Is there a reason to this order? If not, we should alphabetize.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,221 @@
...
+  @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
+  @routes ["Blue", "Orange", "Red", "Mattapan"] ++ @green_line_branches
```
anthonyshull, line 10: Same, unless there is a more communicative ordering, alphabetizing helps with scanning for membership.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,221 @@
...
+      statuses =
+        grouped_alerts
+        |> Map.get(route_id)
+        |> alerts_to_statuses(now)
+        |> consolidate_duplicate_descriptions()
+        |> sort_statuses()
+        |> stringify_times()
+        |> maybe_add_now_text()
+
+      %{route_id: route_id, sub_routes: [], statuses: statuses}
```
anthonyshull, line 26: Should probably be its own private function.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,221 @@
...
+
+  defp stringify_times(statuses) do
+    statuses
+    |> Enum.map(fn status ->
+      case status do
+        %{time: nil} ->
+          status
+
+        %{time: time} ->
+          %{status | time: Timex.format!(time, "%-I:%M%p", :strftime) |> String.downcase()}
```
anthonyshull, line 77: I've seen this time format multiple places. Maybe it is a good time to turn it into a function and move it to a util module.

test/dotcom/system_status/groups_test.exs
```diff
@@ -0,0 +1,401 @@
...
+                 sub_routes: [],
+                 statuses: [%{description: "Normal Service", time: nil}]
+               },
+               %{
+                 route_id: "Green",
+                 sub_routes: [],
+                 statuses: [%{description: "Normal Service", time: nil}]
+               }
+             ] = Groups.groups([], now())
+    end
```
anthonyshull, line 60: The nesting, whitespace, and repetition hint at a cleaner way to do this.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,221 @@
...
+  A module that groups alerts into statuses for the system status widget.
+  """
+
+  alias Alerts.Alert
+
+  @lines ["Blue", "Orange", "Red", "Green"]
+  @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
+  @routes ["Blue", "Orange", "Red", "Mattapan"] ++ @green_line_branches
+
+  def groups(alerts, now) do
```
anthonyshull, line 12: I don't *think* that `now` actually represents now. It should be `date_time` unless it really is `now` in which case it shouldn't be an argument.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,221 @@
...
+        end
+
+      {
+        line_indexes |> Map.get(route_id),
+        base_route_sort_order,
+        description_sort_order,
+        sub_routes
+      }
+    end)
+  end
```
anthonyshull, line 220: This function is both long and deeply nested. It could probably be revisited.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,221 @@
...
+    consolidated_green_line_entries =
+      green_line_entries
+      |> Enum.group_by(& &1.statuses)
+      |> Enum.to_list()
+      |> convert_gl_branches_to_sub_routes()
+
+    other_entries ++ consolidated_green_line_entries
+  end
+
+  defp convert_gl_branches_to_sub_routes([{statuses, _}]) do
```
anthonyshull, line 151: Could probably just call this `convert_branches_to_sub_routes`. Nothing is stopping anyone from calling this function on any other line.

anthonyshull (commented):
Could probably be cleaned up a bit. One suggestion would be to write comments for all of the private functions. I find that it helps add clarity when I have to write narrative descriptions.

lib/dotcom/system_status/groups.ex
joshlarson: This is the order that routes will get sorted in on the widget when they get rendered (I got this order from Figma, but we also discussed changing it to `["Red", "Orange", "Green", "Blue"]` to match the GTFS `route_sort_order`, or `["Red", "Orange", "Blue", "Green"]` to be ordered first by heavy rail before light rail, and then by `route_sort_order`).

I'll add an explanatory comment and sort the other ones.

test/dotcom/system_status/groups_test.exs
joshlarson: I want at least one test that validates not just the existence of those descriptions, but also the absence of other ones, plus at least one test that showcases the entire result struct. 

Basically, if we do a refactor that causes `Mattapan` to suddenly start showing up, or some other extra line that we don't actually want in the end-result, I want at least one test that will catch that.

I think I used the full data structure in two tests (copy-pasted three times), and then used a helper function to have the other tests assert only on the line that they care about.

lib/dotcom/system_status/groups.ex
joshlarson: Huh, that's a good point.

lib/dotcom/system_status/groups.ex
joshlarson: https://github.com/mbta/dotcom/pull/2337

lib/dotcom/system_status/groups.ex
joshlarson: It's long, but is it deeply nested? I think it just has three fairly simple things going on that each take up a few lines of code.

test/dotcom/system_status/groups_test.exs
anthonyshull: But, those are all separate tests and can be communicated more fully.

lib/dotcom/system_status/groups.ex
anthonyshull: To me, it's deeply nested...as indicated by its horizontal shift. Within the function you have a sort_by and then two case statements.

lib/dotcom_web/live/system_status.ex
```diff
@@ -4,21 +4,37 @@ defmodule DotcomWeb.Live.SystemStatus do
   put it into the homepage (and elsewhere).
   """
 
+  alias Dotcom.SystemStatus.Groups
   alias Dotcom.SystemStatus
```
anthonyshull, line 8: Should be ordered the other way.

test/dotcom/system_status/groups_test.exs
```diff
@@ -0,0 +1,565 @@
+defmodule Dotcom.SystemStatus.GroupsTest do
+  use ExUnit.Case, async: true
+
+  alias Dotcom.SystemStatus.Groups
+  alias Test.Support.Factories.Alerts.Alert
+  alias Test.Support.Factories.Alerts.InformedEntity
+
+  @heavy_rail_lines ["Blue", "Orange", "Red"]
+  @all_rail_lines ["Blue", "Green", "Orange", "Red"]
+  @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
```
anthonyshull, line 10: Should be alphabetical in order unless they build off of each other. @all...@green...@heavy

test/dotcom/system_status/groups_test.exs
```diff
@@ -5,177 +5,182 @@ defmodule Dotcom.SystemStatus.GroupsTest do
...
+
+      @heavy_rail_lines
+      |> List.delete(affected_route_id)
+      |> Enum.each(fn route_id ->
+        assert [
+                 %{
+                   statuses: [%{time: nil, description: "Normal Service"}],
+                   sub_routes: [],
+                   route_id: ^route_id
+                 }
```
thecristen, line 94: suggestion(non-blocking): This could read real nice broken out into a function, like `Enum.all?(lines, &shows_normal_service?/1)` which validates the nil time and "Normal Service" text (though I suppose you'd want a separate check to verify the correct route IDs involved, maybe)

test/dotcom/system_status/groups_test.exs
```diff
@@ -5,177 +5,182 @@ defmodule Dotcom.SystemStatus.GroupsTest do
...
 
     test "shows future active time for alerts that will become active later in the day" do
+      affected_route_id = Faker.Util.pick(@heavy_rail_lines)
+
+      time = time_today()
+      alert_start_time = time_after(time)
+      alert_end_time = time_after(alert_start_time)
+
+      alert_start_time_display =
+        alert_start_time |> Timex.format!("{kitchen}") |> String.downcase()
```
thecristen, line 107: Shouldn't this use the Util function that does this?

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,337 @@
+defmodule Dotcom.SystemStatus.Groups do
+  @moduledoc """
+  A module that groups alerts into statuses for the system status widget.
+  """
+
+  alias Alerts.Alert
+
+  # &groups/2 returns an ordered data structure, sorted in the order
+  # given by `@lines`.
```
thecristen, line 9: This comment should be in a `@doc` above the `def` line

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,337 @@
...
+
+  # Given `alerts` and `line_id`, filters out only the alerts
+  # applicable to the given line, using the alert's "informed
+  # entities".
+  defp alerts_for_line(alerts, line_id) do
+    alerts
+    |> Enum.filter(fn %Alert{informed_entity: informed_entity} ->
+      informed_entity
+      |> Enum.any?(fn
+        %{route: ^line_id} -> true
```
thecristen, line 43: confusing: Wait so does this line _not_ mean GTFS line then?

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,337 @@
...
+  alias Alerts.Alert
+
+  # &groups/2 returns an ordered data structure, sorted in the order
+  # given by `@lines`.
+  @lines ["Blue", "Orange", "Red", "Green"]
+  @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
+  @routes ["Blue", "Mattapan", "Orange", "Red"] ++ @green_line_branches
+
+  def groups(alerts, time) do
+    grouped_alerts = Map.new(@routes, &{&1, alerts_for_line(alerts, &1)})
```
thecristen, line 15: That function should be `alerts_for_route` if it operates per-route, no?

I also wonder if something in the `Alerts.Match` module might meet your needs, what if `Alerts.Match.match(alerts, %InformedEntity{route: &1})`?

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,337 @@
...
+
+    @routes
+    |> Enum.map(fn route_id ->
+      statuses =
+        grouped_alerts
```
thecristen, line 145: Why not `Enum.map` over the `grouped_alerts`?

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,337 @@
...
+
+  # Combines statuses that have the same active time and description
+  # into a single pluralized status (e.g. "Station Closures" instead
+  # of "Station Closure").
+  defp consolidate_duplicate_descriptions(statuses) do
+    statuses
+    |> Enum.group_by(fn %{time: time, description: description} -> {time, description} end)
+    |> Enum.map(fn
+      {_, [status]} -> status
+      {_, [status | _]} -> pluralize_description(status)
```
thecristen, line 244: Looks like `Inflex.pluralize/1` is smart enough to ignore existing plurals by the way!

```
iex(2)> ["Delays", "Shuttle Buses", "Station Closure", "Suspension"] |> Enum.map(&Inflex.pluralize/1)
["Delays", "Shuttle Buses", "Station Closures", "Suspensions"]
```

lib/dotcom/system_status/groups.ex
joshlarson: I should restructure this comment to make it clearer that it's actually about `@lines`, not `groups/2`

test/dotcom/system_status/groups_test.exs
joshlarson: Done! (When I first did this work, that Util function hadn't been merged yet)

lib/dotcom/system_status/groups.ex
joshlarson: Ugh yeah - I'm not sure where to draw the line between `route`/`line`/`branch`, etc. This probably isn't it though.

lib/dotcom/system_status/groups.ex
joshlarson: Heh - great question!

I think I refactored back and forth exactly what was happening just inside versus just outside the `Enum.map` call, and then stopped _just short_ of optimal.

lib/dotcom/system_status/groups.ex
joshlarson: > That function should be `alerts_for_route` if it operates per-route, no?

Yep! Done.

> I also wonder if something in the `Alerts.Match` module might meet your needs, what if `Alerts.Match.match(alerts, %InformedEntity{route: &1})`?

Ugh - I agree, but I've been struggling to actually get `Alerts.Match` to work, and I'm not sure exactly why. I still think it's worth using the machinery that already exists, but would you mind if I did that as a follow-up instead?

lib/dotcom/system_status/groups.ex
joshlarson: Extracted!

lib/dotcom_web/live/system_status.ex
joshlarson: Done, I think.

test/dotcom/system_status/groups_test.exs
joshlarson: Question: I've given the tests a pretty thorough once-over (actually more like a twice- or thrice- over). Do you think this is still necessary?

lib/dotcom/system_status/groups.ex
joshlarson: Touché!

lib/dotcom/system_status/groups.ex
```diff
@@ -137,17 +137,33 @@ defmodule Dotcom.SystemStatus.Groups do
...
-      %{route_id: route_id, sub_routes: [], statuses: statuses}
-    end)
+    |> add_alerts(alerts)
+    |> convert_alerts_to_statuses(time)
     |> combine_green_line_branches()
     |> combine_mattapan_with_red_line()
     |> sort_routes_and_sub_routes()
   end
 
+  # Maps the provided `routes` to a struct containing the route ID and
```
thecristen, line 147: nit: technically a map rather than a struct 😅

joshlarson: Heh touché!

joshlarson:
Pulled back into draft because I realized that the data structure isn't quite the right shape to accommodate the desired frontend

test/dotcom/system_status/groups_test.exs
```diff
@@ -0,0 +1,632 @@
...
+      informed_entity:
+        InformedEntitySet.build(:informed_entity_set,
+          route: route_id,
+          entities: [
+            InformedEntity.build(:informed_entity, route: route_id)
+          ]
+        ),
+      active_period: active_period
+    )
+  end
```
anthonyshull, line 631: Any code that goes toward building an alert should go in the alerts factory.

test/dotcom/system_status/groups_test.exs
```diff
@@ -0,0 +1,632 @@
...
+  doctest Dotcom.SystemStatus.Groups
+
+  alias Dotcom.SystemStatus.Groups
+  alias Test.Support.Factories.Alerts.Alert
+  alias Test.Support.Factories.Alerts.InformedEntity
+  alias Test.Support.Factories.Alerts.InformedEntitySet
+
+  @all_rail_lines ["Blue", "Green", "Orange", "Red"]
+  @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
+  @heavy_rail_lines ["Blue", "Orange", "Red"]
```
anthonyshull, line 12: We can export these out of the module.

```
defmodule Dotcom.SystemStatus.Groups do
  @heavy_rail_lines ~w[Blue Orange Red]
  @all_rail_lines @heavy_rails_lines ++ ~w[Green]
  
  def heavy_rail_lines, do: @heavy_rail_lines
  
  def all_rail_lines, do: @all_rail_lines
end

defmodule Dotcom.SystemStatus.GroupsTest do
  describe "groups/2" do
    test "sorts groups by heavy rail first" do
      # Exercise
      groups = Group.groups([], time_today())
      
      # Verify
      route_ids = groups |> Enum.map(&1.route_id)
      heavy_rail_lines = heavy_rail_lines()
      assert Enum.take(route_ids, Kernel.length(heavy_rail_lines)) == heavy_rail_lines
    end
  end
end
```

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,465 @@
...
+  defp alerts_to_statuses_naive(alerts, time) do
+    alerts
+    |> Enum.map(fn alert ->
+      alert_to_status(alert, time)
+    end)
+  end
+
+  # Translates an alert to a status:
+  #  - The effect is humanized into a description for the status.
+  #  - If the alert's already active, `time` is set to `nil`.
```
anthonyshull, line 337: I'm not clear on why setting the time to nil means that it is active.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,465 @@
...
+  # Naively maps a list of alerts to a list of statuses, where a
+  # status is a simple structure with a route, a description, and a
+  # few additional fields that determine how it will render in the
+  # frontend.
+  defp alerts_to_statuses_naive(alerts, time)
+
+  # If there are no alerts, then we want a single status indicating
+  # "Normal Service".
+  defp alerts_to_statuses_naive([], _time) do
+    [%{description: "Normal Service", time: nil}]
```
anthonyshull, line 323: Is there ever a case where the description is "Normal Service" and the time is not nil? Also, I see "Normal Service" in multiple places in this module and in the tests. Why not make it a module attribute and export it?

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,465 @@
...
+  #  - If the alert is in the future, `time` is set to the alert's
+  #    start time
+  defp alert_to_status(alert, time) do
+    description =
+      case alert.effect do
+        :delay -> "Delays"
+        :shuttle -> "Shuttle Buses"
+        :station_closure -> "Station Closure"
+        :suspension -> "Suspension"
+      end
```
anthonyshull, line 347: This could be a module attribute of a map or a function using pattern matching. What happens if the alert.effect isn't in this list?

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,465 @@
...
+  end
+
+  # Returns true if the active period ends before the time given. An
+  # end-time of false indicates an indefinite active period, which
+  # never ends.
+  defp ends_before?({_start_time, nil}, _time), do: false
+  defp ends_before?({_start_time, end_time}, time), do: Timex.before?(end_time, time)
+
+  # Returns true if the active period starts before the time given.
+  defp starts_before?({start_time, _end_time}, time), do: Timex.before?(start_time, time)
```
anthonyshull, line 375: I'm not sure we need these since they just wrap Timex functions.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,465 @@
...
+        status
+        | time: stringify_time(status.time)
+      }
+    end)
+  end
+
+  # Returns a human-readable version of the time given, formatted like
+  # "8:30pm", for example. Leaves nil unchanged.
+  defp stringify_time(nil), do: nil
+  defp stringify_time(time), do: Util.kitchen_downcase_time(time)
```
anthonyshull, line 427: I don't like the fact that stringify_time takes something that isn't a time and returns something that isn't a string.

anthonyshull (dismissed):
Overall, I don't like the `implicit` use of nil as a time in order to try to simplify logic. It makes for using functions that purportedly take a time, but don't really. You could utilize module attributes a lot more to hardcode a lot less.

test/dotcom/system_status/groups_test.exs
joshlarson: I see that that's possible, but what is the benefit?

My problem is that it makes the tests a bit more tautological - instead of the assertion coming from the tests, the assertion comes from a mix of the source code and the tests, which means that the tests don't validate a user-facing behavior so much as validating that the code is interacting with its own internal constants. 

I especially feel that way about this suggestion because the constants you mentioned aren't present in the source file, because I didn't need them. I would be exporting an internal constant that the source file doesn't even need for the sake of the tests, which is the kind of refactor that I've usually considered an antipattern.

I'll think on it for a little longer, but that's my first impression.

lib/dotcom/system_status/groups.ex
joshlarson: The way this will be rendered in the frontend will be something like:
```
case time do
  nil -> description
  _ -> "#{time}: #{description}"
end
```

Thinking about other ways of representing this.... maybe `prefix` instead of `time`, since in a certain sense, that's actually what it is?

lib/dotcom/system_status/groups.ex
joshlarson: > Is there ever a case where the description is "Normal Service" and the time is not nil?

I don't think so.

> Also, I see "Normal Service" in multiple places in this module and in the tests. Why not make it a module attribute and export it?

Touché, although I'm not keen on exporting it for the same reason that I mentioned [here](https://github.com/mbta/dotcom/pull/2330#discussion_r1931388835).

lib/dotcom/system_status/groups.ex
joshlarson: > This could be a module attribute of a map or a function using pattern matching.

True. This felt pretty much equivalent to both of those 🤷‍♂️.

> What happens if the alert.effect isn't in this list?

If the alert effect isn't in this list, that's an error that I'd probably want to see in Sentry. How do we typically handle this kind of thing?

lib/dotcom/system_status/groups.ex
joshlarson: `ends_before?` is needed, but maybe `starts_before?` isn't. It looked a bit odd to me to extract one and not the other, but I guess I can inline `starts_before?`.

lib/dotcom/system_status/groups.ex
joshlarson: Do you have a better name in mind? `maybe_time_to_prefix` (assuming [this suggestion](https://github.com/mbta/dotcom/pull/2330#discussion_r1931390295))?

test/dotcom/system_status/groups_test.exs
joshlarson: I sort of see that, although exactly what belongs in a factory versus just here eludes me at the moment.

The problem is that I don't want to have to copy-pasta the `informed_entity` / `informed_entity_set` nesting over and over again, but I'm not sure that it makes sense to pull this into a factory as-is either.

There probably is a way to do it gracefully, but unless we can get that taken care of _real fast_, I think I'll ask that we tackle it as a follow-up, since this work is blocking the rest of the project.

anthonyshull: You can see in this factory: https://github.com/mbta/dotcom/pull/2336/files#diff-8fbc124c966696c04346223e24b3cc686bb63d07f0f3a35316e93f8dd30c6384 how I separated `_factory` functions from other functions.

lib/dotcom/system_status/groups.ex
anthonyshull: Usually pattern matching and the fallback logs the error with the argument.

test/dotcom/system_status/groups_test.exs
anthonyshull: You don't necessarily have to have the heavy rail lines in a module attribute, though I think you did before. If you're getting them from somewhere else then you can get them there in the test.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,465 @@
+defmodule Dotcom.SystemStatus.Groups do
+  @moduledoc """
+
```
anthonyshull, line 3: We don't have these spaces in the moduledoc of any other modules.

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,465 @@
...
+  alias Alerts.Alert
+
+  # @lines is sorted in the order in which `groups/2` should sort its results
+  @lines ["Blue", "Orange", "Red", "Green"]
+  @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
+
+  @doc """
+
+  Returns a data structure that can be used in the system status
```
anthonyshull, line 17: We don't use this empty line in docs anywhere else. We shouldn't describe this module as being for use in front-end component. It should be useful on its own. Something like `takes alerts groups them in order to communicate system status of each route`.

test/dotcom/system_status/groups_test.exs
joshlarson: > ...though I think you did before. 

This is exactly why I don't want to export the module attributes. I did a pretty hefty refactor to `Groups.groups/2` that changed which module attributes existed, and that refactor would have been harder if I had been exporting them. They're implementation details, and they change during refactors.

> I see that that's possible, but what is the benefit?

Bump on this question, which kind of feels like it makes all the difference.

anthonyshull: The benefit is that the test is more communicative.

lib/dotcom/system_status/groups.ex
joshlarson: This code is gone now - the "current" => nil => "no prefix" logic has been moved into `DotcomWeb` and no longer lives here.

lib/dotcom/system_status/groups.ex
joshlarson: "Normal Service" has been replaced by the atom `:normal` - do you think it's worth extracting that atom to a module attribute, or leave it in-lined?

lib/dotcom/system_status/groups.ex
joshlarson: This code got eaten up by a different refactor, so it's gone now, and the `effect`'s are just passed through without being mapped to strings.

lib/dotcom/system_status/groups.ex
joshlarson: Talked offline and agreed to move all of the presentational stringifying logic into widget/frontend layer, so this is gone now.

lib/dotcom/system_status/subway.ex
```diff
@@ -0,0 +1,391 @@
...
+  A module that groups alerts into statuses for the system status
+  widget. See `Dotcom.SystemStatus` for more information.
+
+  """
+
+  alias Alerts.Alert
+
+  # @lines is sorted in the order in which `groups/2` should sort its results
+  @lines ["Blue", "Orange", "Red", "Green"]
+  @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
```
thecristen, line 10: **suggestion**: call `GreenLine.branch_ids()` for the same output! :sparkles:

lib/dotcom/system_status/subway.ex
```diff
@@ -0,0 +1,371 @@
...
+            branch_ids: ["Mattapan"],
+            status_entries: [
+              %{time: :current, status: :suspension, multiple: false}
+            ]
+          }
+        ],
+        "Green" => [%{branch_ids: [], status_entries: [%{time: :current, status: :normal, multiple: false}]}]
+      }
+  """
+  def subway_status(alerts, time) do
```
thecristen, line 112: **suggestion(blocking):** let's add a `@spec` here! this also gives us a chance to explicitly describe the shape of the output (like, what are all the possible values of `time` or `status` in `status_entries`?).

lib/dotcom/system_status.ex
```diff
@@ -36,4 +36,11 @@ defmodule Dotcom.SystemStatus do
     |> SystemStatus.Alerts.for_day(datetime)
     |> SystemStatus.Alerts.filter_relevant()
   end
+
+  @doc """
+  Returns a map indicating the subway status for each of the subway lines.
+  """
+  def subway_status() do
```
thecristen, line 43: suggestion: let's add a `@spec` here too.

lib/dotcom/system_status/subway.ex
```diff
@@ -0,0 +1,371 @@
...
+    %{
+      branch_ids: branch_ids,
+      status_entries: statuses
+    }
+  end
+
+  # Given `alerts` and `route_id`, filters out only the alerts
+  # applicable to the given route, using the alert's "informed
+  # entities".
+  defp alerts_for_route(route_id, alerts) do
```
thecristen, line 266: suggestion(non-blocking): expanding on my earlier suggestion, this function could potentially be replaced with `Alerts.Match.match(alerts, %InformedEntity{route: route_id})`

thecristen (changes requested):
Awesome. The change I am requesting is just for adding `@spec`s for the public functions. I'd love to see the type definition of what a `SystemStatus` is!

lib/dotcom/system_status/groups.ex
joshlarson: Bye bye empty lines!

I could have sworn I saw it used somewhere else, but now I can't find it, and my editor plays more nicely when there is that empty line, but I also agree that it looks weird (and is inconsistent with the rest of the dotcom world)

lib/dotcom/system_status/subway.ex
joshlarson: Ohhh that reminds me that perhaps the output of this should be a struct instead of a map, for more explicit type-checking, clarity, and obvious errors if it gets populated wrong?

thecristen: https://hexdocs.pm/elixir/1.17.3/typespecs.html#literals you can get pretty descriptive with describing maps, and having this alone would help clarity a lot!  I think a map could be sufficient but a struct would be great too!

lib/dotcom/system_status/subway.ex
joshlarson: I'll give this a try again! I'm pretty sure that this is exactly what I tried and couldn't get it to work, but I will give it another go!

lib/dotcom/system_status/subway.ex
joshlarson: Good call! Will do!

The exact shape of the data is going to change again (which is probably part of why a `@spec` would be a good idea - to communicate these changes! 😅)

joshlarson:
@thecristen follow-up PR to add `@spec`'s:
- https://github.com/mbta/dotcom/pull/2360

I could pull that into this PR if you'd like, or keep it separate - I'm good either way.

lib/dotcom/system_status/groups.ex
joshlarson: Following up on this: Will queue up a task as a follow-up to normalize the data here a bit better.

thecristen (approved):

lib/dotcom/system_status/groups.ex
```diff
@@ -0,0 +1,465 @@
...
+  defp alert_to_status(alert, time) do
+    description =
+      case alert.effect do
+        :delay -> "Delays"
+        :shuttle -> "Shuttle Buses"
+        :station_closure -> "Station Closure"
+        :suspension -> "Suspension"
+      end
+
+    time = future_start_time(alert.active_period, time)
```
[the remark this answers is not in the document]
joshlarson: I don't think so. This function (probably worth renaming it, tbh) gives you either `:current` or `{:future, effect_start_time}`. I think `any_time_match?/2` (and its friend `any_period_match?/2`) return booleans, so I think at best, we would use `any_time_match?/2` to figure out whether to call, effectively, this function.

Open to pairing on whether I've missed something about this, but I'd prefer to punt that to a follow-up if it's not trivially easy to do :)

PR #2333, opened 2025-01-16
feat: Add Lo-Fi system status widget

joshlarson:
Screenshot 2025-01-16 at 6 26 06 PM

---

**Asana Ticket:** [System Status | [Lo-Fi] Component that renders grouped statuses/alerts](https://app.asana.com/0/555089885850811/1209091199210568/f)

---

Depends on:
- https://github.com/mbta/dotcom/pull/2327


joshlarson:
I've made enough progress on the Hi-Fi widget that I don't think this is necessary anymore :)

PR #2336, opened 2025-01-17
Increase coverage and consistency of trip planner tests

anthonyshull:
No ticket. I went through the Trip Planner and wrote tests for everything I think we should cover. Tests are focused on behavior and allow appearance testing to be done in the components.

### DIFF
--- a/lib/dotcom_web/router.ex
+++ b/lib/dotcom_web/router.ex
@@ -265,7 +265,7 @@ defmodule DotcomWeb.Router do
 
   scope "/preview", DotcomWeb do
     import Phoenix.LiveView.Router
-    pipe_through([:browser, :browser_live, :basic_auth_readonly])
+    pipe_through([:browser, :browser_live])
 
     live_session :rider, layout: {DotcomWeb.LayoutView, :preview} do
       live("/trip-planner", Live.TripPlanner)
--- a/test/dotcom_web/live/trip_planner_test.exs
+++ b/test/dotcom_web/live/trip_planner_test.exs
@@ -4,16 +4,9 @@ defmodule DotcomWeb.Live.TripPlannerTest do
   import Mox
   import Phoenix.LiveViewTest
 
-  alias OpenTripPlannerClient.Test.Support.Factory
-  alias Test.Support.Factories.TripPlanner.TripPlanner, as: TripPlannerFactory
+  alias Test.Support.Factories.{MBTA.Api, Stops.Stop, TripPlanner.TripPlanner}
 
   setup :verify_on_exit!
-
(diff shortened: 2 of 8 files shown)


lib/dotcom_web/router.ex
```diff
@@ -265,7 +265,7 @@ defmodule DotcomWeb.Router do
 
   scope "/preview", DotcomWeb do
     import Phoenix.LiveView.Router
-    pipe_through([:browser, :browser_live, :basic_auth_readonly])
+    pipe_through([:browser, :browser_live])
```
anthonyshull, line 268: 📓 Just did this for convenience since basic auth won't be there in the near future. Will remove this from the PR before merge.

test/dotcom_web/live/trip_planner_test.exs
```diff
@@ -4,401 +4,249 @@ defmodule DotcomWeb.Live.TripPlannerTest do
   import Mox
```
anthonyshull, line 4: 📓 Rather than looking at the diff, I would review the file in its entirety since I completely rewrote it.

test/support/factories/trip_planner/trip_planner.ex
```diff
@@ -2,182 +2,201 @@ defmodule Test.Support.Factories.TripPlanner.TripPlanner do
...
+  def ferry_leg_factory do
+    build(:otp_ferry_leg)
+    |> Parser.parse()
+  end
+
+  def groupable_otp_itinerary_factory do
+    [a, b, c] =
+      Enum.map(1..3, fn _ ->
+        Factory.build(:place, stop: Factory.build(:stop))
+      end)
```
thecristen: Would `Factory.build_list(3, :place, stop: Factory.build(:stop))` work just as well?

thecristen: Can happen later, I feel like this should be part of OpenTripPlannerClient's factory

lib/dotcom_web/router.ex
thecristen: Why not remove it now?

thecristen (approved):
Cool!!!!

lib/dotcom_web/router.ex
anthonyshull: I'm going to in a single PR with all of these dangling trip planner tasks.

test/support/factories/trip_planner/trip_planner.ex
anthonyshull: I don't think so because 'grouping' is a dotcom concept not an OTP client concept.

PR #2337, opened 2025-01-17
chore: Add `&Util.kitchen_downcase_time/1`

joshlarson:
No ticket, but would be helpful for cleaning up both the tests and the code in https://github.com/mbta/dotcom/pull/2330.

---

Screenshot 2025-01-17 at 8 09 06 PM


lib/util/util.ex
```diff
@@ -204,6 +204,27 @@ defmodule Util do
 
   def local_tz, do: @local_tz
 
+  @doc """
+
+  Formats a time or a datetime so that it looks right in various
+  places on the site
```
anthonyshull, line 210: Can we come up with a better description? Also, not web app specific since it lives in dotcom.

thecristen (approved):

PR #2339, opened 2025-01-20
Keep input form params in URL

anthonyshull:
https://app.asana.com/0/555089885850811/1209108724880492/f


lib/dotcom_web/live/trip_planner.ex
```diff
@@ -303,6 +311,35 @@ defmodule DotcomWeb.Live.TripPlanner do
...
+
+  # Converts a changeset to a map of params that can be encoded into a URL.
+  defp changeset_to_params(%Ecto.Changeset{changes: changes}) do
+    changes
+    |> Map.update(:from, %{}, &Map.get(&1, :changes))
+    |> Map.update(:to, %{}, &Map.get(&1, :changes))
+    |> Map.update(:modes, %{}, &Map.get(&1, :changes))
+    |> Map.new(fn {k, v} -> {to_string(k), v} end)
+    |> Jason.encode!()
+    |> Jason.decode!()
```
thecristen, line 340: what is this a workaround for, i'm quite curious

lib/dotcom_web/controllers/trip_plan_controller.ex
```diff
@@ -5,305 +5,43 @@ defmodule DotcomWeb.TripPlanController do
...
+  def location(conn, %{"direction" => direction, "query" => query})
+      when direction in ["from", "to"] do
 
-      do_from(conn, destination)
-    else
-      updated_address = check_address(address)
+    case @location_service.geocode(query) do
+      {:ok, [%LocationService.Address{} = location | _]} ->
+        encoded = build_params(direction, location) |> AntiCorruptionLayer.encode()
+        path = live_path(conn, DotcomWeb.Live.TripPlanner)
```
thecristen, line 24: Huh, that's interesting that `live_path` works in this context

**question**: if we merge this before releasing the new trip planner, does this resolve to nil and let this proceed to the current `/trip-planner?plan={stuff}`? or will /from/ and /to/ be broken until the new trip planner lands?

thecristen (changes requested):
This is clean! 

1. One thing I notice is the input displays the coordinates – it should show the name.

For example, `/to/Alewife` shows this:

image

2. I left a comment elsewhere but this blocks release and I wasn't sure if that's what you wanted.

lib/dotcom_web/controllers/trip_plan_controller.ex
```diff
@@ -5,305 +5,43 @@ defmodule DotcomWeb.TripPlanController do
...
-      destination = %NamedPosition{
-        latitude: String.to_float(latitude),
-        longitude: String.to_float(longitude),
-        name: name,
-        stop: nil
+  defp build_params(direction, location) do
+    %{
+      "#{direction}" => %{
+        "latitude" => location.latitude,
+        "longitude" => location.longitude
```
thecristen, line 43: oh, I think if you just add `"name" => query` here that might fix the input

anthonyshull: yeah, i didn't know what that was for. i'll add it in.

lib/dotcom_web/live/trip_planner.ex
anthonyshull: this is because the params come in with strings as keys but the changeset has them as atoms.

lib/dotcom_web/controllers/trip_plan_controller.ex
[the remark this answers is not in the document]
anthonyshull: Yeah, I was under the assumption the new trip planner is being released this week which means it will be realesed before or at the same time as this PR.

thecristen (approved):

PR #2346, opened 2025-01-27
fix(StationInformation): show bike/parking info even for non-stations

thecristen:
#### Summary of changes

**Asana Ticket:** [Stop pages missing facility information](https://app.asana.com/0/555089885850811/1208613560085954/f)

Before, parking info and bike facility info would be suppressed on stop pages representing bus stops or other "not station" stops. Adjusted the logic so this is no longer the case.

Now we can see...
- Parking and bike storage at ferry docks such as Hingham
image
- Parking and bike storage at this bus stop near Watertown Yard
image
- Covered bike racks at Arlington Heights Busway
image

### DIFF
--- a/assets/ts/stop/__tests__/components/StationInformationTest.tsx
+++ b/assets/ts/stop/__tests__/components/StationInformationTest.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { render, screen } from "@testing-library/react";
 import StationInformation from "../../components/StationInformation";
 import { customStop } from "../helpers";
-import { Facility } from "../../../__v3api";
+import { Facility, ParkingLot } from "../../../__v3api";
 
 const stationStop = customStop({
   name: "First Station",
@@ -14,7 +14,8 @@ const stationStop = customStop({
 const busStop = customStop({
   name: "Second Pl",
   "station?": false,
-  bike_storage: ["bike_storage_rack"],
+  bike_storage: [],
+  parking_lots: [],
   accessibility: ["accessible"],
   "has_fare_machine?": false
 });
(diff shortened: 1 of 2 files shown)


assets/css/_spacing.scss
```diff
@@ -128,10 +128,6 @@
   margin-top: -.313rem;
 }
 
-.u-mt-6 {
```
joshlarson, line 131: Nice! 👋

assets/ts/stop/components/amenities/AccessibilityAmenityCard.tsx
```diff
@@ -23,7 +23,7 @@ const accessibilityNames: {
 
 const StationFeatures = (features: AccessibilityType[]): JSX.Element => (
   <>
-    <h2 className="text-2xl">Station Features</h2>
```
joshlarson, line 26: Curiosity: Did you remove this because it wasn't necessary (and is kind of antipattern-y)? Or because it was wrong?

joshlarson: Okay - this _does_ change the font size (makes it a bit smaller). Is that intentional as part of this work?

joshlarson (changes requested):
This PR is actually two changes - one that does what the PR title says, and another unifies the `h2` tag styles and changes the appearance of the amenity cards. I'm a bit worried about merging in visual changes after Chase brought up stylistic instability in the most recent retro.

I'm 👍 on the don't-suppress-parking-and-bikes part, so if you wanted to scope this PR down to just that, and push the `h2` fixes to a different one, I would happily approve.

## Before

Screenshot 2025-01-28 at 9 41 06 AM

## After

Screenshot 2025-01-28 at 9 41 12 AM

---

If there is buy-in from design on the visual changes, then I'm good to merge this, although I do think this could have been split into two PR's.

joshlarson (approved):

PR #2347, opened 2025-01-30
cleanup: Add more alert factories for use in SystemStatus.SubwayTests

joshlarson:
Follow-up to:

### DIFF
--- a/test/dotcom/system_status/subway_test.exs
+++ b/test/dotcom/system_status/subway_test.exs
@@ -5,8 +5,6 @@ defmodule Dotcom.SystemStatus.SubwayTest do
   alias Dotcom.Alerts
   alias Dotcom.SystemStatus.Subway
   alias Test.Support.Factories.Alerts.Alert
-  alias Test.Support.Factories.Alerts.InformedEntity
-  alias Test.Support.Factories.Alerts.InformedEntitySet
 
   @lines_without_branches List.delete(Subway.lines(), "Green")
 
@@ -30,7 +28,11 @@ defmodule Dotcom.SystemStatus.SubwayTest do
       affected_route_id = Faker.Util.pick(@lines_without_branches)
       time = time_today()
       effect = Faker.Util.pick(Alerts.service_impacting_effects())
-      alerts = [current_alert(route_id: affected_route_id, time: time, effect: effect)]
+
+      alerts = [
+        Alert.build(:alert_for_route, route_id: affected_route_id, effect: effect)
+        |> Alert.active_during(time)
+      ]
 
       # Exercise
       groups = Subway.subway_status(alerts, time)
@@ -49,7 +51,11 @@ defmodule Dotcom.SystemStatus.SubwayTest do
       affected_route_id = Faker.Util.pick(@lines_without_branches)
       time = time_today()
       effect = Faker.Util.pick(Alerts.service_impacting_effects())
-      alerts = [current_alert(route_id: affected_route_id, time: time, effect: effect)]
+
+      alerts = [
+        Alert.build(:alert_for_route, route_id: affected_route_id, effect: effect)
+        |> Alert.active_during(time)
+      ]
 
       # Exercise
       groups = Subway.subway_status(alerts, time)
@@ -67,7 +73,11 @@ defmodule Dotcom.SystemStatus.SubwayTest do
       # Setup
       affected_route_id = Faker.Util.pick(@lines_without_branches)
       time = time_today()
-      alerts = [current_alert(route_id: affected_route_id, time: time)]
+
+      alerts = [
+        Alert.build(:alert_for_route, route_id: affected_route_id)
+        |> Alert.active_during(time)
+      ]
 
       # Exercise
       groups = Subway.subway_status(alerts, time)
@@ -86,7 +96,10 @@ defmodule Dotcom.SystemStatus.SubwayTest do
       affected_route_id = Faker.Util.pick(@lines_without_branches)
       time = time_today()
 
-      alerts = [current_alert(route_id: affected_route_id, time: time)]
+      alerts = [
+        Alert.build(:alert_for_route, route_id: affected_route_id)
+        |> Alert.active_during(time)
+      ]
 
       # Exercise
       groups = Subway.subway_status(alerts, time)
@@ -111,7 +124,10 @@ defmodule Dotcom.SystemStatus.SubwayTest do
       time = time_today()
       alert_start_time = time_after(time)
 
-      alerts = [future_alert(route_id: affected_route_id, start_time: alert_start_time)]
+      alerts = [
+        Alert.build(:alert_for_route, route_id: affected_route_id)
+        |> Alert.active_starting_at(alert_start_time)
+      ]
 
       # Exercise
       groups = Subway.subway_status(alerts, time)
@@ -134,7 +150,7 @@ defmodule Dotcom.SystemStatus.SubwayTest do
       alert_start_time = time_before(time)
 
       alerts = [
-        alert(
+        Alert.build(:alert_for_route,
           route_id: affected_route_id,
           effect: effect,
           active_period: [{alert_start_time, nil}]
@@ -165,7 +181,7 @@ defmodule Dotcom.SystemStatus.SubwayTest do
       expired_alert_start_time = time_before(expired_alert_end_time)
 
       alerts = [
-        alert(
+        Alert.build(:alert_for_route,
           route_id: affected_route_id,
           active_period: [
             {expired_alert_start_time, expired_alert_end_time},
@@ -199,8 +215,10 @@ defmodule Dotcom.SystemStatus.SubwayTest do
         |> Enum.sort(:desc)
 
       alerts = [
-        current_alert(route_id: affected_route_id, time: time, effect: effect1),
-        current_alert(route_id: affected_route_id, time: time, effect: effect2)
+        Alert.build(:alert_for_route, route_id: affected_route_id, effect: effect1)
+        |> Alert.active_during(time),
+        Alert.build(:alert_for_route, route_id: affected_route_id, effect: effect2)
+        |> Alert.active_during(time)
       ]
 
       # Exercise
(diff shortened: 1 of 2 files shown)


test/support/factories/alerts/alert.ex
```diff
@@ -28,4 +31,40 @@ defmodule Test.Support.Factories.Alerts.Alert do
...
+      | informed_entity:
+          InformedEntitySet.build(:informed_entity_set,
+            route: route_id,
+            entities: [
+              InformedEntity.build(:informed_entity, route: route_id)
+            ]
+          )
+    }
+  end
+
```
thecristen, line 47: suggestion: If you adjust this function to be its own factory, you wouldn't need to create an alert and pass it to `for_route/2`. Say you want to call `build(:alert_with_route, %{route_id: "Blue"})`, that could work with something like this....

```elixir
def alert_with_route_factory(%{route_id: route_id} = attrs) do
   informed_entity = InformedEntitySet.build(:informed_entity_set,
      route: route_id,
      entities: InformedEntity.build_list(1, :informed_entity, route: route_id)
    )
   build(:alert, informed_entity: informed_entity)
end
```

test/support/factories/alerts/alert.ex
```diff
@@ -28,4 +31,40 @@ defmodule Test.Support.Factories.Alerts.Alert do
...
+  def with_high_priority_effect(alert) do
+    %{alert | effect: Faker.Util.pick(@high_priority_effects)}
+  end
```
thecristen, line 50: Similarly, this could be

```elixir
def high_priority_alert_factory do
  build(:alert, effect: Faker.Util.pick(@high_priority_effects))
end
```

to enable `high_priority_alerts = build_list(4, :high_priority_alert)`

test/support/factories/alerts/alert.ex
```diff
@@ -6,8 +6,11 @@ defmodule Test.Support.Factories.Alerts.Alert do
   use ExMachina
 
   alias Alerts.{Alert, Priority}
+  alias Test.Support.Factories.Alerts.InformedEntity
   alias Test.Support.Factories.Alerts.InformedEntitySet
 
+  @high_priority_effects [:delay, :shuttle, :station_closure, :suspension]
```
thecristen, line 12: It'd be nice to not have to duplicate this from its source of truth.

test/support/factories/alerts/alert.ex
joshlarson: This comment and [this one](https://github.com/mbta/dotcom/pull/2347#discussion_r1935955744) kind of go together to show why I didn't do it that way though, because I wanted to be able to chain them together, like 👇 
```elixir
build(:alert, effect: effect) |> active_during(time) |> for_route(affected_route_id)
```

test/support/factories/alerts/alert.ex
joshlarson: Yeah.... What IS its source of truth?

Also I just realized this point might be moot, since `SystemStatus.Subway.subway_status/2` doesn't actually do anything with the `effect` now. 🤔

test/support/factories/alerts/alert.ex
thecristen: It'd save you a chain or two, e.g. `build(:alert_for_route, route_id: affected_route_id) |> active_during(time)`. But if you like each property being a separate step in the chain that's a choice.... but if you're doing the same steps over and over it might be more useful to make a factory tailored to that use case.

test/support/factories/alerts/alert.ex
thecristen: Wait, what is giving the subway status feature the "high priority" alerts then?? 😅

test/support/factories/alerts/alert.ex
joshlarson: Yeah that's fair - I guess I wasn't sure where to draw the line between what should be chained and what should be "top-level".

I was loosely following [this](https://hexdocs.pm/ex_machina/readme.html#flexible-factories-with-pipes)

test/support/factories/alerts/alert.ex
joshlarson: In `alerts.ex`, the `filter_relevant/1` function [ensures that only certain alerts are returned](https://github.com/mbta/dotcom/blob/fb5974c36f0aae16b5a4e0a6abb78aa6ccbe4563/lib/dotcom/system_status/alerts.ex#L88-L90), and which effects are "relevant" is defined [here](https://github.com/mbta/dotcom/blob/fb5974c36f0aae16b5a4e0a6abb78aa6ccbe4563/lib/dotcom/system_status/alerts.ex#L7). 

In this PR, the results from `alerts.ex` are passed into `subway_status/2`.

That's how it works now - I'm not overly attached to that thought - there's probably a better way.

test/support/factories/alerts/alert.ex
thecristen: Okay, fair enough!

test/support/factories/alerts/alert.ex
thecristen: Ah okay! In that case, the link behind "here" is the source of truth I'm referring to. 

If there were a function there that exposes that list, say `Dotcom.SystemStatus.Alerts.relevant_effects()`, you could call that in this module where needed.

anthonyshull (changes requested):

test/dotcom/system_status/subway_test.exs
```diff
@@ -4,7 +4,7 @@ defmodule Dotcom.SystemStatus.SubwayTest do
 
   alias Dotcom.SystemStatus.Alerts
   alias Dotcom.SystemStatus.Subway
-  alias Test.Support.Factories.Alerts.Alert
+  import Test.Support.Factories.Alerts.Alert
```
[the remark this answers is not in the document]
joshlarson: Yeah - it got mildly annoying to keep calling `Alert.<whatever>`, but that is better than the bizarre "Where did `build` come from setup that using `import` here provides.

test/support/factories/alerts/alert.ex
```diff
@@ -28,4 +29,40 @@ defmodule Test.Support.Factories.Alerts.Alert do
...
+
+  # Returns a random time during the day today before the time provided.
+  defp time_before(time) do
+    between(Timex.beginning_of_day(time), time)
+  end
+
+  # Returns a random time during the day today after the time provided.
+  defp time_after(time) do
+    between(time, Timex.end_of_day(time))
+  end
```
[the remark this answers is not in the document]
joshlarson: Thoughts on where, specifically? A `DateTime` factory?

test/dotcom/system_status/subway_test.exs
```diff
@@ -29,8 +29,13 @@ defmodule Dotcom.SystemStatus.SubwayTest do
       # Setup
       affected_route_id = Faker.Util.pick(@lines_without_branches)
       time = time_today()
+
       effect = Faker.Util.pick(Alerts.service_effects())
-      alerts = [current_alert(route_id: affected_route_id, time: time, effect: effect)]
+
+      alerts = [
+        Alert.build(:alert_for_route, route_id: affected_route_id, effect: effect)
```
joshlarson, line 36: I'm on the fence about the name `route_id` here, since if you look down the factory call stack, it winds up in a field on the `InformedEntity` called `route`, not `route_id`.

I'm torn between two thoughts:
- On the one hand, the thing being passed in is not a route object - it's just a route ID.
  - This is why I started out naming it `route_id`.
- On the other hand, this factory interaction is the tests' entrypoint into the Alert ecosystem, and in the Alert ecosystem, that field is called `route`, not `route_id`, so for consistency, it ought to be called `route` everywhere in the Alert ecosystem. It's kind of arbitrary that (right now), the moment where it switches from `route_id` to `route` is in the `InformedEntity` factory.
  - This is why I'm leaning towards changing the name to `route`, and having the place where it switches from `route_id` to `route` be the place where the factory is first invoked.

thecristen: IMO the fact it's called `route` rather than `route_id` is a flaw of `InformedEntity`/the `Alert` ecosystem 😩

joshlarson: There's no ➕ emoji-react, sadly.

Honestly, it's probably worth fixing that, although definitely not in this PR 😅

test/support/factories/alerts/alert.ex
```diff
@@ -28,4 +28,38 @@ defmodule Test.Support.Factories.Alerts.Alert do
...
+      |> Map.put(
+        :informed_entity,
+        InformedEntitySet.build(:informed_entity_set,
+          entities: [InformedEntity.build(:informed_entity, route: route_id)]
+        )
+      )
+    )
+  end
+
+  def active_during(alert, time) do
```
thecristen, line 47: Is it intentional for it to always output within-a-day active periods, as opposed to ones that span many days? If so, this is better renamed to `active_day` or `active_within_day` IMO.

test/support/factories/alerts/alert.ex
```diff
@@ -28,4 +28,38 @@ defmodule Test.Support.Factories.Alerts.Alert do
       url: Faker.Internet.url()
     }
   end
+
+  def alert_for_route_factory(attrs) do
```
thecristen, line 32: 👍🏼

test/support/factories/alerts/alert.ex
joshlarson: This is gone now 👋

test/support/factories/alerts/alert.ex
joshlarson: Not really - I basically wanted to be able to say "Gimme an alert that's active at `time`", but I don't think it matters a whole lot exactly when.

test/support/factories/alerts/alert.ex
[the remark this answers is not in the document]
thecristen: 💀 Well, now `main` has a `Test.Support.Generators.DateTime.random_time_range_date_time({start, stop})` function that returns a time between start & stop that I suppose would work here

test/dotcom/system_status/subway_test.exs
thecristen: Yeah, keep it `route_id` here.

thecristen (approved):
I like the additional factories.

anthonyshull (approved):

PR #2348, opened 2025-01-30
Maintain locale information when visiting MyCharlie

anthonyshull:
Add locale information to links to MyCharlie so that we can maintain language choices across the two. This causes a small blip where we reload the page when you switch languages, but there isn't anything we can do because Google Translate doesn't work by reloading the page, and the links are rendered on page load.

I considered a JS-only implementation, but that would be completely thrown out the window if and when we move away from Google Translate. It would also be heavy-weight in needing to look at every link on every CMS page after the page has rendered in order to manipulate the href.

So, I figured this is a good compromise.

https://app.asana.com/0/555089885850811/1208595740974445/f


assets/js/translate-analytics.js
```diff
@@ -14,6 +14,10 @@ export default () => {
   if (typeof MutationObserver === "function") {
     // Tell the observer to monitor for changes to HTML attributes
     const config = { attributes: true };
+
+    // Get the language of the page on load.
+    const oldCookieLanguage = getCookieLanguage();
```
thecristen, line 19: The rest of this file's code identifies the language by checking the `lang=` attribute's value on the `<html>` tag. Both approaches appear to work great, I'm just curious whether we need this additional approach as opposed to you doing exactly what you do here but using `document.querySelector("html").getAttribute("lang")` to get the language instead of parsing the cookie.

thecristen (commented):
Solid but I have some comments!

assets/js/translate-analytics.js
anthonyshull: We **have** to check the cookie to get the language change because it won't be in the HTML yet. This just makes it so that I'm checking the same source before and after the change. I could check the `lang` and then check the cookie. But, that might be more confusing.

thecristen (approved):

PR #2349, opened 2025-01-30
When we decode a datetime, we have to cast it to the timezone

anthonyshull:
Even though the timezone information is in the ISO8601 string, it doesn't get decoded by DateTime.

```
DATETIME: #DateTime<2025-01-30 18:01:34.508539-05:00 EST America/New_York>
ENCODED: "2025-01-30T18:01:34.508539-05:00"
DECODED: ~U[2025-01-30 23:01:34.508539Z]
```

So, we have to convert it to a datetime with Timex and specify the timezone.


joshlarson (approved):
Good by me, although the Lint Robot will demand one `mix format` in tribute.

PR #2351, opened 2025-01-31
Fixes the nearest five minute bug and mobile datepicker issue

anthonyshull:
We need to check the current params against the previous params to know that we switched from 'now' to one of the other options.


thecristen (approved):
Thank you!

PR #2353, opened 2025-01-31
feat: Add Hi-Fi system status widget

joshlarson:
## Examples

Screenshot 2025-02-10 at 2 24 09 PM

Screenshot 2025-02-10 at 2 24 18 PM

Screenshot 2025-02-10 at 2 24 31 PM

---

**Asana Ticket:** [System Status | [Hi-Fi] Component that renders grouped statuses/alerts](https://app.asana.com/0/555089885850811/1209091199210570)

### DIFF
--- a/lib/dotcom_web/components/route_pills.ex
+++ b/lib/dotcom_web/components/route_pills.ex
@@ -5,11 +5,24 @@ defmodule DotcomWeb.Components.RoutePills do
 
   use DotcomWeb, :component
 
-  attr :route_id, :string
-  attr :modifier_ids, :list
+  attr :route_id, :string, required: true
+  attr :modifier_ids, :list, default: []
   attr :modifier_class, :string, default: ""
 
-  def route_pill_with_modifiers(assigns) do
+  def route_pill(%{modifier_ids: []} = assigns) do
+    ~H"""
+    <div class={
+      [
+        bg_color_class(@route_id),
+        "w-[3.125rem]"
+      ] ++ shared_icon_classes()
+    }>
+      {pill_text(@route_id)}
+    </div>
+    """
+  end
+
+  def route_pill(assigns) do
     ~H"""
     <span class="flex">
       <.route_pill route_id={@route_id} />
@@ -21,19 +34,6 @@ defmodule DotcomWeb.Components.RoutePills do
         />
       </span>
     </span>
-    """
-  end
-
-  def route_pill(assigns) do
-    ~H"""
-    <div class={
-      [
-        bg_color_class(@route_id),
-        "w-[3.125rem]"
-      ] ++ shared_icon_classes()
-    }>
-      {pill_text(@route_id)}
-    </div>
     """
   end
 
--- a/lib/dotcom_web/components/route_pills.ex
+++ b/lib/dotcom_web/components/route_pills.ex
@@ -26,7 +26,7 @@ defmodule DotcomWeb.Components.RoutePills do
     ~H"""
     <span class="flex">
       <.route_pill route_id={@route_id} />
-      <span :if={Enum.any?(@modifier_ids)} class="h-6 -ml-1 flex gap-0.5">
+      <span class="h-6 -ml-1 flex gap-0.5">
         <.route_modifier
           :for={modifier_id <- @modifier_ids}
           modifier_id={modifier_id}
--- a/lib/dotcom_web/components/route_pills.ex
+++ b/lib/dotcom_web/components/route_pills.ex
@@ -1,0 +1,73 @@
+defmodule DotcomWeb.Components.RoutePills do
+  @moduledoc """
+  Reusable components for displaying subway routes.
+  """
+
+  use DotcomWeb, :component
+
+  attr :route_id, :string
+  attr :modifier_ids, :list
+  attr :modifier_class, :string, default: ""
+
+  def route_pill_with_modifiers(assigns) do
+    ~H"""
+    <span class="flex">
+      <.route_pill route_id={@route_id} />
+      <span :if={Enum.any?(@modifier_ids)} class="h-6 -ml-1 flex gap-0.5">
+        <.route_modifier
+          :for={modifier_id <- @modifier_ids}
+          modifier_id={modifier_id}
+          class={@modifier_class}
+        />
+      </span>
+    </span>
+    """
+  end
+
+  def route_pill(assigns) do
+    ~H"""
+    <div class={
+      [
+        bg_color_class(@route_id),
+        "w-[3.125rem]"
+      ] ++ shared_icon_classes()
+    }>
+      {pill_text(@route_id)}
+    </div>
+    """
+  end
+
+  defp route_modifier(assigns) do
+    ~H"""
+    <div class={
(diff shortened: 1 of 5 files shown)


lib/dotcom_web/components/route_pills.ex
```diff
@@ -0,0 +1,63 @@
...
+  use DotcomWeb, :component
+
+  attr :route_id, :string
+
+  def route_pill_with_modifiers(assigns) do
+    ~H"""
+    <span class="flex">
+      <.route_pill route_id={@route_id} />
+      <span class="h-6 -ml-1 flex gap-0.5">
+        <.route_modifier :for={modifier_id <- @modifier_ids} modifier_id={modifier_id} />
```
thecristen, line 15: Was playing around with this and found that using the existing `DotcomWeb.Components.RouteSymbols.route_icon` component here looks great too:

```
<.route_icon :for={modifier_id <- @modifier_ids} route={%Routes.Route{id: modifier_id}} class="ring-2 ring-white rounded-full" />
```

joshlarson: It looked funny to me in Firefox, because Firefox aligns the text in the route pills slightly differently from how it's aligned in the SVG's.

If we get SVG's for the route pills also, then I'm all in favor of using the existing `<.route_icon />` components, but I'm a bit leery of having the component be partially SVG and partially text-in-a-box.

lib/dotcom_web/live/system_status.ex
```diff
@@ -6,41 +6,89 @@ defmodule DotcomWeb.Live.SystemStatus do
...
+      <div class="flex gap-5">
+        <.system_status_widget routes_with_statuses={example.statuses} />
+        <div class="flex flex-col gap-5">
+          <span class="text-lg font-bold">Alerts</span>
+          <.alert :for={alert <- example.alerts} alert={alert} />
+        </div>
+      </div>
+    </div>
+
+    <h1>Misc Components</h1>
```
joshlarson, line 54: I added these into the preview page for testing convenience, but I'm not attached to keeping them there. Happy to remove this if it's a problem for anyone.

lib/dotcom_web/live/system_status.ex
```diff
@@ -49,4 +97,154 @@ defmodule DotcomWeb.Live.SystemStatus do
     </details>
     """
   end
+
+  defp alerts_examples() do
```
joshlarson, line 101: [This comment](https://github.com/mbta/dotcom/pull/2353/files#r1946663221) applies to this as well.

lib/dotcom_web/components/system_status/widget.ex
```diff
@@ -0,0 +1,135 @@
...
+            prefix={row.status_entry.prefix}
+            plural={row.status_entry.plural}
+          />
+        </div>
+      </div>
+    </div>
+    """
+  end
+
+  defp status_to_rows(data) do
```
joshlarson, line 52: Happy to add documentation-y comments for these private functions, but I wanted to punt that until just before merging.

lib/dotcom_web/components/system_status/widget.ex
```diff
@@ -0,0 +1,150 @@
...
+  """
+
+  use DotcomWeb, :component
+
+  import DotcomWeb.Components.RoutePills
+  import DotcomWeb.Components.SystemStatus.StatusLabel
+
+  @route_ids ["Red", "Orange", "Green", "Blue"]
+
+  def system_status_widget(assigns) do
```
thecristen, line 13: Please add the requisite `attr` here, looks like just `:routes_with_statuses`?

lib/dotcom_web/components/system_status/widget.ex
```diff
@@ -0,0 +1,150 @@
...
+          !row.style.short_bottom_border && "border-gray-lightest border-b-[1px]",
+          "hover:bg-brand-primary-lightest cursor-pointer group/row",
+          "text-black no-underline"
+        ]}
+      >
+        <div class={["pl-2 py-3", row.style.hide_route_pill && "invisible"]}>
+          <.route_pill_with_modifiers
+            route_id={row.route_info.route_id}
+            modifier_ids={row.route_info.branch_ids}
+            modifier_class="group-hover/row:ring-brand-primary-lightest"
```
thecristen, line 36: It looks like we have no reason to expect _different_ `modifier_class` values? Seems like we could remove this and bake `group-hover/row:ring-brand-primary-lightest` directly into the `<.route_pill_with_modifiers />` component?

lib/dotcom_web/components/route_pills.ex
```diff
@@ -0,0 +1,73 @@
+defmodule DotcomWeb.Components.RoutePills do
+  @moduledoc """
+  Reusable components for displaying subway routes.
+  """
+
+  use DotcomWeb, :component
+
+  attr :route_id, :string
```
thecristen, line 8: This should be marked as required

lib/dotcom_web/components/system_status/widget.ex
```diff
@@ -0,0 +1,150 @@
...
+    |> Enum.map(&add_url/1)
+  end
+
+  defp add_url(row) do
+    route_id = route_id_from_route_info(row.route_info)
+    sub_page = if is_normal?(row.status_entry), do: "line", else: "alerts"
+    row |> put_in([:route_info, :url], ~p"/schedules/#{route_id}/#{sub_page}")
+  end
+
+  defp is_normal?(%{status: :normal}), do: true
```
thecristen, line 68: flagged by credo: `Predicate function names should not start with 'is', and should end in a question mark`

thecristen (approved):
This is really effective and thanks for all the examples! My comments are mostly minor, and you'll have to fix a linting error anyway. I do hate the name "widget" for anything, IMO `system_status` or maybe `subway_status` would work fine.

lib/dotcom_web/components/system_status/widget.ex
joshlarson: Ohhhh that reminds me that `<.route_pill_with_modifiers />` itself is redundant, and could just be `<.route_pill modifier_ids={[...]} />`. Should update that before merging.

If you put this component into a context where there shouldn't be any hover behavior, then having `group-hover/row:ring-brand-primary-lightest`... I guess it wouldn't be _wrong_ unless there's a `group/row` somewhere higher up in the DOM, but that feels like a bizarre bug waiting to happen.

And if we ever wanted different hover behavior, or need to render it on a non-white background (when not hovering), then we'd need a different `ring-*` class applied to the modifiers.

Basically, it felt to me like `group-hover/row:ring-brand-primary-lightest` was a statement more about the context that the route pills were being rendered in, than about the route pills themselves.

Honestly, this solution does feel pretty inelegant, and I like the idea of finding something better. Like - if there's a way to get ahold of the parent-div background color, and then apply _that_ to the modifier rings, or something like that?

lib/dotcom_web/components/system_status/widget.ex
joshlarson: Yep!

Also I don't like the name `@routes_with_statuses`. What about `@subway_status`?

PR #2358, opened 2025-02-04
Tell robots to ignore all pages when the domain is mbtace

anthonyshull:
https://app.asana.com/0/555089885850811/1209320056067616/f

Go to dev-green and you'll see the meta tag in the html head.


thecristen (approved):
🚀

PR #2359, opened 2025-02-04
fix: Fix dialyzer/dialyxir config

joshlarson:
A few fixes:
- Uses the default dialyzer ignore file location and format
- Fails the CI task on dialyzer failure

[Example build failure](https://github.com/mbta/dotcom/actions/runs/13142065805/job/36671412593?pr=2359)


.github/workflows/tests.yml
```diff
@@ -159,7 +159,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-all
-      - uses: mbta/actions/dialyzer@v2
+      - uses: mix dialyzer
```
thecristen, line 162: @joshlarson Want `run:` here rather than `uses:` :)

joshlarson: Ha! Thanks! 🤦

thecristen (approved):

PR #2360, opened 2025-02-04
cleanup: Add @spec annotations to Dotcom.SystemStatus.Subway

joshlarson:
Follow-up to:

Specifically, [this comment](https://github.com/mbta/dotcom/pull/2330#pullrequestreview-2584393732).


lib/dotcom/system_status/subway.ex
```diff
@@ -109,6 +119,7 @@ defmodule Dotcom.SystemStatus.Subway do
         "Green" => [%{branch_ids: [], status_entries: [%{time: :current, status: :normal, multiple: false}]}]
       }
   """
+  @spec subway_status([Alert.t()], DateTime.t()) :: %{String.t() => status_entry_group()}
```
thecristen, line 122: Curious if the return type can also be described as `%{Routes.Route.id_t() => status_entry_group()}` or even literally
```
%{
  "Blue" => status_entry_group(),
  "Green" => status_entry_group(),
  "Orange" => status_entry_group(),
  "Red" => status_entry_group()
}
```

lib/dotcom/system_status/subway.ex
```diff
@@ -129,6 +140,10 @@ defmodule Dotcom.SystemStatus.Subway do
   #
   # The exact implementation depends on which line. Green and Red have
   # branches, so they have special implementations.
+  @spec add_nested_statuses_for_line(String.t(), [Alert.t()], DateTime.t()) :: %{
```
thecristen, line 143: Should the first argument be `Routes.Route.id_t()`?

lib/dotcom/system_status/subway.ex
```diff
@@ -202,23 +217,27 @@ defmodule Dotcom.SystemStatus.Subway do
...
   # "Green-C").
+  @spec sort_branches([status_entry_group()]) :: [status_entry_group()]
   defp sort_branches(branches_with_statuses) do
     branches_with_statuses
     |> Enum.sort_by(fn %{status_entries: status_entries, branch_ids: branch_ids} ->
       {status_sort_order(status_entries), branch_ids}
     end)
   end
 
+  @spec status_sort_order([status_entry()]) :: integer()
```
thecristen, line 228:
```suggestion
  @spec status_sort_order([status_entry()]) :: non_neg_integer()
```

Ha I think dialyzer didn't like this one because 0 isn't an integer!

joshlarson: I may have missed it, but is dialyzer complaining? It looks to me like the build is passing. (0 is definitely an integer O_o)

lib/dotcom/system_status/subway.ex
joshlarson: Yes, probably 😇

lib/dotcom/system_status/subway.ex
thecristen: I remember you shared a dialyzer error you were wondering about before, maybe it wasn't this

lib/dotcom/system_status/subway.ex
```diff
@@ -6,6 +6,19 @@ defmodule Dotcom.SystemStatus.Subway do
 
   alias Alerts.Alert
 
+  @type status_time() :: :current | {:future, DateTime.t()}
+
+  @type status_entry() :: %{
+          status: atom(),
```
thecristen, line 12: This atom's`:normal` or `Alerts.Alert.effect()` right? I'd rather that be described explicitly... even better if we could have a type that's limited to the effects we care about.

thecristen (changes requested):
Just one thing, this is great 🎉

joshlarson:
@thecristen PTALA https://github.com/mbta/dotcom/pull/2360/commits/e703c2e37b7816e4d8ef28a79c9415aafd0fc59a 🙏

thecristen (approved):

PR #2361, opened 2025-02-04
feat(autocomplete): add zero state

thecristen:
#### Summary of changes

Related to [TP | Add an error state when location isn't selected](https://app.asana.com/0/555089885850811/1209320056067619) but not technically that! Chase and I decided to split this into two parts, here's one of those parts.

This PR both:
1. Changes the behavior of this input to keep the dropdown open until an item is selected (or the input is cleared).
2. Adds a display for when there're no results.

Both changes apply to this input generally, not just in the trip planner.


anthonyshull (approved):

thecristen:
Not merging because there's a bug where the second input behaves inconsistently to the first!

PR #2364, opened 2025-02-05
feat(Live.TripPlanner): handle OTP timeouts

thecristen:
It turns out we weren't pattern matching to handle the `%Req.TransportError{reason: :timeout}` we were receiving.


joshlarson (approved):
This is a good change. I think we'd probably benefit from improving the error message, since I imagine riders aren't actually concerned about whether we can connect to "Open Trip Planner", so much as being concerned about whether we can plan a trip. 

☝️ is non-blocking because having the error message at all is such a vast improvement over infinite spinner that it's worth rolling forward here.

thecristen:
image

yaaayyyy

PR #2365, opened 2025-02-05
fix(TripPlan.InputForm): improve location validation + add visual error state

thecristen:
#### Summary of changes

**Asana Ticket:** mostly [TP | Add an error state when location isn't selected](https://app.asana.com/0/1209151500209207/1209320056067619) with some [UI/UX | Trip Planner location input error state](https://app.asana.com/0/1209151500209207/1208726277099892), should fix [TP Bugs | Trip planner input location bug](https://app.asana.com/0/1209151500209207/1208025102649692) 

Will be even better with #2364 and #2361 in the mix.

## Key changes
- Validation of the `:to` and `:from` fields has been overhauled, and adds the error where needed
- The location inputs are technically themable now, and we use that to change to the error styling when that form field has errors
- Text that's pasted into the location inputs now gets sent to LiveView (as a pseudo-"selection" that's a name with no lat/lon) and proceeds through our usual pipelines

### More changes
- Fixes the default params in `Dotcom.TripPlan.AntiCorruptionLayer` to not encode missing params as `"nil"`
- Only validates whether locations A != B if each location is itself valid

### DIFF
--- a/test/dotcom/trip_plan/input_form_test.exs
+++ b/test/dotcom/trip_plan/input_form_test.exs
@@ -60,6 +60,23 @@ defmodule Dotcom.TripPlan.InputFormTest do
       refute changeset.errors[:from]
     end
 
+    test "adds error if invalid location" do
+      changeset =
+        InputForm.changeset(%{
+          "to" => %{
+            "latitude" => "",
+            "longitude" => "",
+            "name" => Faker.Cat.breed(),
+            "stop_id" => ""
+          }
+        })
+
+      refute changeset.valid?
+
+      expected_error = InputForm.error_message(:to)
+      assert {^expected_error, _} = changeset.errors[:to]
+    end
+
     test "adds error if from & to are the same" do
       changeset =
         InputForm.changeset(%{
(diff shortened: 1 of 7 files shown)


joshlarson (changes requested):
The UX for this looks great!

I think it'd be worth having a test ensuring that there's an error when a location is invalid, and that there is no error when the location is just fully absent (I think that test would be at the `Dotcom.TripPlan.InputForm` level?)

thecristen:
@joshlarson Great! I've added a test for the first thing; the second thing already has a test for no error when the location is fully absent!

test/dotcom/trip_plan/input_form_test.exs
```diff
@@ -60,6 +60,23 @@ defmodule Dotcom.TripPlan.InputFormTest do
       refute changeset.errors[:from]
     end
 
+    test "adds error if invalid location" do
+      changeset =
+        InputForm.changeset(%{
+          "to" => %{
+            "latitude" => "",
+            "longitude" => "",
+            "name" => Faker.Cat.breed(),
```
joshlarson, line 69: 🐈

joshlarson (approved):
🐈‍⬛

PR #2366, opened 2025-02-05
cleanup: Simplify module attributes in SystemStatus code and tests

joshlarson:
No ticket.

### DIFF
--- a/lib/dotcom/system_status/alerts.ex
+++ b/lib/dotcom/system_status/alerts.ex
@@ -4,7 +4,13 @@ defmodule Dotcom.SystemStatus.Alerts do
   relying on some specific criteria that are specific enough that they don't
   belong in the main `Alerts` module.
   """
-  @relevant_effects [:delay, :shuttle, :suspension, :station_closure]
+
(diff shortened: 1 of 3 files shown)


lib/dotcom/system_status/alerts.ex
```diff
@@ -4,7 +4,13 @@ defmodule Dotcom.SystemStatus.Alerts do
   relying on some specific criteria that are specific enough that they don't
   belong in the main `Alerts` module.
   """
-  @relevant_effects [:delay, :shuttle, :suspension, :station_closure]
+
+  @service_effects [:delay, :shuttle, :suspension, :station_closure]
```
thecristen, line 8: I'm curious whether our teammates might have a clearer idea on what to label to bestow this set of effects.

thecristen (approved):
✨

lib/dotcom/system_status/alerts.ex
joshlarson: Also curious!

For now, I'll get this merged so that other PR's can build on it, but we can re-visit the name whenever we're so inspired!

PR #2368, opened 2025-02-05
Planned disruptions logic and basic display

anthonyshull:
https://app.asana.com/0/555089885850811/1209323323089306/f

### DIFF
--- a/livebooks/planned-disruptions.livemd
+++ b/livebooks/planned-disruptions.livemd
@@ -25,7 +25,7 @@ defmodule Dotcom.Utils.Datetime do
   end
 
   def next_week() do
-    now = Timex.now()
+    now = Util.now()
     end_of_week = Timex.end_of_week(now) |> Util.end_of_service()
     beginning_of_next_week = Timex.shift(end_of_week, seconds: 1)
     end_of_next_week = beginning_of_next_week |> Timex.end_of_week() |> Util.end_of_service()
@@ -43,7 +43,7 @@ defmodule Dotcom.Utils.Datetime do
   end
 
   def this_week() do
-    now = Timex.now()
+    now = Util.now()
     end_of_week = now |> Timex.end_of_week() |> Util.end_of_service()
 
     {now, end_of_week}
@@ -55,7 +55,7 @@ defmodule Dotcom.Utils.Datetime do
   end
 
   def today() do
-    now = Timex.now()
+    now = Util.now()
     end_of_day = now |> Timex.end_of_day() |> Util.end_of_service()
 
     {now, end_of_day}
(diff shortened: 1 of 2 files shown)


livebooks/planned-disruptions.livemd
```diff
@@ -0,0 +1,173 @@
...
+      |> next_week(alert)
+      |> this_week(alert)
+      |> today(alert)
+    end)
+  end
+
+  for {bucket, _} <- @buckets do
+    defp unquote(bucket)(buckets, alert) do
+      if Enum.any?(alert.active_period, fn range ->
+           Datetime.overlap?(Kernel.apply(Datetime, unquote(bucket), []), range)
```
thecristen, line 105: fancy!

livebooks/planned-disruptions.livemd
```diff
@@ -0,0 +1,173 @@
...
+    this_week: [],
+    today: []
+  }
+  @lines ["Blue", "Green", "Orange", "Red"]
+
+  def lines, do: @lines
+
+  def buckets() do
+    @lines
+    |> Alerts.Repo.by_route_ids(Timex.now())
```
thecristen, line 91: Curious whether this would benefit from using `Util.now()`. I'm also curious how this and some of the `Dotcom.Utils.Datetime` do from your timezone/at times like 1am EST that's part of the prior service day/etc —can we see a comparison of results across multiple input DateTimes?

thecristen (commented):
(I haven't had a chance to play with this hands-on yet but wanted to submit my initial comment)

livebooks/planned-disruptions.livemd
anthonyshull: Good call. I think we can just switch to `Util.now()` and make sure to consistently use that in all of our future work.

livebooks/planned-disruptions.livemd
anthonyshull: That's the beauty of having a bunch of small, reusable functions is that you can take on really generic cases without specificity. I could probably figure out a way to do something like this in the pipe of function calls above too.

livebooks/planned-disruptions.livemd
```diff
@@ -0,0 +1,174 @@
...
+    this_week()
+    |> in_range?(datetime)
+  end
+
+  def today() do
+    now = Util.now()
+    end_of_day = now |> Timex.end_of_day() |> Util.end_of_service()
+
+    {now, end_of_day}
+  end
```
thecristen, line 62: Have we confirmed we want "today" to start from **now**, and not from the beginning of today?

thecristen (approved):
Approve the change, have one outstanding question about _today_.

livebooks/planned-disruptions.livemd
anthonyshull: Yeah, I changed that in the actual code.

PR #2372, opened 2025-02-07
DateTime util module and tests

anthonyshull:
This is just some background work to system status and planned disruptions. Most of the logic entails checking the time range of an alert against various service ranges. This elevates the concept of a service range in a similar way that `Timex` has functions like `beginning_of_week` or `end_of_week`. But, instead we have `beginning_of_service_week` and `end_of_service_week`.

The tests introduce a new property-based testing framework which is great for catching edge cases like daylight savings time.


config/config.exs
```diff
@@ -2,6 +2,9 @@ import Config
 
 config :elixir, ansi_enabled: true
 
+config :dotcom, :service_rollover_time, ~T[03:00:00]
+config :dotcom, :timezone, "America/New_York"
+
```
thecristen, line 7: These aren't alphabetized/can be in the bottom. Thanks for adding these!

anthonyshull: That file wasn't ever in alphabetical order :) But, I can order it.

lib/dotcom/utils/date_time.ex
```diff
@@ -0,0 +1,54 @@
...
+  """
+  @spec now() :: DateTime.t()
+  def now(), do: Timex.now(@timezone)
+
+  @doc """
+  Timex can give us ambiguous times during DST transitions; we choose the later time.
+  In the *very* rare case that we are given an {:error, _} tuple, we default to now.
+  In the even rarer case that we are given some other argument type, we also default to now.
+  """
+  @spec coerce_ambiguous_time(DateTime.t() | Timex.AmbiguousDateTime.t() | {:error, term()}) ::
```
thecristen, line 38: phew, I see it's usefulness

lib/dotcom/utils/service_date_time.ex
```diff
@@ -0,0 +1,224 @@
...
+
+  @doc """
+  Get the service date for the given date_time.
+  If the time is before 3am, we consider it to be the previous day.
+  """
+  @spec service_date() :: Date.t()
+  @spec service_date(DateTime.t()) :: Date.t()
+  def service_date(date_time \\ now()) do
+    if date_time.hour < @service_rollover_time.hour do
+      Timex.shift(date_time, hours: -@service_rollover_time.hour) |> Timex.to_date()
```
thecristen, line 42: Would we need `corece_ambiguous_time/1` here too?

lib/dotcom/utils/service_date_time.ex
```diff
@@ -0,0 +1,224 @@
...
+  end
+
+  def in_range?({start, nil}, date_time) do
+    Timex.after?(date_time, start) || Timex.equal?(date_time, start)
+  end
+
+  def in_range?({start, stop}, date_time) do
+    in_range?({start, nil}, date_time) && in_range?({nil, stop}, date_time)
+  end
+
```
thecristen, line 184: Needs a fallback in case both start _and_ stop happen to be `nil`

test/support/factories/utils/date_time.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  @doc "Generate a random date_time between the beginning and end of the time range."
+  def time_range_date_time_generator({start, nil}) do
+    StreamData.repeatedly(fn ->
+      stop = Timex.shift(start, years: 10)
+
+      Faker.DateTime.between(start, stop)
+      |> Timex.to_datetime(@timezone)
+      |> coerce_ambiguous_time()
+    end)
```
thecristen, line 48: Seems this could also be `time_range_date_time_generator({start, Timex.shift(start, years: 10)})`

test/support/factories/utils/date_time.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  @doc "Get a random date_time between the beginning and end of the time range."
+  def random_time_range_date_time({start, stop}) do
+    time_range_date_time_generator({start, stop}) |> Enum.take(1) |> List.first()
+  end
+
+  @doc "Do the two date_times share the same date information?"
+  def same_wall_date?(date_time1, date_time2) do
+    Timex.to_date(date_time1) == Timex.to_date(date_time2)
+  end
```
thecristen, line 32: Not sure this needs a dedicated function as opposed to calling `Timex.equal?(date_time1, date_time2, :day)`

lib/dotcom/utils/service_date_time.ex
anthonyshull: Yeah, good find. Super annoying, but we have to call it after every shift.

test/support/factories/utils/date_time.ex
anthonyshull: I'll play around with it and check if there are any edge cases.

thecristen (approved):
I like the rename to `Test.Support.Generators`. Looks good!

lib/dotcom/utils/date_time.ex
```diff
@@ -0,0 +1,54 @@
...
+  @doc """
+  Get the date_time in the set @timezone.
+  """
+  @spec now() :: DateTime.t()
+  def now(), do: Timex.now(@timezone)
+
+  @doc """
+  Timex can give us ambiguous times during DST transitions; we choose the later time.
+  In the *very* rare case that we are given an {:error, _} tuple, we default to now.
+  In the even rarer case that we are given some other argument type, we also default to now.
```
joshlarson, line 36: `coerce_ambiguous_time/1` will get called with all sorts of times, not just `current_time` or `Timex.now()` or equivalent, right?

I'm nervous about what looks like _almost_ a silent failure along with potentially doing something wrong, rather than just crashing and surfacing an error to the user.

test/dotcom/utils/service_date_time_test.exs
```diff
@@ -0,0 +1,240 @@
...
+      # Setup
+      today = service_range_day() |> random_time_range_date_time()
+
+      # Exercise/Verify
+      assert service_range(today) == :today
+    end
+
+    test "returns :this_week for this week" do
+      # Setup
+      {_, end_of_current_service_week} = service_range_current_week()
```
joshlarson, line 81: What happens if this test runs on the last day of service for a given week?

lib/dotcom/utils/service_date_time.ex
```diff
@@ -0,0 +1,228 @@
...
+  @doc """
+  Get the end of the service day for the given date_time.
+  """
+  @spec end_of_service_day() :: DateTime.t()
+  @spec end_of_service_day(DateTime.t()) :: DateTime.t()
+  def end_of_service_day(date_time \\ now()) do
+    date_time
+    |> service_date()
+    |> Timex.to_datetime(@timezone)
+    |> coerce_ambiguous_time()
```
joshlarson, line 109: Curiosity: Is this first `coerce_ambiguous_time()` necessary, given that there's another one two lines down?

(All of the tests pass even if both `coerce_ambiguous_time()` calls are removed, so I'm actually not sure whether it's that neither one is necessary, or that there's some missing test coverage?)

lib/dotcom/utils/service_date_time.ex
```diff
@@ -0,0 +1,228 @@
...
+
+    {beginning_of_following_week, end_of_following_week}
+  end
+
+  @doc """
+  Get a service range for all time afer the following week of the given date_time.
+  """
+  @spec service_range_later() :: Utils.DateTime.time_range()
+  @spec service_range_later(DateTime.t()) :: Utils.DateTime.time_range()
+  def service_range_later(date_time \\ now()) do
```
joshlarson, line 164: Took me a moment to realize where the term `later` came from here - it comes from the Figma for the planned disruptions work. Outside of that context, I would have no idea what `service_range_later/1` meant! 😅

What about a name like `service_range_after_following_week/1`? I know it's a bulky name, but it's kind of a bulky concept, and `later` doesn't really do it justice.

lib/dotcom/utils/service_date_time.ex
```diff
@@ -0,0 +1,228 @@
...
+
+  def in_range?({start, nil}, date_time) do
+    Timex.after?(date_time, start) || Timex.equal?(date_time, start)
+  end
+
+  def in_range?({start, stop}, date_time) do
+    in_range?({start, nil}, date_time) && in_range?({nil, stop}, date_time)
+  end
+
+  def in_range?(_, _), do: false
```
joshlarson, line 187: This won't work as a fallback for a `{nil, nil}` range, since `def in_range?({nil, stop}, date_time)` will get matched first, and then raise an `:invalid_date` error.

Also, the way I understand these ranges is that `nil` means that there's no bound on that side of the range, so a `{nil, nil}` range would represent all time. If I'm reading that right, shouldn't `in_range?({nil, nil}, _)` return `true`?

Either way, I would recommend adding a test case for this in order to make explicit the assumption we're making here.

lib/dotcom/utils/service_date_time.ex
```diff
@@ -0,0 +1,228 @@
...
+    {_, end_of_following_week} = date_time |> service_range_following_week()
+    beginning_of_later = Timex.shift(end_of_following_week, microseconds: 1)
+
+    {beginning_of_later, nil}
+  end
+
+  @doc """
+  Given a time_range and a date_time, returns true if the date_time is within the time_range.
+  """
+  @spec in_range?(Utils.DateTime.time_range(), DateTime.t()) :: boolean
```
joshlarson, line 174: Idea 💡 (non-blocking)

It feels like there are two concepts going on in this module: Encoding service day/week vagaries into the logic here, and also datetime range math. 

I wonder if it would make sense to pull `in_range?` and the `time_range()` type out into a separate module (along with possibly anything else range related that I may have missed) and keep this one laser-focused on the obnoxious details of service day/week boundaries?

lib/dotcom/utils/date_time.ex
```diff
@@ -0,0 +1,54 @@
...
+  """
+
+  require Logger
+
+  use Timex
+
+  @typedoc """
+  A range of time that can be open ended on either side.
+  """
+  @type time_range() :: {DateTime.t() | nil, DateTime.t() | nil}
```
joshlarson, line 18: Naming: Should this be named `date_time_range()` instead of `time_range()`? When I was looking at the typespecs for the `service_*_range` functions, I was wondering why they were dropping the date part.

Although... per [this comment](https://github.com/mbta/dotcom/pull/2372/files#r1949895730), possibly a better name would be `Utils.DateTimeRange.t()` 🙂

lib/dotcom/utils/service_date_time.ex
```diff
@@ -0,0 +1,228 @@
...
+  @timezone timezone()
+
+  @doc """
+  Returns the time at which service rolls over from 'today' to 'tomorrow'.
+  """
+  def service_rollover_time(), do: @service_rollover_time
+
+  @doc """
+  Get the service date for the given date_time.
+  If the time is before 3am, we consider it to be the previous day.
```
joshlarson, line 36: This is a good use case for doctests and putting some usage examples right here in the doc - that makes the docs a much more powerful reference for how to use this function, and what to expect out of a return value.

(Doctests, rather than just usage examples, because that way if the docs get out of date, the tests will yell at us)

lib/dotcom/utils/service_date_time.ex
```diff
@@ -0,0 +1,228 @@
...
+  @spec service_range_day(DateTime.t()) :: Utils.DateTime.time_range()
+  def service_range_day(date_time \\ now()) do
+    beginning_of_service_day = beginning_of_service_day(date_time)
+    end_of_service_day = end_of_service_day(date_time)
+
+    {beginning_of_service_day, end_of_service_day}
+  end
+
+  @doc """
+  Get a service range for the week of the given date_time.
```
joshlarson, line 128: What is the week boundary? I can't find anything in the tests, docs, or code to tell me O_o.

There are some tests that look like they test the week boundaries, but it looks like they also use the same week boundaries in their test setup, so I'm still not sure how to tell O_o.

test/dotcom/utils/service_date_time_test.exs
joshlarson, line 81: I alluded to this in [a different comment](https://github.com/mbta/dotcom/pull/2372/files#r1949908609), but given that all of the assertions about "current week" also use the implementation of `service_range_current_week()`, I don't have much visibility into what the test is actually asserting.

lib/dotcom/utils/date_time.ex
anthonyshull: That's why we log it and *don't* crash the app because it should only be happening in very strange situations.

lib/dotcom/utils/service_date_time.ex
anthonyshull: Yes, they are both necessary because `to_datetime/2` can return an ambiguous date_time and `shift` does not take an ambiguous date_time. You can be ambiguous both times. Anything that can return a date_time within a daylight shift window has to be disambiguated.

lib/dotcom/utils/service_date_time.ex
anthonyshull: I don't think it is doing two things. It is doing the same kinds of things that Timex does, but within the scope of service times.

lib/dotcom/utils/date_time.ex
joshlarson: Yeah but then the rider experience is that we're confidently telling them something that's not true.

Example: The last trip of the day for a given route ends at a time that's ambiguous and would lead to this error. We discover the error, but instead of telling the rider that there's an unusual error and we don't know when the last trip of the day is, we tell them confidently that the last trip of the day is _now_ (whatever now is). 

IMO, that's a worse experience than a 500.

test/dotcom/utils/service_date_time_test.exs
anthonyshull: It gets the end of the current service week no matter what day it is run.

lib/dotcom/utils/service_date_time.ex
joshlarson: Gotcha - thanks!

Can you add a test that reliably fails for that reason? Otherwise we'll end up refactoring, losing one of the `coerce_ambiguous_time()` calls, and then eventually discarding the existing test as flaky if it fails one in 150 times (or one in ~4000 times).

lib/dotcom/utils/service_date_time.ex
joshlarson: Eh - as I said - non-blocking. But the range type and `in_range?` logic feel pretty self-contained to me.

There's nothing about how `in_range?` is implemented that cares about the arguments being `Service`-related datetimes - only that they're `DateTime`'s at all.

lib/dotcom/utils/date_time.ex
joshlarson: Oh, and if this _isn't_ extracted to a separate module, I think it would make sense to have this type defined in the same module where `in_range?` is implemented, whether that's moving `in_range?` to this module or moving this type to the other module.

test/dotcom/utils/service_date_time_test.exs
anthonyshull: I just changed to the test to accept `:today` or `:this_week` since we don't really have an easy way to pick a day that isn't a specific day.

lib/dotcom/utils/service_date_time.ex
anthonyshull: I can add a guard in there, but if a user calls the function with no start or stop time in the range, they are doing something contrary to all docs. Unfortunately, the type system doesn't have a concept of at least one of them has to be a date_time.

test/dotcom/utils/service_date_time_test.exs
joshlarson: (I'm guessing that the end of the week is Sunday, but my point stands regardless)

Normal operation for this test would be, say, run on Wednesday 2/5:
- `beginning_of_next_service_day` is BoS Thursday 2/6. 
- `end_of_current_service_week` is EoS Sunday 2/9.
- We pick a random time on Thursday thru Sunday.
- That is `:this_week`. Yay.

But say it's run on Sunday 2/9 (I'm just guessing that Sunday is the end of the service week), then end-of-week will be Sunday 
- `beginning_of_next_service_day` is BoS Monday 2/10. 
- `end_of_current_service_week` is EoS Sunday 2/9.
- We pick a random time between BoS Monday 2/10 and EoS Sunday 2/9 (?? range is backwards?? But I guess Faker can handle that??).
- That is, I think, a one-microsecond window, and while I'm not sure exactly how Faker will handle that, I'm guessing it'll either pick the last microsecond of Sunday, which is `:today` or the first microsecond of Monday, which is `:next_week`. Neither of which is `:this_week`.
- That's why I was asking - what will this test do? Will it fail when run on Sundays? Or did I miss something about how the week boundaries are figured?

lib/dotcom/utils/date_time.ex
anthonyshull: changed to `date_time_range()`

lib/dotcom/utils/service_date_time.ex
joshlarson: I bet we could make the type be something like 
```
@type time_range() :: {DateTime.t(), DateTime.t()} | {nil, DateTime.t()} | {DateTime.t(), nil}
```

It's a bit more verbose, but it does make it clear that having the range be unbounded on both ends is not expected.

joshlarson (changes requested):
A whole bunch of little stylistic feedback. I'd like to at least discuss them.

Two things that are blockers for me, at least for now, are:
- [https://github.com/mbta/dotcom/pull/2372#discussion_r1949843892](this thread), about how we'll occasionally randomly tell riders that a time is _now_, when what we actually mean is that we ran into a weird error and can't give accurate time info.
- Some missing test coverage

lib/dotcom/utils/service_date_time.ex
anthonyshull: Not really because we don't have a mock for Timex. We would need to somehow have `to_datetime` return an `{:error, term()}` then also have `shift` return an `{:error, term()}` and that probably isn't ever going to happen given we're feeding it a date. I really think we have this base covered.

lib/dotcom/utils/service_date_time.ex
anthonyshull: The time was also in the moduledoc at the top, but I added the time ranges everywhere you might want them.

lib/dotcom/utils/date_time.ex
anthonyshull: this is an insanely rare case. so rare that I'm not sure how to even trigger it. the way we handle these kinds of issues elsewhere is to log the argument and return some kind of reasonable value. this is not a decision that was made recently as we had multiple instances of the app crashing. this keeps in line with what we've done elsewhere.

lib/dotcom/utils/service_date_time.ex
joshlarson: Oh yeah, no, what I meant was why the term "later" has such a specific and kind of unusual definition.

The reason I'm thinking about this is a distinction that you made not too long ago - that backend logic and core backend data should exist independently of the presentational frontend features that use it. 

When it comes to service dates, it's clear that the _day_ boundary (being 3:00am, and also maybe being dependent on routes) is a core feature of a `ServiceDateTime` module. It's a little less clear to me that week boundaries are - does MBTA service actually operate on week boundaries?

But the fact that anything after the end of "next week" is "later" is exclusively a feature of the planned disruptions section of the alerts page, not a core DateTime concept.

I'm not going to block this PR on this, for two reasons:
1. I'm not sure whether it would make sense to rename this function, or pull the this-week/next-week/later functionality out of this module entirely.
2. Most of the rest of this PR is pretty solid, and deserves to see the light of day, and this isn't a huge deal and can be fixed later (if it is indeed a problem)

But I did want to flag that this does feel like a frontend/feature-specific concept leaking into a core/backend module, which is something that you've done a good job of flagging for me in the past 🙂

test/dotcom/utils/date_time_test.exs
```diff
@@ -0,0 +1,117 @@
...
+  use ExUnit.Case
+  use ExUnitProperties
+
+  import Dotcom.Utils.DateTime
+  import ExUnit.CaptureLog
+  import Test.Support.Generators.DateTime
+
+  describe "timezone/0" do
+    test "returns a valid timezone for the application" do
+      # Verify/Verify
```
joshlarson, line 11: (Non-blocking) Should this be `Exercise/Verify`? Or is there a convention that I missed?

lib/dotcom/utils/service_date_time.ex
```diff
@@ -0,0 +1,213 @@
...
+  require Logger
+
+  use Timex
+
+  import Dotcom.Utils.DateTime,
+    only: [coerce_ambiguous_time: 1, in_range?: 2, now: 0, timezone: 0]
+
+  alias Dotcom.Utils
+
+  @type service_range() :: :past | :today | :this_week | :next_week | :later
```
joshlarson, line 25: (Non-blocking) This name (and the name of the function below that's also called `service_range/1`) feels like a potential source of confusion, since it refers to an atom, not an actual range.

lib/dotcom/utils/service_date_time.ex
anthonyshull: I don't really agree. We are trying to organize disruptions by time. `later` is clearly defined in the module. We could come up with some other term for it, but `later = later than next week`

anthonyshull: The point of this module, and why it is separate from DateTime, is that we want to elevate service ranges to a first-class citizen. Just like Timex has `beginning_of_week` we want to have `beginning_of_service_week.` As noted in the moduledoc, for now, we just have a general case of looking at time, but the intention is to have route-specific information later. I'm not even sure what moving this to *another* module would even look like.

lib/dotcom/utils/service_date_time.ex
joshlarson: I'm a little confused - I think you marked this as resolved, but I missed what the resolution actually was?

lib/dotcom/utils/service_date_time.ex
joshlarson: Wait a minute... this function can take a `date_time` as an argument, so it doesn't need to mock Timex. Just pass in that one time in March (or November?) where the first coercion matters, and then in a different test, pass in that one time an hour later where the other one matters.

lib/dotcom/utils/service_date_time.ex
joshlarson: Okay - I see a comment now that says that Monday is the service week boundary.

(Non-blocking) A doctest would be helpful here, as executable verification that the code actually does this.

lib/dotcom/utils/date_time.ex
```diff
@@ -0,0 +1,86 @@
...
+
+  Timex will return an error if the time occurs when we "spring-forward" in DST transitions.
+  That is because one hour a day does not occur--2am to 3am.
+  In that case, we set the time to 3am.
+
+  If we are given something thatis not a DateTime, AmbiguousDateTime, or an error tuple, we log the input and return `now`.
+  """
+  @spec coerce_ambiguous_time(DateTime.t() | Timex.AmbiguousDateTime.t() | {:error, term()}) ::
+          DateTime.t()
+  def coerce_ambiguous_time(%DateTime{} = date_time), do: date_time
```
thecristen, line 49: There're lots of spots in Dotcom working with `%NaiveDateTime{}`... we probably want to make this work with those too.

lib/dotcom/utils/date_time.ex
```diff
@@ -31,24 +32,55 @@ defmodule Dotcom.Utils.DateTime do
...
+    |> Timex.to_datetime(@timezone)
+    |> coerce_ambiguous_time()
+    |> Timex.shift(hours: 2)
+    |> coerce_ambiguous_time()
   end
 
   def coerce_ambiguous_time(arg) do
     Logger.error("#{__MODULE__} failed to coerce ambiguous time: #{inspect(arg)}")
 
     now()
```
thecristen, line 64: I see what Josh means about the error returning now thing. Definitely weird in the context of a pipeline working on some entirely different date. Feels like the more Elixir-y thing to do here would be to return an `:error` tuple of our own, as annoying as handling that would be.

lib/dotcom/utils/date_time.ex
anthonyshull: The idea behind the library is that the caller is responsible for handling that. One of the reasons that the old DateTime code is such a mess is because it was trying to do too much in the sense of trying to accept everything that client code could throw at it. The only reason we have this function is because Timex can return these things and we need to make default choices about how to handle them.

lib/dotcom/utils/date_time.ex
anthonyshull: In that case, we should just do nothing and have the error be that the function was called with an unsupported argument.

lib/dotcom/utils/date_time.ex
anthonyshull: I.e., I believe we have so much handling of naive date times because of integrations with front-end code. Rather than saying...sure, just give us a naive date time, we make that code give us a date time. Otherwise, every function we write now has to handle naive date times and try to do the conversion.

test/dotcom/utils/service_date_time_test.exs
anthonyshull: All of our functions except this one *do* take a date_time and use now as a default. The reason this one doesn't is because it's entire purpose is to give you an atom that expresses the relation of the date_time you do pass to a range of time relative to now. Avoiding mocking isn't a good strategy, but, the question is whether it is worth mocking Timex in order to test this one function when we can be pretty sure it is behaving as we expect by reading the code. There are tradeoffs.

lib/dotcom/utils/date_time.ex
```diff
@@ -0,0 +1,77 @@
...
+
+  Timex can give us ambiguous times when we "fall-back" in DST transitions.
+  That is because the same hour occurs twice.
+  In that case, we choose the later time.
+
+  Timex will return an error if the time occurs when we "spring-forward" in DST transitions.
+  That is because one hour does not occur--02:00:00am to 03:00:00am.
+  In that case, we set the time to 03:00:00am.
+
+  If we are given something tha tis not a DateTime, AmbiguousDateTime, or an error tuple, we log the input and return `now`.
```
joshlarson, line 43: (Non-blocking): It looks like the `now` fallback was removed (thank you 🙂), so unless I missed something, I think we can also drop this part of the doc comment?

joshlarson: Also a small nit, but if I _didn't_ miss something then here's a tiny spelling fix. 

```suggestion
  If we are given something that is not a DateTime, AmbiguousDateTime, or an error tuple, we log the input and return `now`.
```

lib/dotcom/utils/date_time.ex
thecristen: Okay, fair enough!

lib/dotcom/utils/date_time.ex
thecristen: I think that's a good call, thanks!

thecristen (approved):

test/dotcom/utils/service_date_time_test.exs
```diff
@@ -0,0 +1,221 @@
...
+      # Setup
+      today = service_range_day() |> random_time_range_date_time()
+
+      # Exercise/Verify
+      assert service_range(today) == :today
+    end
+
+    test "returns :today if today is the last day of the service week" do
+      # Setup
+      expect(Dotcom.Utils.DateTime.Mock, :now, 2, fn ->
```
joshlarson, line 87: Suggestion: Instead of mocking `now` twice in this test, pass `Dotcom.Utils.DateTime.now() |> Timex.beginning_of_week() |> beginning_of_service_day()` into `service_range_later_this_week()` as an argument. 

I think that in order for that to work fully, we'd need to replace the `expect` calls with `stub` in this test.

test/dotcom/utils/service_date_time_test.exs
```diff
@@ -0,0 +1,221 @@
...
+
+      # Exercise / Verify
+      assert service_range(end_of_service_week) == :today
+    end
+
+    test "returns :later_this_week for later this week" do
+      # Setup
+      {beginning_of_service_week, end_of_service_week} = service_range_later_this_week()
+
+      stub(Dotcom.Utils.DateTime.Mock, :now, fn ->
```
joshlarson, line 105: Curiosity: How come we use `stub` sometimes in this suite, and sometimes we use `expect`?

joshlarson (approved):
Left a few more comments, but you'll _have_ to address them as follow-ups if you feel they're worth addressing at all. Ha. Ha!

(Maybe I'll address them myself if I think they're worth addressing - anything can happen!)

But seriously, this is great - big improvement over the mishmash of `Utils.service_*` and `Timex` calls, and nice to have a unified place for core service and core date-time stuff.

PR #2373, opened 2025-02-10
Planned disruptions logic

anthonyshull:
https://app.asana.com/0/555089885850811/1209323323089311/f

This is the backend logic for planned disruptions. It also includes today's disruptions for the system status component.


lib/alerts.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Dotcom.Alerts do
```
joshlarson, line 1: Should this live in `lib/dotcom/alerts.ex` instead of `lib/alerts.ex`?

thecristen (approved):
There's a lot here! Thanks for adding the tests and supporting behaviours and mocks.

lib/dotcom/alerts/disruptions/subway.ex
```diff
@@ -0,0 +1,55 @@
...
+  @spec future_disruptions() :: %{Utils.ServiceDateTime.named_service_range() => [Alert.t()]}
+  def future_disruptions() do
+    disruption_groups() |> Map.take([:later_this_week, :next_week, :after_next_week])
+  end
+
+  @doc """
+  Disruptions that occur during today's service range.
+  """
+  @spec todays_disruptions() :: %{Utils.ServiceDateTime.named_service_range() => [Alert.t()]}
+  def todays_disruptions() do
```
joshlarson, line 27: I may be reading this wrong, but it looks like this will include alerts from earlier today, even if their effect periods have ended. If so, we should double-check the requirements, because I had thought that we only want to include alerts if they're either currently active or will be active later in the day.

Either way, I think it would be good to have a test explicitly highlighting the "earlier-today" alerts behavior.

lib/dotcom/routes.ex
```diff
@@ -0,0 +1,13 @@
...
+  @moduledoc """
+  A collection of functions that help to work with routes in a unified way.
+  """
+
+  @subway_route_ids ["Blue", "Green", "Orange", "Red"]
+
+  @doc """
+  Returns a list of route ids for all subway routes.
+  """
+  @spec subway_route_ids() :: [String.t()]
```
joshlarson, line 11: **Praise** This is a better place for this than where it was - nice to see this attr gradually migrating to its optimal home 🏠

lib/dotcom/routes.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Dotcom.Routes do
+  @moduledoc """
+  A collection of functions that help to work with routes in a unified way.
+  """
+
+  @subway_route_ids ["Blue", "Green", "Orange", "Red"]
```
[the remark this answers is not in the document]
joshlarson: Couple of thoughts on this line of code in particular:

Idea: Would it be worth having a type for line ID that only allows these four values? Then the type of `subway_route_ids` would be `[Dotcom.Routes.line_id()]`.

Also, the name `subway_route_ids()` is misleading, because `"Green"` is not a Route ID - the subway Route ID's are "Red", "Mattapan", "Orange", "Green-B", "Green-C", "Green-D", "Green-E", and "Blue". I'd call this `subway_line_ids` or `subway_lines`.

And finally - love extracting this and putting it somewhere clean and universal!

lib/dotcom/system_status/alerts.ex
```diff
@@ -92,7 +88,7 @@ defmodule Dotcom.SystemStatus.Alerts do
   }
   """
   def filter_relevant(alerts) do
-    alerts |> Enum.filter(fn %{effect: effect} -> effect in @service_effects end)
+    alerts |> Enum.filter(&service_impacting_alert?/1)
```
joshlarson, line 91: Nice!

I could eventually see this whole module getting lumped into `Dotcom.Alerts`, but that's a project for a future day.

lib/dotcom/utils/date_time.ex
```diff
@@ -39,6 +39,8 @@ defmodule Dotcom.Utils.DateTime do
   Timex will return an error if the time occurs when we "spring-forward" in DST transitions.
   That is because one hour does not occur--02:00:00am to 03:00:00am.
   In that case, we set the time to 03:00:00am.
+
+  If we are given something tha tis not a DateTime, AmbiguousDateTime, or an error tuple, we log the input and return `now`.
```
joshlarson, line 43:
```suggestion
```

Begone!

livebooks/planned-disruptions.livemd
```diff
@@ -0,0 +1,124 @@
+# Planned Disruptions
```
joshlarson: Did you mean to include this?

lib/dotcom/routes.ex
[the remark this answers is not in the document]
anthonyshull: Eventually, we won't need this because we'll just be able to query routes by type.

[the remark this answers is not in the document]
anthonyshull: And then query alerts by whatever attributes.

lib/dotcom/alerts/disruptions/subway.ex
```diff
@@ -0,0 +1,55 @@
...
+    disruption_groups() |> Map.take([:today])
+  end
+
+  # Groups all disruption alerts by service range.
+  #
+  # 1. Gets all alerts for subway routes.
+  # 2. Filters out non-service-impacting alerts
+  # 3. Groups them according to service range.
+  defp disruption_groups() do
+    subway_route_ids()
```
joshlarson, line 37: I don't think this will work correctly (unless I missed a change in `by_route_ids/2`. Green line alerts aren't associated with the route ID `"Green"` - they're associated with the branches - `"Green-B"`/etc. This will also miss Mattapan alerts, I think.

joshlarson (changes requested):
A lot of this looks great! Nice to see a bunch of kind of universal concerns moving into better homes.

I have two comments that are blocking, related to the actual functionality:
- It looks like `todays_disruptions` includes alerts that shouldn't be included, and the tests don't make it clear what happens in that circumstance.
- The way `subway_route_ids()` looks like it will miss all green line and Mattapan alerts.

livebooks/planned-disruptions.livemd
joshlarson: I guess not!

lib/dotcom/alerts/disruptions/subway.ex
anthonyshull: This is grabbing todays disruptions. If the view doesn't want ones that are earlier than `now` they can remove them. The intention here isn't to do all of the work for the only current consumer. They will still have to do grouping and sorting by line if they want it.

anthonyshull: And, it will only get those alerts if the effect period ended after the start of the service day. So, the disruption effected 'today'.

lib/dotcom/alerts/disruptions/subway.ex
anthonyshull: I was pulling Green Line alerts with this code. I can check to see if we want Mattapan alerts to show up in this component. Right now it only shows the main lines. Easy to add if we need/want them.

PR #2374, opened 2025-02-10
refactor: alerts page layout

thecristen:
#### Summary of changes

**Asana Ticket:** Precursor to [Service Alerts page | Refine Station and Service Alerts section](https://app.asana.com/0/555089885850811/1209323323089304)

Refactored the alerts "show" template into a layout. I used named slots to control what goes into the sidebar, since this will change for the subway service alerts (the timeframe filters will be removed). Anyways, I'm hoping it'll make it easier to swap in/out the differing content between subway and the other modes (which will remain the same). Visually, this should be a no-op.

### DIFF
--- a/lib/dotcom_web/components/components.ex
+++ b/lib/dotcom_web/components/components.ex
@@ -6,6 +6,12 @@ defmodule DotcomWeb.Components do
   """
   use Phoenix.Component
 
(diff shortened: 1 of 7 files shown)


anthonyshull (approved):
I like the templating and the slots.

PR #2380, opened 2025-02-12
feat(DotcomWeb.Components): add two components

thecristen:
#### Summary of changes

Add/documents/tests two new generic components that can be reused between the Subway Status & Planned Work components (and uses them for the current subway status component).

1. `<.bordered_container />` is a lightly bordered rounded box with a heading. Keeping it simple. The dividing line between heading and content is toggleable since subway status doesn't need it but I think planned work will.
2. `<.lined_list />` is a "list" with faint lines between items, structured in such a way that it's possible to adjust the lines per-item, mostly because of the exception needed by the subway status. I know there's a similar component in Metro, but this one is more bare-bones (doesn't even dictate padding or other styling) and doesn't even handle proper unordered or ordered lists (wrapping each child in a `<li>` would remove our ability to easily control those line-styles between items). This component probably won't see much use perhaps.

With the exception of the subway status component now expanding to fill its container width, this should visually be a no-op — let me know if I missed anything.

image

### DIFF
--- a/lib/dotcom_web/components/components.ex
+++ b/lib/dotcom_web/components/components.ex
@@ -92,4 +92,103 @@ defmodule DotcomWeb.Components do
     </div>
     """
   end
+
+  slot(:heading, required: false, doc: "Large title shown at top of container.")
+  slot(:inner_block, required: true)
+  attr(:show_divider?, :boolean, required: false, default: true)
+
+  @doc """
+  A generic "card" sort of component, with an optional large header block.
+
+  Example usage:
+
+  ```elixir
+  <.bordered_container>
+    <:heading>Title on Top</:heading>
+    <p class="text-yellow-600">
+      <strong>Whatever</strong> else inside.
+    </p>
+  </.bordered_container>
+  ```
+
+  Can omit the divider between the heading and content using
+  `show_divider?={false}`.
+
+  ```elixir
+  <.bordered_container show_divider?={false}>
+    <:heading>
+      <div class="flex justify-between">
+        Slightly elaborate
+        <mark>Thing</mark>
+      </div>
+    </:heading>
+    <p class="border-t-4 border-t-brand-primary p-6 bg-brand-primary-lightest">
+      Got my own dividing line
+    </p>
+  </.bordered_container>
+  ```
+  """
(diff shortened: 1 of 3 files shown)


test/dotcom_web/components_test.exs
```diff
@@ -55,4 +55,66 @@ defmodule DotcomWeb.ComponentsTest do
              """) =~ assigns.content
     end
   end
+
+  describe "bordered_container" do
+    test "optionally shows a dividing line" do
```
joshlarson, line 60: **Suggestion:** Split this into two tests, one for the dividing line - one for no dividing line.

lib/dotcom_web/components/components.ex
```diff
@@ -92,4 +92,103 @@ defmodule DotcomWeb.Components do
     </div>
     """
   end
+
+  slot(:heading, required: false, doc: "Large title shown at top of container.")
+  slot(:inner_block, required: true)
+  attr(:show_divider?, :boolean, required: false, default: true)
```
joshlarson, line 98: **Suggestion** Drop the `?` from `show_divider`, and have the default be `false` instead. That way, usages will look like

```heex
<.bordered_container />
```

and

```heex
<.bordered_container show_divider />
```

Or if you prefer the default to be that the divider is shown, then have the parameter be `hide_divider`, which defaults to `false`, so that the non-default case is

```heex
<.bordered_container hide_divider />
```

lib/dotcom_web/components/system_status/subway_status.ex
```diff
@@ -129,17 +130,6 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatus do
     ]
   end
 
-  defp add_short_intermediate_borders([entry]) do
```
joshlarson, line 132: Nice! Bye bye!

joshlarson (approved):
Left a few non-blocking comments, but this looks good!

(Please forgive me - I once again forgot to lump the comments together into a review)

anthonyshull:
Should this be in Metro?

thecristen:
@anthonyshull Not expecting them to be used outside Dotcom, so (at least for this moment) nope!

PR #2381, opened 2025-02-13
fix: Green line grouping bug

joshlarson:
Changes the way that green line statuses get grouped, so that they line up more closely with the alerts themselves. Previously, branches would only be grouped together if their full set of alerts was the same. Now, branches will be grouped together if they're impacted by the same alert, even if they also have other statuses.

## Before
Screenshot 2025-02-12 at 9 21 05 PM

## After
Screenshot 2025-02-12 at 9 21 19 PM

---

**Asana Ticket:** [System Status | 🐞 Green line grouping bug](https://app.asana.com/0/555089885850811/1209323504575454)

---

Depends on:
- #2347

### DIFF
--- a/test/dotcom/system_status/subway_test.exs
+++ b/test/dotcom/system_status/subway_test.exs
@@ -324,11 +324,10 @@ defmodule Dotcom.SystemStatus.SubwayTest do
       effect = Faker.Util.pick(Alerts.service_impacting_effects())
 
       alerts =
-        GreenLine.branch_ids()
-        |> Enum.map(fn route_id ->
-          Alert.build(:alert_for_route, route_id: route_id, effect: effect)
+        [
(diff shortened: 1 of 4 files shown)


test/dotcom/system_status/subway_test.exs
```diff
@@ -513,68 +652,4 @@ defmodule Dotcom.SystemStatus.SubwayTest do
...
-  # Returns a random alert that will be active at the time given by
-  # the required `:time` opt.
-  #
-  # Required opts:
-  #  - route_id
-  #  - time
-  #
-  # Optional opts:
-  #  - effect (default behavior is to choose an effect at random)
-  defp current_alert(opts) do
```
thecristen, line 526: non-blocking: I liked these functions you removed, as they made the tests themselves more concise 😅 Miss them.

thecristen (approved):
Seems like it works, neat! Please fix the merge conflicts

PR #2383, opened 2025-02-13
fix file download link

anthonyshull:
File downloads with a comma were not working. `URI.encode` doesn't encode commas, but `URI.encode_www_form` does.


thecristen (approved):
✨

PR #2386, opened 2025-02-14
refactor(Components.RouteSymbols): move route_pills to subway_route_pill

thecristen:
No ticket, just showing my idea for the subway route pill.

image


anthonyshull (approved):
Love it, I think we'll have some conflicts in routes.ex, but I can figure those out.

PR #2387, opened 2025-02-14
Planned disruptions display

anthonyshull:
https://app.asana.com/0/555089885850811/1209323327282330/f

`/preview/system-status#planned-disruptions`


lib/dotcom_web/components/components.ex
```diff
@@ -10,6 +10,8 @@ defmodule DotcomWeb.Components do
     endpoint: DotcomWeb.Endpoint,
```
anthonyshull, line 10: 📓 These were copied over from @thecristen's work. So, these will shake out of this PR.

lib/dotcom_web/components/alerts.ex
```diff
@@ -0,0 +1,36 @@
+defmodule DotcomWeb.Components.Alerts do
+  @moduledoc """
+  Function components for rendering different styles of alerts.
+  """
+
+  use DotcomWeb, :component
+
+  attr :alert, :map, required: true
```
thecristen, line 8: I think we can specify the struct here in lieu of `:map` (unless there's some compelling reason not to do that)

lib/dotcom_web/components/alerts.ex
```diff
@@ -0,0 +1,36 @@
...
+      <%= if @alert.image do %>
+        <img
+          class="w-full mb-4"
+          src={@alert.image}
+          alt={if @alert.image_alternative_text, do: @alert.image_alternative_text, else: ""}
+        />
+      <% end %>
+      <p class="my-0">{@alert.header}</p>
+      <hr class="h-px my-4 bg-gray-200 border-0" />
+      <p class="my-0">{@alert.description}</p>
```
thecristen, line 26: So I'm not sure what magic we do to get this description formatting... but it'd be really nice if we could figure out how to replicate it here. Also IIRC people like seeing the "updated by" because it's reassurance that the data is fresh.  
image

as opposed to 
image

lib/dotcom_web/components/planned_disruptions.ex
```diff
@@ -0,0 +1,103 @@
...
+    periods = Enum.sort_by(active_period, fn {start, _} -> start end)
+
+    {start, _} = List.first(periods)
+    {_, stop} = List.last(periods)
+
+    {start, stop}
+  end
+
+  # Extracts the subway lines from the alert's informed entity.
+  defp alert_lines(%Alert{informed_entity: %{entities: entities}}) do
```
thecristen, line 78: Would an alert with no subway line (e.g. the informed entity is only "Mattapan" or only one/some of the Green Line branch IDs) return nothing for the `:route_id` key and break your component?

lib/dotcom_web/components/alerts.ex
anthonyshull: We already found the function that does that. I'll add it in.

lib/dotcom_web/components/planned_disruptions.ex
anthonyshull: Yeah, it did break it. I didn't know they could do Mattapan only since the Mattapan branch effects the Red Line by definition. But, your route_pill work will solve this as I'll just pass in a list instead of trying to parse it out myself.

lib/dotcom_web/components/planned_disruptions.ex
```diff
@@ -0,0 +1,95 @@
...
+    </.bordered_container>
+    """
+  end
+
+  # A single disruption rendered as an accordion with a header and the alert as the content.
+  defp disruption(assigns) do
+    ~H"""
+    <.unstyled_accordion
+      summary_class="flex gap-sm p-2 items-center hover:bg-slate-600 focus:bg-slate-600 hover:text-white focus:text-white cursor-pointer group/row"
+      chevron_class="ml-auto mr-2 w-3 h-3"
```
thecristen, line 39:
```suggestion
      summary_class="flex gap-sm p-2 items-center hover:bg-brand-primary-lightest cursor-pointer group/row"
      chevron_class="fill-gray-lighter ml-auto mr-2 w-3 h-3"
```

fancy, but this'll be more consistent with the others

thecristen (approved):
Not sure if you got additional design feedback, but I added one consistency-related proposed change.

lib/dotcom_web/components/planned_disruptions.ex
thecristen: @anthonyshull TIL you didn't actually resolve this

PR #2388, opened 2025-02-18
Subway status in alerts page display

thecristen:
#### Summary of changes

**Asana Ticket:** [Service Alerts page | Add "Current Status" section for subway](https://app.asana.com/0/555089885850811/1209323323089299)

The component's got a few differences to its homepage counterpart! There's a lot of silly code things I'm working around so this implementation is on the slightly "dumb" side, but it works while building off the existing logic.

The preview page is modified to show both component versions side by side. 

Note: this PR uses the existing alert component for now, will be modified to use the same alert format from the planned work widget in #2387 

image


lib/dotcom/system_status/subway.ex
```diff
@@ -11,6 +11,7 @@ defmodule Dotcom.SystemStatus.Subway do
   @type status_t() :: :normal | Dotcom.SystemStatus.Alerts.service_effect_t()
 
   @type status_entry() :: %{
+          alerts: Alert.t(),
```
joshlarson, line 14: Question: Should this be a list of alerts, rather than a single alert?
```suggestion
          alerts: [Alert.t()],
```

(I think [Dialyzer might be complaining about this as well](https://github.com/mbta/dotcom/actions/runs/13397606493/job/37420443978?pr=2388), hence the failing build)

lib/dotcom_web/components/system_status/subway_status.ex
```diff
@@ -65,6 +131,12 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatus do
     |> Enum.map(&add_url/1)
   end
 
+  defp status_to_alerts_rows(subway_status) do
+    @route_ids
+    |> Enum.map(&{&1, subway_status |> Map.get(&1)})
+    |> Enum.flat_map(&rows_for_route(&1, false))
```
joshlarson, line 137: **Suggestion** (non-blocking): Use a keyword argument to make it clearer what the boolean value that's being passed in is there for.

```suggestion
    |> Enum.flat_map(&rows_for_route(&1, collapse: false))
```

Another option would be finding a different way of accomplishing the combining, for instance, pushing the combine-like-alerts step to a homepage-only post-processing step (kind of like the other collapsing bits), but honestly I don't think that's necessary for this PR.

lib/dotcom_web/live/system_status.ex
```diff
@@ -87,27 +61,31 @@ defmodule DotcomWeb.Live.SystemStatus do
     """
   end
 
-  defp alert(assigns) do
+  defp example_table(assigns) do
```
joshlarson, line 64: This is such a nice presentation for this!

lib/dotcom_web/components/system_status/subway_status.ex
```diff
@@ -57,6 +57,72 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatus do
...
+              ]}>
+                <.status_label
+                  status={row.status_entry.status}
+                  prefix={row.status_entry.prefix}
+                  plural={row.status_entry.plural}
+                />
+              </div>
+            </:heading>
+            <:content>
+              {Phoenix.View.render_one(row.alert, DotcomWeb.AlertView, "_item.html",
```
joshlarson, line 96: Annoyingly I think we'll need a component here that's slightly different from the regular alert, because this creates a kind of accordion-within-an-accordion situation, which is different from what's in the designs.

## Figma
Screenshot 2025-02-18 at 2 46 08 PM


## Current branch
Screenshot 2025-02-18 at 2 46 20 PM

joshlarson (changes requested):
This looks so good! I love how simple the solution is, and how clearly it exposes how the original version of this can be improved even more! 😅 

I added some non-blocking thoughts. The two that are blocking are:
- I'm pretty sure the type in `Subway.subway_status` should be a list, rather than a single alert (I guess Dialyzer will also block approval on that 😅).
- The way the alerts are rendering doesn't look quite right.

lib/dotcom_web/components/system_status/subway_status.ex
thecristen: Yeah, the actual plan is to use the alert component from Anthony's branch in #2387

lib/dotcom_web/components/system_status/subway_status.ex
thecristen: Oh, I like this!

lib/dotcom_web/components/system_status/subway_status.ex
joshlarson: Ahaaa - gotcha!

test/dotcom/system_status_test.exs
```diff
@@ -0,0 +1,90 @@
...
+    ]
+
+  import Mox
+  import Test.Support.Factories.Alerts.Alert
+
+  setup :verify_on_exit!
+
+  describe "subway_alerts_for_today/0" do
+    test "requests alerts for all subway lines @ today datetime" do
+      today = Faker.DateTime.forward(1)
```
anthonyshull, line 20: You can use the generator `random_date_time()`.

test/dotcom/system_status_test.exs
```diff
@@ -0,0 +1,90 @@
...
+        assert Enum.sort(route_ids) == Dotcom.Routes.subway_route_ids() |> Enum.sort()
+        []
+      end)
+
+      _ = subway_alerts_for_today()
+    end
+
+    # does Alerts.Repo.by_route_ids/2 not already scope results to the given day?
+    test "filters alerts for falling within a day" do
+      stub_with(Dotcom.Utils.DateTime.Mock, Dotcom.Utils.DateTime)
```
anthonyshull, line 36: You can move these up into a `setup` block and they will be overridden only if you do an `expect` in the test.

lib/dotcom/alerts/disruptions/subway.ex
```diff
@@ -36,7 +37,7 @@ defmodule Dotcom.Alerts.Disruptions.Subway do
   # 4. Sorts the alerts within the group by start time.
   defp disruption_groups() do
     subway_route_ids()
-    |> @alerts_repo.by_route_ids(Utils.DateTime.now())
+    |> @alerts_repo.by_route_ids(@date_time_module.now())
```
anthonyshull, line 40: 👍

lib/dotcom/system_status.ex
```diff
@@ -7,33 +7,16 @@ defmodule Dotcom.SystemStatus do
...
-      "Green-D",
-      "Green-E",
-      "Mattapan"
-    ]
-    |> @alerts_repo.by_route_ids(datetime)
-    |> SystemStatus.Alerts.for_day(datetime)
-    |> SystemStatus.Alerts.filter_relevant()
+    Dotcom.Alerts.Disruptions.Subway.todays_disruptions()
+    |> Map.get(:today, [])
+    |> SystemStatus.Alerts.for_day(@date_time_module.now())
```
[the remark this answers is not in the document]
anthonyshull: I don't think you'll need that extra call. Is there a difference between:

```
Dotcom.Alerts.Disruptions.Subway.todays_disruptions()
 |> Map.get(:today, [])
```
and
```
Dotcom.Alerts.Disruptions.Subway.todays_disruptions()
|> Map.get(:today, [])
|> SystemStatus.Alerts.for_day(@date_time_module.now())
```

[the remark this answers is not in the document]
anthonyshull: Maybe if it just filters out alerts that are today, but earlier today, that can just be a private function in this module.

lib/dotcom/system_status/alerts.ex
```diff
@@ -54,24 +54,24 @@ defmodule Dotcom.SystemStatus.Alerts do
...
     end)
   end
 
   @doc """
-  Given a list of alerts, filters only the ones that are active today, as defined in `active_on_day?/2`.
+  Given a list of alerts, filters only the ones that are active at the, as defined in `active_now_or_later_on_day?/2`.
   """
   def for_day(alerts, datetime) do
-    Enum.filter(alerts, &active_on_day?(&1, datetime))
+    Enum.filter(alerts, &active_now_or_later_on_day?(&1, datetime))
```
anthonyshull, line 74: Unless these are used anywhere else, I *think* they can be moved to the system status module as a private function.

anthonyshull (approved):

test/dotcom/system_status_test.exs
```diff
@@ -1,21 +1,70 @@
...
+      _ = subway_status()
+    end
+
+    test "returns status without alerts ending earlier that day" do
+      route_id_with_alerts = Dotcom.Routes.subway_route_ids() |> Faker.Util.pick()
+      line = Dotcom.Routes.line_name_for_subway_route(route_id_with_alerts)
+      {day_start, day_end} = service_range_day()
+      earlier_end = Timex.shift(day_start, minutes: 1)
+      alert_today = disruption_alert({day_start, day_end}, route_id_with_alerts)
+      alert_earlier = disruption_alert({day_start, earlier_end}, route_id_with_alerts)
```
joshlarson, line 41: It took me a bit to understand what these variables were. What about names like 👇?

```suggestion
      currently_active_alert = disruption_alert({day_start, day_end}, route_id_with_alerts)
      expired_alert = disruption_alert({day_start, earlier_end}, route_id_with_alerts)
```

test/dotcom/system_status_test.exs
```diff
@@ -120,58 +169,31 @@ defmodule Dotcom.SystemStatus.AlertsTest do
...
+  end
 
-    test "does not include alerts with other effects" do
-      assert Alerts.filter_relevant([
-               build(:alert, effect: :policy_change),
-               build(:alert, effect: :extra_service),
-               build(:alert, effect: :stop_closure)
-             ]) == []
-    end
+  defp disruption_alert(active_period, route_id) do
```
joshlarson, line 192: Question: will this test file ever want to interact with alerts that are for multiple route_ids (other than the entire green line)?

Non-blocking, because this can very easily be a later question, and can be addressed if/when it's needed.

joshlarson (approved):
Looks great!

One non-blocking comment and one non-blocking question, but I think this looks ready to go!

test/dotcom/system_status_test.exs
thecristen: Yeah, probably!

test/dotcom/system_status_test.exs
thecristen: I'll make this change

PR #2389, opened 2025-02-19
String match bug cause 'href' can be a list.

anthonyshull:
https://mbtace.sentry.io/issues/6268836106/?environment=prod&project=4507102206820352&referrer=project-issue-stream

### DIFF
--- a/lib/dotcom/content_rewriters/links.ex
+++ b/lib/dotcom/content_rewriters/links.ex
@@ -48,11 +48,12 @@ defmodule Dotcom.ContentRewriters.Links do
     attr_map = Enum.into(attrs, %{})
     locale = String.split(languages, "/") |> List.last()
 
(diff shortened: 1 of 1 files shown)


lib/dotcom/content_rewriters/links.ex
```diff
@@ -49,7 +49,7 @@ defmodule Dotcom.ContentRewriters.Links do
     href = Map.get(attr_map, "href", "")
```
thecristen, line 49: Looks like when it's a list it's overwhelmingly a single string in the list. I think we should use this string instead of ignoring it with your change below. Maybe something

```elixir
href = case Map.get(attr_map, "href") do
  href when is_binary(href) -> href
  [href] -> href
  _ -> ""
end
```

thecristen (approved):

PR #2395, opened 2025-02-21
feat(AlertController): put new features on the subway alerts page

thecristen:
#### Summary of changes

**Asana Ticket:** [Service Alerts page | Refine Station and Service Alerts section](https://app.asana.com/0/555089885850811/1209323323089304) and [Service Alerts page | Add "Current Status" section for subway](https://app.asana.com/0/555089885850811/1209323323089299)

image

Most of the changes here were adding an additional sub-layout for the content of the alert page. Everything's the same between modes except for: 
- the icon shown next to each route (for subway it's the `subway_route_pill`, all others `route_icon`)
- the `<h2>` shown above the list of alerts, as there's no heading shown for non-subway
- I changed the verbiage of the "no alerts here" message for subway to read more like "no other alerts here", since there might be other alerts covered by the Current Status or Planned Disruptions sections
- and of course, the new sections are only included for subway

Outside of the main content, the timeframe filters are also omitted from the sidebar for subway only.

Hopefully the structure is clear enough. Please omit the first commit from review, as it is already covered in #2388

### DIFF
--- a/lib/dotcom_web/live/system_status.ex
+++ b/lib/dotcom_web/live/system_status.ex
@@ -18,23 +18,26 @@ defmodule DotcomWeb.Live.SystemStatus do
   @date_time_module Application.compile_env!(:dotcom, :date_time_module)
 
   def render(assigns) do
+    now = @date_time_module.now()
+
     live_alerts =
       subway_route_ids()
-      |> Alerts.Repo.by_route_ids(@date_time_module.now())
-      |> Enum.filter(&SystemStatus.status_alert?(&1, @date_time_module.now()))
+      |> Alerts.Repo.by_route_ids(now)
+      |> Enum.filter(&SystemStatus.status_alert?(&1, now))
 
     assigns =
       assigns
       |> assign(:alerts, live_alerts)
       |> assign(:examples, alerts_examples())
+      |> assign(:now, now)
 
     ~H"""
     <h1>Live Data</h1>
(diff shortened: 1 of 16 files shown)


thecristen:
Also added a small change forgotten from [System Status | Put it on the homepage](https://app.asana.com/0/555089885850811/1209091204225418) - changing the name of the homepage tab from "Schedules" to "Routes"

lib/dotcom_web/live/system_status.ex
```diff
@@ -18,23 +18,26 @@ defmodule DotcomWeb.Live.SystemStatus do
...
 
     ~H"""
     <h1>Live Data</h1>
-    <.example_table alerts={@alerts} />
+    <.example_table alerts={@alerts} now={@now} />
 
     <h1>Examples</h1>
     <div :for={example <- @examples} class="mb-4">
-      <.example_table alerts={example.alerts} />
+      <.example_table alerts={example.alerts} now={@now} />
```
joshlarson, line 40: Minor: the way it was set up before guaranteed that the 8:30PM scheduled alert would always show up as `8:30 PM`. As this code is, if we look at this after 8:30 on a given day, then it'll render as current.

Minor because if we're looking at the preview page after 8:30 PM, maybe that's a sign that it's time to sign off for the day 😅

anthonyshull (approved):

PR #2396, opened 2025-02-21
Add stop headsign to repo condensed

anthonyshull:


thecristen (approved):
Let's try it, pending tests passing

PR #2397, opened 2025-02-21
feat(RouteSymbols): improve subway_route_pill readout

thecristen:
#### Summary of changes

**Asana Ticket:** [System Status | 🐞 Voice Over should work even when the route pill is suppressed](https://app.asana.com/0/555089885850811/1209412979464961)

Technically also improves the subway route pill so it'll read like "Red Line" instead of "RL".

Able to confirm by inspecting the accessibility tree on the /preview/subway-status page in your favorite browser dev tools.


joshlarson (approved):
Looks (and sounds) great! Good find on `opacity-0` - that does exactly what we need here!

PR #2398, opened 2025-02-24
Use stop headsign info on stop pages

anthonyshull:
https://app.asana.com/0/555089885850811/1209276281980955/f

Screenshot Capture - 2025-02-24 - 14-36-09


assets/ts/stop/components/DepartureCard.tsx
```diff
@@ -76,33 +107,43 @@ const DepartureCard = ({
...
       </a>
-      {sortedRoutePatternsByHeadsign.map(
+      {Object.entries(departures).map(
         ([
           headsign,
-          { direction_id: directionId, route_patterns: routePatterns }
+          departures
         ]) => {
+          let alerts : Alert[] = [];
+          let directionId = 0;
```
anthonyshull, line 116: 📓 This is the only thing I am wondering about. Does this matter if there are no actual departures?

assets/ts/stop/__tests__/DeparturesAndMapTest.tsx
```diff
@@ -71,39 +71,6 @@ describe("DeparturesAndMap", () => {
     });
   });
 
-  it("toggles departure list on click", async () => {
```
anthonyshull, line 74: 📓 This clearly works when using the app. I'm not sure why the test is failing other than some hardcoding changing. Not sure it's worth fixing this test.

assets/ts/stop/__tests__/DepartureCardTest.tsx
```diff
@@ -149,10 +204,8 @@ describe("DepartureCard", () => {
     );
 
     await waitFor(() => {
-      const suspensionBadges = screen.getAllByText("No Service");
-      expect(suspensionBadges.length).toBe(Object.keys(routePatterns).length);
-      expect(screen.queryByText("Shuttle Service")).toBeNull();
-      expect(screen.queryByText("Detour")).toBeNull();
```
anthonyshull, line 155: These are no longer applicable as they only occur when there are no departures. Because we have a departure, the matching alert is 'Detour'.

assets/ts/stop/components/DeparturesAndMap.tsx
```diff
@@ -157,11 +157,15 @@ const DeparturesAndMap = ({
...
+        const { route, schedule, trip } = departure;
         return (
           route.id === activeRow.route.id &&
           trip.direction_id === activeRow.directionId &&
-          trip.headsign === activeRow.headsign
+          // We have to match the headsign from the schedule or the trip.
+          // If the schedule doesn't have a headsign, we use the trip's headsign.
+          (schedule?.stop_headsign === activeRow.headsign ||
+            (schedule?.stop_headsign === null &&
+              trip.headsign === activeRow.headsign))
```
thecristen, line 168:
```suggestion
          (schedule?.stop_headsign === activeRow.headsign ||
            trip.headsign === activeRow.headsign)
```
FYI if `schedule` is undefined (this may happen for predictions with `schedule_relationship=added`) then `schedule?.stop_headsign === null` will be false as `schedule?.stop_headsign` will resolve to `undefined`. Anyway, am not convinced you need to check that in this line anyway.

thecristen (commented):
This mostly seems reasonable, but it's showing _so_ many more polylines and I'm banging my head here triyng to figure out why... any idea what's behind that?

For example, https://dev-blue.mbtace.com/stops/place-sstat?route=CR-Worcester&direction_id=0&headsign=Framingham

image

Whereas before: 

image

thecristen:
@anthonyshull Not sure you need it at this point but FYSA there's background docs on the design and requirements of this page (incl. links to relevant Figma parts) https://www.notion.so/mbta-downtown-crossing/Stop-Page-Designs-Logic-5e3d88244d33464aa9adcf44c7986d73?pvs=4

anthonyshull:
> This mostly seems reasonable, but it's showing _so_ many more polylines and I'm banging my head here triyng to figure out why... any idea what's behind that?
> 
> For example, https://dev-blue.mbtace.com/stops/place-sstat?route=CR-Worcester&direction_id=0&headsign=Framingham

I'm only seeing one line:
Screenshot Capture - 2025-03-03 - 08-07-11

assets/ts/stop/__tests__/DepartureCardTest.tsx
anthonyshull: I don't think so, I think that because everything was route pattern based previously, they just didn't pass in any departures in the tests. But, now that everything is departure based, you have to pass in a departure or it falls back to the default route pattern behavior of listing it as no departures.

thecristen: Have checked on it and confirmed what I wanted to see 👍🏼

thecristen:
@anthonyshull Thanks for checking, I think it just depends on the particular headsign. [https://dev-blue.mbtace.com/stops/place-sstat?route=CR-Franklin&direction_id=0&headsign=Forge+Park%2F495+via+Fairmount](https://dev-blue.mbtace.com/stops/place-sstat?route=CR-Franklin&direction_id=0&headsign=Forge+Park%2F495+via+Fairmount)

image

Maybe it's just returning all the route's polylines since it's failing to match on the headsign... actually yeah I bet that's it, since the stop_headsign doesn't necessarily match the shape's associated headsign, which is derived from a trip headsign. I don't think there's a straightforward resolution to _that_, if you agree I'm happy propose that be left to a future improvement.

thecristen (approved):

PR #2400, opened 2025-02-26
fix: `SystemStatus` QA improvements 

thecristen:
#### Summary of changes

In order:

1. **Asana Ticket:** [[PW/SS] Fix text wrapping](https://app.asana.com/0/555089885850811/1209437138451266)
2. Hiding the status icon from screen readers, as the text alone is sufficient
3. _Part_ of [Tweaks to delay handling, homepage + /alerts/subway system status](https://app.asana.com/0/555089885850811/1209498039973244)
  - Add "Expect " to all future delays (`Expect Delay`)
  - For all future alerts, adjust time display:
    - Capitalize AM/PM 
    - Add space between time & AM/PM
    - Hide :00 if even hour
    - `8:00pm → 8 PM`

### DIFF
--- a/lib/dotcom_web/components/system_status/status_label.ex
+++ b/lib/dotcom_web/components/system_status/status_label.ex
@@ -6,6 +6,8 @@ defmodule DotcomWeb.Components.SystemStatus.StatusLabel do
 
   use DotcomWeb, :component
 
+  alias Alerts.Alert
+
   attr :prefix, :string, default: nil
   attr :plural, :boolean, default: false
   attr :status, :atom, required: true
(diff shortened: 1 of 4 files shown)


lib/util/util.ex
```diff
@@ -225,6 +225,36 @@ defmodule Util do
     time |> Timex.format!("{kitchen}") |> String.downcase()
   end
 
+  @doc """
+  A very concise representation of time, with period (AM/PM).
+
+  ## Examples
```
joshlarson, line 231: Nice!

lib/dotcom_web/components/system_status/status_label.ex
```diff
@@ -37,13 +40,18 @@ defmodule DotcomWeb.Components.SystemStatus.StatusLabel do
...
   end
 
   defp status_icon(assigns) do
     ~H"""
-    <.icon class="h-[1.125rem] w-[1.125rem]" type="icon-svg" name={status_icon_name(@status)} />
+    <.icon
+      class="h-[1.125rem] w-[1.125rem] shrink-0"
+      type="icon-svg"
+      name={status_icon_name(@status)}
+      aria-hidden={true}
```
joshlarson, line 53: Yeah this makes the VoiceOver readout a lot better.

lib/dotcom_web/components/system_status/status_label.ex
```diff
@@ -22,11 +22,14 @@ defmodule DotcomWeb.Components.SystemStatus.StatusLabel do
     ~H"""
     <span class={[status_classes(@status), "flex items-center gap-2"]}>
       <.status_icon status={@status} />
-      {@rendered_prefix} {description(@status, @plural)}
+      {@rendered_prefix} {expectation(@status, @prefix)} {description(@status, @plural)}
     </span>
     """
   end
 
+  defp expectation(:delay, prefix) when is_binary(prefix), do: "Expect"
```
joshlarson, line 30: Non-blocking: This is a weird and wacky enough use case that it probably deserves both a test and an explanatory comment (or at least one or the other of those 😅).

lib/dotcom_web/components/system_status/status_label.ex
```diff
@@ -22,11 +22,14 @@ defmodule DotcomWeb.Components.SystemStatus.StatusLabel do
     ~H"""
     <span class={[status_classes(@status), "flex items-center gap-2"]}>
       <.status_icon status={@status} />
-      {@rendered_prefix} {description(@status, @plural)}
+      {@rendered_prefix} {expectation(@status, @prefix)} {description(@status, @plural)}
```
joshlarson, line 25: Non-blocking: I think that this would be easier to understand if the whole decision tree of `"Delay"` versus `"Expect Delays"` versus some other status like `"Shuttle"` happened purely inside the `description` function. That means that `description` would need to just take the whole glob of data at this point - either `defp description(@status, @prefix, @plural)` or just `defp description(@status_entry)`, but we wouldn't have the distinction between `"Delay"` and `"Expect Delay"` be a top-level concern.

I'm happy to pair on this, and/or treat it as a follow-up, since it does clearly work, and the feature here is worth getting merged, but I would like to clean this up at some point.

joshlarson (approved):
Looks great!

Left a few non-blocking comments, and disabled auto-merge so that the Github 🤖 wouldn't scoop you if you wanted to respond to them.

lib/dotcom_web/components/system_status/status_label.ex
thecristen: > happened purely inside the description function

Ha that's how I started! Though I wasn't clever enough to try `description(@status, @prefix, @plural)`, which I like here -- will try that

PR #2402, opened 2025-02-26
fix: Combine contiguous active periods

joshlarson:
### Changes
Screenshot 2025-02-26 at 2 42 27 PM

Before is on the left, After is on the right. Note that:
- The OL disruptions have narrower date ranges now, based on not combining the _whole_ active period range anymore. (For instance, the "This week" OL disruption only goes until Feb 28)
- The BL disruptions got split up into each weekend, instead of spanning the whole week.
- Disruptions are still not sorted correctly within their buckets - that's out of scope for this PR.

---

**Asana Ticket:** [[PW] Fix contiguous active period logic](https://app.asana.com/0/555089885850811/1209437138451265)

---

TODO
- [x] Tests are a bit flaky - I goofed something in my Faker setup
- [ ] Docs have some sentence fragments
- [ ] Doctest or regular test for `DateTime.shift/2`


lib/dotcom/alerts/disruptions/subway.ex
```diff
@@ -46,6 +48,38 @@ defmodule Dotcom.Alerts.Disruptions.Subway do
     |> Enum.into(%{})
   end
 
+  # Given an alert with multiple active periods, some of which may be
+  # contiguous with one another (next one starts the same day or the
+  # day after the previous one), combines the contiguous active
+  # periods, and then returns
+  defp separate_active_periods(alert) do
```
thecristen, line 55: **suggestion**: I love the structure of this function but not the name because I'm taking it too literally 😅  .. maybe something like `reconfigure_active_periods` or `split_discontiguous_active_periods` ...?

test/dotcom/alerts/disruptions/subway_test.exs
```diff
@@ -119,6 +270,7 @@ defmodule Dotcom.Alerts.Disruptions.SubwayTest do
       assert %{today: [^alert_today]} = todays_disruptions()
     end
 
+    @tag :skip
```
thecristen, line 279: Does this need to be skipped?

lib/dotcom/utils/date_time.ex
```diff
@@ -55,6 +55,16 @@ defmodule Dotcom.Utils.DateTime do
     |> coerce_ambiguous_date_time()
   end
 
+  @doc """
+  This works very much like `Timex.shift/2`, except that it calls
+  `coerce_ambiguous_date_time/1` in order to get a pure `DateTime` as
+  a result.
+  """
+  def shift(date_time, opts) do
```
thecristen, line 63: **suggestion**: I was going to suggest renaming this to avoid looking like Elixir's own `DateTime.shift/3`, but perhaps you'd rather use that directly instead of making this new function anyway 😇

test/dotcom/alerts/disruptions/subway_test.exs
```diff
@@ -79,20 +82,168 @@ defmodule Dotcom.Alerts.Disruptions.SubwayTest do
              } = future_disruptions()
     end
 
-    test "handles alert with more than one active_period" do
```
thecristen, line 82: Can we keep the existing tests as-is?

thecristen: (I just mean to not delete/rewrite them)

test/dotcom/alerts/disruptions/subway_test.exs
joshlarson: Ohhh woops! No, this needs to be repaired, I think

lib/dotcom/alerts/disruptions/subway.ex
```diff
@@ -46,6 +48,38 @@ defmodule Dotcom.Alerts.Disruptions.Subway do
...
+    end)
+  end
+
+  # Given a list of active periods
+  defp combine_contiguous_active_periods([active_period1, active_period2 | rest]) do
+    {start1, end1} = active_period1
+    {start2, end2} = active_period2
+
+    if Timex.before?(
+         ServiceDateTime.service_date(end1) |> Timex.shift(days: 1),
```
thecristen, line 69: I think you meant to use the other `shift` here?

lib/dotcom/alerts/disruptions/subway.ex
```diff
@@ -46,6 +48,38 @@ defmodule Dotcom.Alerts.Disruptions.Subway do
...
+         ServiceDateTime.service_date(start2)
```
thecristen, line 70: Thanks for comparing by service dates here!

test/dotcom/alerts/disruptions/subway_test.exs
joshlarson: This test can't be preserved exactly as-is, because an alert with more than one `active_period` won't make it all the way into the output. The new test (`"splits the active_periods of alerts with more than one into different buckets"`) is actually as close as I was able to get with two separate `active_period`'s if they don't get combined.

But I suppose I could also restore a test with a setup closer to this one where the assert is updated to match what will happen in the new world (`long_alert` won't wind up exactly as-is in the output, but its `active_period`'s will get combined).

lib/dotcom/alerts/disruptions/subway.ex
```diff
@@ -46,6 +48,38 @@ defmodule Dotcom.Alerts.Disruptions.Subway do
...
 
+  # Given an alert with multiple active periods, some of which may be
+  # contiguous with one another (next one starts the same day or the
+  # day after the previous one), combines the contiguous active
+  # periods, and then returns
+  defp separate_active_periods(alert) do
+    alert.active_period
+    |> combine_contiguous_active_periods()
+    |> Enum.map(fn active_period ->
+      %Alerts.Alert{alert | active_period: [active_period]}
```
thecristen, line 59: question: if the original alert has `active_period: []` does this result in `%Alerts.Alert{alert | active_period: [nil]}`? or maybe just `[]`?!

(I actually don't know how possible/probable such an alert is, but I did find other code in Dotcom which presumes it 💀 I'm theorizing it might happen for cancellations. In any case I'd expect an output with an alert with a `[]` active_period rather than a `[nil]` one) 😅

lib/dotcom/alerts/disruptions/subway.ex
joshlarson: [PTALA](https://github.com/mbta/dotcom/pull/2402/commits/421311d2481f70ad9d404dcd6dda9f59e1e013f5)

test/dotcom/alerts/disruptions/subway_test.exs
joshlarson: [PTALA](https://github.com/mbta/dotcom/pull/2402/commits/a9352bfdad7ddb19104e2ab988ed873597436dee)

lib/dotcom/utils/date_time.ex
joshlarson: [PTALA](https://github.com/mbta/dotcom/pull/2402/commits/7ae76ea506abada69345a3f556de34e06fb98d0d)

test/dotcom/alerts/disruptions/subway_test.exs
```diff
@@ -79,20 +81,174 @@ defmodule Dotcom.Alerts.Disruptions.SubwayTest do
...
+        )
+
+      {beginning_of_week_after_next, _} = service_range_after_next_week()
+
+      active_period_2_start =
+        Generators.DateTime.random_time_range_date_time({beginning_of_week_after_next, nil})
+
+      active_period_2_end =
+        Generators.DateTime.random_time_range_date_time({active_period_2_start, nil})
+
```
thecristen, line 105: question: Why not have the second active period be just `service_range_after_next_week()`? Since you modified the first active period it won't get combined right?

test/dotcom/alerts/disruptions/subway_test.exs
thecristen: Thanks for explaining! It took me too long to realize that's what the first test was doing 😓

joshlarson: [PTALA](https://github.com/mbta/dotcom/pull/2402/commits/dee78bd7903bd8db58f62b7f325da6e8e6ed14f3)

(Annoyingly, Git is rendering the diff weird, so it's showing the restored test as being fully added, and then its removal diff kind of piecemeal through the rest of the diff, but I did restore it as best I could.)

lib/dotcom/alerts/disruptions/subway.ex
joshlarson: I think this would drop the alert, so

```elixir
[%Alert{active_period: []}]
|> split_by_discontiguous_active_periods()
```

would return
```elixir
[]
```

While not intentional, I think that's the right behavior??

(I also suspect that there's code earlier on that will either crash or prevent any alerts with `[]` active_periods from getting to this spot.)

test/dotcom/alerts/disruptions/subway_test.exs
joshlarson: I admit that I'm new to Faker/communicative-tests/etc, but my understanding is that using `service_range_after_next_week()` would imply that there's something special about the active period starting right at the service range boundary, while this test doesn't actually care about that level of detail. 

(Also at least for right now, using `service_range_after_next_week()` would force me to deal with the indefinite-alert use case, which I didn't want to do yet 😇)

test/dotcom/alerts/disruptions/subway_test.exs
```diff
@@ -81,18 +83,191 @@ defmodule Dotcom.Alerts.Disruptions.SubwayTest do
...
+      active_period_1_end =
+        Generators.DateTime.random_time_range_date_time(
+          {active_period_1_start, DateTime.shift(end_of_week, day: -2)}
+        )
+
+      active_period_2_start =
+        Generators.DateTime.random_time_range_date_time(
+          {active_period_1_end, ServiceDateTime.end_of_service_day(active_period_1_end)}
+        )
```
thecristen, line 241: randomness on top of randomness!

I think these tests might all be more readable if we leaned into `shift`ing more and generating bespoke random times less. 

What if we pick _one_ datetime from the `random_time_range` as a start, shift that by a day or hour or whatever for the end datetime, shift _that_ by N days to define the start of the next period, and so on.... IMO this would let the values still be random, but the chosen datetimes would be easier to understand in relation to each other (and it'd probably be a lot fewer lines to write here).

lib/dotcom/alerts/disruptions/subway.ex
thecristen: I like it!

test/dotcom/alerts/disruptions/subway_test.exs
joshlarson: Haha I started out doing it the way you suggested, and then ran into a flaky test. The randomness solved it, believe it or not.

The day right before daylight savings time starts, there is a one-hour period of time where 23 hours later (aka `Timex.shift(days: 1)` pushes us not one but two service days forward.

I can see if `DateTime.shift(day: 1)` avoids that problem, but it would surprise me a bit if it did.

lib/dotcom/alerts/disruptions/subway.ex
joshlarson: Well, uh... yes I did 👀 

In this case, I've decided that I want `Date.shift/2`, which I had not known about until about 30 seconds ago.

joshlarson: [PTALA](https://github.com/mbta/dotcom/pull/2402/commits/c5a9bcf4f52ee3cd6282a10df441054ef78adb89)

lib/dotcom/alerts/disruptions/subway.ex
thecristen: OK, I think that makes sense. And if a new scenario of alert proves to be an issue, we can cross that bridge when we get to it

test/dotcom/alerts/disruptions/subway_test.exs
thecristen: It's fair, though we could also wonder whether there's something special about the active periods _not_ being at the boundary for these tests :P

test/dotcom/alerts/disruptions/subway_test.exs
thecristen: Pesky times! Okay then.

thecristen (approved):
Thanks!!

lib/dotcom/alerts/disruptions/subway.ex
```diff
@@ -7,6 +7,7 @@ defmodule Dotcom.Alerts.Disruptions.Subway do
   import Dotcom.Routes, only: [subway_route_ids: 0]
   import Dotcom.Utils.ServiceDateTime, only: [all_service_ranges: 0, service_range: 1]
 
+  alias Dotcom.Utils.ServiceDateTime
```
anthonyshull, line 10: Should be alphabetical with the ones below.

joshlarson: Woops - good catch thanks!

PR #2403, opened 2025-02-26
hotfix(Alerts.Disruptions.Subway): process alerts with no end datetime

thecristen:
I _think_ this is everything needed to handle alerts configured as "until further notice", but I didn't want to recreate that alert and break dev-green/blue while people were using it to review :)

- `service_range_range` has been updated to gracefully handle `nil`, and tests have been added for that
- More creative formatting has been added to handle active periods with missing start or stop times.

### DIFF
--- a/test/dotcom/utils/service_date_time_test.exs
+++ b/test/dotcom/utils/service_date_time_test.exs
@@ -226,6 +226,38 @@ defmodule Dotcom.Utils.ServiceDateTimeTest do
     end
   end
 
+  describe "service_range_range/2" do
+    test "returns all service ranges between two datetimes" do
+      # Setup
+      today = service_range_day() |> random_time_range_date_time()
+      next_week = service_range_next_week() |> random_time_range_date_time()
+
+      # Exercise / Verify
+      assert service_range_range(today, next_week) == [:today, :this_week, :next_week]
+    end
+
+    test "handles open end" do
+      # Setup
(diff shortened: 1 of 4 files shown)


thecristen:
Closing a stop on the Blue Line from today/9:34AM until further notice looks like this across the three components:

image
image

test/dotcom/utils/service_date_time_test.exs
```diff
@@ -226,6 +226,38 @@ defmodule Dotcom.Utils.ServiceDateTimeTest do
...
+    test "returns all service ranges between two datetimes" do
+      # Setup
+      today = service_range_day() |> random_time_range_date_time()
+      next_week = service_range_next_week() |> random_time_range_date_time()
+
+      # Exercise / Verify
+      assert service_range_range(today, next_week) == [:today, :this_week, :next_week]
+    end
+
+    test "handles open end" do
```
anthonyshull, line 239: This could be more descriptive in line with the actual assertion. Also, we should call it 'stop' rather than 'end.'

anthonyshull (approved):

test/dotcom/utils/service_date_time_test.exs
thecristen: Good call, I've [changed this](https://github.com/mbta/dotcom/pull/2403/commits/5f2c060c769962c97362ec478dfdb0e489e7e265) to describe more directly what's happening.

PR #2406, opened 2025-02-28
fix(DotcomWeb.PredictionsChannel): let skipped/cancelled with no departure times through

thecristen:
#### Summary of changes

**Asana Ticket:** [Stop Pages don't display cancelled trips with strikethrough](https://app.asana.com/0/555089885850811/1207956121805040)

...it turns out that skipped/cancelled predictions without departure times (e.g. probably all of them) were never making it to the stops page in the first place (on account of we were rejecting every prediction having `nil` departure_time). This PR should fix that. 

And this should be enough to resolve the cited issue, as everything's in place in the frontend to give skipped/cancelled trips the strikethrough treatment they deserve.

### DIFF
--- a/lib/dotcom_web/channels/predictions_channel.ex
+++ b/lib/dotcom_web/channels/predictions_channel.ex
@@ -72,12 +72,9 @@ defmodule DotcomWeb.PredictionsChannel do
   end
 
   # For predictions which have a departure time, filter out ones that are past
-  defp departure_exists_in_past?(prediction) do
-    if prediction.departure_time do
-      Timex.before?(prediction.departure_time, now())
-    else
-      # it doesn't exist
-      false
-    end
+  defp departure_exists_in_past?(%{departure_time: nil}), do: false
+
+  defp departure_exists_in_past?(%{departure_time: departure_time}) do
+    Timex.before?(departure_time, now())
   end
 end
--- a/lib/dotcom_web/channels/predictions_channel.ex
+++ b/lib/dotcom_web/channels/predictions_channel.ex
@@ -7,6 +7,8 @@ defmodule DotcomWeb.PredictionsChannel do
 
   require Routes.Route
 
+  import Dotcom.Utils.DateTime, only: [now: 0]
+
   alias Routes.Route
   alias Phoenix.{Channel, Socket}
   alias Predictions.Prediction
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/channels/predictions_channel.ex
```diff
@@ -46,24 +48,36 @@ defmodule DotcomWeb.PredictionsChannel do
...
 
-  defp in_future_seconds?(prediction) do
-    Timex.compare(prediction.departure_time, Util.now(), :seconds) >= 0
+  # For predictions which have a departure time, filter out ones that are past
+  defp departure_exists_in_past?(prediction) do
+    if prediction.departure_time do
+      Timex.before?(prediction.departure_time, now())
+    else
+      # it doesn't exist
+      false
```
anthonyshull, line 80: Might be cleaner with pattern matching.

test/dotcom_web/channels/predictions_channel_test.exs
```diff
@@ -89,13 +114,15 @@ defmodule DotcomWeb.PredictionsChannelTest do
 
     test "filters out past predictions", context do
       # Setup
-      now = Timex.now() |> Timex.shift(seconds: 1)
-      past = Timex.shift(now, seconds: -15)
-      future = Timex.shift(now, seconds: 15)
+      now = now() |> DateTime.shift(second: 1)
+      past = DateTime.shift(now, second: -15)
+      future = DateTime.shift(now, second: 15)
```
anthonyshull, line 119: Why are we shifting it 1 second and then 15 seconds? I remember something about predictions being off or needing to be rounded. But, if we're just looking for before and after we should shift by the smallest increment. Otherwise, why not 1 minute or 1 hour, etc.

test/dotcom_web/channels/predictions_channel_test.exs
```diff
@@ -2,9 +2,10 @@ defmodule DotcomWeb.PredictionsChannelTest do
   use DotcomWeb.ChannelCase, async: false
 
   import Mox
+  import Dotcom.Utils.DateTime, only: [now: 0]
```
anthonyshull, line 5: Should credo flag this as not being in alphabetical order?

thecristen: :( credo isn't smart enough to alphabetize sadly, but thanks for calling it out here

test/dotcom_web/channels/predictions_channel_test.exs
thecristen: That's a great point, why _not_ 1 minute or hour...

PR #2410, opened 2025-03-03
Add stop headsigns to rail pages

anthonyshull:
https://app.asana.com/0/555089885850811/1209276227078169/f


assets/ts/schedule/components/line-diagram/StopPredictions.tsx
```diff
@@ -55,13 +55,24 @@ const StopPredictions = ({
...
+
       return (
-        <div key={`${headsign.name}-cr`}>
+        <div key={`${headsignCopy.name}-cr`}>
           <div className="m-schedule-diagram__cr-prediction">
             {timeForCommuterRail(predictedOrScheduledTime, predictionTimeClass)}
           </div>
           <div className="m-schedule-diagram__cr-prediction-details">
-            {predictionTexts(headsign).map((text: string) => (
+            {predictionTexts(headsignCopy).map((text: string) => (
```
thecristen, line 75: question: Isn't `headsign.name` already corrected via the backend change? Not sure we need any of this...

anthonyshull: Unfortunately, it isn't for every case. This is actually effecting three different places rather than two.

1. The line diagram
2. The list of trips in the schedule finder
3. The predictions that sometimes appear at the top of the schedule finder

anthonyshull: I figured changing it like this is less risky and time-consuming than trying to figure out when and where the backend data is getting overwritten since there are functions that are like...if commuter rail do X, if bus do Y.

thecristen (approved):
wacky, thanks for explaining!

PR #2414, opened 2025-03-04
hotfix(TimetableController): handle no route patterns

thecristen:
The timetable logic assumes canonical route patterns exist for every route. When that doesn't happen to be the case, we should gracefully handle that. This PR stops `Enum.EmptyError` from ruining everything, by setting an initial accumulator for `Enum.reduce/3`! 

The test is a little silly at this point but it did help me clearly see the fail vs success state!

### DIFF
--- a/test/dotcom_web/controllers/schedule/timetable_controller_test.exs
+++ b/test/dotcom_web/controllers/schedule/timetable_controller_test.exs
@@ -214,6 +214,14 @@ defmodule DotcomWeb.ScheduleController.TimetableControllerTest do
 
       assert ["1", "2", "3", "4", ^special_stop_id, "5", "6"] = Enum.map(trip_stops, & &1.id)
     end
+
+    test "doesn't crash when missing canonical route patterns", %{conn: conn} do
+      expect(RoutePatterns.Repo.Mock, :by_route_id, fn _, _ ->
+        []
+      end)
+
+      assert build_timetable(conn, @schedules)
+    end
   end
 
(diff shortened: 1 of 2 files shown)


test/dotcom_web/controllers/schedule/timetable_controller_test.exs
```diff
@@ -214,6 +214,14 @@ defmodule DotcomWeb.ScheduleController.TimetableControllerTest do
 
       assert ["1", "2", "3", "4", ^special_stop_id, "5", "6"] = Enum.map(trip_stops, & &1.id)
     end
+
+    test "doesn't crash when missing canonical route patterns", %{conn: conn} do
+      expect(RoutePatterns.Repo.Mock, :by_route_id, fn _, _ ->
+        []
+      end)
+
+      assert build_timetable(conn, @schedules)
```
joshlarson, line 223: Non-blocking: Is there a way to make a non-trivial relevant empty-state assertion for this use case? Like that it pulls a route pattern from somewhere else, or that it has a reasonable empty state (empty list, `"No route pattern exists you bozo"` string, or something like that?)

joshlarson (approved):
📆

test/dotcom_web/controllers/schedule/timetable_controller_test.exs
thecristen: Yeah... a followup to test the "no trips here" version of the page will do

PR #2404, opened 2025-02-27
tests: Add tests validating homepage-subway-status collapsing behavior

joshlarson:
[This PR](https://github.com/mbta/dotcom/pull/2385) was introduced without tests for the homepage component, but upon further reflection, there's enough non-trivial stuff going on in there, that it does deserve some tests.

I'm doing this for two reasons:
- It's better to have this test coverage late than never
- There are tickets that I'm about to work on that affect how to collapse [Mattapan trolley status](https://app.asana.com/0/555089885850811/1209412979464965) and [green line status](https://app.asana.com/0/555089885850811/1209412979464963), which I think will involve substantially changing the collapsing logic, so I want to reduce the risk of introducing a regression.

---

No ticket.

---

Follow-up to:
- https://github.com/mbta/dotcom/pull/2385

Blocking:
- https://github.com/mbta/dotcom/pull/2415

PR #2415, opened 2025-03-05
fix: Avoid collapsing green line and Mattapan line rows unless needed

joshlarson:
This PR fixes a few edge cases in the homepage subway status component where it was unnecessarily collapsing green line rows together, or collapsing Mattapan alerts into Red line (which we decided it should never do).

#### Before
Screenshot 2025-03-05 at 11 50 27 AM

#### After
Screenshot 2025-03-05 at 11 47 38 AM

#### Before
Screenshot 2025-03-05 at 11 49 54 AM

#### After
Screenshot 2025-03-05 at 11 49 11 AM

### Asana Tickets
- [System Status | 🐞 Don't needlessly collapse Mattapan alerts into Red](https://app.asana.com/0/555089885850811/1209412979464965)
- [System Status | 🐞 Don't needlessly collapse GL alerts](https://app.asana.com/0/555089885850811/1209412979464963)

Blocked by:
- https://github.com/mbta/dotcom/pull/2404

### DIFF
--- a/lib/dotcom_web/components/system_status/subway_status.ex
+++ b/lib/dotcom_web/components/system_status/subway_status.ex
@@ -151,39 +151,63 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatus do
   defp if_too_large(rows, collapse_fun) when length(rows) > @max_rows, do: collapse_fun.(rows)
   defp if_too_large(rows, _collapse_fun), do: rows
 
-  defp collapse_identical_route_rows([
-         %{route_info: %{route_id: route_id1, branch_ids: branch_ids1}} = row1,
-         %{route_info: %{route_id: route_id2, branch_ids: branch_ids2}} = _row2
-         | rest_of_rows
-       ])
-       when route_id1 == route_id2 and branch_ids1 == branch_ids2 do
-    combined_row =
-      row1 |> Map.put(:status_entry, see_alerts_status())
-
-    [combined_row | rest_of_rows] |> collapse_identical_route_rows()
+  defp collapse_identical_route_rows(rows) do
+    collapse_collapsible_rows(rows, &combine_identical_route_rows?/2)
   end
 
-  defp collapse_identical_route_rows([row1 | rest_of_rows]) do
-    [row1 | collapse_identical_route_rows(rest_of_rows)]
+  defp collapse_disrupted_green_line_rows(rows) do
+    collapse_collapsible_rows(rows, &combine_disrupted_green_line_rows?/2)
   end
 
-  defp collapse_identical_route_rows([]) do
-    []
+  defp collapse_all_green_line_rows(rows) do
+    collapse_collapsible_rows(rows, &combine_all_green_line_rows?/2)
   end
 
-  defp collapse_disrupted_green_line_rows([
-         %{route_info: %{route_id: "Green"}} = row1,
-         %{route_info: %{route_id: "Green"}, status_entry: %{status: :normal}} = row2
-         | rest_of_rows
-       ]) do
-    [row1 | collapse_disrupted_green_line_rows([row2 | rest_of_rows])]
+  defp collapse_collapsible_rows([row1, row2 | rest_of_rows], combine_fun) do
+    if combine_fun.(row1, row2) do
+      collapse_collapsible_rows([combine_rows(row1, row2) | rest_of_rows], combine_fun)
+    else
+      [row1 | collapse_collapsible_rows([row2 | rest_of_rows], combine_fun)]
+    end
   end
 
-  defp collapse_disrupted_green_line_rows([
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/components/system_status/subway_status.ex
```diff
@@ -141,18 +141,45 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatus do
     |> Enum.flat_map(&rows_for_route(&1, include_alert: true))
   end
 
-  defp maybe_collapse_rows(rows) when length(rows) > @max_rows, do: collapse_rows(rows)
-  defp maybe_collapse_rows(rows), do: rows
+  defp collapse_rows_if_needed(rows) do
+    rows
+    |> if_too_large(&collapse_identical_route_rows/1)
+    |> if_too_large(&collapse_disrupted_green_line_rows/1)
+    |> if_too_large(&collapse_all_green_line_rows/1)
```
joshlarson, line 148: Note: This implicitly preserves Mattapan rows because none of ☝️ functions will ever collapse a Mattapan row into a Red row.

lib/dotcom_web/components/system_status/subway_status.ex
```diff
@@ -141,18 +141,45 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatus do
     |> Enum.flat_map(&rows_for_route(&1, include_alert: true))
   end
 
-  defp maybe_collapse_rows(rows) when length(rows) > @max_rows, do: collapse_rows(rows)
-  defp maybe_collapse_rows(rows), do: rows
+  defp collapse_rows_if_needed(rows) do
```
thecristen, line 144: suggestion(non-blocking): I like this! But you probably don't need this wrapping function and can put the three lines in there directly :D

test/dotcom_web/components/system_status/subway_status_test.exs
```diff
@@ -177,6 +177,148 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatusTest do
...
+    end
+
+    test "collapses Green line rows if there is a Mattapan alert" do
+      # Setup
+      alerts = [
+        Factories.Alerts.Alert.build(:alert_for_route,
+          route_id: Faker.Util.pick(GreenLine.branch_ids()),
+          effect: Faker.Util.pick(service_impacting_effects())
+        )
+        |> Factories.Alerts.Alert.active_during(Timex.now()),
```
thecristen, line 236: question: didn't you recently make an `active_now()` helper which doesn't use `Timex.now`? :D use it!

thecristen (changes requested):
The before and after is soooo good!

I almost feel like there might be a meta-programming-ish thing one could implement around the three collapse functions, since they all work so similarly. But I didn't leave a blocking comment on it, since _this_ implementation is probably clearer anyway, despite the repetition.

Requesting changes for adjusting the "active now" part of your test code.

test/dotcom_web/components/system_status/subway_status_test.exs
joshlarson: Yes I did! 

You caught me - I originally branched off an old version of [this PR](https://github.com/mbta/dotcom/pull/2404), before [this commit](https://github.com/mbta/dotcom/pull/2404/commits/16b3220cea6515df89f4b066b7718fc7fd8fa636) made it on, and forgot to apply that change here as well 😅

lib/dotcom_web/components/system_status/subway_status.ex
joshlarson: I'm on the fence, but I think I prefer it as-is. As it currently is, it keeps the "should we think about collapsing" separate from the "what does collapsing mean", which makes `status_to_rows/1` read more cleanly (to me, at least 😇)

test/dotcom_web/components/system_status/subway_status_test.exs
joshlarson: [Here we go](https://github.com/mbta/dotcom/pull/2415/commits/6e64f17cc7bd13fb7849a6183711dfa32a75c11b)

lib/dotcom_web/components/system_status/subway_status.ex
```diff
@@ -131,7 +131,7 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatus do
     @route_ids
     |> Enum.map(&{&1, subway_status |> Map.get(&1)})
     |> Enum.flat_map(&rows_for_route/1)
-    |> maybe_collapse_rows()
+    |> collapse_rows_if_needed()
```
joshlarson, line 134: I guess I could name this back to `maybe_collapse_rows()`. I can't remember now why I changed it.

thecristen (approved):
🚀

PR #2416, opened 2025-03-05
deps(mix): March 2025 dependency upgrades

thecristen:
#### Summary of changes

**Asana Ticket:** [[Monthly] Upgrade Erlang, Elixir & dependencies](https://app.asana.com/0/555089885850811/1209312810091834)

The second commit's a bit more involved, as it involves a major upgrade. The unit tests still pass, but I want to put it on dev-blue and test submitting the customer support form, and verify we get no surprises there.


mix.exs
```diff
@@ -96,14 +97,15 @@ defmodule DotCom.Mixfile do
...
+      {:hackney, "1.23.0"},
+      {:hammer, "7.0.1"},
       {:html_sanitize_ex, "1.4.3"},
-      {:httpoison, "2.2.1"},
+      {:httpoison, "2.2.2"},
       {:inflex, "2.1.0"},
       {:jason, "1.4.4", override: true},
       {:kino_live_component, "0.0.5"},
       {:logster, "1.1.1"},
+      # reverted from 0.4
```
thecristen, line 108: I'd add a comment as to why but I don't remember and I don't want to revisit the trouble I had with that upgrade when I tried it a number of months ago (and forgot to document it then)

mix.exs
```diff
@@ -73,7 +73,8 @@ defmodule DotCom.Mixfile do
       {:address_us, "0.4.3"},
       {:aws, "1.0.4"},
       {:aws_credentials, "0.3.2", optional: true},
-      {:castore, "1.0.11"},
+      {:castore, "1.0.12"},
+      {:cowlib, "2.14.0", override: true},
```
thecristen, line 77: technically not sure this override is essential, but I had a heck of a time trying to get mix to obey me with reconciling the cowboy/cowlib requirements.

thecristen:
I submitted a customer support ticket in dev-blue and saw a confirmation of success logged in [Splunk](https://mbta.splunkcloud.com/en-US/app/search/show_source?sid=1741195599.8118&offset=0&latest_time=).

mix.exs
anthonyshull: It's usually because some other dependency depends on this one at a different version number.

anthonyshull (approved):

PR #2418, opened 2025-03-06
Map stop ids to Google place ids

anthonyshull:
https://app.asana.com/0/555089885850811/1206002702609152/f

I was aiming for 90% confidence and a 10% margin of error. For a population of 7180 stops, that's a sample size of 68. Out of the 68, 5 were incorrect which is 7%. We are 90% confident that the number of incorrect place ids is between 0% and 17%. Spot checking those 5 incorrect place ids, 1 corresponded to there being no stop in Google. The other 4 look like issues where our latitude and longitude are actually closer to the wrong stop. Stop 157 was incorrectly identified as the place id for stop 46. They are across the street from each other, but the latitude and longitude for both are farther up the street. So, it's a downside to this approach that it is reliant on us having accurate data. But, it's also a positive because we know how to fix it. I'll look at those 39 cases where no place id could be identified and report back on those.

Of the 39 stops where we couldn't find a place id, I'm pretty sure we'll find everything falls in one of those three categories: the stop doesn't exist in Google, the stop should be removed from GTFS, or our lat/lon are very inaccurate.


lib/mix/tasks/write_stop_place_ids.ex
```diff
@@ -0,0 +1,159 @@
...
+    |> :json.encode()
+  end
+
+  # Get an individual place id and match it with its stop id.
+  defp get_place_id({id, latitude, longitude}) do
+    body = body(latitude, longitude)
+    headers = headers()
+
+    # We have to wait so we don't exceed our rate limit of 10 requests per second.
+    :timer.sleep(1000)
```
thecristen, line 62: **suggestion(non-blocking)**: we have `Hammer` and I see a nail... could [use it](https://hexdocs.pm/hammer/tutorial.html#example-basic-rate-limit-check) to rate limit this script & programmatically wait the required ms so as to more cleverly avoid exceeding our Google's rate limits.

lib/mix/tasks/write_stop_place_ids.ex
```diff
@@ -0,0 +1,159 @@
...
+
+    {:ok, {{_, 200, _}, _, response}} =
+      :httpc.request(:post, {@google_url, headers, ~c"application/json", body}, [], [])
+
+    places =
+      response
+      |> Kernel.to_string()
+      |> :json.decode()
```
thecristen, line 70: suggestion(non-blocking): using `Req.post/2` would save you the body encoding and response decoding calls...

lib/mix/tasks/write_stop_place_ids.ex
```diff
@@ -0,0 +1,159 @@
...
+  # Get an individual place id and match it with its stop id.
+  defp get_place_id({id, latitude, longitude}) do
+    body = body(latitude, longitude)
+    headers = headers()
+
+    # We have to wait so we don't exceed our rate limit of 10 requests per second.
+    :timer.sleep(1000)
+
+    {:ok, {{_, 200, _}, _, response}} =
+      :httpc.request(:post, {@google_url, headers, ~c"application/json", body}, [], [])
```
thecristen, line 65: should add a fallback for the non-success condition.

lib/mix/tasks/write_stop_place_ids.ex
```diff
@@ -0,0 +1,159 @@
...
+
+    :ok = write_footer()
+  end
+
+  # Given a latitude and longitude, construct a request body for Google.
+  # We don't use the name of the location.
+  # Rather, we use the latitude and longitude, restrict the types of places returned,
+  # order them by distance, and take the first.
+  # This means that accuracy is based on the stop latitude and longitude accuracy
+  # which should be more accurate than trying to match names.
```
thecristen, line 37: question(non-blocking): Would adding the name here help (hence attempting matching on lat/long _and_ name), or would that make the success rate worse?

lib/mix/tasks/write_stop_place_ids.ex
```diff
@@ -0,0 +1,159 @@
...
+  defp write_footer() do
+    default = "\tdef stop_place_id(_), do: nil\n"
+    module_end = "end\n"
+
+    File.write(@file_path, default <> module_end, [:append])
+  end
+
+  # Write the module header.
+  # Includes the module name, module doc, function doc, and function spec.
+  defp write_header() do
```
thecristen, line 131: idea: could we add a timestamp to indicate date of last generation?

thecristen (changes requested):
✨ something about how this 
- doesn't have to be worked around or accommodated in any special way during testing
- doesn't have to involve configuration for a deployment environment

makes me quite happy.

The main change I'd like to request:
- Please write some directions for the next person who's going to run this and set them up for success:
  - [ ] Add a small section in the README covering when to run this task and why we'd ever even want to do that. If you'd prefer to put it in the mix task doc directly that's fine too.
  - [ ] Add an Asana task to our Dev Backlog Inbound section with a due date around year from now, where the description lists the mix task and requirements (e.g. whoever runs it needs their Google API key handy)

lib/mix/tasks/write_stop_place_ids.ex
```diff
@@ -0,0 +1,159 @@
...
+    [
+      {~c"accept", ~c"application/json"},
+      {~c"X-Goog-Api-Key", String.to_charlist(@google_api_key)},
+      {~c"X-Goog-FieldMask", ~c"places.id"}
+    ]
+  end
+
+  # Convert the CSV row to a simple tuple.
+  defp row_to_tuple(row) do
+    {row["stop_id"], row["stop_lat"], row["stop_lon"]}
```
joshlarson, line 100: **Suggestion**: Keep the keys `latitude` and `longitude` as named keys. Representing geo coordinates as a pair of numbers leads to confusion about which one is latitude versus longitude (for instance, in [trip planner](https://github.com/mbta/dotcom/blob/067337387ded24c83246fc553beb10b1b387b6f2/lib/dotcom_web/live/trip_planner.ex#L470), we're using `[long, lat]` rather than `[lat, long]`.

Non-blocking because the `lat, long` convention introduced here does stay fully inside of this file, so it's not going to leak out and conflict with anything anywhere else.

lib/mix/tasks/write_stop_place_ids.ex
```diff
@@ -0,0 +1,159 @@
...
+  Creates a `Dotcom.StopPlaceIds` module with a single function `stop_place_id/1`.
+  The function is a simple one to one mapping of stop ids to Google place ids.
+  """
+
+  @shortdoc "Create the `Dotcom.StopPlaceIds` module."
+
+  use Mix.Task
+
+  @file_path "lib/dotcom/stop_place_ids.ex"
+  @google_api_key System.get_env("GOOGLE_API_KEY", "")
```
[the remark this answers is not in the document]
joshlarson: Oh another suggestion on the same line - can we have this fail noisily if `GOOGLE_API_KEY` isn't defined? There's no way that the script will run without that variable, so might as well fail fast and obvious once it notices that the key isn't there.

lib/mix/tasks/write_stop_place_ids.ex
```diff
@@ -0,0 +1,159 @@
...
+    "transit_station"
+  ]
+
+  @impl Mix.Task
+  def run(_) do
+    :ok = write_header()
+
+    :ok = write_place_ids()
+
+    :ok = write_footer()
```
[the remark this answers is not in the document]
thecristen: I like this too!

lib/mix/tasks/write_stop_place_ids.ex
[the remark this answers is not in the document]
thecristen: Oh right, good catch -- module attributes in Elixir are evaluated at compile time, meaning they cannot directly use runtime environment variables.

test/dotcom/stop_place_ids_test.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Dotcom.StopPlaceIdsTest do
+  use ExUnit.Case
+
+  import Dotcom.StopPlaceIds
+
+  describe "stop_place_id/1" do
+    test "returns a string when the id exists" do
+      # Exercise/Verify
+      assert stop_place_id("1") |> is_binary()
```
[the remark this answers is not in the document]
anthonyshull: Not without reading the file and using a regex to find the defs with actual ids. The idea of randomness is to indicate that things don't matter. Here, we're saying it does matter. But, it also doesn't matter very much. So, writing a test that parses the file to get an id to prove that it's defined in the file, doesn't seem to be useful.

lib/mix/tasks/write_stop_place_ids.ex
```diff
@@ -0,0 +1,159 @@
...
+  end
+
+  # Given a latitude and longitude, construct a request body for Google.
+  # We don't use the name of the location.
+  # Rather, we use the latitude and longitude, restrict the types of places returned,
+  # order them by distance, and take the first.
+  # This means that accuracy is based on the stop latitude and longitude accuracy
+  # which should be more accurate than trying to match names.
+  defp body(latitude, longitude) do
```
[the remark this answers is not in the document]
anthonyshull: I always alphabetize private functions. Or, at least I try to.

lib/mix/tasks/write_stop_place_ids.ex
anthonyshull: I think it makes it worse for some places where a) the name is vague, and/or b) the name is very close to another name.

lib/mix/tasks/write_stop_place_ids.ex
anthonyshull: The script will just fail which is fine.

lib/mix/tasks/write_stop_place_ids.ex
anthonyshull: Yeah, but using httpc we don't have to use a nimble pool...which we don't really need for a script.

lib/mix/tasks/write_stop_place_ids.ex
anthonyshull: Hammer really isn't going to provide much there as we would still have to check the rate limit at some pre-defined period. Something more sophisticated would be some kind of pull-based mechanism where it tells the script it is ready to look up more. But, all of those are overkill for what we're trying to do here.

joshlarson (approved):
Very nice!

thecristen (approved):

PR #2419, opened 2025-03-06
feat(TimetableController): support new rating trip messages

thecristen:
#### Summary of changes

**Asana Ticket:** [SCR | Update "Via Fairmount line" message to match new train numbers](https://app.asana.com/0/555089885850811/1209144152994478)

Will show the "via Fairmount line" message for current and next rating's (after 3/24) trip numbers, visible on dev-green.

### DIFF
--- a/test/dotcom_web/controllers/schedule/timetable_controller_test.exs
+++ b/test/dotcom_web/controllers/schedule/timetable_controller_test.exs
@@ -224,11 +224,17 @@ defmodule DotcomWeb.ScheduleController.TimetableControllerTest do
     end
   end
 
-  describe "trip_messages/2" do
-    test "returns proper messages for a CR Franklin train running via Fairmount" do
+  describe "trip_messages/3" do
+    setup do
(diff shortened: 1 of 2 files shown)


test/dotcom_web/controllers/schedule/timetable_controller_test.exs
```diff
@@ -224,11 +224,17 @@ defmodule DotcomWeb.ScheduleController.TimetableControllerTest do
     end
   end
 
-  describe "trip_messages/2" do
-    test "returns proper messages for a CR Franklin train running via Fairmount" do
+  describe "trip_messages/3" do
+    setup do
+      # hardcoded date before the rating switchover that'll change the relevant trip numbers
```
joshlarson, line 229: Honestly, this is a good comment. I do have a hazy idea about a way to solve the problem raised in the comment, but this is high priority enough not to need to be blocked on it. I wouldn't have even been thinking about it without the comment though.

joshlarson (approved):
🚆

PR #2422, opened 2025-03-10
feat: add nonce to content security policy

thecristen:
#### Summary of changes

**Asana Ticket:** [CSP | Web Application Security - www.mbta.com:443](https://app.asana.com/0/555089885850811/1208252819442819)

The bulk of this PR is refactoring the existing stuff to use [`ContentSecurityPolicy`](https://hexdocs.pm/content_security_policy/getting-started.html) library instead, and also make use of [`Phoenix.Controller.put_secure_browser_headers/2`](https://hexdocs.pm/phoenix/Phoenix.Controller.html#put_secure_browser_headers/2) instead of rolling my own!

Then adding the [script nonce](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) is easy thanks to `ContentSecurityPolicy`. This in turn allows us to follow Google's [recommended configuration for using Google Tag Manager with a Content Security Policy](https://developers.google.com/tag-platform/security/guides/csp). Which in turn finally makes the [CSP evaluator](https://csp-evaluator.withgoogle.com/) stop yelling at our site -- feel free to point it to dev-blue and point it to dev or prod and see the difference.


config/dev.exs
```diff
@@ -1,6 +1,7 @@
 import Config
 
 config :dotcom, :cache, Dotcom.Cache.Multilevel
+config :dotcom, dev_server?: true
```
anthonyshull, line 4: Can't you use: `Mix.env() == :dev`?

integration/e2e_tests/security-headers.spec.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+test.describe("Security headers are present", () => {
+  [
+    ["get", "/"],
+    ["post", "/"],
+    ["get", "/favicon.ico"],
+    ["post", "/schedules/742/line"],
+    ["get", "/icon-svg/icon-map-station-marker.svg"],
+    ["post", "/schedules/36/line"],
+    ["get", "/error/unexpected/page"],
```
anthonyshull, line 36: Could clean this up a little by making this a file const outside of the test.

integration/e2e_tests/security-headers.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        "content-security-policy",
+        "referrer-policy",
+        "strict-transport-security",
+        "x-content-type-options",
+        "x-xss-protection"
+      ] : [
+        "cache-control",
+        "strict-transport-security",
+        "x-content-type-options",
+        "x-xss-protection"
```
anthonyshull, line 56: Since there is overlap...and a lot of lines in a ternary...could turn this into two file consts. One has the shorter set and the other concats the set difference.
```
const STATIC_HEADERS = [...]
const HTML_HEADERS = STATIC_HEADERS ++ [...]
```

lib/dotcom_web/endpoint.ex
```diff
@@ -72,5 +72,116 @@ defmodule DotcomWeb.Endpoint do
...
+      [host: host, port: port] -> "#{host}:#{port}"
+      _ -> nil
+    end
+
+  host = Keyword.get(endpoint_config, :url)[:host]
+
+  webpack_path =
+    Application.compile_env(:dotcom, :webpack_path, "") |> String.replace("http://", "")
+
+  default_policy = %ContentSecurityPolicy.Policy{
```
anthonyshull, line 90: Might be a good idea to move this to its own module since it is pretty lengthy.

thecristen:
I'll redeploy this branch somewhere once a staging server frees up... probably tomorrow earliest 🤷🏼‍♀️

anthonyshull (approved):
I would think about moving the config out of config since it doesn't change relative to environment. Could just be in a module that gets included at compile time with:

```
@secure_headers secure_headers()
```

PR #2430, opened 2025-03-17
refactor: remove React component from timetable

thecristen:
#### Summary of changes

**Asana Ticket:** [Uncaught exit - {:timeout, {Task, :await, [%Task{mfa: {:erlang, :apply, 2}, owner: #PID<0.26396768.0>, pid: #PID<0.26396822.0>, ref: #Reference<0.0.425996307.1154070796.3517251586.135124>}, 5000]}}](https://app.asana.com/0/555089885850811/1209703855878428/f) (links to [this Sentry issue](https://mbtace.sentry.io/issues/5767367004/))

Since the crowding icon was initially made in React, it was later added to the timetable via `Dotcom.React.render/2`... which was cool and all but turns out performs terribly, and also contributes to many of the 500 errors we were seeing post-deploy. This PR removes that.

Also:

- Adds a simple functional component for rendering tooltips
- Refactors the React component to remove the commuter rail-specific logic that's no longer needed

### DIFF
--- a/lib/dotcom_web/components/components.ex
+++ b/lib/dotcom_web/components/components.ex
@@ -229,4 +229,38 @@ defmodule DotcomWeb.Components do
     </details>
     """
   end
+
+  slot(:inner_block, required: true)
+  attr(:class, :string, default: "")
+
+  attr(:placement, :atom,
+    default: :left,
+    values: [:left, :right, :top, :bottom],
+    doc: "Where the tooltip appears relative to the hovered content."
+  )
+
+  attr(:title, :string,
+    doc:
+      "A preferably short bit of content which will appear on hover. Supports rendering of HTML."
+  )
+
+  @doc """
+  Basic tooltip to display content on hover.
+
+  This uses Bootstrap's JS plugin under the hood.
+  """
+  def tooltip(assigns) do
+    ~H"""
+    <span
+      class={@class}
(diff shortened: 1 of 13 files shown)


lib/dotcom_web/components/components.ex
```diff
@@ -229,4 +229,38 @@ defmodule DotcomWeb.Components do
...
+
+  attr(:title, :string,
+    doc:
+      "A preferably short bit of content which will appear on hover. Supports rendering of HTML."
+  )
+
+  @doc """
+  Basic tooltip to display content on hover.
+
+  This uses Bootstrap's JS plugin under the hood.
```
anthonyshull, line 250: Is there an alternative to Bootstrap that we can easily plugin? Just thinking that we want to remove Bootstrap so this is adding another place we would eventually have to remove it.

thecristen: > easily

No. Tooltips are one of the trickier things to get right, accessibility-wise. 

Part of my thinking is that, at the very least, the more places that use this reusable component, the fewer spots across our templates we'll have to touch to migrate from Bootstrap later. Keep in mind this isn't adding new Bootstrap tooltips where there isn't, just swapping out the template.

anthonyshull (approved):

PR #2432, opened 2025-03-17
feat: Add PathParamsRedirector plug that preserves path params

joshlarson:
Add a `keep_path_params` option to the `Redirector` plug so that paths like `/schedules/SL1/alerts` redirect to `/schedules/741/alerts` instead of either `/schedules/741` or just 404'ing.

### SL5 - Alerts

Local: http://localhost:4001/schedules/SL5/alerts
Prod: https://www.mbta.com/schedules/SL5/alerts

**Before:** 404
**After:** Redirects to `/schedules/749/alerts`

### SL5 - Root

Local: http://localhost:4001/schedules/SL5
Prod: https://www.mbta.com/schedules/SL5

**Before:** Redirects to `/schedules/749/line`
**After:** Same (we don't break existing functionality)

### CT1

Local: http://localhost:4001/schedules/CT1/alerts
Prod: https://www.mbta.com/schedules/CT1/alerts

**Before:** Redirects to `/projects/better-bus-project/CT1`
**After:** Same (we don't break existing functionality

---

No Ticket, but is a prerequisite for:
- https://github.com/mbta/dotcom/pull/2433


test/dotcom/lib/redirector_test.exs
```diff
@@ -18,23 +18,50 @@ defmodule DotcomWeb.RedirectorTest do
   end
 
   test "route redirected to internal route", %{conn: conn} do
-    conn = Redirector.call(conn, to: "/miami")
+    conn = call(conn, to: "/miami")
```
anthonyshull, line 21: Is there something special about Miami? If not, this should be a random string. Same with all others below.

joshlarson: Agreed - let's do that as a follow-up.

anthonyshull: Why a follow-up? This has a week before it needs to be launched, right?

anthonyshull:
Have you thought about having two different Redirectors instead of one where you pass in an argument?

test/dotcom/lib/redirector_test.exs
anthonyshull: https://www.notion.so/mbta-downtown-crossing/The-Four-Legs-of-Quality-Programming-b35248799da84e02b93887361d5bf343

lib/dotcom_web/plugs/path_params_redirector.ex
```diff
@@ -0,0 +1,39 @@
+defmodule DotcomWeb.Plugs.PathParamsRedirector do
+  import Plug.Conn, only: [put_status: 2, halt: 1]
+  import Phoenix.Controller, only: [redirect: 2]
+
+  @behaviour Plug
+
+  @impl true
```
anthonyshull, line 7: Need to move this down. Also, I like to specify what it is implementing.

```
@impl Plug
```

Is especially useful if you ever implement two behaviors.

lib/dotcom_web/plugs/path_params_redirector.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  @spec init(Keyword.t()) :: Keyword.t()
+  def init(opts) do
+    opts
+    |> validate_to_keyword()
```
anthonyshull, line 12: Could just move this to the init.

```
def init([to: _ = opts), do: opts

def init(_), do: raise(...)
```

anthonyshull (approved):
A few small suggestions.

lib/dotcom_web/plugs/path_params_redirector.ex
```diff
@@ -0,0 +1,39 @@
...
+      |> add_path_params(conn)
+      |> append_query_string(conn)
+
+    conn
+    |> put_status(:moved_permanently)
+    |> redirect(to: new_path)
+    |> halt()
+  end
+
+  defp add_path_params(to, %Plug.Conn{path_params: %{"path_params" => path_params}}),
```
anthonyshull, line 29: Might want to just make this multi-line:

```
def .... do
  ...
end
```

lib/dotcom_web/plugs/path_params_redirector.ex
joshlarson: [Fair](https://github.com/mbta/dotcom/pull/2432/commits/9b6d5ab6571ffd0a0e24a06f1cb7e9d41d01c397)

lib/dotcom_web/plugs/path_params_redirector.ex
joshlarson: Yeah - I like it when the simple single-line implementations can just get stacked together, but that already wasn't happening here :(

[PTALA](https://github.com/mbta/dotcom/pull/2432/commits/ef6710843655372b184501766f002ad20c38022a)

lib/dotcom_web/plugs/path_params_redirector.ex
```diff
@@ -0,0 +1,34 @@
...
+  import Phoenix.Controller, only: [redirect: 2]
+
+  @behaviour Plug
+
+  @impl Plug
+  @spec init(Keyword.t()) :: Keyword.t()
+  def init([to: _] = opts), do: opts
+  def init(_opts), do: raise("Missing required to: option in redirect")
+
+  @impl true
```
anthonyshull, line 12: This is a Plug implemented function too.

joshlarson: [Yes it is - thanks for catching!](https://github.com/mbta/dotcom/pull/2432/commits/6fff69862b804b3dec8044389b566501372861d6)

Good thing I [also forgot to add a moduledoc](https://github.com/mbta/dotcom/pull/2432/commits/1457f18a4a4af995d0744608b31d8bf03095904b), so that this didn't get scooped by auto-merge 😅

PR #2437, opened 2025-03-19
feat(AppStoreController): Controller to direct to appropriate app store w/ campaign info

KaylaBrady:
#### Summary of changes

**Asana Ticket:** [Set up marketing URLs](https://app.asana.com/0/1205732265579288/1209584638067002/f)

This PR adds a new controller that redirects to the apple app store, google play store, or MBTA Go project page based on where a user is visiting from. This will be used in marketing campaigns that promote downloading the app. 

I added some unit tests and tested locally by manually setting each condition to `true`. If there is a free dev environment, I'll deploy there too for easier testing on actual devices!

---

#### General checks


mix.exs
```diff
@@ -73,6 +73,7 @@ defmodule DotCom.Mixfile do
       {:address_us, "0.4.3"},
       {:aws, "1.0.4"},
       {:aws_credentials, "0.3.2", optional: true},
+      {:browser, "~> 0.5.4"},
```
KaylaBrady, line 76: h/t to @thecristen for pointing out the existence of this library in [an earlier banner PR ](https://github.com/mbta/dotcom/pull/2175#pullrequestreview-2334036329) that I found while looking for prior art here!

test/dotcom_web/controllers/app_store_controller_test.exs
```diff
@@ -0,0 +1,32 @@
...
+
+        conn = get(conn, app_store_path(conn,:redirect_mbta_go, %{"pt" => "fake pt", "ct" => "fake ct", "mt" => "fake-mt"}))
+     assert redirected_to(conn, 302) =~ "/default_project_page"
+  end
+
+    test "redirects to app store for ios browser", %{conn: conn} do
+          reassign_env(:dotcom, :mbta_go_app, ios_store_base_url: "fake_app_store_url?")
+        conn =
+          conn
+          |> put_req_header("user-agent", "Mozilla/5.0 (iPhone; U; CPU iPhone OS 3_0 like Mac OS X; en-us) AppleWebKit/420.1 (KHTML, like Gecko) Version/3.0 Mobile/1A542a Safari/419.3")
```
KaylaBrady, line 17: Pulled from Browser's own test fixtures

lib/dotcom_web/controllers/app_store_controller.ex
```diff
@@ -0,0 +1,53 @@
...
+
+    cond do
+      Browser.ios?(conn) ->
+        redirect(conn, external: ios_campaign_url(config[:ios_store_base_url], params))
+
+      Browser.android?(conn) ->
+        redirect(conn, external: android_campaign_url(config[:android_store_base_url], params))
+
+      true ->
+        redirect(conn, to: config[:default_project_page])
```
thecristen, line 26: **question**: Why omit the URL params for the default project page case?

lib/dotcom_web/controllers/app_store_controller.ex
```diff
@@ -0,0 +1,53 @@
+defmodule DotcomWeb.AppStoreController do
+  @moduledoc """
+  Redirects users to the appropriate app store (ios or android) based on
+  their brower. When neither store is appropriate, redirects to a project page.
+  """
+  use DotcomWeb, :controller
+  import Browser
```
thecristen, line 7: **nit**: you're using `Browser` as an `alias` down there 😅

test/dotcom_web/controllers/app_store_controller_test.exs
```diff
@@ -0,0 +1,63 @@
...
+    reassign_env(:dotcom, :mbta_go_app, default_project_page: "/default_project_page")
+
+    conn =
+      get(
+        conn,
+        app_store_path(conn, :redirect_mbta_go, %{
+          "pt" => "fake pt",
+          "ct" => "fake ct",
+          "mt" => "fake-mt"
+        })
```
thecristen, line 15: **suggestion**: This relates to my question in the code about whether it's necessary to strip URL parameters for the default case. If it _is_ needed to remove them, I think it'd be less confusing here to use a different arbitrary URL parameter here (instead of the ones that happen to be preserved for the iOS case).

lib/dotcom_web/controllers/app_store_controller.ex
```diff
@@ -0,0 +1,53 @@
...
+    android_query_encoded =
+      params
+      |> Map.take(["referrer", "utm_source", "utm_campaign"])
+      |> URI.encode_query()
+
+    base_url
+    |> URI.parse()
+    |> URI.append_query(android_query_encoded)
+    |> URI.to_string()
+  end
```
thecristen, line 52: **question**: Is it necessary to be so careful with the URL params, as opposed to encoding any params that happen to come with the original request?

Just thinking this could all be simpler if you wanted to say `redirect(conn, to: config[:desired_url] <> "?#{URI.encode_query(params))}"`

test/dotcom_web/controllers/app_store_controller_test.exs
```diff
@@ -0,0 +1,63 @@
...
+      )
+
+    assert redirected_to(conn, 302) =~ "/default_project_page"
+  end
+
+  test "redirects to app store for ios browser", %{conn: conn} do
+    reassign_env(:dotcom, :mbta_go_app, ios_store_base_url: "fake_app_store_url?")
```
thecristen, line 22: suggestion: we tend to use `Faker` to generate fake values and avoid having the "magic value" issue. `Faker.Internet.url()` can give you a random URL and random values can be used for the param values too.

config/config.exs
```diff
@@ -14,6 +14,11 @@ config :dotcom, :location_service, LocationService
 
 config :dotcom, :mbta_api_module, MBTA.Api
 
+config :dotcom, :mbta_go_app,
+  android_store_base_url: "https://play.google.com/store/apps/details?id=com.mbta.tid.mbta_app",
+  ios_store_base_url: "https://apps.apple.com/app/apple-store/id6472726821?",
+  default_project_page: "/goapp"
+
```
thecristen, line 21: suggestion: I think you can leave this out the config and put these in `DotcomWeb.AppStoreController` directly, since they aren't needed anywhere else. I'm not convinced it's worth the trouble mocking these values in tests either -- the iOS test could look for `"apps.apple.com"` in the link, for example.

test/dotcom_web/controllers/app_store_controller_test.exs
thecristen: Nice!

lib/dotcom_web/controllers/app_store_controller.ex
KaylaBrady: That's fair! I wanted to make sure there was no way to accidentally / maliciously pass a different app id in the URL param & redirect to an app other than Go for the Google play store, but confirmed by testing the below link that only the first id is kept, which we have hardcoded.

https://play.google.com/store/apps/details?id=com.mbta.tid.mbta_app&id=com.mbta.mobileapp

test/dotcom_web/controllers/app_store_controller_test.exs
KaylaBrady: Oh cool! Opted to move the values out of config anyway, but good tip!

PR #2439, opened 2025-03-21
fix(trip-planner): Nudge MA State House locations to a public entrance

joshlarson:
## Summary

This is a very hacky stop-gap solution to a kind of thorny problem. The problem is that updating OSM data isn't an option, because then OSM will tell people that the inside of the State House is publicly accessible, which it's not. And we don't have fine-grained control over what lat/long corresponds to specific search terms.

The solution here is that any location in the `To` or `From` fields in trip planner whose name/address indicates that its address is `24 Beacon St, Boston, MA` with a zip code of either `02133` or `02108` will be nudged to the security booth where visiting members of the public can actually go.

## Example

[Local testing link](http://localhost:4001/trip-planner?plan=hsQVX3VudXNlZF9kYXRldGltZV90eXBlxADEEl91bnVzZWRfd2hlZWxjaGFpcsQAxAhkYXRldGltZcQgMjAyNS0wMy0yMVQxNDozNTo1Ny4zNjQ3NjgtMDQ6MDDEBGZyb22ExAhsYXRpdHVkZctARSy0jTrmhsQJbG9uZ2l0dWRly8BRxQQtjCpFxARuYW1lxC5Cb3N0b24gUHVibGljIExpYnJhcnksIENpdHkgb2YgQm9zdG9uLCBNQSwgVVNBxAdzdG9wX2lkxADEBW1vZGVzicQDQlVTxAR0cnVlxAVGRVJSWcQEdHJ1ZcQEUkFJTMQEdHJ1ZcQGU1VCV0FZxAR0cnVlxA5fcGVyc2lzdGVudF9pZMQBMMQLX3VudXNlZF9CVVPEAMQNX3VudXNlZF9GRVJSWcQAxAxfdW51c2VkX1JBSUzEAMQOX3VudXNlZF9TVUJXQVnEAMQCdG-ExAhsYXRpdHVkZctARS3sgMc6vcQJbG9uZ2l0dWRly8BRxBV-7UXpxARuYW1lxD9NYXNzYWNodXNldHRzIFN0YXRlIEhvdXNlLCAyNCBCZWFjb24gU3QsIEJvc3RvbiwgTUEsIDAyMTA4LCBVU0HEB3N0b3BfaWTEAA==)

### Before
Screenshot 2025-03-21 at 2 35 16 PM

### After
Screenshot 2025-03-21 at 2 34 02 PM

#### Summary of changes

**Asana Ticket:** [TP Bugs | 🐞 Trip Planner (New and Old) | Can't plan trips to Massachusetts State House](https://app.asana.com/0/555089885850811/1208887683514495/f)

### DIFF
--- a/lib/dotcom/trip_plan/location_nudger.ex
+++ b/lib/dotcom/trip_plan/location_nudger.ex
@@ -1,0 +1,49 @@
+defmodule Dotcom.TripPlan.LocationNudger do
+  @moduledoc """
+  There are some locations that are fairly close to transit stops, but
+  for which all of the OpenStreetMap paths are not public. The
+  Massachusetts State House is a good example of this. The building
+  isn't publicly accessible according to OpenStreetMaps data, and
+  that's correct, because all of the entries into it are gated with
+  security guards. That said, plenty of people need to plan trips to
+  the State House.
+
+  This location nudger is responsible for nudging locations that
+  aren't publicly accessible, but that should still be possible to
+  plan trips to and from, to a sensible lat/long that can be used in
+  trip plans, e.g. the public side of a security entrance.
+  """
+
+  @state_house_address_string "24 Beacon St"
+  @state_house_zip_codes ["02108", "02133"]
+  @state_house_location %{latitude: 42.35861, longitude: -71.06297}
+
+  def nudge(%{from: from, to: to} = data) do
+    %{data | from: nudge_location(from), to: nudge_location(to)}
+  end
+
+  def state_house_location, do: @state_house_location
+  def state_house_zip_codes(), do: @state_house_zip_codes
+
+  def nudge_location(location) do
+    if state_house?(location) do
+      location |> Map.merge(@state_house_location)
+    else
+      location
+    end
+  end
+
+  defp state_house?(%{name: name}) do
+    state_house_zip_code_and_city?(name) &&
+      state_house_street_address?(name)
+  end
+
+  defp state_house_zip_code_and_city?(name) do
+    @state_house_zip_codes
+    |> Enum.any?(fn zip_code -> name |> String.contains?("Boston, MA, #{zip_code}") end)
+  end
+
+  defp state_house_street_address?(name) do
+    name |> String.contains?(@state_house_address_string)
+  end
+end
--- a/test/dotcom/trip_plan/location_nudger_test.exs
+++ b/test/dotcom/trip_plan/location_nudger_test.exs
@@ -1,0 +1,178 @@
+defmodule Dotcom.TripPlan.LocationNudgerTest do
+  alias Dotcom.TripPlan.LocationNudger
+  alias Test.Support.Generators
+  use ExUnit.Case
+
+  import Dotcom.TripPlan.LocationNudger, only: [nudge: 1, nudge_location: 1]
+
+  @state_house_address_strings ["24 Beacon St", "24 Beacon Street"]
+  @state_house_location LocationNudger.state_house_location()
+  @state_house_zip_codes LocationNudger.state_house_zip_codes()
+
+  describe "nudge_location/1" do
+    test "does not alter a typical location" do
+      # Setup
+      location = %{
+        name: Generators.Address.address(),
+        latitude: Faker.Address.latitude(),
+        longitude: Faker.Address.longitude()
+      }
+
+      # Exercise / Verify
+      assert nudge_location(location) == location
+    end
+
+    test "nudges addresses associated with the MA State House to the MA State House lat and long" do
+      street_address = Faker.Util.pick(@state_house_address_strings)
+      zip_code = Faker.Util.pick(@state_house_zip_codes)
+
+      # Setup
+      location = %{
+        name: "#{Faker.Company.name()}, #{street_address}, Boston, MA, #{zip_code}, USA",
+        latitude: Faker.Address.latitude(),
+        longitude: Faker.Address.longitude()
+      }
+
+      # Exercise
+      nudged_location = nudge_location(location)
+
+      # Verify
+      assert @state_house_location = nudged_location
+    end
+
+    test "does not nudge addresses to the State House if the zip code isn't one of the State House zip codes" do
+      street_address = Faker.Util.pick(@state_house_address_strings)
+
+      non_state_house_zip_code =
+        fake_with_blocklist(
+          fn -> Faker.Address.zip_code() end,
+          @state_house_zip_codes
+        )
+
+      # Setup
+      location = %{
+        name:
+          "#{Faker.Company.name()}, #{street_address}, Boston, MA, #{non_state_house_zip_code}, USA",
+        latitude: Faker.Address.latitude(),
+        longitude: Faker.Address.longitude()
+      }
+
+      # Exercise
+      nudged_location = nudge_location(location)
+
+      # Verify
+      assert nudged_location == location
+    end
+
+    test "does not nudge addresses to the State House if they're not in Boston" do
+      # Setup
+      street_address = Faker.Util.pick(@state_house_address_strings)
+      zip_code = Faker.Util.pick(@state_house_zip_codes)
+
+      non_boston_city =
+        fake_with_blocklist(
+          fn -> Faker.Address.city() end,
+          ["Boston"]
+        )
+
+      location = %{
+        name:
+          "#{Faker.Company.name()}, #{street_address}, #{non_boston_city}, MA, #{zip_code}, USA",
+        latitude: Faker.Address.latitude(),
+        longitude: Faker.Address.longitude()
+      }
+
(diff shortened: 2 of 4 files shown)


lib/dotcom/trip_plan/location_nudger.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Dotcom.TripPlan.LocationNudger do
```
joshlarson, line 1: Open to suggestions for better names than this.

test/dotcom/trip_plan/location_nudger_test.exs
```diff
@@ -0,0 +1,178 @@
...
+      # Exercise
+      nudged_form_data = nudge(form_data)
+
+      # Verify
+      assert @state_house_location = nudged_form_data.to
+      assert nudged_form_data.from == form_data.from
+    end
+  end
+
+  defp fake_with_blocklist(fake_fun, blocklist) do
```
joshlarson, line 169: This feels general-purpose enough that it ought to live somewhere in `Test.Support`. I could make a special-purpose module just for it, or put it with other like things (I couldn't think of any OTOH).

Also open to suggestions for better names for this.

test/support/factories/trip_planner/trip_planner.ex
```diff
@@ -3,6 +3,7 @@ defmodule Test.Support.Factories.TripPlanner.TripPlanner do
   Provides generated test data via ExMachina and Faker.
   """
 
+  alias Test.Support.Generators
```
joshlarson, line 6: Ooops. Will fix.

Although I thought we had set up Credo to catch this type of thing?

test/dotcom/trip_plan/location_nudger_test.exs
joshlarson: I moved it :)

lib/dotcom/trip_plan/location_nudger.ex
thecristen: `LocationOSMReconciler` ?

lib/dotcom/trip_plan/location_nudger.ex
```diff
@@ -0,0 +1,58 @@
...
+  @state_house_location %{latitude: 42.35861, longitude: -71.06297}
+
+  def nudge(%{from: from, to: to} = data) do
+    %{data | from: nudge_location(from), to: nudge_location(to)}
+  end
+
+  def state_house_location, do: @state_house_location
+  def state_house_zip_codes(), do: @state_house_zip_codes
+
+  def nudge_location(location) do
```
thecristen, line 28: can we please add  `@spec` to this and probably other functions here to help me understand what's going in and out of these!

lib/dotcom/trip_plan/location_nudger.ex
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  defp state_house_street_address?(name) do
+    name |> String.contains?(@state_house_address_string)
+  end
+
+  defp state_house_title_and_city?(name) do
+    name |> String.contains?("Massachusetts State House") &&
+      name |> String.contains?("City of Boston, MA, USA")
+  end
```
thecristen, line 57: This is very granular and I'm wondering if a simpler approach might work just as well:


```suggestion
  defp state_house?(%{name: name}) do
     in_state_house_zips? = Enum.any?(@state_house_zip_codes, fn zip_code -> 
       name =~ "Boston, MA, #{zip_code}" 
     end)
     
     name =~ "Massachusetts State House" || (in_state_house_zips? and name =~ @state_house_address_string)
  end
```

lib/dotcom/trip_plan/location_nudger.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  This location nudger is responsible for nudging locations that
+  aren't publicly accessible, but that should still be possible to
+  plan trips to and from, to a sensible lat/long that can be used in
+  trip plans, e.g. the public side of a security entrance.
+  """
+
+  @state_house_address_string "24 Beacon St"
+  @state_house_zip_codes ["02108", "02133"]
+  @state_house_location %{latitude: 42.35861, longitude: -71.06297}
```
thecristen, line 19: Curious if you've considered a "dumber" solution that just defines before and after pairs that can be iterated through. The bonus would be easy to add new values which aren't state house specific.

```
@reconciliations [
  {"24 Beacon St, Boston, MA, 02108, USA", %{latitude: 42.35861, longitude: -71.06297}},
  {"24 Beacon St, Boston, MA, 02133, USA", %{latitude: 42.35861, longitude: -71.06297}},
  {"Massachusetts State House, 24 Beacon St, Boston, MA, 02108, USA", %{latitude: 42.35861, longitude: -71.06297}},
  {"Massachusetts State House, City of Boston, MA, USA", %{latitude: 42.35861, longitude: -71.06297}},
  {"State House News Service, 24 Beacon St, Boston, MA, 02133, USA", %{latitude: 42.35861, longitude: -71.06297}}
]
```

lib/dotcom/trip_plan/location_nudger.ex
joshlarson: How about `OpenStreetMapReconciler`?

lib/dotcom/trip_plan/location_nudger.ex
joshlarson: I did consider that, and I'm not totally against it.

My reasoning was that I didn't want to have to list out each individual search query that would need to get nudged, and also if a new location crops up in search results like `State House Cafeteria, 24 Beacon St...`, then it wouldn't get nudged properly.

I'm honestly not sure what the best approach is. Matching on the zip may not be necessary. I did that a bit defensively to protect against a name like `Boston, MA Fake House, 24 Beacon St, Nowhere, MA, 02000`, which would get matched if we were just checking for `24 Beacon St` and `Boston, MA`. Admittedly that is a bit farfetched...

lib/dotcom/trip_plan/location_nudger.ex
thecristen: I like it!

lib/dotcom/trip_plan/location_nudger.ex
joshlarson: Had I not gone with your/our other idea of using geometry, this suggestion probably would have been cleaner 😅!

lib/dotcom/trip_plan/location_nudger.ex
joshlarson: Talked offline and decided to use geometry to figure out whether the coordinates are inside of the State House instead of basing it on the name.

test/dotcom/trip_plan/open_street_map_reconciler_test.exs
```diff
@@ -0,0 +1,82 @@
+defmodule Dotcom.TripPlan.OpenStreetMapReconcilerTest do
+  use ExUnit.Case
+
+  import Dotcom.TripPlan.OpenStreetMapReconciler, only: [reconcile: 1, reconcile_location: 1]
+
+  alias Dotcom.TripPlan.OpenStreetMapReconciler
+  alias Test.Support.Factories.TripPlanner.TripPlanner
+
+  @location_inside_state_house %{latitude: 42.35878, longitude: -71.063812}
```
joshlarson, line 9: Didn't use Faker for this because I couldn't figure out a way to do this without having some hard-coded values, and wound up in a kind of complicated mess when I tried. 

I'm open to ideas and especially open to pairing on a solution here, although I want to balance that against the fact that I've already spent a lot of time on this, and it would be good to get the rider-facing bugfix out there sooner rather than later.

lib/dotcom/trip_plan/location_nudger.ex
joshlarson: After much trials and tribulations, `@spec`'s have been added 😳

lib/dotcom/trip_plan/open_street_map_reconciler.ex
```diff
@@ -0,0 +1,97 @@
...
+  """
+
+  alias Dotcom.TripPlan.InputForm
+
+  @state_house_location %{latitude: 42.35861, longitude: -71.06297}
+
+  @state_house_geofence %Geo.Polygon{
+    coordinates: [
+      [
+        {-71.06428959524202, 42.35929686684236},
```
thecristen, line 24:
```
{-71.06429, 42.35930}
```
Please trim to fewer significant figures 😅 Matching your state house point seems good!

thecristen: Kinda wish you'd built a simpler polygon so this wasn't sooo long!

thecristen (approved):
Oh this ended up pretty elegant! Just some comments on geometry things

PR #2440, opened 2025-03-21
refactor(ScheduleController.Green): slightly simplified logic

thecristen:
Trying to fix `Plug.Conn.AlreadySentError: the response was already sent` ([Sentry issue](https://mbtace.sentry.io/issues/6067756726/?environment=prod&project=4507102206820352&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=2)) and while I can't guarantee this fixes that I _can_ guarantee that this PR also removes a lot of stuff that isn't needed!

Most of the code removed had been added in the following PR from May 2017 as a way of caching the list of stops on the Green Line:
- https://github.com/mbta/dotcom-legacy/pull/1158

The other main change is a slight refactor of how predictions are fetched for the Green Line. I was having some troubles with timeouts on the `/schedules/Green` page which seemingly disappeared with this change. Who knows, it might be more of a lateral change.


lib/dotcom_web/controllers/schedule/green.ex
```diff
@@ -75,25 +69,19 @@ defmodule DotcomWeb.ScheduleController.Green do
...
-
-        {flat_map_results(predictions_stream), vehicle_predictions}
-      else
-        {[], []}
-      end
+      conn
+      |> conn_with_branches
+      |> Task.async_stream(
+        fn branch_conn ->
+          call_plug(branch_conn, DotcomWeb.ScheduleController.Predictions)
```
thecristen, line 76: The `DotcomWeb.ScheduleController.Predictions.should_fetch_predictions?` logic is executed in this plug itself, so it's not needed to be run again in this module.

joshlarson (approved):
Everything looks like it's still working, and losing all that code is going to be pretty nice!

Left a few minor suggestions, but looks good either way.

PR #2443, opened 2025-03-24
refactor(RealtimeSchedule): optimize

thecristen:
This is the change for which I showed my before/after flame graphs describing how `Dotcom.RealtimeSchedule.stop_data/3` computes its data.

flame_foxboro_before
flame_foxboro_after

This function is used in Transit Near Me. For every stop surfaced, run this function every minute to get updated results!
There's probably a lot which can be overhauled about how this feature is implemented altogether. We're not ready to dive into rearchitecting this whole thing like it deserves. That aside, this PR...

- remove extra arguments & dependency injection (weren't really used anyway)
- implement sane setup of awaiting lists of async tasks
- stop using JsonHelpers to encode things (not needed anymore)
- remove bespoke DateTime sorter
- add stop_id to the schedule request (dramatically reduces number of schedules returned)
- more Streams


lib/dotcom/transit_near_me.ex
```diff
@@ -197,6 +194,10 @@ defmodule Dotcom.TransitNearMe do
 
   def format_prediction_time(_, _, :bus, _), do: ["arriving"]
 
+  def format_prediction_time(%DateTime{} = time, _now, _, _) do
```
thecristen, line 197: If we ever add support for ferry predictions, this'll have us covered (and format them the same as commuter rail ones, which is a pattern we follow elsewhere when showing schedules)!

anthonyshull (approved):

PR #2448, opened 2025-03-25
deps(npm): sync Playwright version to 1.48.0

thecristen:
This is thanks to Artillery's Playwright plugin versioning.
Note: There is an even newer version of Artillery available, but it requires upgrading our version of Node, which is a more involved effort (I tried... a Cypress dependency broke).

This got the monitor testing scenarios again when I ran it locally.


thecristen:
https://github.com/mbta/dotcom/actions/runs/14067171636/job/39392513765 Pushing it from this branch because what's the worst that'll happen? It's not like it'll break even more 😆

anthonyshull (approved):

PR #2452, opened 2025-03-26
hotfix(ContentSecurityPolicy): corrected websocket host for iOS15

thecristen:
Well, I'm hoping that works -- will deploy and test.
Also adds a missing directive for Fullstory, and fixes the Sentry URL (which was broken :()


joshlarson (approved):
☎️

PR #2453, opened 2025-03-26
fix(PageController): filter index alerts to active on datetime

thecristen:
#### Summary of changes

**Asana Ticket:** [Only currently-active accessibility alerts should show up on alerts tab](https://app.asana.com/0/555089885850811/1209539521192559/f)

This will affect other things consuming `@alerts` on the index.html page... which as far as I can tell is only the recently visited buttons (to determine whether to show the exclamation triangle).

### DIFF
--- a/lib/dotcom_web/controllers/page_controller.ex
+++ b/lib/dotcom_web/controllers/page_controller.ex
@@ -36,7 +36,15 @@ defmodule DotcomWeb.PageController do
     |> async_assign_default(:homepage_fares, fn -> fares end)
     |> async_assign_default(:promoted_items, fn -> promoted end)
     |> async_assign_default(:whats_happening_items, fn -> remainder end)
-    |> async_assign_default(:alerts, fn -> Alerts.Repo.all(conn.assigns.date_time) end)
+    |> async_assign_default(
+      :alerts,
+      fn ->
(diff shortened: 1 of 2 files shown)


lib/dotcom_web/controllers/page_controller.ex
```diff
@@ -36,7 +36,15 @@ defmodule DotcomWeb.PageController do
     |> async_assign_default(:homepage_fares, fn -> fares end)
     |> async_assign_default(:promoted_items, fn -> promoted end)
     |> async_assign_default(:whats_happening_items, fn -> remainder end)
-    |> async_assign_default(:alerts, fn -> Alerts.Repo.all(conn.assigns.date_time) end)
+    |> async_assign_default(
+      :alerts,
+      fn ->
+        conn.assigns.date_time
+        |> Alerts.Repo.all()
+        |> Enum.filter(&Alerts.Match.any_time_match?(&1, conn.assigns.date_time))
```
joshlarson, line 44: **issue (non-blocking):** This should probably have a test associated with it in order to help us make sure we don't regress this next time we update the page.

Non-blocking because this is a fairly annoying bug, and it looks like the existing tests for the homepage are in a bad state right now 😬

joshlarson (approved):

PR #2455, opened 2025-03-27
Tooltip offset

anthonyshull:
Moves the tooltip down 40px so that it sits on top of the marker. I'm not sure why this is necessary. My guess is that it has something to do with the map's CSS positioning within the page.

https://app.asana.com/0/555089885850811/1209641407424374/f

Screenshot Capture - 2025-03-27 - 09-35-54

### DIFF
--- a/assets/ts/leaflet/components/Map.tsx
+++ b/assets/ts/leaflet/components/Map.tsx
@@ -134,11 +134,7 @@ const Component = ({
             keyboard={false}
             onclick={marker.onClick}
           >
-            {marker.tooltip && (
-              <Popup offset={[0, 40]}>
-                {marker.tooltip}
-              </Popup>
-            )}
+            {marker.tooltip && <Popup offset={[0, 40]}>{marker.tooltip}</Popup>}
           </Marker>
         ))}
       </Map>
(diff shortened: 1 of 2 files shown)


assets/ts/leaflet/components/Map.tsx
```diff
@@ -134,7 +134,7 @@ const Component = ({
             keyboard={false}
             onclick={marker.onClick}
           >
-            {marker.tooltip && <Popup maxHeight={175}>{marker.tooltip}</Popup>}
+            {marker.tooltip && <Popup offset={[0, 40]}>{marker.tooltip}</Popup>}
```
joshlarson, line 137: **Question:** Just double-checking that we're confident we don't actually need `maxHeight`? I thought maybe it would have an effect for long stop names, but with 5 minutes of searching, the longest stop name I could find was `Washington St @ Melnea Cass Blvd - Silver Line` on the SL4/SL5, and that looked fine 🤷‍♂️

joshlarson (approved):

assets/ts/leaflet/components/Map.tsx
anthonyshull: I couldn't see any effect after removing it. My guess is that it was there for some other implementation of the map. But, afaik, this is the only implementation used since we replaced the map in trip planner.

PR #2456, opened 2025-03-27
Add a mobile banner

anthonyshull:
https://app.asana.com/0/555089885850811/1209466814065153/f


lib/dotcom_web/templates/layout/root.html.heex
```diff
@@ -76,6 +76,21 @@
...
+        id="mobile-app-banner"
+        style="background: #DCD3E8; display: none; margin: 0; padding: 10px; text-align: center"
+      >
+        <a
+          href="https://mbta.com/app-store?pt=117998862&ct=Dotcom%20Banners&mt=8&referrer=utm_source%3Ddotcom%26utm_campaign%3Dbanners"
+          rel="noreferrer"
+          style="color: black; display: block"
+          target="_blank"
+        >
+          Download <strong>MBTA Go</strong> to <span></span> &#x2192;
```
thecristen, line 90:
```suggestion
          Download <strong>MBTA Go</strong> to <span>track your trip</span> &#x2192;
```
Let's offer a fallback that'll be adjusted by the JS. Just in case the script doesn't connect.

thecristen (approved):

joshlarson (commented):
**issue:** In the Figma, the banner is below the breadcrumbs, and in this PR, the banner is above the breadcrumbs. It's possible that it doesn't matter that much (although anecdotally, I don't really notice the banner when it's further from the page content), but just wanted to flag it.

### Figma
Screenshot 2025-03-27 at 5 58 01 PM

### PR
Screenshot 2025-03-27 at 5 57 53 PM

thecristen (approved):
Thanks for your watchful design eye!

PR #2457, opened 2025-03-27
chore: delete old IE warning banner

thecristen:
It was implemented as a temporary thing a few years ago, when we'd decided to stop supporting IE11 and wanted to offer a banner message to help them phase out. 

ie

It's been enough years... and our latest stats (featured in #2451) show no more IE11 users!

image


lib/dotcom_web/templates/layout/root.html.heex
```diff
@@ -90,17 +90,6 @@
         {content_tag(:main, @inner_content, id: "main", tabindex: -1)}
       </div>
       {render(__MODULE__, "_footer.html", conn: @conn)}
-
-      <%= if Application.get_env(:dotcom, :dev_server?) do %>
-        <script defer src={"#{Application.get_env(:dotcom, :webpack_path)}/iewarning.js"}>
-        </script>
-      <% else %>
-        <script defer {static_attributes("/js/react.js")}>
```
joshlarson, line 98: **Question:** Just double-checking that the `js/react.js` invocation here is also safe to remove?

Clicking around, I wasn't able to find any visible effect of removing it (even on react-heavy pages like the Green Line line-map page), but just wanted to be sure.

joshlarson (approved):
This is truly an incredible amount of code for an IE warning banner that nobody sees 😂 

Love this! 👋

lib/dotcom_web/templates/layout/root.html.heex
thecristen: Yep! It's just that that thing was a React component, so needed to make sure React was available too (it's not loaded on every page). The other React-ified pages have their own `<script>` to load React

PR #2460, opened 2025-03-28
feat: Have the subway-status header link to /alerts/subway

joshlarson:
Screenshot 2025-03-28 at 11 18 48 AM

Change the `Subway Status` header to be a link that directs to `/alerts/subway`.

This change also includes a small change to the `subway_status` component tests, because, those tests were assuming that the only `<a>` tags in the component would be the status rows, but that's not true anymore.

#### Summary of changes

**Asana Ticket:** [QF | Homepage subway status goes to /alerts/subway](https://app.asana.com/0/555089885850811/1209601118457610/f)


thecristen (approved):
Cool!

PR #2463, opened 2025-03-31
fix: show mobile app banner in schedule finder for additional modes

thecristen:
- moves rendering the banner into the `<ScheduleModalContent />` component so it could potentially render in any mode (instead of just subway)
- instead of hard-coding the banner in the React component, copies it from the DOM (thus leveraging the logic used in `js/mobile-app-banner.js`). Thereby, if the banner isn't on the schedule page anyway, it doesn't get copied and rendered in the schedule finder, either.
- adds a new Playwright test! This tests against our expectations, and also runs the tests in triplicate - on desktop, iPhone, and Android - so we can verify presence on mobile and absence on desktop.

### DIFF
--- a/integration/e2e_tests/mobile-app-banner.spec.js
+++ b/integration/e2e_tests/mobile-app-banner.spec.js
@@ -1,0 +1,69 @@
+const { expect, test } = require("@playwright/test");
+
+const baseURL = process.env.HOST
+  ? `https://${process.env.HOST}`
+  : "http://localhost:4001";
+
+test.use({ baseURL, headless: true, screenshot: { mode: "only-on-failure" }, video: "off" });
+
+/**
+ * One-off test suite to verify the mobile app banner is configured to show where it's supposed to.
+ * 
+ * Defaults to running against localhost, but can run against any site using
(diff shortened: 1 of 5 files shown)


integration/e2e_tests/mobile-app-banner.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  test.describe(`${modeName} schedule`, () => {
+    tabs.forEach(tabPath => {
+      test(tabPath, async ({ page, isMobile }) => {
+        await page.goto(`/schedules/${routeId}${tabPath}`);
+        await checkBannerVisibility(page, isMobile);
+        if (tabPath == "/line") {
+          const scheduleFinderLinkName = modeName == "subway" ? "View upcoming departures" : "View schedule";
+          await page.getByRole("button", { name: scheduleFinderLinkName }).first().click();
+          const modal = await page.getByRole("dialog");
+          await checkBannerVisibility(modal, isMobile);
```
anthonyshull, line 33: ⚡ Could turn this into two async functions to clean this up.

anthonyshull (approved):

PR #2466, opened 2025-04-01
fix: Respect user agent font size set in browser

thecristen:
#### Summary of changes

**Asana Ticket:** [Respect user agent font size set in browser](https://app.asana.com/0/555089885850811/1209458974384398/f)

h/t to @firestack for proposing and elaborating the steps to reproduce.

This was mostly an easy change via setting `$font-size-root: 100%;` for Bootstrap's "reboot" CSS, but there were a smattering of `px` values throughout the codebase that needed improvement such that the website wouldn't look totally insane on base font size changes. My rationale for making these changes roughly go as follows:
- change all `px` font sizing to equivalent `rem`
- change `px` line heights to use equivalent `rem` for now (should probably be a plain number honestly)
- change `px` margins to equivalent `rem`
- change pixel-based `width`, `height`, and similar to `rem`
- try to change `px` paddings to `rem` too, some of these were weirdly specific though and many were kept as pixels
- try to change `px` positioning (`top`/`left`/etc) to `rem`
- left the following as `px`: border and outline widths, border radius, box shadow anything


firestack:
Just a drive by note, you may not want to change _every_ instance of `px` to `rem`. Definitely for `font-size`/typography related things should be `rem`/`em`, but padding/margin/borders/position may still be best served via `px`. But that is more intention/design/instance specific.

for a specific example see the demo under [Strategic unit deployment](https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/#strategic-unit-deployment-6).


The whole article is good, so I'll link that as well https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/

anthonyshull (approved):

PR #2467, opened 2025-04-01
Add service change to list of diversions if severity level greater than info

anthonyshull:
https://app.asana.com/0/555089885850811/1209715928503652/f

Screenshot Capture - 2025-04-01 - 13-52-16

Screenshot Capture - 2025-04-01 - 13-51-04

Screenshot Capture - 2025-04-01 - 13-51-32

### DIFF
--- a/test/dotcom/alerts/subway_test.exs
+++ b/test/dotcom/alerts/subway_test.exs
@@ -7,6 +7,7 @@ defmodule Dotcom.Alerts.SubwayTest do
   import Test.Support.Generators.DateTime,
     only: [random_date_time: 0, random_time_range_date_time: 1]
 
+  alias Alerts.Alert
   alias Test.Support.Factories
 
   setup :verify_on_exit!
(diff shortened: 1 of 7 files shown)


test/dotcom/alerts/subway_test.exs
```diff
@@ -158,10 +167,12 @@ defmodule Dotcom.Alerts.SubwayTest do
       effects =
         Enum.find(groups, fn {group, _effects} -> group == random_group end) |> Kernel.elem(1)
 
-      random_effect = Enum.random(effects)
+      {random_effect, random_severity} = Enum.random(effects)
+
+      alert = %Alert{effect: random_effect, severity: random_severity}
```
thecristen, line 172: nit: why not `Factories.Alerts.Alert.build(:alert, effect: random_effect, severity: random_severity)`?

thecristen (commented):
Hm, getting complex...but seemingly only as complex as it needs to be. 🤔

test/dotcom/alerts/subway_test.exs
anthonyshull: Just because we don't need any other attributes.

thecristen (approved):

PR #2472, opened 2025-04-03
Monthly Upgrade 2025-04

anthonyshull:
I decided to fix some warnings and errors that were driving us nuts.

I just removed the Endpoint init because it didn't seem to be doing anything that wasn't already done in config.

I fixed a test that was leading to errors in server_sent_event_stage.


test/util/async_assign_test.exs
```diff
@@ -1,7 +1,7 @@
 defmodule Util.AsyncAssignTest do
   use ExUnit.Case, async: true
-  use Plug.Test
 
+  import Plug.Test
```
anthonyshull, line 4: 📓 Deprecation warning that the api had changed to use `import`.

test/dotcom_web/endpoint_test.exs
```diff
@@ -1,27 +0,0 @@
-defmodule DotcomWeb.EndpointTest do
```
anthonyshull, line 1: 📓 No need for this test since it just covers the function I removed.

mix.exs
```diff
@@ -128,37 +128,37 @@ defmodule DotCom.Mixfile do
...
       },
       {:parallel_stream, "1.1.0"},
-      {:phoenix, "1.7.20"},
+      {:phoenix, "1.7.21"},
       {:phoenix_ecto, "4.6.3"},
       {:phoenix_html_helpers, "1.0.1"},
       {:phoenix_live_dashboard, "0.8.6"},
       {:phoenix_live_reload, "1.5.3", only: [:dev, :test]},
-      {:phoenix_live_view, "1.0.5", override: true},
+      {:phoenix_live_view, "1.0.9", override: true},
```
thecristen, line 136: Can you update the corresponding JS library too to keep the versions in sync? 😅

config/deps/endpoint.exs
```diff
@@ -1,9 +1,12 @@
 import Config
 
+default_key_base = "yK6hUINZWlq04EPu3SJjAHNDYgka8MZqgXZykF+AQ2PvWs4Ua4IELdFl198aMvw0"
+secret_key_base = System.get_env("SITE_SECRET_KEY_BASE", default_key_base)
```
thecristen, line 4: this should be handled in the `runtime.exs` config so it's not just set at compile time

thecristen (approved):

PR #2473, opened 2025-04-03
fix: Format long trip durations with hr/min instead of just min

joshlarson:
## Before

Screenshot 2025-04-03 at 6 22 06 PM

## After

Screenshot 2025-04-03 at 6 21 46 PM

---

**Asana Ticket:** [QF | TP | Reformat trip time for times > 1 hour to show hr/min](https://app.asana.com/0/385363666817452/1209391396072991/f)

### DIFF
--- a/lib/util/util.ex
+++ b/lib/util/util.ex
@@ -256,6 +256,29 @@ defmodule Util do
   end
 
   @doc """
+
+  Formats a number of minutes as a string. If the number of minutes is
+  greater than 60, then split it out into hours and minutes
+
+  ## Examples
+      iex> Util.format_minutes_duration(59)
+      "59 min"
+
+      iex> Util.format_minutes_duration(61)
+      "1 hr 1 min"
+  """
+  @spec format_minutes_duration(integer) :: String.t()
+  def format_minutes_duration(duration) do
+    case duration do
+      duration when duration >= 60 ->
+        "#{div(duration, 60)} hr #{rem(duration, 60)} min"
+
+      _ ->
+        "#{duration} min"
+    end
+  end
+
+  @doc """
   Converts an `{:error, _}` tuple to a default value.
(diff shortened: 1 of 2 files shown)


lib/util/util.ex
```diff
@@ -255,6 +255,29 @@ defmodule Util do
...
+      "59 min"
+
+      iex> Util.format_minutes_duration(61)
+      "1 hr 1 min"
+  """
+  @spec format_minutes_duration(integer) :: String.t()
+  def format_minutes_duration(duration) do
+    case duration do
+      duration when duration >= 60 ->
+        "#{div(duration, 60)} hr #{rem(duration, 60)} min"
```
thecristen, line 274: question: If `rem(duration, 60)` happens to equal 0, would we want to omit the `0 min`?

joshlarson: Woops! Thanks for catching that 😅

joshlarson: [Fixed here, I think](https://github.com/mbta/dotcom/pull/2473/commits/67697deab55272b81fb109c8f6cf2b8fafe4698a)

joshlarson: Actually, what DO we want here? I copy-pasta'ed the [original code from Ye Olde Trippe Planner](https://github.com/mbta/dotcom/blob/ebb97fc10975f2160c986e02c52ec32570048201/lib/dotcom_web/views/trip_plan_view.ex#L395-L404), so the previous version _did_ have the `0 min`. 

I asked in the standup thread.

thecristen (approved):

lib/util/util.ex
joshlarson: Confirmed that we want `1 hr`, not `1 hr 0 min`, so the PR as it is now should be ready!

PR #2474, opened 2025-04-03
Hotfix: forgot to add service change to diversions

anthonyshull:
I thought that diversions were drawing from the same place as service impacting alerts, but they are subtly different. This makes it so that diversions also look at service changes where the severity >= 3.

### DIFF
--- a/lib/dotcom/alerts.ex
+++ b/lib/dotcom/alerts.ex
@@ -50,16 +50,12 @@ defmodule Dotcom.Alerts do
 
   @doc """
   Does the alert have an effect/severity that is considered a diversion?
-  And, is the first active period *after* the created at time?
-  That is, was/is it planned rather than reactive?
+  And, was it planned?
   """
   @spec diversion_alert?(Alert.t()) :: boolean()
   def diversion_alert?(alert) do
     effects_match?(@diversion_effects, alert) &&
-      alert.active_period
-      |> List.first()
-      |> Kernel.elem(0)
-      |> Timex.after?(alert.created_at)
+      planned_alert?(alert)
   end
 
   @doc """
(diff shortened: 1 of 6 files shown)


lib/dotcom/alerts.ex
```diff
@@ -8,9 +8,22 @@ defmodule Dotcom.Alerts do
...
 
+  # A keyword list of effects and the severity level necessary to make an alert a 'diversion.'
+  @diversion_effects [
+    detour: 1,
+    service_change: 3,
+    shuttle: 1,
+    station_closure: 1,
+    stop_closure: 1,
+    suspension: 1
+  ]
```
thecristen, line 25: Not sure it's more helpful but we could probably abstract this to `  @diversion_effects @service_impact_effects ++ [detour: 1, stop_closure: 1]`
🫠 It almost adds a tiny bit of product clarity, because as far as I remember this was primarily for bus problems (detours and stop closures are more of a bus thing, hence the spec for describing subway/rail problems wouldn't need to consider them...

lib/dotcom/alerts.ex
```diff
@@ -86,6 +115,14 @@ defmodule Dotcom.Alerts do
     effect == alert.effect && alert.severity >= severity
   end
 
+  # We consider an alert planned if the first active period starts after the alert was created.
```
thecristen, line 118: I enjoy this description

thecristen (changes requested):
You're still using the logic around the diversion alert being created before/after active period, so you should leave those tests in too :)

lib/dotcom/alerts.ex
anthonyshull: Why I think it would be more confusing to try to derive them.

thecristen (approved):

PR #2478, opened 2025-04-03
cleanup: Remove unused partial `_recommended_routes`

joshlarson:
No ticket, but it looks like this partial isn't being used anymore, and the thing it used to do is now being done by [the recently-visited partial](https://github.com/mbta/dotcom/blob/f87b4bef79bf18e2484da3794117b5d29b60294c/lib/dotcom_web/templates/partial/_recently_visited.html.eex) instead.


thecristen (approved):
Great find!

PR #2479, opened 2025-04-04
fix: don't set default fontWeight on links

thecristen:
#### Summary of changes

**Asana Ticket:** [Style issues | Link bolding](https://app.asana.com/0/555089885850811/1209237709550402/f)

This reverts the Tailwind-related change.

I feel like this might unbold some links that we might want more weight on, but I figure we can find and fix those on a case-by-case basis.


joshlarson (approved):
✔️

PR #2483, opened 2025-04-04
feat(tailwind): new component and base heading styles

thecristen:
#### Summary of changes

**Asana Ticket:** [Style issues | Spacing issues (H3, bullet points)](https://app.asana.com/0/555089885850811/1209204070486332/f)

This PR technically goes far beyond the ticket requirements... :)

Our Sass mixins had all the magic which made these right, and I felt like it was time to embrace them fully! I've refactored them into Tailwind components and used them to help correct the issue noted in the ticket, as well as a few other style regressions around the codebase that I introduced in #2194 . The main point is that headings have better styling, including the aspects of styling which affect adjacent elements.

Before/after from the cited page:

image

Homepage

image

Fares

image

image

image


amaisano (approved):
Amazing - glad to have the mixin magic back! Thank you @thecristen

joshlarson (approved):
As far as I can tell, this is a net improvement, and thus I love it!

I do want to flag that one of the catalyzing issues mentioned in the ticket was how the `Guides` page renders on mobile. This is better than it was for that too, but the margins on the headers on that page still push the headers too close to the wrong articles, and too far from the right ones (e.g. in the screenshot below, the circles at the top is the real "Boston Visitor's Guide to The T", but the margins make it look like that header applies to the slightly uneven bookshelf instead.

Screenshot 2025-04-07 at 12 27 15 PM

Approving anyway because this is still a net improvement, and I think that fixing that issue might be more bespoke for that page (since most headers apply to the content below them, but these headers apply to the content above instead), which means that I'd rather have a separate PR focused only on that.

PR #2487, opened 2025-04-07
hotfix(Cache.KeyGenerator): don't hash single binaries

thecristen:
Minor note: this changes handling of a single argument which _isn't_ a binary - instead of hashing it directly it'll be wrapped in a list (e.g. `:erlang.phash2([arg])` instead of `:erlang.phash2(arg)`)

But I also took the suggestion to increase the range via the second argument of `:erlang.phash2/2`

### DIFF
--- a/lib/dotcom/cache/key_generator.ex
+++ b/lib/dotcom/cache/key_generator.ex
@@ -17,7 +17,7 @@ defmodule Dotcom.Cache.KeyGenerator do
   end
 
   def generate(mod, fun, args) do
-    "#{clean_mod(mod)}|#{fun}|#{:erlang.phash2(args)}"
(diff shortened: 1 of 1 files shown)


anthonyshull (approved):
We could also think about using a stronger hash in the future:

```
> :crypto.hash(:sha3_512, "10590") |> Base.encode16()
"EA5504962775343FC4A835AEB4C19FA2D99AF9ECE1107740EFCAFA79169780FF221F8AC7DD0D08450954CE0E94B823AE0B2B8C1BD1AB6D315A4D7EA0A3A994BF"
> :crypto.hash(:sha3_512, "2783") |> Base.encode16()
"11FFE740A2720960975F271FF01FE42B617884EFF1BF1059C596A45F9766952D15CE9AEEB497A1645F2FB00A18D7B0BC6CFAC4383ABB1261E259AE2F15B158CB"
```

PR #2490, opened 2025-04-07
hotfix: Only show `<OtherStopList />` for bus routes

joshlarson:
This should fix the fact that https://www.mbta.com/schedules/CR-Providence/line (and some other commuter rail line-diagram pages) aren't loading right now.

---

No ticket.


joshlarson:
Superceded by #2491

PR #2493, opened 2025-04-07
cleanup: Remove unused `RouteCard` component

joshlarson:
I think this RouteCard component is not being used. I wasn't able to find any usages of it, and the tests all passed even after I deleted it.

Not to be confused with [this other RouteCard component](https://github.com/mbta/dotcom/blob/dca8afeabbf70d5e8705fd53dad70b02b0bc7d70/assets/ts/tnm/components/RouteCard.tsx), which is being used!

---

No ticket.


thecristen (approved):
Am shocked. Thanks!

PR #2497, opened 2025-04-08
refactor feedback form category/service selection

thecristen:
#### Summary of changes

**Asana Ticket:** [Feedback form | Form refresh breaks subject selection](https://app.asana.com/0/555089885850811/1208914203038402/f)

After consultation with some design-savvy folk, I've decided to adjust the form this way. 

Instead of a dynamically-populated `<select>`, which was subject to data loss on refresh and probably wasn't particularly accessible anyway, we use a simpler static `<select>` with `<optgroup>` to separate the groups.

image

It's a bit long, but this was agreed upon to be the least bad quick option.

One side effect is that we can remove the "Category" radio buttons, since we can now infer the selected category from this big dropdown.

And this all needed some downstream changes, to ensure validation still works and that the right data is sent.

> [!CAUTION]
> The Cypress test only mostly works: the mechanism for retrieving the "sent" data is flawed when many tests are happening together, I think. But I think fixing this is out of scope 👀

### DIFF
--- a/lib/dotcom_web/controllers/customer_support_controller.ex
+++ b/lib/dotcom_web/controllers/customer_support_controller.ex
@@ -448,7 +448,21 @@ defmodule DotcomWeb.CustomerSupportController do
   end
 
   defp set_service_options(conn, _) do
-    assign(conn, :service_options, Feedback.Message.service_options())
+    service_options =
+      Feedback.Message.service_options()
+      |> Enum.map(fn {category_label, category_value, options} ->
+        {category_label, Enum.map(options, &option(&1, category_value))}
+      end)
+
+    assign(conn, :service_options, service_options)
+  end
+
+  defp option(value, category) do
+    Keyword.new()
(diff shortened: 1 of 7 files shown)


lib/dotcom_web/controllers/customer_support_controller.ex
```diff
@@ -448,7 +448,21 @@ defmodule DotcomWeb.CustomerSupportController do
...
+      end)
+
+    assign(conn, :service_options, service_options)
+  end
+
+  defp option(value, category) do
+    Keyword.new()
+    |> Keyword.put(:key, value)
+    |> Keyword.put(:value, value)
+    |> Keyword.put(:"aria-label", value)
```
thecristen, line 464: Adding aria-labels here per the advice [here](https://html5accessibility.com/stuff/2024/01/12/options-for-optgroup-labeling-of-options/) to improve screen reader handling,

anthonyshull (approved):

PR #2495, opened 2025-04-08
fix(trip-planner): Render itinerary details for rail replacement shuttles correctly

joshlarson:
### Before 

Screenshot 2025-04-08 at 4 07 13 PM

### After

Screenshot 2025-04-08 at 4 04 27 PM

---

**Asana Ticket:** [QF | TP | Fix shuttle icon size, line color, map color in details view](https://app.asana.com/0/1205718271156548/1209501772392193/f)

PR #2499, opened 2025-04-09
fix(trip-planner): Fix rail-replacement-shuttle colors on the map

joshlarson:
### Before

Screenshot 2025-04-09 at 11 20 13 AM

### After
Screenshot 2025-04-09 at 11 19 44 AM

Follow-up to:
- https://github.com/mbta/dotcom/pull/2495

I had forgotten to update the line color on the map. But this is slightly complicated from the fact that we get color for the map from GTFS (while the other places we use color come from [the Rider Design System](https://github.com/mbta/rider-design-system/blob/c9589af1a5c1b6734ffb786f5f8b9ba993662f9c/dist/variables.Light.css#L62-L72)), and our GTFS sets rail-replacement-shuttles as having the bus color.

This is a _very_ inelegant solution for now. If there's a better way to do this, I am so eager for it.

---

**Asana Ticket:** [QF | TP | Fix shuttle icon size, line color, map color in details view](https://app.asana.com/0/555089885850811/1209501772392193/f)

### DIFF
--- a/lib/dotcom/trip_plan/map.ex
+++ b/lib/dotcom/trip_plan/map.ex
@@ -175,6 +175,20 @@ defmodule Dotcom.TripPlan.Map do
   def stop_icon_size(_), do: %{icon_size: [22, 22], icon_anchor: [0, 0]}
 
   @spec leg_color(Leg.t()) :: String.t()
+  defp leg_color(%Leg{
+         mode: %TransitDetail{
+           route: %Route{description: :rail_replacement_bus, name: name, color: color}
+         }
+       }) do
+    case name do
+      "Blue" <> _ -> "#003DA5"
+      "Green" <> _ -> "00843D"
+      "Orange" <> _ -> "#ED8B00"
+      "Red" <> _ -> "#DA291C"
+      _ -> color
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


lib/dotcom/trip_plan/map.ex
```diff
@@ -175,6 +175,20 @@ defmodule Dotcom.TripPlan.Map do
...
+         mode: %TransitDetail{
+           route: %Route{description: :rail_replacement_bus, name: name, color: color}
+         }
+       }) do
+    case name do
+      "Blue" <> _ -> "#003DA5"
+      "Green" <> _ -> "00843D"
+      "Orange" <> _ -> "#ED8B00"
+      "Red" <> _ -> "#DA291C"
+      _ -> color
```
joshlarson, line 188: **Question:** Should this fallback to the commuter rail color? Or stick with whatever GTFS says (which will be the bus color for now).

thecristen (approved):
oof, this'll do for now

lib/dotcom/trip_plan/map.ex
joshlarson: [Yeah, me too](https://github.com/mbta/dotcom/pull/2499/commits/afa0d7169669e3ed454cc23e9bff0c0bed23ec3e)

PR #2500, opened 2025-04-09
feat: Suppress a specific alert so that it doesn't show up anywhere

joshlarson:
**Asana Ticket:** [Hide alert 636777 for the RL Ashmont Diamond disruption](https://app.asana.com/0/555089885850811/1209938689385280/f)

### DIFF
--- a/lib/alerts/cache/fetcher.ex
+++ b/lib/alerts/cache/fetcher.ex
@@ -48,7 +48,10 @@ defmodule Alerts.Cache.Fetcher do
   def handle_info(:fetch, {update_fn, api_mfa, repeat_ms} = state) do
     case api_result(api_mfa) do
       %{data: data} ->
-        alerts = Enum.map(data, &Parser.Alert.parse/1)
+        alerts =
+          data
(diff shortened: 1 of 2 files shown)


lib/alerts/cache/fetcher.ex
```diff
@@ -48,7 +48,10 @@ defmodule Alerts.Cache.Fetcher do
   def handle_info(:fetch, {update_fn, api_mfa, repeat_ms} = state) do
     case api_result(api_mfa) do
       %{data: data} ->
-        alerts = Enum.map(data, &Parser.Alert.parse/1)
+        alerts =
+          data
+          |> Enum.map(&Parser.Alert.parse/1)
+          |> Enum.reject(&suppressed_alert?/1)
```
thecristen, line 54: kinda feels not worth it to parse alerts before rejecting, what're your thoughts on discarding the bad alerts before parsing?

thecristen (approved):
The silliest PR 2500

thecristen (approved):

PR #2503, opened 2025-04-10
fix(TripPlan.Parser): interpret MBTA buses as accessible

thecristen:
From a recent discussion during my work towards [TP | Update trip accessibility display](https://app.asana.com/1/15492006741476/project/555089885850811/task/1209572743277407?focus=true) it's come to light that we'd like MBTA buses to be considered accessible in the trip planner! So, we can now account for that while parsing OTP output.


joshlarson (approved):
Looks great! I left two style comments, which you can take-em-or-leave-em (I disabled auto-merge so that you'd have a chance to see them).

PR #2511, opened 2025-04-11
fix(TripPlan.ItineraryGroups): correct representative itinerary

thecristen:
For the "arrive by" situation, actually chooses the representative itinerary correctly (after limiting the itineraries in the group).

image


joshlarson (approved):
It might be worth adding a test for this too.

PR #2512, opened 2025-04-15
fix: Alert-table line height sizes on mobile

joshlarson:
Update the alert-table line height sizes so that on mobile, they're the same height on subway pages as on bus pages.

Screenshot 2025-04-15 at 10 49 51 AM

---
Below is a before/after comparison on mobile (Before is on the left; after is on the right)

Screenshot 2025-04-15 at 10 50 34 AM

---
This also slightly shrinks the line height on desktop

(Before on the left; after on the right)
Screenshot 2025-04-15 at 10 53 28 AM

---

But on desktop, they're still slightly bigger on subway pages than on bus pages.

Screenshot 2025-04-15 at 10 54 51 AM

---

**Asana Ticket:** [QF | Adjust subway alerts tab link height on mobile](https://app.asana.com/1/15492006741476/project/1204352509486278/task/1209775659080578?focus=true)

### DIFF
--- a/lib/dotcom_web/components/alerts/subway.ex
+++ b/lib/dotcom_web/components/alerts/subway.ex
@@ -41,11 +41,7 @@ defmodule DotcomWeb.Components.Alerts.Subway do
 
     ~H"""
     <div class="m-alerts__time-filters">
-      <a
-        :for={group <- group_order()}
-        href={"#" <> anchor(group)}
-        class="m-alerts__time-filter leading-[2]"
-      >
+      <a :for={group <- group_order()} href={"#" <> anchor(group)} class="m-alerts__time-filter">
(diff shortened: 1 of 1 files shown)


lib/dotcom_web/components/alerts/subway.ex
```diff
@@ -41,11 +41,7 @@ defmodule DotcomWeb.Components.Alerts.Subway do
 
     ~H"""
     <div class="m-alerts__time-filters">
-      <a
-        :for={group <- group_order()}
-        href={"#" <> anchor(group)}
-        class="m-alerts__time-filter leading-[2]"
```
joshlarson, line 47: The only non-formatting change is removing `leading-[2]` from the class list.

thecristen (approved):

PR #2515, opened 2025-04-15
feat: Update Schedule Finder Layout

joshlarson:
(Before is to the left; After is to the right)

Screenshot 2025-04-16 at 10 34 16 AM

Specific things to note:
- No more special subway banner at the top
- Daily Schedule is now below Upcoming Departures (and now matches other modes)
- No more "Plan Your Trip" button
- "Train Frequency" box is no longer an accordion, but otherwise looks the same as it ever did
- This PR also removes some CSS that's no longer used and clean up some imports

--- 

**Asana Ticket:** [SR | Update subway Schedule Finder layout](https://app.asana.com/1/15492006741476/project/555089885850811/task/1209964779942178?focus=true)


thecristen (approved):
😍 so much nice cleanup here

PR #2516, opened 2025-04-15
chore: remove /admin page

thecristen:
Since we're not really using it at the moment anyway.

When we revive it, we should likely be using Keycloak to handle authentication.


joshlarson (commented):
I think we need to delete [the admin layout file](https://github.com/mbta/dotcom/blob/00ae4c1de584852a19e3246e4fc29c894f73e2c4/lib/dotcom_web/templates/layout/admin.html.heex) as well, since that has a link to the admin page, which is now broken, causing Dialyzer to get mad about it.

That makes me sad, because I really like the "🚌 Admins Only 🚌" marquee 😭

PR #2518, opened 2025-04-18
refactor(OpenTripPlanner): use updated PlanParams

thecristen:
Updated that in https://github.com/mbta/open_trip_planner_client/pull/7


joshlarson (changes requested):
Screenshot 2025-04-23 at 1 51 24 PM

Mostly looks good (and IMO the code in this PR is organized better than when I tried it out on my own last week 🎉). The one thing is that we do need to fetch more than OTP-client's default of 5 itineraries in order to capture enough trips. The rider-facing effect is screenshotted above.

PR #2520, opened 2025-04-22
Remove real-time predictions from line diagram

anthonyshull:
https://app.asana.com/1/15492006741476/project/555089885850811/task/1209964790389173?focus=true


thecristen (commented):
image
Removed just a little too much CSS I think, haha (crowding icon used to be adjacent to the minutes instead of below)

lib/dotcom_web/controllers/schedule/line_api.ex
```diff
@@ -90,134 +77,4 @@ defmodule DotcomWeb.ScheduleController.LineApi do
...
-  end
-
-  defp do_realtime(route_id, direction_id, date, now, tooltips) do
-    headsigns_by_stop =
-      TransitNearMe.time_data_for_route_by_stop(
-        route_id,
-        String.to_integer(direction_id),
-        date: date,
-        now: now
-      )
```
thecristen, line 137: I believe this was the only place this function was used, so you can kill it too :)

lib/dotcom_web/controllers/schedule/line_api.ex
```diff
@@ -90,134 +77,4 @@ defmodule DotcomWeb.ScheduleController.LineApi do
   end
 
   defp other_route_stops(_), do: []
-
-  @doc """
-  Provides predictions and vehicle information for a given route and direction, organized by stop.
-  The line diagram polls this endpoint for its real-time data.
-  """
-  @spec realtime(Plug.Conn.t(), map()) :: Plug.Conn.t()
-  def realtime(conn, %{"id" => route_id, "direction_id" => direction_id}) do
```
thecristen, line 99: Pretty sure there's an associated unit test you can delete too

assets/ts/schedule/components/line-diagram/StopCard.tsx
```diff
@@ -166,37 +132,23 @@ const StopCard = ({
...
+            <div className="m-schedule-diagram__alert">
+              {effectNameForAlert(diversionAlert!)}
+            </div>
+          ) : null}
         </div>
-
-        {(stopTree
-          ? hasUpcomingDeparturesIfSubway(stopTree, stopId, liveData)
-          : true) && (
+        {!isEnd ? (
```
joshlarson, line 141: **Issue (blocking):** We want to show the `View Departures` link even for the last item in a list of stop cards.

joshlarson (changes requested):
Left one functional comment - everything else (minus what Cristen found) looks good, I think!

thecristen (commented):
I thought I submitted these but forgot! 

IMO The only blocking things are 
1. the newly misplaced crowding icon from the Upcoming Departures modal, and 
2. the erroneously excluded "View departures" terminus link which Josh pointed out.

joshlarson (approved):
This PR has me seeing RED 📉

PR #2522, opened 2025-04-24
feat: Have the back button close schedule finder, rather than navigating to a different page

joshlarson:
This fixes a bug/weird-UX where if you open schedule finder from a schedule page, and then hit the back button, instead of closing the schedule finder, it navigates to whatever page you were on _before_ schedule finder was opened. On desktop, that's _fine_ (maybe not ideal, but not awful), but on mobile, schedule finder takes up the whole screen and looks like a separate page, so it looks from the user's perspective like hitting the back button takes you back two pages.

This fix comes in two parts:
- Use `pushState` instead of `replaceState` when opening the schedule finder modal.
  - This is what allows the back button to do something _other_ than navigate to the pre-schedule-page part of your history.
- Add a `popstate` event listener whenever `modalOpen` is true.
  - Without this, because the only thing that changes is query params, the browser won't actually reload the page, which means that we need to manually handle closing the modal when it's open.

---

**Asana Ticket:** [SR | Fix Schedule Finder close/back behavior](https://app.asana.com/1/15492006741476/project/555089885850811/task/1209964790389162?focus=true)

### DIFF
--- a/assets/ts/schedule/components/SchedulePage.tsx
+++ b/assets/ts/schedule/components/SchedulePage.tsx
@@ -35,7 +35,7 @@ const updateURL = (origin: SelectedOrigin, direction?: DirectionId): void => {
     };
     const newLoc = updateInLocation(newQuery, window.location);
     // newLoc is not a true Location, so toString doesn't work
-    window.history.replaceState({}, "", `${newLoc.pathname}${newLoc.search}`);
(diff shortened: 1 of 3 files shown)


anthonyshull (approved):

joshlarson:
Closed for now, because we decided that this back-button behavior introduces more unpredictability and confusion than it resolves.

PR #2526, opened 2025-04-24
Bus stop accessibility

anthonyshull:
https://app.asana.com/1/15492006741476/project/555089885850811/task/1209140471569429?focus=true

I just rewrote the header in Elixir and removed the React one. Incidentally, this has the effect of making the stop page actually match the line diagram.

Screenshot Capture - 2025-04-24 - 17-07-43

Screenshot Capture - 2025-04-24 - 17-09-06


assets/ts/stop/__tests__/StopPageTest.tsx
```diff
@@ -63,13 +63,6 @@ describe("StopPage", () => {
     cleanup();
   });
 
-  it("should render", async () => {
```
anthonyshull, line 66: 📓 Since the header is no longer React, this is no longer applicable.

assets/ts/stop/components/amenities/AccessibilityAmenityCard.tsx
```diff
@@ -82,17 +82,21 @@ const AccessibilityAmenityCard = ({
   const stationOrStop = isStation ? "station" : "stop";
   const features = without(accessibleFeatures, "accessible");
   const hasAccessibleFeatures = accessibleFeatures.includes("accessible");
-  const badge = !hasAccessibleFeatures && (
+
+  const accessibleOverride = document.querySelector("div[data-mbta-accessible]")?.getAttribute("data-mbta-accessible") === "true";
```
anthonyshull, line 86: 📓 This is kind of a hacky way of getting this information into React without having to duplicate the logic that looks at both the stop and the routes for that stop...the latter of which we don't even have in this React code.

lib/dotcom_web/components/transit_icons.ex
```diff
@@ -0,0 +1,50 @@
+defmodule DotcomWeb.Components.TransitIcons do
```
anthonyshull, line 1: 📓 I put these in here for now. They aren't really customizable. We should figure out how we want to do that on a broader scale later. E.g., we could offer three different sizes.

lib/dotcom_web/components/transit_icons.ex
```diff
@@ -0,0 +1,50 @@
...
+  use DotcomWeb, :component
+
+  import MbtaMetro.Components.Icon, only: [icon: 1]
+
+  @doc """
+  Icon for accessibility.
+  """
+  def accessibility(assigns) do
+    ~H"""
+    <.badge class="bg-cobalt-30 h-6 !px-1.5 !min-w-6 max-w-6 !rounded-sm inline-flex" variant="square">
```
joshlarson, line 15: **Suggestion:** The fact that this is specifying all of its styling, and a bunch of the styles it specifies are `!`-force-overrides of `<.badge />` defaults suggests that `<.badge />` isn't the right tool for this job. If you replace `.badge` with `div`, then the only thing you have to add to keep it looking right is `items-center`.

There are two advantages that I see of switching to `div`:
- Using `.badge` here means that if the design system ever specifies how actual badges should look (actual badges like the `Upcoming` badge on alert components, for instance) in a way that causes Metro's badge implementation to change, that could cause surprising changes in these icons as well. 
- It would be a slight reduction in code, because we wouldn't need to override a bunch of the `<.badge />` defaults.

```elixir
 ~H"""
 <div class="bg-cobalt-30 h-6 px-1.5 w-6 rounded-sm inline-flex items-center">
   <.icon name="wheelchair" class="h-3.5 w-3.5 fill-white" aria-label="Accessible" />
 </div>
 """
```

(*Note: This comment applies to all three icons in this file, so even if you apply my other suggestion to use the O.G. SVG for the accessibility icon, you might still find this worth doing for the parking and zone icons.*)

---

Even if you prefer to stick with `<.badge />` for this, is there an advantage to separately specifying `!min-w-6 max-w-6` rather than just `w-6` (or maybe `!w-6`)?

joshlarson (commented):
This looks so great! I'm super intrigued by this way of moving stuff out of React in smaller pieces - seems like it worked pretty well for this task!

The only thing that's giving me pause on approving is the wheelchair icon thing - I'm not convinced that this change is a good one - maybe we should loop in Chase?

assets/ts/stop/components/StopPageHeader.tsx
```diff
@@ -1,44 +0,0 @@
-import React, { ReactElement } from "react";
-import { Route, Stop } from "../../__v3api";
-import StopFeatures from "./icons/StopFeatures";
```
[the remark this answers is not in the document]
anthonyshull: Pretty sure it is still being used in the line diagram.

lib/dotcom_web/components/transit_icons.ex
```diff
@@ -0,0 +1,45 @@
...
+    """
+  end
+
+  @doc """
+  Icon for parking.
+  """
+  def parking(assigns) do
+    ~H"""
+    <div class="bg-gray-light text-white h-6 px-1.5 w-6 rounded-sm inline-flex items-center">
+      <span arial-label="Parking">P</span>
```
joshlarson, line 25: Typo 😇 

```suggestion
      <span aria-label="Parking">P</span>
```

lib/dotcom_web/components/transit_icons.ex
```diff
@@ -0,0 +1,45 @@
...
+    <.icon type="icon-svg" name="icon-accessible-small" class="h-6 w-6" aria-label="Accessible" />
+    """
+  end
+
+  @doc """
+  Icon for parking.
+  """
+  def parking(assigns) do
+    ~H"""
+    <div class="bg-gray-light text-white h-6 px-1.5 w-6 rounded-sm inline-flex items-center">
```
joshlarson, line 24: **Issue:** Ahh - my bad - I think this *was* picking up `font-bold` and `justify-center` from `<.badge />`, so we need to add that in here as well.

Screenshot 2025-05-01 at 1 28 52 PM

versus

Screenshot 2025-05-01 at 1 31 46 PM

And I think it probably does need a `notranslate`, [like the route symbols](https://github.com/mbta/dotcom/blob/a933416552d60f0db96623c72064c9de450fb895/lib/dotcom_web/components/route_symbols.ex#L211), so that Google translate doesn't try to translate "P" into something.

lib/dotcom_web/components/transit_icons.ex
```diff
@@ -0,0 +1,45 @@
...
+    """
+  end
+
+  @doc """
+  Icon for displaying the zone of a stop.
+  """
+  def zone(assigns) when not is_nil(assigns.stop.zone) do
+    ~H"""
+    <div
+      class="border-[1px] border-commuter-rail text-commuter-rail max-h-6 rounded-sm px-1.5"
```
joshlarson, line 36: **Issue:** This needs `font-bold` and `inline-flex justify-center` in order to look right.

Screenshot 2025-05-01 at 1 35 51 PM

versus

Screenshot 2025-05-01 at 1 36 03 PM


---

Or honestly, maybe I was wrong about swapping the zone icon away from a badge 🤔... it does look more *badge*-like than the other icons. (I think the parking one still makes more sense as a plain old div though)

lib/dotcom_web/components/transit_icons.ex
```diff
@@ -0,0 +1,45 @@
...
+      variant="square"
```
joshlarson, line 37:
```suggestion
```

Don't need the variant for a `<div />` 😇

joshlarson (changes requested):
I wasn't thorough enough in my suggestions before! Some of the icons don't look right now - should be easy fixes though.

joshlarson (approved):
🚌 ♿

PR #2527, opened 2025-04-30
Group commuter rail alerts

anthonyshull:
Group commuter rail alerts nearly like subway, but with some small differences. I pulled the logic from subway into a behaviour module and now you just have to define the grouping and sorting and it just works.

https://app.asana.com/1/15492006741476/project/555089885850811/task/1210070462275466?focus=true

### DIFF
--- a/lib/dotcom/alerts/commuter_rail.ex
+++ b/lib/dotcom/alerts/commuter_rail.ex
@@ -8,22 +8,14 @@ defmodule Dotcom.Alerts.CommuterRail do
   import import Dotcom.Alerts, only: [service_impacting_effects: 0]
 
   @impl Dotcom.Alerts.Group
-  def effects do
-    %{
-      "Bike" => [bike_issue: 1],
-      "Elevator & Escalator" => [
-        elevator_closure: 1,
-        escalator_closure: 1
-      ],
-      "Other" => [],
-      "Parking" => [parking_issue: 1],
-      "Service" => service_impacting_effects(),
-      "Track Change" => [track_change: 1]
-    }
-  end
-
-  @impl Dotcom.Alerts.Group
-  def group_order do
-    ["Service", "Track Change", "Elevator & Escalator", "Bike", "Parking", "Other"]
+  def effect_groups do
+    [
+      {"Service", service_impacting_effects()},
+      {"Track Change", [track_change: 1]},
+      {"Elevator & Escalator", [elevator_closure: 1, escalator_closure: 1]},
+      {"Bike", [bike_issue: 1]},
+      {"Parking", [parking_issue: 1]},
+      {"Other", []}
+    ]
   end
 end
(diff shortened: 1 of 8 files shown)


lib/dotcom/alerts/commuter_rail.ex
```diff
@@ -0,0 +1,29 @@
...
+      "Parking" => [parking_issue: 1],
+      "Service" => service_impacting_effects(),
+      "Track Change" => [track_change: 1]
+    }
+  end
+
+  @impl Dotcom.Alerts.Group
+  def group_order do
+    ["Service", "Track Change", "Elevator & Escalator", "Bike", "Parking", "Other"]
+  end
```
thecristen, line 28: Kinda like what your `group()` function outputs, I wonder if we should just write `effects` and `group_order` together in the first place.

```elixir
@impl Dotcom.Alerts.Group
def grouped_effects do
  [
     {"Service", service_impacting_effects()},
     {"Track Change", [track_change: 1]},
     {"Elevator & Escalator", [elevator_closure: 1, escalator_closure: 1],
     {"Bike", [bike_issue: 1]},
     {"Parking", [parking_issue: 1]},
     {"Other", []}
   ]
end
```

Then you'd only need to iterate through this one thing, it's already in order, etc

lib/dotcom_web/controllers/schedule/alerts_controller.ex
```diff
@@ -12,8 +12,16 @@ defmodule DotcomWeb.ScheduleController.AlertsController do
   plug(:tab_name)
 
   def show(conn, _) do
+    mode =
+      conn
+      |> Map.get(:assigns, %{})
+      |> Map.get(:breadcrumbs, %{})
+      |> Enum.at(1, %{})
+      |> Map.get(:text)
+
```
thecristen, line 21: Interesting, I would've first tried to get this via `Routes.Route.type_atom(conn.assigns.route)`, or maybe `path_atom` or `type_name` from that same module. To me that seems more direct, and that wouldn't be affected if we ever remove breadcrumbs or change what they say.

lib/dotcom/alerts/group.ex
```diff
@@ -0,0 +1,104 @@
...
+      end
+
+      @doc """
+      Sort alerts by station and then by start time.
+      """
+      @spec sort_alerts([Alert.t()]) :: [Alert.t()]
+      def sort_alerts(alerts) do
+        alerts
+        |> Enum.sort(&sort_by_station_sorter/2)
+        |> Enum.sort(&sort_by_start_time_sorter/2)
```
thecristen, line 85: I'm not sure this achieves the intended sorting, might want to test that. `sort_by/3` with a tuple can be used to sort by multiple things though ([doc](https://hexdocs.pm/elixir/1.17.3/Enum.html#sort_by/3-examples))

lib/dotcom_web/controllers/schedule/alerts_controller.ex
anthonyshull: Oh, if there is a better place to get the info, I'll take it!

lib/dotcom/alerts/group.ex
anthonyshull: This was just copied to a new module; it's already in production.

lib/dotcom/alerts/commuter_rail.ex
anthonyshull: I like that idea :)

PR #2530, opened 2025-05-02
fix(JourneyList): don't overwrite schedule map entries with later schedules

thecristen:
#### Summary of changes

**Asana Ticket:** [Fix schedule finder daily schedule to handle trips visiting the same stop multiple times](https://app.asana.com/1/15492006741476/project/555089885850811/task/1210103449936938?focus=true)

It's a bit of a hack, but it works beautiful for the ferry case (as seen in the dev-green API data).

### DIFF
--- a/lib/journey_list.ex
+++ b/lib/journey_list.ex
@@ -6,6 +6,7 @@ defmodule JourneyList do
 
   alias PredictedSchedule.Group
   alias Predictions.Prediction
+  alias Routes.Route
   alias Schedules.{Schedule, Trip}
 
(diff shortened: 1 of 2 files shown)


joshlarson (approved):
Looks good! 

I won't really pretend to understand how this fixes the bug (or why we only want to do this for ferry), but it does look right.

Would it be possible to add a test for this? (Non-blocking because this bugfix seems to work, and is blocking other teams, but we really should get this logic tested as it gets more and more complicated and edge-casey)

PR #2535, opened 2025-05-05
cleanup: Remove unused `lined_list` imports

joshlarson:
Oops - [missed these](https://github.com/mbta/dotcom/pull/2534).

These were actually the last usages of `lined_list`, so we could, in theory, remove that entire component as well. Thoughts?


thecristen (changes requested):
Sure, remove the component

thecristen (approved):

PR #2536, opened 2025-05-06
refactor: Change status row header from flex to grid

joshlarson:
Re-implement the `<.status_row_heading />` component as a CSS `grid` instead of its previous `flex` implementation. This is in order to support the kind of layout where there is sub-heading text directly below the status description (**Shuttle** or **Station Closure** or whatever). 

This was intended to be a pure refactor, but while working on this, I noticed a few small details that I got wrong when originally implementing the subway status feature, so I went ahead and fixed those.

Screenshot 2025-05-06 at 9 57 38 AM

- The font size for disrupted statuses is now slightly larger, as well as being bold.
- The padding to the left of the root pills is now 4px rather than 8px.

---

**Asana Subtask:** [[Endpoints] Refactor status row heading to use grid instead of flex](https://app.asana.com/1/15492006741476/project/1205718271156548/task/1210172306293798?focus=true)
**Asana Parent Ticket:** [Add endpoints of disruption to system status and planned work](https://app.asana.com/1/15492006741476/project/1205718271156548/task/1209641123314696?focus=true)

### DIFF
--- a/lib/dotcom_web/components/system_status/status_label.ex
+++ b/lib/dotcom_web/components/system_status/status_label.ex
@@ -22,11 +22,8 @@ defmodule DotcomWeb.Components.SystemStatus.StatusLabel do
     assigns = assigns |> assign(:rendered_prefix, rendered_prefix)
 
     ~H"""
-    <span class={[status_classes(@status), "flex items-center gap-2"]}>
-      <.status_icon status={@status} />
-      <span data-test="status_label_text">
-        {@rendered_prefix} {description(@status, @prefix, @plural)}
-      </span>
+    <span data-test="status_label_text" class={status_classes(@status)}>
+      {@rendered_prefix} {description(@status, @prefix, @plural)}
     </span>
     """
   end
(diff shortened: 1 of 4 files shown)


lib/dotcom_web/components/system_status/status_label.ex
```diff
@@ -22,11 +22,8 @@ defmodule DotcomWeb.Components.SystemStatus.StatusLabel do
     assigns = assigns |> assign(:rendered_prefix, rendered_prefix)
 
     ~H"""
-    <span class={[status_classes(@status), "flex items-center gap-2"]}>
-      <.status_icon status={@status} />
-      <span data-test="status_label_text">
-        {@rendered_prefix} {description(@status, @prefix, @plural)}
-      </span>
+    <span data-test="status_label_text" class={status_classes(@status)}>
```
joshlarson, line 25: **Note:** This is a functional-ish change to this component, because now `<.status_label />` will only have the text, not the icon.

PR #2539, opened 2025-05-07
chore: update Elixir and Erlang versions

thecristen:


joshlarson (approved):
This looks good and seems to work! (I got spooked because some predictions in dev-green are messed up, but that looks like it's a dev-green data problem, not an Elixir 1.18 problem)

The new Elixir version introduces a bunch of new warnings, mostly (entirely?) about invalid types in assignments and pattern matches. I'm on the fence about whether it's worth it to clean up those warnings in this PR, or do it in follow-ups.

thecristen:
@joshlarson it seems there are very few warnings for Dotcom (yes, there are various for the dependencies, but we've limited access to fix those). For Dotcom's, I was able to fix one of them, but the other turns out to be more involved so I'm gonna leave that one in for now.

lib/dotcom_web/controllers/transit_near_me_controller.ex
```diff
@@ -31,7 +31,7 @@ defmodule DotcomWeb.TransitNearMeController do
     data = TransitNearMe.build(location, [])
 
     case data do
-      {:stops, {:error, _}} -> :error
+      %{stops: {:error, _}} -> :error
```
joshlarson, line 34: **Suggestion:** Given that this condition wasn't previously match-able before, that means it wasn't actually being used. I bet that means this condition could be deleted.

PR #2541, opened 2025-05-09
fix: only rewrite file links in prod

thecristen:
#### Summary of changes

**Asana Ticket:** [CMS resources link to incorrect environment](https://app.asana.com/1/15492006741476/project/385363666817452/task/1210203210481595?focus=true)

We have an [existing line in our router](https://github.com/mbta/dotcom/blob/2025.05.14.01/lib/dotcom_web/router.ex#L320) to handle paths such as `/sites/default/files/...` which link to various files across the site (mostly images and PDFs, all from Drupal). So to achieve this task, this PR adjusts a bunch of links such that requests are handled by this controller, instead of doing whatever other redirecting/hostname rewriting it was doing before.

However, we still want these to point to the cdn.mbta.com host on mbta.com, so there's a little logic in place (accessed via checking `Application.get_env(:dotcom, :is_prod_env?)`) to maintain that.

### DIFF
--- a/lib/dotcom/map_helpers.ex
+++ b/lib/dotcom/map_helpers.ex
@@ -2,7 +2,7 @@ defmodule Dotcom.MapHelpers do
   @moduledoc """
   Map-related helper functions.
   """
-  import DotcomWeb.Router.Helpers, only: [static_url: 2]
+  import CMS.Helpers, only: [rewrite_url: 1]
   import DotcomWeb.ViewHelpers, only: [cms_static_page_path: 2]
 
   @spec map_pdf_url(atom) :: String.t()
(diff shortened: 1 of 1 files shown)


amaisano:
schedule_view.ex needs updates too -- the PDF URL output is still using the old static_path, which I think causes it to skip the `/sites/` path prefix. Could be other places too - still looking (will update comment with more).

lib/dotcom_web/controllers/helpers.ex
```diff
@@ -122,11 +122,11 @@ defmodule DotcomWeb.ControllerHelpers do
   """
   @spec forward_static_file(Conn.t(), String.t()) :: Conn.t()
   def forward_static_file(conn, url) do
-    case client() |> @req.get(url: url) do
-      {:ok, %{status: 200, body: body, headers: headers}} ->
+    case client() |> @req.get(url: url, cache: true) do
```
amaisano, line 125: What's the cache timing on this?

lib/dotcom_web/templates/redirect/show.html.eex
amaisano, line 1: I think this gif is actually local to dotcom, not the CMS. So we wanna make sure it's not trying to get it from CMS in that case.

lib/dotcom_web/controllers/helpers.ex
thecristen: N/A, it uses the `if-modified-since` header! ([doc](https://hexdocs.pm/req/0.5.1/Req.Steps.html#cache/1))

lib/dotcom_web/templates/redirect/show.html.eex
thecristen: ahhh my bad, thanks for catching!

amaisano:
Not sure if this is a local image, but the little thubmnails in these buttons seem to be linking to the CDN. If that's actually a CMS image, I think it could use the same logic as the larger images you've updated.

image

PR #2545, opened 2025-05-13
fix: Don't include connecting stops when determining station accessibility

joshlarson:
Don't include `connecting_stops` when considering whether a stop/station is accessible.

This was leading to an issue where stations were being given a ♿ icon if they were connected to a bus stop (which is always considered accessible), even if the station itself is not accessible. This was impacting some Commuter Rail stations as well as a few inaccessible Green Line stations, including Hynes Convention Center and Boylston.

### Screenshots

(Before on the left; After on the right)

[Belmont](https://www.mbta.com/stops/place-FR-0064)
Screenshot 2025-05-13 at 1 01 08 PM

[Boylston](https://﻿﻿www.mbta.com/stops/place-boyls)
Screenshot 2025-05-13 at 1 04 43 PM
(This also works for stations whose names start with letters other than `B`)

---

It also fixes the `Accessibility` amenity card for affected stations.

Screenshot 2025-05-13 at 1 02 57 PM

---

No ticket.

### DIFF
--- a/lib/dotcom_web/components/stops/header.ex
+++ b/lib/dotcom_web/components/stops/header.ex
@@ -32,7 +32,7 @@ defmodule DotcomWeb.Components.Stops.Header do
         <div class="flex items-end justify-items-end gap-2 flex-wrap">
           <.mode_icons routes_by_stop={@routes_by_stop} />
           <.zone stop={@stop} />
-          <.accessibility :if={accessible?(assigns)} />
+          <.accessibility :if={@accessible?} />
           <.parking :if={parking?(assigns)} />
         </div>
         <div :if={all_bus_routes?(assigns) and not @stop.station?} class="text-sm">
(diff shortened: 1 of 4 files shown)


anthonyshull (approved):

thecristen (approved):
I'm confused how this fixed the modals but otherwise looks great!

joshlarson:
@thecristen Ahahaha yeah it's subtle (I had to go back and double-check that it actually worked when I saw your comment).

The confusing bit is that we re-calculated whether a station was accessible twice, rather than just passing the result into the header component. So one change was to have the new stop page header [take in](https://github.com/mbta/dotcom/pull/2545/files#diff-8db3f12addec772998b668aaec720f04b6cada2750741fe408603c72d5c38569R5-R9) and [use](https://github.com/mbta/dotcom/pull/2545/files#diff-a73d56cf14191e5be66fd42d246e42cd6a793f34026b401aef1460614ab29e83R35) the calculated `accessible?` field, same as [the react part of the stop page does](https://github.com/mbta/dotcom/blob/11cb38405463f7717cb2cb5f1bbfa090918f0815/lib/dotcom_web/templates/stop/show.html.heex#L13). With that change, the accessibility amenity card and the stop features at the top of the page will both get the same `accessibility?` field.

The other change was to [update how we're computing](https://github.com/mbta/dotcom/pull/2545/files#diff-5876bc4a528bdc80fbf22e1ae7b9acecf827de961b83841a0c9b336f6205a211R390-R394) `accessible?` so that it doesn't look at connecting stops. Because of the first change, this fixes both the amenity card and the icon at the top of the page.

PR #2546, opened 2025-05-13
hotfix: Increase default Finch connection pool size

joshlarson:
No ticket.

We saw some issues that included error messages like

```
** (RuntimeError) Finch was unable to provide a connection
within the timeout due to excess queuing for connections.
Consider adjusting the pool size, count, timeout or reducing
the rate of requests if it is possible that the downstream
service is unable to keep up with the current rate.
```

So we figured we would bump the Finch pool size from its default of 50 up to 200.

[Finch docs.](https://hexdocs.pm/finch/Finch.html#module-usage)

### DIFF
--- a/lib/dotcom/application.ex
+++ b/lib/dotcom/application.ex
@@ -31,7 +31,7 @@ defmodule Dotcom.Application do
       ] ++
         if Application.get_env(:dotcom, :env) != :test do
           [
-            {Finch, name: Telemetry.Finch},
+            {Finch, name: Telemetry.Finch, pools: %{default: [size: 200]}},
             {Dotcom.Telemetry, []},
             {Dotcom.Cache.Telemetry, []},
             {DotcomWeb.Telemetry, []},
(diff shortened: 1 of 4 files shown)


thecristen:
From Req's docs, requests are made using a default Finch pool set up by Req. Here we can start our own Finch pool with custom settings, and instruct Req to use that instead!

lib/dotcom/application.ex
```diff
@@ -27,11 +27,12 @@ defmodule Dotcom.Application do
     children =
       [
         {Application.get_env(:dotcom, :cache, Dotcom.Cache.Multilevel), []},
-        {Dotcom.RateLimit, [clean_period: 60_000 * 10]}
+        {Dotcom.RateLimit, [clean_period: 60_000 * 10]},
+        {Finch, name: Dotcom.Finch, pools: %{default: [size: 300, count: 10]}}
```
joshlarson, line 31: Silly question: Do we want this for the `:test` environment too, or just deployed envs?

anthonyshull (approved):

PR #2551, opened 2025-05-16
chore: Upgrade faker to version 0.19.0-alpha.1

joshlarson:
Faker released a version that doesn't warn in Elixir 1.17+!


thecristen (approved):
I don't really like using alpha versions, but this changelog seems modest enough.

PR #2555, opened 2025-05-19
feat: set up deep linking for MBTA Go

boringcactus:
#### Summary of changes

**Asana Ticket:** [Create link that opens the app if you have it installed and the store otherwise](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209949928014406?focus=true)

As described in the ticket, sets `/go` to redirect to the App Store / Play Store / MBTA Go project page in the same way that #2437’s `/app-store` does, and creates the `/.well-known/` files that are needed to have MBTA Go open links under `/go` if it’s installed.

It’d be theoretically possible to have every instance of dotcom attached to every instance of the mobile app, which would let this be self-contained like #2437 is, but for people with multiple instances of the app installed, that’d make it difficult to control which instance of the mobile app is actually opened. Instead, per discussion with the MBTA Go team, I’ve set this to pull information about a single MBTA Go instance from the environment, so that URLs for dotcom dev can always open in MBTA Go staging, prod can always open in prod, and dev-blue and dev-green can always open in dev-orange.

Devops PR: https://github.com/mbta/devops/pull/2894

---

#### General checks

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -152,6 +152,11 @@ config :dotcom, OpenTripPlanner,
   timezone: System.get_env("OPEN_TRIP_PLANNER_TIMEZONE", "America/New_York"),
   otp_url: System.get_env("OPEN_TRIP_PLANNER_URL")
 
+config :dotcom, :mbta_go,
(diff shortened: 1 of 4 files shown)


thecristen (approved):
Cool! Looks pretty solid to me.

PR #2557, opened 2025-05-20
test: explore if certain paths into MBTA Go deep link work

boringcactus:
Apparently, opening links from other apps to the `/go` URL established in #2555 will correctly open MBTA Go, but manually entering that URL in a Web browser will open that URL in the Web browser instead of opening MBTA Go, on both Android and iOS. As such, we need to figure out how we can get people from a manually entered URL to the deep link, and there are some things that may work differently when they’re on the same domain vs moving across domains, so we need to test a redirect and a link that are from dotcom to dotcom, rather than redirecting or linking from some other domain to dotcom.

If I can borrow a dotcom dev environment for a couple of minutes, I should be able to figure out which things work and which things don’t. This isn’t urgent, but there are a couple of different things that it’s blocking.

### DIFF
--- a/lib/dotcom_web/controllers/mbta_go_deep_link_test_controller.ex
+++ b/lib/dotcom_web/controllers/mbta_go_deep_link_test_controller.ex
@@ -1,0 +1,13 @@
+defmodule DotcomWeb.MbtaGoDeepLinkTestController do
+  @moduledoc false
+
+  use DotcomWeb, :controller
+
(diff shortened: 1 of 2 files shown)


thecristen:
@boringcactus you may borrow dev-green, have at it!

boringcactus:
It seems like, on both iOS and Android, neither redirects nor links work. This is good to know now so we aren’t discovering it three weeks later. Thanks!

PR #2574, opened 2025-05-29
Push updates to commuter rail status

anthonyshull:
This is a POC showing how we can cache system status and push new results out whenever the cache is busted.


lib/dotcom_web/live/system_status.ex
```diff
@@ -1,436 +1,32 @@
...
 
-  alias Dotcom.Alerts.Subway.Disruptions
-  alias Dotcom.SystemStatus
+  import DotcomWeb.Components.SystemStatus.CommuterRailStatus,
+    only: [alerts_commuter_rail_status: 1]
 
-  @date_time_module Application.compile_env!(:dotcom, :date_time_module)
+  def mount(_params, _session, socket) do
+    if connected?(socket) do
+      Phoenix.PubSub.subscribe(Dotcom.PubSub, "commuter-rail-system-status")
```
thecristen, line 15: Nice. How about using a naming scheme like `"system_status:commuter_rail"`, which gives us a clear template for adding to later (e.g. `"system_status:subway"` or `"system_status:Red"`)

lib/dotcom_web/templates/iframe/root.html.heex
```diff
@@ -0,0 +1,25 @@
...
+      <script defer src={"#{webpack_path()}/core.js"}>
+      </script>
+      <script defer src={"#{webpack_path()}/iframe.js"}>
+      </script>
+    <% else %>
+      <link rel="stylesheet" {static_attributes("/css/core.css")} />
+      <link rel="stylesheet" {static_attributes("/css/app.css")} />
+      <script defer {static_attributes("/js/vendors.js")}>
+      </script>
+      <script defer {static_attributes("/js/iframe.js")}>
```
thecristen, line 18: Not sure if this is happening in other browsers, but I do get this in Firefox 

image

config/config.exs
```diff
@@ -84,4 +84,13 @@ config :sentry,
   root_source_code_paths: [File.cwd!()],
   context_lines: 5
 
+
+config :dotcom, Dotcom.Scheduler,
+  jobs: [
+    commuter_rail_status: [
+      schedule: {:extended, "*/30"}, # Runs every two seconds
+      task: {Dotcom.SystemStatus.CommuterRail, :commuter_rail_status, []}
```
thecristen, line 92: I think it might be simpler to have this trigger a new function which _computes and broadcasts_  the commuter rail status (perhaps with the last updated time) every 5 or whatever minutes (and revert `commuter_rail_status/0` to the simpler, uncached version which doesn't itself handle broadcasting)

thecristen (commented):
a few comments!

PR #2576, opened 2025-05-29
refactor: Reduce duplication in CR status component

joshlarson:
In preparation for more commuter-rail status work, I noticed that there was a lot of similarity between how different types of rows were defined in the Commuter Rail Status component. 

This aims to reduce that duplication, with the eventual goal of extracting shared visual components to re-usable components (for instance, pulling the icon / bold-text combo out as a "status label" that's more generic than the current `<.status_label />` component, which is pretty much only suited for subway.

As far as I've been able to tell with visual testing, this is a pure refactor that leaves the end result completely unchanged.

### DIFF
--- a/lib/dotcom_web/components/system_status/commuter_rail_status.ex
+++ b/lib/dotcom_web/components/system_status/commuter_rail_status.ex
@@ -26,7 +26,7 @@ defmodule DotcomWeb.Components.SystemStatus.CommuterRailStatus do
       end)
       |> Enum.sort_by(& &1.sort_order)
 
-    assigns = Map.put(assigns, :status_for_line, status_for_line)
+    assigns = assigns |> assign(:status_for_line, status_for_line)
 
     ~H"""
     <.bordered_container hide_divider>
--- a/lib/dotcom_web/components/system_status/commuter_rail_status.ex
+++ b/lib/dotcom_web/components/system_status/commuter_rail_status.ex
@@ -161,7 +161,7 @@ defmodule DotcomWeb.Components.SystemStatus.CommuterRailStatus do
       }
       disrupted
       label={label}
-      row_name={(index == 0 && row_name(@status)) || nil}
+      row_name={if index == 0, do: row_name(@status)}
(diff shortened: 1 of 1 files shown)


lib/dotcom_web/components/system_status/commuter_rail_status.ex
```diff
@@ -27,7 +26,7 @@ defmodule DotcomWeb.Components.SystemStatus.CommuterRailStatus do
       end)
       |> Enum.sort_by(& &1.sort_order)
 
-    assigns = Map.put(assigns, :rows, rows)
+    assigns = Map.put(assigns, :status_for_line, status_for_line)
```
thecristen, line 29:
```suggestion
    assigns = assign(assigns, :status_for_line, status_for_line)
```

lib/dotcom_web/components/system_status/commuter_rail_status.ex
```diff
@@ -131,127 +130,79 @@ defmodule DotcomWeb.Components.SystemStatus.CommuterRailStatus do
...
-          <span>{row_name(@row)}</span>
-        </div>
+    <.row
+      :for={
+        {{status, label}, index} <-
+          @status.alert_counts |> combine_alert_counts() |> Enum.with_index()
+      }
+      disrupted
+      label={label}
+      row_name={(index == 0 && row_name(@status)) || nil}
```
thecristen, line 164: nit: could also do `if(index == 0, do: row_name(@status))`

thecristen (approved):
This reads quite nicely!

lib/dotcom_web/components/system_status/commuter_rail_status.ex
joshlarson: Huh, yeah I was trying that and confusing myself about where the parens() should go.

[This seems to work!](https://github.com/mbta/dotcom/pull/2576/commits/70416f86032642c3988ab5c79193e47b93fc10ae)

PR #2579, opened 2025-06-02
feat(TimetableController): use CSV for Boat-F6/7/8

thecristen:
#### Summary of changes

**Asana Ticket:** [Fix Winthrop and Quincy timetables](https://app.asana.com/1/15492006741476/project/555089885850811/task/1210136714832172?focus=true)

The changes are organized per commit. I've transcribed the PDF schedules for the Winthrop and Quincy ferries into CSVs, and read from that to populate the timetable.

This also disables changing the direction (these routes only serve one direction per the GTFS data).


anthonyshull (approved):
It would be nice to create these CSVs via some automated process in the future.

PR #2580, opened 2025-06-03
refactor: Make `<.status_label />` component more re-usable

joshlarson:
Two major changes:
- Move icon and subheading layout and logic into the component for consistency
- Move prefix and plural logic out of the component for flexibility

So now `<.status_label />` will render with an icon, and will take text as an argument, as well as an optional subheading. Below are some example `<.status_label />` components on the preview page, before on the left, after on the right.

Screenshot 2025-06-03 at 10 12 45 AM

Some additional thoughts:
- From a rider perspective, this should be a pure refactor. The change is entirely in what components are available to `<.status_row_header />` and other possible consumers of `<.status_label />`.
- CR system status doesn't use this yet - that will be a follow-up (there's some fiddling with wrapping behavior that needs to happen in order for that to be fully right, and nothing is blocked on CR system status using this.)

---

No ticket, but in anticipation for adding detailed status on the CR timetable pages.

### DIFF
--- a/lib/dotcom_web/live/system_status.ex
+++ b/lib/dotcom_web/live/system_status.ex
@@ -43,13 +43,19 @@ defmodule DotcomWeb.Live.SystemStatus do
     <h1>Misc Components</h1>
     <h2>Status Labels</h2>
     <div class="flex flex-col gap-2">
-      <.status_label status={:normal} />
-      <.status_label status={:shuttle} />
-      <.status_label status={:shuttle} plural />
-      <.status_label status={:shuttle} prefix="8:30pm" />
-      <.status_label status={:shuttle} prefix="Wed Feb 12 - Fri Feb 14" />
-      <.status_label status={:station_closure} />
-      <.status_label status={:delay} />
+      <.status_label status={:normal} description="Normal Service" />
+      <.status_label status={:no_scheduled_service} description="No Scheduled Service" />
+      <.status_label status={:shuttle} description="Shuttles" />
+      <.status_label status={:suspension} description="Suspension" />
+      <.status_label status={:shuttle} description="8:30pm: Shuttles" />
+      <.status_label status={:shuttle} description="Wed Feb 12 - Fri Feb 14: Shuttles" />
+      <.status_label status={:delay} description="Delay" />
+      <.status_label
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/live/system_status.ex
```diff
@@ -43,13 +43,19 @@ defmodule DotcomWeb.Live.SystemStatus do
...
-      <.status_label status={:normal} />
-      <.status_label status={:shuttle} />
-      <.status_label status={:shuttle} plural />
-      <.status_label status={:shuttle} prefix="8:30pm" />
-      <.status_label status={:shuttle} prefix="Wed Feb 12 - Fri Feb 14" />
-      <.status_label status={:station_closure} />
-      <.status_label status={:delay} />
+      <.status_label status={:normal} description="Normal Service" />
+      <.status_label status={:no_scheduled_service} description="No Scheduled Service" />
+      <.status_label status={:shuttle} description="Shuttles" />
```
joshlarson, line 48: Bikeshed time!

---

I definitely want the flexibility to provide whatever text, so that we can do things like "2 Delays / Cancellations", or "8:30PM: Suspension", or "2 Station Closures" but should we have `<.status_label />` provide a default description if one isn't provided, so that if you pass in `:shuttle` and nothing else, then it says "Shuttles"?

I was thinking favorably about that in the abstract, but I couldn't think of a nice way to make that work, given that every use case here has some way that it's a bit different (for instance, in subway status, we would have to override the description whenever there's a plural, unless we re-introduce `plural` as an argument, and the way that CR status handles counts means that we would be overriding that as well).

---

Another option would be to go back to having the `:status` control the text, and have all sorts of possible customizations, like `time`, `date_range`, `plural`, and `count` (if `count` is present, it'll say something like `2 Delays`, and if it's `nil`, it'll just say `Delays` or `Delay`). That feels overly elaborate, but maybe it's still a good idea?

---

I suppose another option could be to make more disciplined use of HEEx's ability to use slots, and build a little template - something like 👇 

```elixir
~H"""
<.status_label :let={description} status={effect}>
  2 {Inflex.pluralize(description)}
</.status_label>
"""
```

anthonyshull (approved):

PR #2581, opened 2025-06-03
Fixes the naive in effect today

anthonyshull:
I was stupidly just checking the start and stop to see if they were today. I should have been checking the range. This PR fixes that.

### DIFF
--- a/lib/dotcom_web/controllers/alert_controller.ex
+++ b/lib/dotcom_web/controllers/alert_controller.ex
@@ -3,7 +3,7 @@ defmodule DotcomWeb.AlertController do
 
   use DotcomWeb, :controller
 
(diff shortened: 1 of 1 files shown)


lib/dotcom/alerts.ex
```diff
@@ -90,7 +93,10 @@ defmodule Dotcom.Alerts do
   @spec in_effect_today?(Alerts.Alert.t()) :: boolean()
   def in_effect_today?(%Alerts.Alert{active_period: active_period}) do
```
thecristen, line 94: I almost think we might want to rename this `in_effect_current_service_day`

thecristen (approved):
yay!!

joshlarson (approved):
Very nice!

I wouldn't hate adding a test for this, but happy to have that happen as a fast-follow (I'm happy to add that test if needed 🙂)

PR #2593, opened 2025-06-09
chore: remove react_renderer

thecristen:
we've whittled away at its uses over the years so much, there're barely any of it left working now anyhow. we're removing `TransitNearMe`, and the `ProjectsPage`, `AdditionalLineInfo`, and `ScheduleFinder` configurations either aren't invoked or are invoked in a way that doesn't even work.


anthonyshull (approved):
Thank god.

PR #2595, opened 2025-06-09
chore: alternate of `Dotcom.Utils.Time` module cleanup

thecristen:
Was struggling to parse the changes in #2594 and then had an idea which felt, at least to me, a lot simpler to understand. @joshlarson what do you think? I did keep your test changes so I could verify it still works. :D


joshlarson:
It wounds me a little bit, because technically this solution won't work for years before the year 0, or after the year 10,000, but I guess we don't need to worry about service planning outside of a thousand year window 😭

PR #2609, opened 2025-06-18
feat: new autocomplete on projects page

thecristen:
#### Summary of changes

**Asana Ticket:** [SR/TNM | Remove location search from project page search](https://app.asana.com/1/15492006741476/project/555089885850811/task/1210583202080124?focus=true)
**Bonus Asana ticket:** [AutocompleteJS v1.0 | Projects page](https://app.asana.com/1/15492006741476/project/1205739908126608/task/1205902725457088?focus=true)

- adds a new `<.algolia_autocomplete />` config for searching only for projects, uses on `/projects` page

**Bonus changes**
- adds a new `<.algolia_autocomplete />` config for searching only for stops, uses on `/stops` page
- bumps the count of each to 10 so we can show plenty of results
- gives each a custom no results error message because why not
- pressing `Enter` not only takes you to the `/search` page, but adds URL parameters to scope the result down to projects or stops
- removes location searching from the `/search` page, which I probably should've done back in #2591 
- removes a bunch of no longer needed code due to removing the old search inputs from the `/projects` and `/stops` pages


joshlarson:
Following up on ☝️...

- The `/projects` search results discrepancy was caused by two things:
  - Non-prod uses a different index, with slightly different heuristics than prod.
  - Some of the results I was noticing as "discrepancy" were location results, whose removal was intentional!
- The `/stops` one is cosmetic and not a big deal.

lib/dotcom_web/templates/project/index.html.heex
```diff
@@ -19,21 +19,12 @@
...
 
 <div class="container">
-  {content_tag(
-    :div,
-    [
-      DotcomWeb.PartialView.render(
-        "_search_input.html",
-        Map.merge(assigns, %{
-          placeholder: "Enter keyword(s)",
-          aria_label: "Enter keywords",
```
joshlarson, line 29: Do we want to keep this aria label?

lib/dotcom_web/templates/stop/_search_bar.html.eex
```diff
@@ -1,15 +0,0 @@
-<div id="stop-search" class="hidden-no-js c-search-bar c-search-bar--embedded">
-  <%= DotcomWeb.PartialView.render("_search_input.html",
-        Map.merge(assigns, %{
-          placeholder: "Search for a station or location",
-          aria_label: "Search for a station or location",
```
joshlarson, line 5: Do we want to keep this aria label?

joshlarson (approved):
Two questions about aria labels, but otherwise LGTM!

PR #2619, opened 2025-06-24
fix: add host to check_origin

thecristen:
No ticket, just noticed a lot of `Could not check origin for Phoenix.Socket transport.` errors logged in prod.


thecristen:
I'm not seeing the error logged in dev-green's logs so I think I achieved what I wanted to

joshlarson (approved):

PR #2626, opened 2025-06-26
Remove node from final docker image

anthonyshull:
**Old** 942 sec to build 739 mb image
**New** 718 sec to build 648 mb image


thecristen (approved):
Hey this looks pretty good! I feel like the end result is easier to understand, too. Thanks for taking this on.

PR #2635, opened 2025-07-07
wip: update open_trip_planner_client

thecristen:
(will open a PR there for adding Gettext support)


thecristen:
I made a trip plan using the updated dependency and nothing bad happened!

PR #2644, opened 2025-07-15
Set up libre translate

anthonyshull:
```
%> docker run -it --rm -p 9999:5000 libretranslate/libretranslate --load-only en,es
```

Change the text in `lib/dotcom_web/controllers/test_controller.ex` to something new.

```
%> mix gettext.extract
%> mix gettext.translate
%> mix phx.server
```

```
%> curl http://localhost:4001/_test
%> curl http://localhost:4001/_test?locale=es
```


thecristen (approved):
I think this is nice! Some suggestions:
- Can we have the mix task give feedback if there are no changes detected? Basically something to cue the user that they forgot to run `mix gettext.extract` first.

PR #2649, opened 2025-07-16
Fix date range issue

anthonyshull:
Because the end of rating is given as a date and not a date_time, we have to add a day in order to get the max date to show up correctly. I could make this cleaner by creating service_date_time functions that take a date or date_time, but that seems like a lot of work for this one use case.

https://app.asana.com/1/15492006741476/project/555089885850811/task/1209757107608482?focus=true

### DIFF
--- a/lib/dotcom_web/components/trip_planner/input_form.ex
+++ b/lib/dotcom_web/components/trip_planner/input_form.ex
@@ -157,10 +157,10 @@ defmodule DotcomWeb.Components.TripPlanner.InputForm do
 
   defp datepicker_config do
     %{
-      default_date: ServiceDateTime.service_date(),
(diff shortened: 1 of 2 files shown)


lib/dotcom/utils/date_time.ex
```diff
@@ -29,6 +29,12 @@ defmodule Dotcom.Utils.DateTime do
   @impl Behaviour
   def now(), do: Timex.now(@timezone)
 
+  @doc """
+  Get the date in the set @timezone.
+  """
+  @impl Behaviour
+  def today(), do: Timex.today(@timezone)
+
```
anthonyshull, line 37: 📓 I didn't end up using this, but I figured it is convenient enough to keep in there.

PR #2653, opened 2025-07-22
Identify which templates are used when loading a route

anthonyshull:
```
%> iex -S mix phx.server
```
```
%> curl http://localhost:4001
```
```
iex> DotcomWeb.Translations.templates("/")
```

This will only run in `dev`, so it should be fine to merge and deploy without affecting anything.

For posterity, this is the code that writes the tracker into each template file.

```elixir
defmodule TranslationTracker do
  @base_directory "lib/dotcom_web/templates"

  def write_all_trackers() do
    Enum.each(get_all_templates(), &write_tracker/1)
  end

  defp get_all_templates(reference \\ @base_directory) do
    if File.dir?(reference) do
      File.ls!(reference)
      |> Enum.map(&(reference <> "/" <> &1))
      |> Enum.map(&get_all_templates/1)
      |> List.flatten()
    else
      reference
    end
  end

  defp write_tracker(reference) do
    current_content = File.read!(reference)
    tracked_content = "<% track_template() %>\n"

    new_content = tracked_content <> current_content

    File.write!(reference, new_content)
  end
end
```

### DIFF
--- a/lib/dotcom/application.ex
+++ b/lib/dotcom/application.ex
@@ -41,6 +41,13 @@ defmodule Dotcom.Application do
         else
           []
         end ++
+        if Application.get_env(:dotcom, :env) === :dev do
+          [
+            {DotcomWeb.Translations, %{}}
+          ]
+        else
+          []
+        end ++
         if Application.get_env(:dotcom, :start_data_processes) do
           [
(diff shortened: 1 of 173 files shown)


thecristen (approved):
Nice

PR #2657, opened 2025-07-23
Use custom terminologies in translations

anthonyshull:
No ticket. I just noticed that without custom terminologies, we would have some wonky translations or have to *not* wrap words we didn't want translated. This adds a layer to the translation task that checks our custom terminology and uses the manual translation. It *should* seamlessly slide that custom terminology back into a larger body of translated text.


thecristen (approved):
Seems good

PR #2661, opened 2025-07-24
Translation progress banner

anthonyshull:
Screenshot Capture - 2025-07-24 - 11-04-08

Adds a progress banner that indicates what language you're viewing the site in. Also adds some progress information when you run translations.


lib/dotcom_web/templates/layout/root.html.heex
```diff
@@ -66,6 +66,22 @@
   </head>
   <%= content_tag(:body, class: Dotcom.BodyTag.class_name(@conn)) do %>
     <div class="body-wrapper" id="body-wrapper">
+      <%= if !Application.get_env(:dotcom, :is_prod_env?) && Gettext.get_locale(Dotcom.Gettext) != "en" do %>
+        <% completion_percent = Mix.Tasks.Gettext.Translate.completion_percent() %>
+        <% completion_ratio = Mix.Tasks.Gettext.Translate.completion_ratio() %>
+        <% language =
+          Dotcom.Gettext |> Gettext.get_locale() |> Dotcom.Locales.locale() |> Map.get(:endonym) %>
```
thecristen, line 73: nit: Spitting out the current endonym could be a helper function

lib/dotcom_web/templates/layout/root.html.heex
```diff
@@ -66,6 +66,22 @@
   </head>
   <%= content_tag(:body, class: Dotcom.BodyTag.class_name(@conn)) do %>
     <div class="body-wrapper" id="body-wrapper">
+      <%= if !Application.get_env(:dotcom, :is_prod_env?) && Gettext.get_locale(Dotcom.Gettext) != "en" do %>
+        <% completion_percent = Mix.Tasks.Gettext.Translate.completion_percent() %>
+        <% completion_ratio = Mix.Tasks.Gettext.Translate.completion_ratio() %>
```
thecristen, line 71: nit: Feels like this UI shouldn't be the responsibility of the mix task.

thecristen (approved):

lib/dotcom_web/templates/layout/root.html.heex
anthonyshull: Yeah, I am also using it in the translation task. There just isn't a better place to put things that we don't plan on sticking around for long. That's why I put multiple modules in that task file.

anthonyshull: Also, if you want something to be processed at compile time, you have to define the function in a different module and then set it as a module attribute somewhere else. This we want to define at compile time because it won't change and we don't want to recheck the template files every time we load a page.

PR #2663, opened 2025-07-25
chore: Replace subway alerts page with a LiveView

joshlarson:
This has a number of benefits:
- Separates the subway and commuter rail alerts pages, which are already fairly separate from the other modes.
- Allows us to easily do live-updating (so that riders won't have to refresh the page when status or alerts change) as a follow-up
  - _Live-updating is not part of this PR, but LiveView is a requirement for it._
- Resolves part of the Asana ticket linked below.

---

**Asana Ticket:** [Update CR, subway alerts summary + tab pages to avoid time filtering](https://app.asana.com/1/15492006741476/project/555089885850811/task/1210732979952245?focus=true)
**Asana Subtask:** [Replace subway alerts page with a LiveView](https://app.asana.com/1/15492006741476/task/1210893459312442?focus=true)


lib/dotcom_web/live/subway_alerts.ex
```diff
@@ -0,0 +1,159 @@
...
+      |> Map.delete("alerts_timeframe")
+      |> URI.encode_query()
+
+    new_path =
+      uri
+      |> URI.parse()
+      |> set_query(new_query)
+      |> to_path()
+
+    {:noreply, socket |> push_patch(to: new_path)}
```
thecristen, line 56: I'm kinda confused that you have to mess with encoding/parsing at all here, would something like this work?

```elixir
new_path = live_path(socket, __MODULE__, Map.drop(params, ["alerts_timeframe"]))
{:noreply, push_patch(socket, to: new_path)}
```

lib/dotcom_web/live/subway_alerts.ex
```diff
@@ -0,0 +1,159 @@
...
+     )
+     |> assign(:breadcrumbs, [Breadcrumb.build("Alerts")])
+     |> assign_now()
+     |> assign_route_type()
+     |> assign_routes()
+     |> assign(:subway_status, Dotcom.SystemStatus.subway_status())
+     |> assign(:disruption_groups, Disruptions.future_disruptions())
+     |> assign_alerts()
+     |> assign_banner_alert()
+     |> assign_alert_groups()}
```
thecristen, line 41: Do we need _all_ of these as separate steps in the pipeline here? I'd recommended either consolidating the ones that are dependent on each other, or extract ones that'd be useful in other views into a separate [on_mount hook](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#on_mount/1) (or perhaps some combination of both).

lib/dotcom_web/live/subway_alerts.ex
joshlarson: [Ah yeah, that is much better!](https://github.com/mbta/dotcom/pull/2663/commits/15457054b7a992f81068b2bf9270ae7cffe7f9e6)

I tricked myself into thinking that `live_path` had been removed, because I couldn't figure out where it was defined. Looking again, I found it!

lib/dotcom_web/components/t_alerts.ex
```diff
@@ -0,0 +1,39 @@
...
+  use DotcomWeb, :component
+
+  def t_alerts(assigns) do
+    ~H"""
+    <div class="m-alerts-ad">
+      <h3>Get T-Alerts</h3>
+      <div class="m-alerts-ad__content">
+        <div class="m-alerts-ad__buttons">
+          <p>
+            Tell us about your regular trips and receive text or email alerts that are relevant to you, at times when you want them.
```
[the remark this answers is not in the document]
joshlarson: [Done](https://github.com/mbta/dotcom/pull/2663/commits/4345191a56453201f19d5d14b918704a04660612) and [done](https://github.com/mbta/dotcom/pull/2663/commits/57d930d2a4e15f6ee8e81ff7aea77b8c792a500a)!

Thank goodness for that handy translation guide!

test/dotcom_web/live/subway_alerts_test.exs
```diff
@@ -0,0 +1,169 @@
...
+      refute rendered_section
+             |> Floki.find("[data-test=\"route_pill:#{route_id}\"]")
+             |> Enum.empty?()
+    end
+
+    test "does not put service-impacting alerts in the station_and_service section", %{conn: conn} do
+      route_id = Faker.Util.pick(Subway.lines())
+      route_type = Faker.Util.pick([0, 1])
+
+      stub(Alerts.Repo.Mock, :all, fn date ->
```
anthonyshull, line 97: Is there a reason you're stubbing these in the test instead of mocking them? With the `verify_on_exit!` above, I'm pretty sure it verifies mocks. Stubs don't get verified. The only reason I think we would want to use stubs in a test is if we don't know how many times the mock will be called.

joshlarson: LiveView actually calls `mount` and `render` twice, once when the HTML is loaded and once again when the socket connection is established, so `expect` fails because the mocks are called twice. I could assert that the mocks are called twice, but I didn't for two reasons:
- It's a little confusing to see a random "this should be called twice", or "2 times however many times you actually expect it to be called" in the middle of an otherwise unrelated test.
- I wasn't 100% confident that the number of calls will always reliably be 2, and I wasn't sure it was worth the effort to confirm one way or the other.


I don't feel that strongly about it though - happy to switch it to `expect(yadda yadda, 2, :blah)` and then figure something else out if it turns out to be flaky or overly confusing.

test/dotcom_web/live/subway_alerts_test.exs
```diff
@@ -0,0 +1,169 @@
+defmodule DotcomWeb.Live.SubwayAlertsTest do
+  use DotcomWeb.ConnCase, async: true
+
+  import Dotcom.Alerts, only: [service_impacting_effects: 0]
+  import DotcomWeb.Router.Helpers, only: [live_path: 2]
+  import Mox
+  import Phoenix.LiveViewTest
+
+  alias Test.Support.FactoryHelpers
```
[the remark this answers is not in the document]
joshlarson: [Done!](https://github.com/mbta/dotcom/pull/2663/commits/c3ae13330e2d23f48bf07486a1603b59025d5676)

lib/dotcom_web/live/subway_alerts.ex
```diff
@@ -0,0 +1,137 @@
...
+    Enum.reject(alerts, &(&1.id == alert_banner.id))
+  end
+
+  defp excluding_banner(_, alerts), do: alerts
+
+  def render(assigns) do
+    ~H"""
+    <.alerts_layout mode={:subway}>
+      <:sidebar_bottom>
+        <.t_alerts />
```
thecristen, line 111: **suggestion**: reuse the existing component instead of making a new copy, I think this should work
```suggestion
        {DotcomWeb.AlertView.render("_t-alerts.html")}
```

lib/dotcom_web/live/subway_alerts.ex
```diff
@@ -0,0 +1,137 @@
...
+      routes
+      |> attach_alert_lists(non_banner_alerts)
+      |> drop_empty_groups()
+    )
+  end
+
+  defp assign_alerts(%{assigns: %{date_time: date_time}} = socket) do
+    alerts = @alerts_repo.all(date_time) |> Enum.reject(&service_impacting_alert?/1)
+
+    socket |> assign(:alerts, alerts)
```
thecristen, line 68: **question**: Given the whole this-code-runs-twice thing as per discussion in the test cases, I wonder if this (and others) should use `assign_new/3` instead of `assign/3`, such that it only runs if it's not already in the socket ([doc](hexdocs.pm/phoenix_live_view/Phoenix.Component.html#assign_new/3)). 🤔

joshlarson: I'm still open to it, but it doesn't actually solve the runs-twice thing, because the first invocation is before the socket is connected, which means that it still has to run twice.

It would (I think) still have the effect of causing the various functions to only be called twice, rather than twice plus however many additional times `render` is called, but that might actually be a bug rather than a feature, because it could get in the way of live-updating.

lib/dotcom_web/live/subway_alerts.ex
```diff
@@ -0,0 +1,115 @@
...
+    |> drop_empty_groups()
+  end
+
+  defp with_alert_lists(routes, alerts) do
+    routes
+    |> Enum.map(fn route ->
+      entity = %InformedEntity{
+        route_type: route.type,
+        route: route.id
+      }
```
thecristen, line 73: suggestion: there's a function for that :D

```suggestion
      entity = InformedEntity.from_keywords(
         route_type: route.type,
         route: route.id
      )
```

lib/dotcom_web/live/subway_alerts.ex
```diff
@@ -0,0 +1,115 @@
...
+  @date_time Application.compile_env!(:dotcom, :date_time_module)
+
+  embed_templates "layouts/*"
+
+  def mount(_params, _session, socket) do
+    {:ok,
+     socket
+     |> assign(
+       :meta_description,
+       "Live service alerts for all MBTA transportation modes, including subway, bus, Commuter Rail, and ferry. " <>
```
thecristen, line 30: don't forget to translate this too!

lib/dotcom_web/live/subway_alerts.ex
```diff
@@ -0,0 +1,115 @@
...
+         "Updates on delays, construction, elevator outages, and more."
+     )
+     |> assign(:breadcrumbs, [Breadcrumb.build("Alerts")])
```
thecristen, line 33: another translation opportunity

lib/dotcom_web/router.ex
```diff
@@ -279,6 +288,12 @@ defmodule DotcomWeb.Router do
     end
   end
 
+  scope "/", DotcomWeb do
+    pipe_through([:browser, :browser_live])
+
+    live "/demo", Live.Demo
+  end
+
```
thecristen, line 296: Can you remove this please?

test/dotcom_web/live/subway_alerts_test.exs
```diff
@@ -0,0 +1,165 @@
...
+  setup _ do
+    stub(Dotcom.Utils.DateTime.Mock, :coerce_ambiguous_date_time, fn date_time ->
+      Dotcom.Utils.DateTime.coerce_ambiguous_date_time(date_time)
+    end)
+
+    stub(Dotcom.Utils.DateTime.Mock, :now, fn ->
+      Dotcom.Utils.DateTime.now()
+    end)
+
+    stub(Alerts.Repo.Mock, :all, fn _ -> [] end)
```
thecristen, line 32: I think all your tests already stub this, so you might not need it here

lib/dotcom_web/live/subway_alerts.ex
joshlarson: [Cool!](https://github.com/mbta/dotcom/pull/2663/commits/ec95e52b7bfd7dba68b15f0c3886eabb5262532a)

lib/dotcom_web/live/subway_alerts.ex
joshlarson: [Done](https://github.com/mbta/dotcom/pull/2663/commits/1e0cc0631f775ad915bc2f425059f4b7da1c26bf)

lib/dotcom_web/live/subway_alerts.ex
joshlarson: [Done](https://github.com/mbta/dotcom/pull/2663/commits/1e0cc0631f775ad915bc2f425059f4b7da1c26bf)

lib/dotcom_web/router.ex
joshlarson: [Woops](https://github.com/mbta/dotcom/pull/2663/commits/13f238371dc720192e64a079fab6eef5e1e8aa5a)

thecristen (approved):

PR #2669, opened 2025-07-30
Tracks template usage so we can remove unused templates

anthonyshull:
Converts the local template tracker to obtain daily template usage. Any template that exists but is never called will be logged.

We will want to add the tracker back in to every template file:

```elixir
defmodule TranslationTracker do

  import Dotcom.Utils.File, only: [list_all_files: 0]

  @base_directory "lib/dotcom_web/templates"

  def write_all_trackers() do
    Enum.each(get_all_templates(), &write_tracker/1)
  end

  defp get_all_templates() do
    @base_directory
    |> list_all_files()
    |> Enum.filter(fn ref -> String.match?(ref, ~r/\eex$/) end)
  end
  
  defp write_tracker(reference) do
    current_content = File.read!(reference)
    tracked_content = "<% track_template() %>\n"

    new_content = tracked_content <> current_content

    File.write!(reference, new_content)
  end
end
```


thecristen (approved):
Why not? 👯

PR #2674, opened 2025-07-31
cleanup: Don't mock pure `DateTime` functions

joshlarson:
I noticed that every time `coerce_ambiguous_date_time` appears in the call path, that requires a stub like
```elixir
stub(Dotcom.Utils.DateTime.Mock, :coerce_ambiguous_date_time, fn date_time ->
  Dotcom.Utils.DateTime.coerce_ambiguous_date_time(date_time)
end)
```

...which is just noise - `coerce_ambiguous_date_time` is a pure function, and there's no need to stub it to its own implementation.

This removes them from the behavior and just sets them up as pure functions, so that we don't need to stub them in seemingly random tests.

### DIFF
--- a/lib/dotcom/utils/date_time.ex
+++ b/lib/dotcom/utils/date_time.ex
@@ -68,7 +68,10 @@ defmodule Dotcom.Utils.DateTime do
   @doc """
   Given a date_time_range and a date_time, returns true if the date_time is within the date_time_range.
   """
-  @impl Behaviour
+  @spec in_range?(
+          Dotcom.Utils.DateTime.date_time_range(),
+          DateTime.t()
+        ) :: boolean
   def in_range?({nil, nil}, _), do: false
 
   def in_range?({nil, %DateTime{} = stop}, %DateTime{} = date_time) do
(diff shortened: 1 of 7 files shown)


lib/dotcom/utils/date_time.ex
```diff
@@ -48,8 +48,12 @@ defmodule Dotcom.Utils.DateTime do
...
   If we are given something tha tis not a DateTime, AmbiguousDateTime, or an error tuple, we log the input and return `now`.
   """
-  @impl Behaviour
-  def coerce_ambiguous_date_time(%DateTime{} = date_time), do: date_time
+  @spec coerce_ambiguous_date_time(
+          DateTime.t()
+          | Timex.AmbiguousDateTime.t()
+          | {:error, term()}
+        ) :: DateTime.t()
+  def(coerce_ambiguous_date_time(%DateTime{} = date_time), do: date_time)
```
thecristen, line 56: nit: these extra parens are funny!

thecristen (approved):

PR #2682, opened 2025-08-05
Tracks functions by way of a decorator

anthonyshull:


lib/dotcom_web/controllers/mode/hub_behavior.ex
```diff
@@ -1,7 +1,28 @@
 defmodule DotcomWeb.Mode.HubBehavior do
   @moduledoc "Behavior for mode hub pages."
 
-  use DotcomWeb, :controller
+  use Dotcom.Gettext.Sigils
```
anthonyshull, line 4: 📓 I just moved what was in `DotcomWeb, :controller` into this module because it broke when trying to add the decorator.

deploy/dotcom/prod/Dockerfile
```diff
@@ -46,6 +46,7 @@ RUN apk add curl
 RUN apk add libssl3
 
 COPY --from=builder /home/_build /home/_build
+COPY --from=builder /tmp/.functions /tmp/.functions
```
anthonyshull, line 49: 📓 We have to build the list of functions at compile time. We use the file in the `tmp` directory so that it survives the image build process.

lib/dotcom_web/controllers/mode/hub_behavior.ex
thecristen: The mode hub pages are `/schedules/(bus|ferry|commuter-rail|subway)`, each using a separate controller which implements this behavior!

PR #2684, opened 2025-08-06
Fix small translation issues

anthonyshull:
Fixes three issues. One, html tags were creating breaking translations. I fixed that by doing a first pass to change all double quotes to single quotes. Two, we were still reporting completion ratio of template translation based on the tracking tag. We are done translating templates and the tag has been repurposed. So, I just removed all of that code. Three, we don't want to run the gettext merge command. So, I wrote a module with the same name that tells you to use the translate task.

### DIFF
--- a/lib/mix/tasks/gettext/translate.ex
+++ b/lib/mix/tasks/gettext/translate.ex
@@ -1,40 +1,3 @@
-defmodule Mix.Tasks.GetText.Translate.Progress do
-  @moduledoc """
-  A support module for tracking progress of our translation efforts.
-  """
-
-  import Dotcom.Utils.File, only: [list_all_files: 1]
-
-  @directory "lib/dotcom_web/templates"
-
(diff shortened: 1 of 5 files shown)


lib/mix/tasks/gettext/translate.ex
```diff
@@ -169,8 +107,11 @@ defmodule Mix.Tasks.Gettext.Translate do
   end
 
   # Gets the content of msgid of a `.pot` file.
+  # Replaces " with ' so that the output doesn't break.
   defp domain_line(line) do
-    Regex.scan(~r/(?<=\")(.*?)(?=\")/, line)
+    text = String.replace(line, ~r/\\\"/, "'")
```
anthonyshull, line 112: 📓 This fixes the translation issue. In our first pass, we replace all instances of `\"` with `'`.

PR #2689, opened 2025-08-12
remove unnecessary fetching of route pdfs

anthonyshull:
I just noticed while looking at cache values that all of these PDFs were outdated. Turns out, we don't even use them; we filter them all out. This removes the code that loads them.


thecristen (approved):
Huh, neat!!!

PR #2690, opened 2025-08-13
Replace fare cards with links

anthonyshull:
Screenshot Capture - 2025-08-13 - 09-12-52


lib/dotcom_web/templates/page/_find_a_location.html.heex
```diff
@@ -7,7 +7,7 @@
   </div>
```
thecristen, line 7: you probably want to remove `track_template()` from this file too.

thecristen (approved):
Looks nice!

PR #2693, opened 2025-08-14
feat: localized trip planning strings

thecristen:
#### Summary of changes

**Asana Ticket:** [routing error messages](https://app.asana.com/1/15492006741476/project/1210709545912061/task/1210710269085151?focus=true)
**Asana Ticket:** [itinerary group alternatives text](https://app.asana.com/1/15492006741476/project/555089885850811/task/1210710269085153?focus=true)
**Asana Ticket:** [walk_summary output](https://app.asana.com/1/15492006741476/project/555089885850811/task/1210710269085149?focus=true)

The related PR is needed for changing the error messages:
- https://github.com/mbta/open_trip_planner_client/pull/19


config/deps/deps.exs
```diff
@@ -21,5 +21,4 @@ if config_env() == :test do
 end
 
 config :open_trip_planner_client,
-  timezone: "America/New_York",
-  fallback_error_message: "Please try again or send us feedback at mbta.com/customer-support"
+  timezone: "America/New_York"
```
anthonyshull, line 24: Should actually get this from the config we already set: https://github.com/mbta/dotcom/blob/65bd4ecfd5145b605b7a96997da36e96938d8e4c/config/config.exs#L46

anthonyshull: Probably worth looking for everywhere that string is used and always use the same config.

PR #2695, opened 2025-08-14
Homepage Image from CMS

amaisano:
#### Summary of changes

**Asana Ticket:** [TICKET_NAME](TICKET_LINK)

---

#### General checks

### DIFF
--- a/lib/cms/repo.ex
+++ b/lib/cms/repo.ex
@@ -17,6 +17,7 @@ defmodule CMS.Repo do
   alias CMS.Partial.{
     Banner,
     Paragraph,
+    Photo,
     RoutePdf,
     Teaser,
     WhatsHappeningItem
@@ -130,6 +131,27 @@ defmodule CMS.Repo do
   def do_banner do
     case @cms_api.view("/cms/important-notices", []) do
       {:ok, [api_data | _]} -> Banner.from_api(api_data)
+      {:ok, _} -> :empty
+      {:error, _} -> :error
+    end
+  end
+
+  @spec photo() :: Photo.t() | nil
+  def photo do
+    cached_value = do_photo()
+
+    if cached_value == :empty || cached_value == :error, do: nil, else: cached_value
+  end
+
+  @decorate cacheable(
+              cache: @cache,
+              key: "cms.repo|photos",
+              on_error: :nothing,
+              opts: [ttl: @ttl]
+            )
+  def do_photo do
+    case @cms_api.view("/cms/photos", []) do
+      {:ok, [api_data | _]} -> Photo.from_api(api_data)
       {:ok, _} -> :empty
       {:error, _} -> :error
     end
--- a/lib/dotcom_web/controllers/page_controller.ex
+++ b/lib/dotcom_web/controllers/page_controller.ex
@@ -9,6 +9,7 @@ defmodule DotcomWeb.PageController do
   alias CMS.Partial.{
     Banner,
     Paragraph,
+    Photo,
     Teaser,
     WhatsHappeningItem
   }
@@ -23,6 +24,7 @@ defmodule DotcomWeb.PageController do
   def index(conn, _params) do
     {promoted, remainder} = whats_happening_items()
     banner = banner()
+    photo = photo()
     fares = fares(conn.query_params)
 
     conn
@@ -33,6 +35,7 @@ defmodule DotcomWeb.PageController do
     )
     |> async_assign_default(:news, &news/0, [])
     |> async_assign_default(:banner, fn -> banner end)
+    |> async_assign_default(:photo, fn -> photo end)
     |> async_assign_default(:homepage_fares, fn -> fares end)
     |> async_assign_default(:promoted_items, fn -> promoted end)
     |> async_assign_default(:whats_happening_items, fn -> remainder end)
@@ -72,6 +75,14 @@ defmodule DotcomWeb.PageController do
     case Repo.banner() do
       nil -> nil
       banner -> add_utm_url(banner)
+    end
+  end
+
+  @spec photo :: Photo.t() | nil
+  defp photo do
+    case Repo.photo() do
+      nil -> nil
+      photo -> photo
     end
   end
 
--- a/lib/dotcom_web/templates/page/_photo.html.heex
+++ b/lib/dotcom_web/templates/page/_photo.html.heex
@@ -1,0 +1,1 @@
+<img src={"#{ @photo.image.url }"} class="h-120 w-full object-cover max-sm:hidden -z-10 relative" />
--- a/lib/dotcom_web/templates/page/index.html.heex
+++ b/lib/dotcom_web/templates/page/index.html.heex
@@ -1,5 +1,6 @@
 <div class="index">
   <h1 class="sr-only">{~t(MBTA Homepage)}</h1>
+  <img src="/images/header-image-ol-green-st.jpg" class="h-120 w-full object-cover max-sm:hidden -z-10 relative" />
   {render("_tabbed_nav.html", assigns)}
   <div class="container page-section">
     {render("_find_a_location.html", conn: @conn)}
--- a/lib/dotcom_web/templates/page/index.html.heex
+++ b/lib/dotcom_web/templates/page/index.html.heex
@@ -1,6 +1,8 @@
 <div class="index">
   <h1 class="sr-only">{~t(MBTA Homepage)}</h1>
-  <img src="/images/header-image-ol-green-st.jpg" class="h-120 w-full object-cover max-sm:hidden -z-10 relative" />
+  {if(assigns[:photo],
+    do: render(DotcomWeb.PageView, "_photo.html", conn: @conn, photo: @photo)
+  )}
   {render("_tabbed_nav.html", assigns)}
   <div class="container page-section">
     {render("_find_a_location.html", conn: @conn)}
--- a/assets/css/_tabbed-nav.scss
+++ b/assets/css/_tabbed-nav.scss
@@ -2,20 +2,26 @@
   @include make-container();
   @include make-container-max-widths();
   background-color: $brand-primary;
+  border-top-left-radius: $large-border-radius;
+  border-top-right-radius: $large-border-radius;
+  box-shadow: 0px -4px 4px 0px rgb(0 0 0 / 0.25);
   display: block;
-  margin-top: $space-32;
+  margin-top: -$space-120;
   padding-left: 0;
   padding-right: 0;
 
   @include media-breakpoint-down(sm) {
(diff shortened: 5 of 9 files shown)


lib/cms/repo.ex
```diff
@@ -135,6 +136,27 @@ defmodule CMS.Repo do
...
+  @decorate cacheable(
+              cache: @cache,
+              key: "cms.repo|photos",
+              on_error: :nothing,
+              opts: [ttl: @ttl]
+            )
+  def do_photo do
+    case @cms_api.view("/cms/photos", []) do
+      {:ok, [api_data | _]} -> Photo.from_api(api_data)
+      {:ok, _} -> :empty
```
thecristen, line 155: oh interesting, I hadn't quite realized the banner follows this pattern. 🤔 

`:empty` will be cached, is that what we want? Alternatively we could return `nil` (which won't be cached - this is governed by the default match function in the cacheable decorator function ([doc](https://hexdocs.pm/nebulex/Nebulex.Caching.Decorators.html#module-shared-options))).

lib/dotcom_web/templates/page/_photo.html.heex
```diff
@@ -0,0 +1 @@
+<img src={"#{ @photo.image.url }"} class="h-120 w-full object-cover max-sm:hidden -z-10 relative" />
```
thecristen, line 1: suggestion: If we're not using alt text here, we can add `alt=""` to indicate to screen readers that this image is decorative.

thecristen, line 1: nit: `src={@photo.image.url}` should work too :)

lib/dotcom_web/templates/page/index.html.heex
```diff
@@ -1,5 +1,8 @@
 <div class="index">
   <h1 class="sr-only">{~t(MBTA Homepage)}</h1>
+  {if(assigns[:photo],
+    do: render(DotcomWeb.PageView, "_photo.html", conn: @conn, photo: @photo)
+  )}
```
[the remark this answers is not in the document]
amaisano: I did move the conditional to the template, but it won't be as clean. I want to avoid spitting out the style block (and everything really) if the photo is absent, so I enclosed it all inside a classical `<%= if...`.

lib/cms/repo.ex
```diff
@@ -130,7 +131,28 @@ defmodule CMS.Repo do
...
+              cache: @cache,
+              key: "cms.repo|photos",
+              on_error: :nothing,
+              opts: [ttl: @ttl]
+            )
+  def do_photo do
+    case @cms_api.view("/cms/photos", []) do
+      {:ok, [api_data | _]} -> Photo.from_api(api_data)
+      {:ok, _} -> nil
       {:error, _} -> :error
```
anthonyshull, line 156: I feel like there is a better way to handle this. We are using :error and :empty and nil when I think we can limit it to two different responses. Either {:ok, value} or {:error, reason}.

lib/dotcom_web/controllers/page_controller.ex
```diff
@@ -23,6 +23,7 @@ defmodule DotcomWeb.PageController do
   def index(conn, _params) do
     {promoted, remainder} = whats_happening_items()
     banner = banner()
+    photo = Repo.photo()
```
anthonyshull, line 26: For both of these, unless we're actually doing something with the variable, they can just be inlined in the assigns.

```
import CMS.Repo, only: [photo: 0]

|> async_assign_default(:photo, &photo/0)
```

priv/cms/cms/photos.json
```diff
@@ -0,0 +1,131 @@
...
+      }
+    ],
+    "field_vanity_photo": [
+      {
+        "url": "https://live.pantheonsite.io/sites/default/files/styles/vanity_photo/public/photos/2025-08/2025-06-24-jackson-square-crews-prepare-platform.jpg?itok=fhWzrEeu",
+        "alt": ""
+      }
+    ]
+  }
+]
```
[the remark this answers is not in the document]
amaisano: What linters are you using? I have a bunch and also used mix format. I might be missing other commands, but nothing like this has been tripping warnings locally.

test/cms/partial/photo_test.exs
```diff
@@ -0,0 +1,29 @@
+defmodule CMS.Partial.PhotoTest do
```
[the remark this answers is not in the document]
amaisano: I will need more guidance on this - not sure if we should block on it right now?

priv/cms/cms/photos.json
```diff
@@ -0,0 +1,131 @@
+[
```
anthonyshull, line 1: Shouldn't need this file when using Mox.

lib/cms/api/static.ex
```diff
@@ -21,6 +21,10 @@ defmodule CMS.Api.Static do
     parse_json("cms/banners.json")
   end
 
+  def photos_response do
```
anthonyshull, line 24: Shouldn't need this when using Mox.

lib/cms/api/static.ex
```diff
@@ -310,6 +314,10 @@ defmodule CMS.Api.Static do
     {:ok, banners_response()}
   end
 
+  def view("/cms/photos", _) do
```
anthonyshull, line 317: Same, shouldn't need with Mox.

test/cms/partial/photo_test.exs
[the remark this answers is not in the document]
anthonyshull: This is a good example showing Mox: test/dotcom/trip_plan/open_trip_planner_test.exs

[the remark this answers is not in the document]
joshlarson: Just going to flag that `Mox` and the fixture problem are orthogonal for the tests in this file as they currently are. Simply switching to `Mox` would involve something like:
```elixir
test "it parses the API response into a CMS.Photo struct", %{api_notices: [raw | _]} do
  expect(CMS.Api.Mock, :view, fn "/cms/photos", [] ->
    {:ok, raw}
  end)

  assert %Photo{
           image: %Image{url: url, alt: alt}
         } = CMS.Repo.photo()

  assert url =~
           "/sites/default/files/styles/vanity_photo/public/photos/2025-08/2025-06-24-jackson-square-crews-prepare-platform.jpg?itok=fhWzrEeu"

  assert alt == ""
end
```

...which still uses the fixture. 

This still may have some benefits, like allowing us to test the other parts of the `case @cms_api.view("/cms/photos", [])` statement, but also (this is a pro or a con, depending on how you look at this kind of thing) this would bump the test coverage up to a higher layer - the `CMS.Repo` layer, rather than the `CMS.Partial.Photo` layer - which means that we'd want to rename the test module and move the file as well.

If we want to get rid of the fixture, we can do that just as easily by doing something like:

```elixir
test "it parses the API response into a CMS.Photo struct" do
  expected_image_url = Faker.Internet.image_url()
  expected_alt_text = Faker.Lorem.words(10)

  raw =
    %{
      "field_vanity_photo" => [
        %{
          "alt" => expected_alt_text,
          "url" => expected_image_url
        }
      ]
    }

  assert %Photo{
           image: %Image{url: actual_image_url, alt: actual_alt_text}
         } = Photo.from_api(raw)

  assert actual_image_url == expected_image_url
  assert actual_alt_text == expected_alt_text
end
``` 

...which does have the advantage of co-locating the image URL and alt text definitions with their assertions, and also doesn't use Mox.

(I am pro-this, because I don't like having to dig through fixtures to figure out where `itok=fhWzrEeu` or an empty-string alt-text came from, although when we paired on it earlier, Adamo did point out that the fixtures have the advantage of being easily-generated from Drupal directly, which means that if the response format changes for whatever reason, it'll be pretty easy to get the code and tests up-to-date with Drupal-reality.)

---

TL;DR - It's not quite as simple as just replacing the fixture with Mox 😅, unless Adamo and I misunderstood what you were actually suggesting. Combining that with the fact that the comment isn't marked as blocking, and the fact that this PR is time-sensitive (must be merged on Monday morning, I think), I vote that we deal with this as a post-merge follow-up when we can all talk about it in a more leisurely way.

(That'll also give us breathing room to figure out an effective way to get rid of the fixtures not just for `photos.json`, but for the others too, since it would be kind of confusing to have multiple patterns for how to mock/stub/fake CMS data in tests.)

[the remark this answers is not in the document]
joshlarson: Oh, also as it turns out, the CMS API Mock isn't already configured quite right to just drop in - [there is a mock defined](https://github.com/mbta/dotcom/blob/957682cb7a5334193dc809700edcdf68a4066a31/test/support/mocks.ex#L9) for `CMS.Api.Mock`, but [the test environment is currently configured](https://github.com/mbta/dotcom/blob/957682cb7a5334193dc809700edcdf68a4066a31/config/test.exs#L18) to use `CMS.Api.Static` as its `cms_api_module`, which means that in order to make this change, we'd need to change pretty much every test that interacts with the CMS API.

I think it's likely that that's worth doing, but we're not going to get that done before Monday, so I'm doubling down on my "let's do this as a follow-up".

[Asana ticket](https://app.asana.com/1/15492006741476/project/385363666817452/task/1211126421107450?focus=true)

lib/cms/repo.ex
```diff
@@ -130,7 +131,28 @@ defmodule CMS.Repo do
...
+      {:ok, _} -> nil
+      {:error, _} -> :error
+    end
+  end
+
+  @spec photo() :: Photo.t() | nil
+  def photo do
+    cached_value = do_photo()
+
+    if cached_value == :empty || cached_value == :error, do: nil, else: cached_value
```
joshlarson, line 143: This may be made obsolete depending on how you want to address [this comment](https://github.com/mbta/dotcom/pull/2695/files#r2291920948), but if the implementation of `do_photo()` is kept as it currently is, then `cached_value` can never be equal to `:empty`

lib/cms/repo.ex
joshlarson: Pulling this back out of resolved 😁

Per [this comment](https://github.com/mbta/dotcom/pull/2695#discussion_r2294828860), I think we actually _do_ want `:empty` to be cached, so that we're not bombarding the CMS re-requesting the image if there isn't one.

I think not caching `nil`/`:error`/etc is supposed to avoid caching things like connectivity errors or timeouts, but `:empty` in this context is a successful response - it's the CMS saying "there's no image right now".

thecristen: Now reverted back to what it was in the initial PR, thanks for raising this Josh!

thecristen (approved):
While there's plenty of refactor opportunity around these parts, we've got enough here to have this ready to release.

PR #2706, opened 2025-08-28
Combine legs in loop routes

anthonyshull:
https://app.asana.com/1/15492006741476/project/555089885850811/task/1208039959742771?focus=true


lib/dotcom/trip_plan/helpers.ex
```diff
@@ -18,7 +18,6 @@ defmodule Dotcom.TripPlan.Helpers do
   For an OpenTripPlanner stop, return the MBTA commuter rail zone
   """
   @spec mbta_zone_id(Stop.t()) :: String.t() | nil
-  def mbta_zone_id(%Stop{zone_id: "CR-zone-" <> zone_id}), do: zone_id
```
anthonyshull, line 21: ❓ This seems to have changed in the trip planner client?

thecristen (changes requested):
Let's revert the `mbta_zone_id` change, unless there's something else going on there

PR #2715, opened 2025-09-04
Translate planned disruptions

anthonyshull:
https://app.asana.com/1/15492006741476/project/555089885850811/task/1210709781286118?focus=true

Fixed the libretranslate docker compose so that works now too.

### DIFF
--- a/priv/gettext/dotcom.pot
+++ b/priv/gettext/dotcom.pot
@@ -3069,3 +3069,23 @@ msgstr ""
 #, elixir-autogen, elixir-format
 msgid "between %{first} and %{last}"
 msgstr ""
+
+#: lib/dotcom_web/components/planned_disruptions.ex:106
+#, elixir-autogen, elixir-format
+msgid "%{date} until further notice"
+msgstr ""
+
+#: lib/dotcom_web/components/planned_disruptions.ex:38
+#, elixir-autogen, elixir-format
+msgid "Planned Work"
+msgstr ""
+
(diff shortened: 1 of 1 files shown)


lib/dotcom_web/components/planned_disruptions.ex
```diff
@@ -139,7 +142,7 @@ defmodule DotcomWeb.Components.PlannedDisruptions do
     if Timex.after?(service_date_datetime, service_date_today) do
       service_date_datetime |> Timex.format!("%a, %b %-d", :strftime)
```
joshlarson, line 143: We probably need to do something about this too. For example, `Hoy – Thu, Sep 11: Shuttles` should probably be `Hoy – Jue, Sep 11: Shuttles`

joshlarson (dismissed):

joshlarson (dismissed):
Let's use [Clrd Dates & Times](https://hexdocs.pm/ex_cldr_dates_times/readme.html) to translate things like `Thu, Sep 11` as well, since that's also part of this component.

lib/dotcom_web/components/planned_disruptions.ex
thecristen: Yeah, I've localized a few things for the trip planner, mostly in https://github.com/mbta/dotcom/blob/main/lib/dotcom_web/components/trip_planner/helpers.ex... can use custom format a la [this example](https://github.com/mbta/dotcom/blob/main/lib/dotcom_web/components/trip_planner/helpers.ex#L170).

PR #2716, opened 2025-09-04
Translate status row heading

anthonyshull:
https://app.asana.com/1/15492006741476/project/555089885850811/task/1210709781286124?focus=true

### DIFF
--- a/lib/dotcom_web/components/system_status/status_row_heading.ex
+++ b/lib/dotcom_web/components/system_status/status_row_heading.ex
@@ -4,6 +4,7 @@ defmodule DotcomWeb.Components.SystemStatus.StatusRowHeading do
   optional prefix and a route pill for the given route ID's.
   """
 
+  use Dotcom.Gettext.Sigils
   use DotcomWeb, :component
 
   import DotcomWeb.Components.RouteSymbols, only: [subway_route_pill: 1]
@@ -105,7 +106,7 @@ defmodule DotcomWeb.Components.SystemStatus.StatusRowHeading do
   defp decorations(%{status: :delay, alerts: alerts}) do
     all_single_tracking? = alerts |> Enum.all?(&(&1.cause == :single_tracking))
 
-    subheading_text = if all_single_tracking?, do: "Due to Single Tracking"
+    subheading_text = if all_single_tracking?, do: ~t"Due to Single Tracking"
 
     %{
       subheading_text: subheading_text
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/components/system_status/status_row_heading.ex
```diff
@@ -127,17 +128,17 @@ defmodule DotcomWeb.Components.SystemStatus.StatusRowHeading do
...
-  # Special case for delays - when displayed with a future date, say 
+  # Special case for delays - when displayed with a future date, say
   # "Expect Delay" (or Expect Delays) rather than simply "Delay". A
   # prefix of "Now" should still display as "Delay", rather than
   # "Expect Delay".
-  defp description(:delay, "Now"), do: "Delays"
-  defp description(:delay, prefix) when is_binary(prefix), do: "Expect Delays"
-  defp description(:delay, _), do: "Delays"
-  defp description(:shuttle, _), do: "Shuttles"
+  defp description(:delay, "Now"), do: ~t"Delays"
```
joshlarson, line 138: Once we translate [the string "Now"](https://github.com/mbta/dotcom/blob/9c911bf366b98cda16693ac2eadd3feed1c839ff/lib/dotcom_web/components/system_status/subway_status.ex#L312), this match won't work anymore outside of English, which means that translations will look like `Now: Expect Delays` (e.g. `Ahora: Demoras esperadas up to 10 min`).

I don't think that needs to block this PR, but I did want to flag that.

priv/gettext/es/LC_MESSAGES/dotcom.po
```diff
@@ -87,6 +87,9 @@ msgstr "Regístrate para T-Alerts"
 msgid "Boston Visitor's Guide to The T"
 msgstr "Guía del visitante de Boston para el T"
 
+msgid "Delays"
+msgstr "Delays"
```
joshlarson, line 91: I know this isn't the point, but I suspect that this isn't right 😅

joshlarson (approved):

lib/dotcom_web/components/system_status/status_row_heading.ex
thecristen: ehhh I think this should be blocking

priv/gettext/es/LC_MESSAGES/dotcom.po
anthonyshull: i think it struggles with single capitalized words probably thinking they are proper nouns

PR #2743, opened 2025-09-26
feat(Components): CR survey banner

thecristen:
#### Summary of changes

**Asana Ticket:** [Stand up banners for CR survey](https://app.asana.com/1/15492006741476/project/385363666817452/task/1211453626386948?focus=true)

image

image

### DIFF
--- a/lib/dotcom_web/components/components.ex
+++ b/lib/dotcom_web/components/components.ex
@@ -4,6 +4,7 @@ defmodule DotcomWeb.Components do
   events -- those should utilize LiveView/LiveComponents or some other
   technology.
   """
+  use Dotcom.Gettext.Sigils
   use Phoenix.Component
 
   use Phoenix.VerifiedRoutes,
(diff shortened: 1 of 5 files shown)


lib/dotcom_web/components/components.ex
```diff
@@ -261,9 +262,24 @@ defmodule DotcomWeb.Components do
...
   def promo_banner(assigns) do
     ~H"""
-    <a href={@href} class={"block bg-amethyst-90 text-black text-center #{@class}"} target="_blank">
+    <a href={@href} class={"block bg-amethyst-80 text-black text-center #{@class}"} target="_blank">
       {render_slot(@inner_block)}
     </a>
     """
   end
+
+  def cr_survey_promo_banner(assigns) do
```
joshlarson, line 271: Style nit: I would put this component into a separate module, so that when it's time to remove it, it's a cleaner excision.

lib/dotcom_web/templates/mode/hub.html.heex
```diff
@@ -1,5 +1,6 @@
 <% mode = mode_atom(@mode_name) %>
 <% grid_template = if mode == :bus, do: "_grid_buttons_bus.html", else: "_grid_buttons.html" %>
+<.cr_survey_promo_banner :if={mode == :commuter_rail} />
```
joshlarson, line 3: HEEx makes simple things look so simple 🤩

priv/gettext/es/LC_MESSAGES/dotcom.po
```diff
@@ -1378,7 +1384,7 @@ msgid "Cancellations"
 msgstr "Cancelaciones"
 
 msgid "Lost & Found"
-msgstr "Perdido &amp; encontrado"
+msgstr "Lost &amp; Found"
```
joshlarson, line 1387: Noooooo!

joshlarson (approved):

lib/dotcom_web/components/components.ex
thecristen: eh, _would_ it be cleaner? you'd have to remove the new module and its references...

joshlarson: My thought was that then you can just delete the file and clean up the compiler errors and warnings.

But I suppose that's true of having the component in this file as well.

PR #2744, opened 2025-09-26
fix: Alert counts for multi-trip alerts in commuter rail status

joshlarson:
## Before

Screenshot 2025-09-26 at 1 37 45 PM

## After

Screenshot 2025-09-26 at 1 36 33 PM

The fix is deceptively simple, and also a bit inelegant. We already have `Dotcom.SystemStatus.CommuterRail.commuter_rail_route_status/1`, which figures out the correct delay/cancellation count, so we can just use the data from that to populate `commuter_rail_status/0` as well.

The inelegance comes from the fact that the `<.alerts_commuter_rail_status />` frontend component relies on the data structure returned from `commuter_rail_status/0` being structured a particular way, which is different from how `commuter_rail_route_status/1` does it. So I added a bit of "adapter" logic to `route_info/1` to convert the data structure `commuter_rail_route_status/1` presents to the data structure `<.alerts_commuter_rail_status /> expects.

(This is especially inelegant because `commuter_rail_route_status/1` already makes a distinction between train impacts (delays and cancellations) and service impacts (everything else), and `<.alerts_commuter_rail_status />` does care about that distinction, but has to re-figure it out on its own, because `commuter_rail_status/0` treats them all the same.)

Medium/longer-term, I think the thing to do is to update `<.alerts_commuter_rail_status />` to accept `commuter_rail_route_status/1`-shaped data. I'm actively working on that now, and can add that to this PR. But this fixes a rider-facing bug, so I figured I'd share this commit as a standalone fix. We can decide whether we want to merge as-is and treat a more elegant approach as a follow-up, versus combining it all together and holding this PR until the refactor is done.

**Asana Ticket:** [Fix how we count cancellations/delays on CR system status](https://app.asana.com/1/15492006741476/project/555089885850811/task/1211265393748828?focus=true)

### DIFF
--- a/lib/dotcom/system_status/commuter_rail.ex
+++ b/lib/dotcom/system_status/commuter_rail.ex
@@ -259,20 +259,6 @@ defmodule Dotcom.SystemStatus.CommuterRail do
     |> Enum.filter(&Routes.Route.commuter_rail?/1)
   end
 
-  # Returns a map where the key is the effect of the alert
-  # and the value is the number of alerts with that effect.
-  # For example, if there are 2 delays and 1 cancellation,
-  # the map would be `%{delay: 2, cancellation: 1}`.
-  @spec alert_counts([Alerts.Alert.t()]) :: map()
-  defp alert_counts(alerts) do
-    alerts
-    |> Enum.group_by(& &1.effect)
-    |> Enum.map(fn {effect, alerts} ->
-      {effect, %{count: Kernel.length(alerts), next_active: next_active_time(alerts)}}
-    end)
-    |> Map.new()
-  end
-
   # Returns a boolean indicating whether or not the route has a schedule
   # for today. This is used to determine if the route is running service today.
   @spec service_today?(String.t()) :: boolean()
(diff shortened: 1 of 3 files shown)


lib/dotcom/system_status/commuter_rail.ex
```diff
@@ -295,9 +295,27 @@ defmodule Dotcom.SystemStatus.CommuterRail do
...
+        } ->
+          service_impacts
+          |> Enum.group_by(& &1.alert.effect)
+          |> Map.new(fn {effect, impact_list} ->
+            {effect, impact_summary(impact_list)}
+          end)
+          |> Enum.into(%{
+            cancellation: impact_summary(cancellations),
+            delay: impact_summary(delays)
+          })
```
thecristen, line 312: I wonder if this should use `Enum.into/3`, like (untested)

```elixir
|> Enum.into(%{
  cancellation: impact_summary(cancellations),
  delay: impact_summary(delays)
}, fn {effect, impact_list} ->
  {effect, impact_summary(impact_list)}
end)
```

thecristen (approved):
This seems pretty reasonable~

PR #2749, opened 2025-10-01
Localization for the datepicker in dotcom

anthonyshull:
https://app.asana.com/1/15492006741476/project/555089885850811/task/1210709546068805?focus=true


lib/dotcom_web/components/trip_planner/input_form.ex
```diff
@@ -75,6 +75,7 @@ defmodule DotcomWeb.Components.TripPlanner.InputForm do
             :if={show_datepicker?(f)}
             module={DatePicker}
             config={datepicker_config()}
+            locale={Gettext.get_locale(Dotcom.Gettext)}
```
thecristen, line 78: nit: Might be better to pass this in as a prop from the parent LiveView (which has the locale already in its assigns via a server-side hook). Just not sure how this version would behave after we get around to implementing language switching.

thecristen (approved):
🎉

PR #2751, opened 2025-10-06
fix: Don't crash the planned disruptions page if an alert is recurring and "until further notice"

joshlarson:
While testing https://github.com/mbta/dotcom/pull/2675, we discovered that the `/alerts/subway` page was crashing on both [dev-blue](https://dev-blue.mbtace.com/alerts/subway) and [dev-green](https://dev-green.mbtace.com/alerts/subway). That's because there's an alert in the test environment that is recurring and also has an effect period of `until further notice`.

Screenshot 2025-10-06 at 11 25 35 AM

Our [logic to combine contiguous active periods](https://github.com/mbta/dotcom/blob/d5a4a370f08037acbda9f731d2cdfd572a58ecb3/lib/dotcom/alerts/subway/disruptions.ex#L83) croaks if the active period list looks like `[{date1, nil}, {date2, _}]`, because it tries to get the service date of `nil`, which isn't a sensible thing to do. This PR fixes that by just short-circuiting when it sees `nil` in that spot.

---

This is a scenario that shouldn't happen in the real world, because the units of recurring alerts almost definitionally should be bounded in time. But Alerts UI allows you to do this, so we need to handle it.

---

No ticket.

### DIFF
--- a/lib/dotcom/alerts/subway/disruptions.ex
+++ b/lib/dotcom/alerts/subway/disruptions.ex
@@ -75,6 +75,15 @@ defmodule Dotcom.Alerts.Subway.Disruptions do
   # after, so active periods of {Mon, Wed} and {Thu, Fri} would be
   # combined into {Mon, Fri}, as would {Mon, Wed} and {Wed, Fri}, but
   # {Mon, Tue} and {Thu, Fri} would be kept separate.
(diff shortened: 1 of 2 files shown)


thecristen (approved):
:shipit:

PR #3095, opened 2026-04-08
SF2.0[Fix]: Handle predictions with nil times and nil schedule

lvachon1:
## Scope

https://mbta.slack.com/archives/GRB64NPHS/p1775669968697359

## Implementation
Updated cancelled?() logic to return true if there are no prediction times nor schedule for a prediction
Added logic to handle display time for predictions without a schedule nor any predicted arrival or departure times and to treat them as cancelled/skipped.

### This branch
Screenshot 2026-04-08 at 3 06 02 PM

### Dev Green
Screenshot 2026-04-08 at 3 06 22 PM

## How to test

Visit SF2.0 when stops are being skipped and trips are added(?)  THere's one now on dev-green
http://localhost:4001/departures/?route_id=Red&direction_id=0&stop_id=place-sstat

### DIFF
--- a/lib/dotcom/schedule_finder/trip_details.ex
+++ b/lib/dotcom/schedule_finder/trip_details.ex
@@ -145,7 +145,7 @@ defmodule Dotcom.ScheduleFinder.TripDetails do
     if route_type in [0, 1] && status do
       {:status, status}
     else
-      if(display_time) do
+      if display_time do
         {:time, display_time |> truncate(:minute)}
       end
     end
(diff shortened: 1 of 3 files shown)


joshlarson (changes requested):
I think this looks good, but I really do want a test for this

joshlarson (changes requested):
Tiny change - the test is a bit flaky!

joshlarson (approved):

PR #3096, opened 2026-04-08
chore: remove `DotcomWeb.Components.Prototype`

thecristen:
## Scope

**Asana Ticket:** [📅🔎🚀 | QF | Pre-release, remove the testing route/direction/stop selector from the top](https://app.asana.com/1/15492006741476/project/555089885850811/task/1213526833787104?focus=true)

## Implementation

Removed the component, then realized we don't need the `handle_params/3` callback, as that's useful when navigating within a LiveView. 
- Instead, we can do all of our data loading in `mount/3`. So I did this refactor, which let me remove some of the smaller intermediate functions like `assign_route/2` or `assign_stop/2`.
- I also tweaked the "go to 404" handling slightly to handle this in the `mount/3` as well. Turns out you can raise an exception.
- I removed a few `:if={@stop}` from the template, since by the time we render we've already fetched the stop.

The diff looks more dramatic than it actually is, because nothing else was changed here...

## Screenshots

It looks the same, but without the buttons for switching any of the URL params.

image

## How to test

Manually try different URLs 😩 
- /departures should redirect to /schedules
- /departures with a combination of URL params that doesn't have a valid `route_id`, `direction_id` or `stop_id` will return the 404 page


joshlarson (approved):
Goodbye Prototype Picker! It was a good run 🥲

PR #3099, opened 2026-04-09
fix(alerts): Refactor alerts to use semantic elements

jlucytan:
## Scope

**Asana Ticket:** [Don't flash open alert when first going to any page ](https://app.asana.com/1/15492006741476/project/1213039586590742/task/1209962729125888)

## Implementation
Changed alerts to use details/summary elements instead of relying on JS to operate as an expandable. Stripped out the now-unnecessary JS. Alerts no longer animate as the animation was not consistent after the change to using details/summary.

Also found and corrected a typo so now the React alerts have the correct coloring 😆 

## Screenshots
Before 
image

After 
image

No other static visual change; alerts should appear the same as before, other than the lack of animation.

## How to test
Go to a schedule with expandable alerts. Observe that none of the alerts start open. [Local Green-B](http://localhost:4001/schedules/Green-B/alerts) for convenience. You may need to change the MBTA API base URL to dev-green to get the test alerts from staging. 

For React alerts, see a stop page with an alert ([Copley](http://localhost:4001/stops/place-coecl?route=Green-B&direction_id=1&headsign=Government+Center) has one going at time of writing)


assets/ts/components/__tests__/AlertsTest.tsx
```diff
@@ -58,24 +58,6 @@ const alertUrlDesc: Alert = {
...
-
-  wrapper.find(id).simulate("click");
-
-  expect(wrapper.find(id).prop("aria-expanded")).toEqual(true);
-
-  wrapper.find(id).simulate("keypress", { key: "Enter" });
-
-  expect(wrapper.find(id).prop("aria-expanded")).toEqual(false);
-});
-
```
jlucytan, line 78: Not using JS to handle expand/collapse anymore so removed this test

lvachon1 (approved):
Looks great!

PR #3104, opened 2026-04-13
Reduce upcoming departures polling rate from 1hz to 0.2hz

lvachon1:
## Scope

I think implementing this change now could help our CPU load.

## Implementation

Reduced upcoming departures polling rate from 1hz (1/s) to 0.2hz (1/5s)

There's still more work to be done, but this is an easy start

### Dev-blue (no change)
Screenshot 2026-04-13 at 1 44 32 PM

### Dev-green (this change)
Screenshot 2026-04-13 at 1 44 43 PM

*note the different vertical scale!

## How to test

~~I think we'll have to send this to dev-blue or dev-green to properly test it.~~


thecristen (approved):
I don't see any harm in moving forward with this!

PR #3106, opened 2026-04-13
chore: remove mbta-site Sentry project

thecristen:
As discussed in retro -- we'll keep mbta-dotcom (reports Elixir errors to Sentry) and stop sending JS errors to mbta-site.
**Note**: I didn't remove the `@sentry/react` frontend dependency, because we're using its `ErrorBoundary` component in the stop page, and I didn't consider it worth refactoring for now!


lvachon1 (approved):
Looks good to me, dev-blue seems to be working as expected.  Hopefully improves the signal-to-noise ratio in splunk.

PR #3107, opened 2026-04-13
A11Y: Improve semantics of visual tab panels that are links

lvachon1:
## Scope

**Asana Ticket:** [♿ Improve semantics of visual tab panels that are links](https://app.asana.com/1/15492006741476/project/555089885850811/task/1212906222065889?focus=true)

## Implementation

Two different implementations for two different behaviors:

For the tabs on the main page (Routes, Trip Planner, Alerts) I updated the existing javascript that currently hides the appropriate content and sets the css classes to also set the aria attributes and respect keyboard navigation.

For the tabs on the other pages that are links to new pages I created a generic keyboard event hander to enabled keyboard navigation and to update the aria-selected attribute accordingly.

Users should be able to use the arrow keys to switch which tab is focused, and then either Space or Enter to activate that tab.

### MBTA.com index page tabs
https://github.com/user-attachments/assets/1848db50-9a9b-4a30-afc7-f050733cfa0d

### Alerts page mode tabs
https://github.com/user-attachments/assets/fbca6029-c018-4787-bcd8-9dbad5734441

### Schedule/Line page tabs
https://github.com/user-attachments/assets/e97fe47b-e166-48a6-ad9f-8c321716a41d

### Stops page tabs
https://github.com/user-attachments/assets/6fe3c33c-51d6-40dd-86e0-70a6de1a1190

## How to test

http://localhost:4001/
http://localhost:4001/alerts
http://localhost:4001/schedules/Orange/line
http://localhost:4001/stops/subway

---
  - https://app.asana.com/0/0/1211525768292437

### DIFF
--- a/assets/js/tabbed-nav.js
+++ b/assets/js/tabbed-nav.js
@@ -1,16 +1,25 @@
-const updateTabs = (tabType, navTabs, contentTabs) => e => {
+const updateTabs = (tabType, navTabs, contentTabs, updateContent=true) => e => {
+
   e.preventDefault();
+  e.stopPropagation();
 
   navTabs.forEach(tab => {
     const clickedTab = tab.dataset.tabType === tabType;
     tab.classList.toggle("active", clickedTab);
     tab.setAttribute("aria-selected", clickedTab);
+    tab.removeAttribute("aria-current")
+    if(clickedTab){
+      if(updateContent){tab.setAttribute("aria-current","page");}
+      tab.focus();
+    }
   });
 
-  contentTabs.forEach(tab => {
-    const clickedTabContentType = tab.dataset.tabContentType === tabType;
-    tab.classList.toggle("active", clickedTabContentType);
-  });
+  if(updateContent){
+    contentTabs.forEach(tab => {
+      const clickedTabContentType = tab.dataset.tabContentType === tabType;
+      tab.classList.toggle("active", clickedTabContentType);
+    });
+  }
 };
 
 const tabbedNavSetup = () => {
@@ -22,9 +31,19 @@ const tabbedNavSetup = () => {
     const callback = updateTabs(tabType, navTabs, contentTabs);
 
     item.addEventListener("click", callback);
-    item.addEventListener("keyup", e => {
+    item.addEventListener("keydown", e => {
       if (e.key === "Enter") {
         callback(e);
+      }
+      const selectedTab = Array.from(navTabs).findIndex(tab=>{tab.getAttribute("aria-selected")=="true"});
+      if(e.key === "ArrowLeft" && selectedTab > 0){
+        updateTabs(navTabs[selectedTab-1].dataset.tabType, navTabs, contentTabs, false)(e);
+      }
+      if(e.key === "ArrowRight" && selectedTab < navTabs.length - 1){
+        updateTabs(navTabs[selectedTab+1].dataset.tabType, navTabs, contentTabs, false)(e);
+      }
+      if(e.key === " "){
+        updateTabs(navTabs[selectedTab].dataset.tabType, navTabs, contentTabs, true)(e);
       }
     });
   });
--- a/assets/js/tabbed-nav.js
+++ b/assets/js/tabbed-nav.js
@@ -1,5 +1,9 @@
-const updateTabs = (tabType, navTabs, contentTabs, updateContent=true) => e => {
-
+const updateTabs = (
+  tabType,
+  navTabs,
+  contentTabs,
+  updateContent = true
+) => e => {
   e.preventDefault();
   e.stopPropagation();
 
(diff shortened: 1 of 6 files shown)


lib/dotcom/components/tabs/tab_selector/component.html.eex
```diff
@@ -1,13 +1,13 @@
...
+  <nav class="<%= args.class %>" role="navigation" aria-label="stations mode tabs">
     <div class="tab-select-btn-group">
       <div class="stacked-tab-label"><strong><%= args.stacked_label %></strong></div>
       <%= for {tab_item_name, title, href} <- args.links do %>
         <%
         selected_class = if selected?(tab_item_name, args.selected), do: "tab-select-btn-selected", else: ""
         slug = slug(title)
         %>
-        <%= PhoenixHTMLHelpers.Link.link to: "#{href}\##{slug}", data: [scroll: "false"], id: slug, class: "btn #{selected_class}" do %>
+        <%= PhoenixHTMLHelpers.Link.link to: "#{href}\##{slug}", data: [scroll: "false"], id: slug, aria_selected: args.selected, aria_current: (if args.selected, do: "page"), class: "btn #{selected_class} tabnav" do %>
```
thecristen, line 10: I think this might not be working as intended, given what I see in the generated markup on `/stops/` (all three links have the same `aria-current` and `aria-selected` values)

assets/js/tabbed-nav.js
```diff
@@ -22,16 +37,84 @@ const tabbedNavSetup = () => {
...
+      }
+      if (e.key === "ArrowRight" && selectedTab < navTabs.length - 1) {
+        updateTabs(
+          navTabs[selectedTab + 1].dataset.tabType,
+          navTabs,
+          contentTabs,
+          false
+        )(e);
+      }
+      if (e.key === " ") {
```
joshlarson, line 93: **Issue (non-blocking):** On the homepage, this is treating `Tab`/`Shift-Tab` differently from `Right Arrow`/`Left Arrow`.

👇 `Tab` 👇 
Image

(Pressing `Space` does not switch the view to Trip Planner. Pressing `Enter` does.)


👇 `Right Arrow` 👇 
Image

(Both `Enter` and `Space` switch to Trip Planner.)

---

(Putting this comment here because I _think_ the lines of code here and around here are responsible, but I'm not certain... sorry if this is the wrong place for this!)

joshlarson (commented):
The homepage behavior still doesn't look quite right - if that's an easy fix, I think it's probably worth doing.

Otherwise this looks great! I always found the space-bar-scrolls-down-a-bit behavior so annoying!

assets/js/tabbed-nav.js
```diff
@@ -22,16 +37,84 @@ const tabbedNavSetup = () => {
     const callback = updateTabs(tabType, navTabs, contentTabs);
 
     item.addEventListener("click", callback);
-    item.addEventListener("keyup", e => {
+    item.addEventListener("keydown", e => {
```
thecristen, line 70: **suggestion(blocking)**: I think we need one more clause here -- according to the [ARIA authoring guidance on manual tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/), when pressing `Tab` the focus should move from the tab to the next element in the page tab sequence outside the tablist. e.g. if I'm focused on the "Trip Planner" tab, I'd expect `Tab` to take me to the "From" input, instead of to the next tab button.

assets/js/tabbed-nav.js
```diff
@@ -1,16 +1,31 @@
...
+  e.stopPropagation();
 
   navTabs.forEach(tab => {
     const clickedTab = tab.dataset.tabType === tabType;
     tab.classList.toggle("active", clickedTab);
     tab.setAttribute("aria-selected", clickedTab);
+    tab.removeAttribute("aria-current");
+    if (clickedTab) {
+      if (updateContent) {
+        tab.setAttribute("aria-current", "page");
```
thecristen, line 17: These tabs aren't pages, and I think we don't need `aria-current` on the homepage tabs at all. Excerpt from the WAI-ARIA doc (https://w3c.github.io/aria/#aria-current):

> Authors SHOULD NOT use the [aria-current](https://w3c.github.io/aria/#aria-current) attribute as a substitute for [aria-selected](https://w3c.github.io/aria/#aria-selected) in widgets where [aria-selected](https://w3c.github.io/aria/#aria-selected) has the same meaning. For example, in a [tablist](https://w3c.github.io/aria/#tablist), [aria-selected](https://w3c.github.io/aria/#aria-selected) is used on a [tab](https://w3c.github.io/aria/#tab) to indicate the currently-displayed [tabpanel](https://w3c.github.io/aria/#tabpanel)

assets/js/tabbed-nav.js
lvachon1: Good catch, I've updated the JS to fix that.

thecristen (changes requested):
So work for two tickets got combined here, but those two tickets were separate for a reason I will call out -- the homepage tabs exhibit tab behavior (switching content views without navigating away) while all the other "tabs" in MBTA.com behave like lists of links (supporting navigation, right-click to open, loading new pages), thus they **should** have different semantics.

- This PR does a great job at adding the necessary semantics and keyboard navigation to the homepage tabs, with the exception of the incomplete behavior of pressing `Tab` from the `tablist` (noted in a comment)
- Navigation links don't need `aria-selected`, but setting `aria-current="page"` IS helpful
- Using a `<nav>` is appropriate for the navigation links, so thanks for that
- The navigation links don't need the fancy keyboard behavior expected for tabbed interfaces -- likely the default HTML behavior is fine.

Hoping this makes sense. Happy to discuss if you have any questions, this was interesting to ponder over back when I was reading NCAM's feedback and writing these tickets.

assets/js/tabbed-nav.js
lvachon1: I've added some JS so that [tab] no longer focuses the next tab, but instead focuses the next focusable item.  [shift]+[tab] does the same, but in the other direction.  I'm not sure if I like it though, it seems sorta janky.  Part of me feels like completely remaking these tabs so we don't have to rely on so much JS.

lvachon1:
Closing this PR to break it out into two different ones.  Here's the first: https://github.com/mbta/dotcom/pull/3114

PR #3108, opened 2026-04-13
fix: Filter out empty params in QueryReducer

jlucytan:
## Scope

No asana ticket. @thecristen mentioned this in Slack.
> https://www.mbta.com/departures/?route_id=Red&&direction_id=0&stop_id=place-sstat prod seems to hang indefinitely when I fail at typing and insert an extra & heheh

See [slack thread](https://mbta.slack.com/archives/C06TP48CJ9W/p1776114081358089) for more details.

## Implementation
QueryReducer fails when the URI param has an undefined value (e.g. `mbta.com/?&&` and `mbta.com/?foo&`). This PR filters out cases where the key name is empty and adds basic handling for `undefined` values (simply passes it through as `undefined`) 

## Screenshots
Before - page hangs due to the JS param parsing erroring 
image

After - normal display
image

## How to test
Visit a schedule finder 2.0 page with various unexpected params: 
- && (empty key, no val) [local link](http://localhost:4001/departures/?route_id=Red&&direction_id=0&stop_id=place-sstat)
- &foo& (present key, no val) [local link](http://localhost:4001/departures/?route_id=Red&foo&direction_id=0&stop_id=place-sstat)
- &foo=& (present key, empty val) [local link](http://localhost:4001/departures/?route_id=Red&foo=&direction_id=0&stop_id=place-sstat)
  - this should not hang even before this PR 
- &=bar& (empty key, present val) [local link](http://localhost:4001/departures/?route_id=Red&=bar&direction_id=0&stop_id=place-sstat)
  - this should not hang even before this PR, but the QueryReducer now strips this entry from the params
  - it may be good to go to other places where we use params to make sure they haven't been affected

verify that they no longer hang

### DIFF
--- a/assets/ts/helpers/query.ts
+++ b/assets/ts/helpers/query.ts
@@ -31,6 +31,10 @@ const queryReducer = (decoder: EncoderDecoder): QueryReducer => (
   kvString: string
 ) => {
   const [key, val] = kvString.split("=");
+
+  if (key === "") {
+    return { ...params };
+  }
   return { ...params, [decoder(key)]: decoder(val.replace(/\+/g, " ")) };
 };
 
--- a/assets/ts/helpers/query.ts
+++ b/assets/ts/helpers/query.ts
@@ -33,7 +33,7 @@ const queryReducer = (decoder: EncoderDecoder): QueryReducer => (
   const [key, val] = kvString.split("=");
 
   if (key === "") {
-    return { ...params };
+    return params;
   }
 
   if (val === undefined) {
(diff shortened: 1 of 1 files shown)


lvachon1 (approved):
Works great!

assets/ts/helpers/query.ts
```diff
@@ -31,6 +31,15 @@ const queryReducer = (decoder: EncoderDecoder): QueryReducer => (
   kvString: string
 ) => {
   const [key, val] = kvString.split("=");
+
+  if (key === "") {
+    return { ...params };
```
lvachon1: I think we can just return params here, but the difference is probably only a handful of bytes to the user.

jlucytan: oop, was definitely just on auto-pilot copy-pasting and modifying from the other lines 🤣 thanks for catching!

PR #3110, opened 2026-04-14
A11Y: Ensure Trip Planner form errors notify screen readers

lvachon1:
## Scope

**Asana Ticket:** [♿ Ensure Trip Planner form errors notify screen readers](https://app.asana.com/1/15492006741476/project/555089885850811/task/1211545530335853?focus=true)

## Implementation

Added aria roles and form labels to location error messages so that screen readers will announce them.

It does not seem possible to cause an error on the mode input

The DatePicker is an MBTA_Metro component, so that'll need it's own PR (https://github.com/mbta/mbta_metro/pull/36)

## Screenshots

https://github.com/user-attachments/assets/e62b1192-b7d4-45ac-9094-c104e76d69de

## How to test

https://support.apple.com/guide/voiceover/get-started-with-voiceover-vo4be8816d70/10/mac/26

~~Observe~~ listen for the alert chime and error message text to be read after entering an invalid origin and/or destination location


joshlarson (commented):
My experience so far using VoiceOver on Chrome is that the `Please select...` message and the `Clear button` message enter into a race condition where whichever message starts first gets interrupted by the other one, and then doesn't play.

I'm honestly not sure whether that's a bug in my version of VoiceOver, or a problem with how the aria-live is being placed into the DOM (or both?)

Apparently this is not happening on other people's machines. I'm going to keep digging and noodling, but just wanted to flag why I haven't approved this yet!

joshlarson (approved):
It seems like everyone else's VO'es handle this the way I would expect - this can be one of the things we ask NCAM about on Tuesday!

PR #3115, opened 2026-04-15
A11Y: Implement tab behavior for homepage tabs

lvachon1:
## Scope

**Asana Ticket:** [♿ Implement tab behavior for homepage tabs](https://app.asana.com/1/15492006741476/project/555089885850811/task/1211525768292437?focus=true)

## Implementation

Changed the JS for the main page tabs to adhere to a11y guidelines:

- Only the active tab is focusable with the [tab] key, the arrow keys are used to focus other tabs, [space] or [enter] is used to active the focused tab
- Hitting [tab] from a tab group focuses the next focusable item in the tab's related content, not the next tab
- `aria-selected` attribute is set to true for the currently focused tab
- `aria-current` attribute is set to "page" for the currently active tab

Updated the template to wrap things in a <nav> instead of a <div> and setup the initial aria attributes and tabindexes (tabindices?)

## Screenshots

https://github.com/user-attachments/assets/256ddc5c-4c5d-4253-83c0-c456c78bf25f

## How to test

http://localhost:4001/

Use keyboard navigation and inspect the main page tabs to ensure the aria attributes are being set correctly, and the keyboard behavior works as expected.

### DIFF
--- a/assets/js/tabbed-nav.js
+++ b/assets/js/tabbed-nav.js
@@ -1,17 +1,57 @@
-const updateTabs = (tabType, navTabs, contentTabs) => e => {
+const updateTabs = (
+  tabType,
+  navTabs,
+  contentTabs,
+  updateContent = true
+) => e => {
   e.preventDefault();
+  e.stopPropagation();
 
   navTabs.forEach(tab => {
     const clickedTab = tab.dataset.tabType === tabType;
     tab.classList.toggle("active", clickedTab);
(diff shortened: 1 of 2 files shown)


joshlarson (approved):
With the ~grain~ _mound_ of salt that I'm a non-expert on tabbed nav behavior, this seems like it works the way I was expecting it to!

(I wish that proper tabbed behavior didn't require a bunch of bespoke Javascript code 😵‍💫)

PR #3118, opened 2026-04-16
fix: Remove `kino_live_component` and `kino` deps

joshlarson:
## Scope

No ticket.

I recently started running into 👇 error, which was preventing me from starting dotcom, and which I could only resolve by `rm -rf _build`, which then meant that I have to recompile _everything_ 😩.

```elixir
** (Mix) the application :elixir has a different value set for key :dbg_callback during runtime compared to compile time. Since this application environment entry was marked as compile time, this difference can lead to different behavior than expected:

  * Compile time value was set to: {Kino.Debug, :dbg, [{Macro, :dbg, []}]}
  * Runtime value was set to: {Macro, :dbg, []}

To fix this error, you might:

  * Make the runtime value match the compile time one

  * Recompile your project. If the misconfigured application is a dependency, you may need to run "mix deps.clean elixir --build"

  * Alternatively, you can disable this check. If you are using releases, you can set :validate_compile_env to false in your release configuration. If you are using Mix to start your system, you can pass the --no-validate-compile-env flag
```

Since we weren't really using `kino_live_component` or `kino` in Dotcom, removing it felt like a reasonably straightforward solution.


lvachon1 (approved):
Thanks for taking care of that annoying issue!

lvachon1:
I noticed an unrelated flaky test on these CI runs, so I wrote a PR to fix it: https://github.com/mbta/dotcom/pull/3120

PR #3120, opened 2026-04-16
Fix flaky upcoming_departure trip name test

lvachon1:
## Scope

I noticed a test was flaky, so I fixed it.

## Implementation

~~Fixed flaky test that was sometimes breaking when the factory returned a rail replacement bus instead of a train.  Updated assertion so it will accept either "Bus #name" or "Train #name"~~

Updated commuter_rail_route_factory so it always returns a train and never a rail_replacement_bus.  

## Screenshots

Errors caught in CI: 
Screenshot 2026-04-16 at 2 42 25 PM

## How to test

Run the unit test a few times


test/dotcom/schedule_finder/upcoming_departures_test.exs
```diff
@@ -198,7 +198,9 @@ defmodule Dotcom.ScheduleFinder.UpcomingDeparturesTest do
       # Verify
       assert [departure] = departures
 
-      assert departure.trip_name == "Train #{trip.name}"
+      assert departure.trip_name == "Train #{trip.name}" or
```
joshlarson: **Issue:** Would it be possible to update the test setup so that it's reliably a train instead? Having the tests that reliably tell us `Bus ___` versus `Train ___` is actually helpful.

lvachon1: It turns out Cristen already wrote a test for "Bus #{name}", I just needed to tweak the original commuter_rail_route_factory to always return a train.

joshlarson (approved):

PR #3124, opened 2026-04-17
perf(SF2.0): Dynamically load trip details on expand

joshlarson:
## Scope

**Asana Ticket:** [📅🔎🎭 Improve SF2.0 performance](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214093591659854?focus=true)

## Implementation

- Add a handler to the LiveView that loads trip details into the socket when their associated row is expanded.
- Lift all upcoming-departures related fields (that is, fields that are visible when upcoming departures rows are collapsed) to the `UpcomingDepartures` struct.
- Remove `trip_details` from `UpcomingDepartures`, in favor of calculating that out of band (on expand).
- Reduce the scope of `Schedules.Repo` queries to avoid iterating over as many `%Schedule{}` structs.

## Screenshots

Screenshot 2026-04-17 at 3 08 02 PM

This is a mostly invisible change (except that hopefully things will load faster!) This state is kind of the exception - it's not possible to see a loading spinner in that spot on `main` right now.

## How to test

The real test is to deploy this to dev-{green,blue}, and [compare CPU utilization](https://mbta.splunkcloud.com/en-US/app/search/dotcom_dev_cpu_comparison?form.global_time_tok.earliest=-1h&form.global_time_tok.latest=now) between dev-{green,blue} and dev-{blue,green} while having like 20 SF2.0 tabs open.

### DIFF
--- a/lib/dotcom/schedule_finder/upcoming_departures.ex
+++ b/lib/dotcom/schedule_finder/upcoming_departures.ex
@@ -40,7 +40,8 @@ defmodule Dotcom.ScheduleFinder.UpcomingDepartures do
       :stop_sequence,
       :trip_details,
       :trip_id,
-      :trip_name
+      :trip_name,
+      :vehicle_name
     ]
 
     @type realtime_arrival_status_t ::
@@ -76,7 +77,8 @@ defmodule Dotcom.ScheduleFinder.UpcomingDepartures do
             stop_sequence: non_neg_integer(),
             trip_details: __MODULE__.UpcomingTripDetails.t(),
             trip_id: Trip.id_t(),
-            trip_name: String.t()
+            trip_name: String.t(),
+            vehicle_name: String.t() | nil
           }
 
     defmodule UpcomingTripDetails do
@@ -319,7 +321,8 @@ defmodule Dotcom.ScheduleFinder.UpcomingDepartures do
         if(route_type == :commuter_rail,
           do: trip_name(predicted_schedule_route, trip.name),
           else: nil
-        )
+        ),
+      vehicle_name: vehicle_name(vehicle, route_type)
     }
   end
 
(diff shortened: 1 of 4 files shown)


lvachon1:
Screenshot 2026-04-17 at 4 25 15 PM

Certainly seems to help so far.  20 tabs on each, green never went above 50%

joshlarson:
Finally ready for review!

I recommend reviewing commit-by-commit.

lib/dotcom/schedule_finder/upcoming_departures.ex
```diff
@@ -127,11 +127,27 @@ defmodule Dotcom.ScheduleFinder.UpcomingDepartures do
...
-    predicted_schedules = predicted_schedules(route.id, direction_id, now)
+
+    predictions =
+      @predictions_repo.all(
+        route: route.id,
+        direction_id: direction_id,
+        include_terminals: true,
+        discard_past_subway_predictions: false
+      )
+      |> Enum.filter(&(&1.stop.id == stop_id))
```
lilybarrett, line 138: I could be reading this wrong, but are we fetching all the predictions that meet this criteria for `route`, `direction` etc. and then filtering them? I also ask this while still being an Elixir newbie (and a newbie to this codebase), but is there any way to filter by `stop_id` directly rather than fetching all the predictions first?

lvachon1 (approved):
Everything looks good to me code wise, and we've already seen the performance gains this brings.  I spent a good hour or so testing manually, everything works great!  Even in weird situations like trains being taken out of service, shuttles, different destinations than normal, things performed as expected.

The only oddity I saw is when an orange line train was crossing the main line to get into the yard at Wellington, it's status went from Boarding back to Arriving until it was clear of the tracks and properly disappeared.  I'm not sure this is something we should worry about though, seems more like a rare prediction issue, and I definitely don't think it has anything to do with this PR.

lib/dotcom/schedule_finder/upcoming_departures.ex
joshlarson: That's a great question, and I actually tried it out in [this draft PR](https://github.com/mbta/dotcom/pull/3149). For reasons I won't pretend to fully understand, that change made performance substantially worse, both in terms of number of API calls and CPU usage, so I dropped it.

PR #3126, opened 2026-04-21
add initial tests for ScheduleFinderLive

thecristen:
We punted on writing unit tests for this particular file earlier, because we have lots of other unit tests covering the more critical functions being called in `ScheduleFinderLive`.

This is just an initial crack at it -- it only covers the file somewhere between 30-40% so there'll be more tests to come!

- There's a little refactoring on how services are handled, because it was too hard to test otherwise (the underlying functions weren't being called a consistent number of times). The service picker _should_ work the same way as before, though
- There're other unrelated test changes I just noticed while running tests over and over

---
  - https://app.asana.com/0/0/1212779092699529

### DIFF
--- a/test/dotcom_web/live/schedule_finder_live_test.exs
+++ b/test/dotcom_web/live/schedule_finder_live_test.exs
@@ -1,0 +1,344 @@
+defmodule DotcomWeb.ScheduleFinderLiveTest do
+  use DotcomWeb.ConnCase, async: true
+
+  import DotcomWeb.Router.Helpers, only: [live_path: 2, live_path: 3]
+  import Mox
+  import Phoenix.LiveViewTest
+  import Test.Support.Generators.DateTime, only: [random_date_time: 0]
+
+  alias DotcomWeb.ScheduleFinderLive
+  alias Test.Support.Factories
+  alias Test.Support.FactoryHelpers
+
+  setup :verify_on_exit!
+
+  setup _ do
+    cache = Application.get_env(:dotcom, :cache)
+    cache.flush()
+
+    stub_with(Dotcom.Utils.DateTime.Mock, Dotcom.Utils.DateTime)
+
+    :ok
+  end
+
+  test "loads, fetching route info", %{conn: conn} do
+    route_id = FactoryHelpers.build(:id)
+    stop_id = FactoryHelpers.build(:id)
+    direction_id = FactoryHelpers.build(:direction_id)
+
+    # The LiveView lifecycle calls mount/3 twice, hence these are run twice
+    expect(Dotcom.ScheduleFinder.Mock, :current_alerts, 2, fn stop, route ->
+      assert stop.id == stop_id
+      assert route.id == route_id
+      []
+    end)
+
+    expect(Routes.Repo.Mock, :get, 2, fn ^route_id ->
+      Factories.Routes.Route.build(:route, %{id: route_id})
+    end)
+
+    expect(Services.Repo.Mock, :by_route_id, 2, fn ^route_id ->
+      Factories.Services.Service.build_list(5, :service)
+    end)
+
+    expect(Stops.Repo.Mock, :get, 2, fn ^stop_id ->
+      Factories.Stops.Stop.build(:stop, %{id: stop_id})
+    end)
+
+    path =
+      live_path(conn, ScheduleFinderLive,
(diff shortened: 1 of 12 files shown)


test/dotcom_web/live/schedule_finder_live_test.exs
```diff
@@ -0,0 +1,344 @@
...
+          route: route
+        )
+
+      assert html =~
+               ~s(<a href=\"/schedules/#{route.id}?schedule_direction[direction_id]=#{direction_id}\")
+
+      assert html =~ route.name
+      assert html =~ route.direction_names[direction_id]
+      assert html =~ "towards"
+      assert html =~ route.direction_destinations[direction_id]
```
joshlarson: **Issue:** String-matching on HTML directly makes tests brittle. We have [prior art for subway status tests](https://github.com/mbta/dotcom/blob/47f6cb6120684f8a7c43a6b20645544f8bd4b7de/test/dotcom_web/components/system_status/subway_status_test.exs) that show how we can use Floki and `data-test` attributes to make the tests target the specific parts of the page that they're interested in.

test/dotcom_web/live/schedule_finder_live_test.exs
```diff
@@ -0,0 +1,344 @@
...
+
+    # It will get called again!
+    expect(Dotcom.ScheduleFinder.Mock, :current_alerts, 1, fn _, _ ->
+      Factories.Alerts.Alert.build_list(1, :alert_for_informed_entity,
+        informed_entity: %{direction_id: nil}
+      )
+    end)
+
+    # Send out alert update
+    DotcomWeb.Endpoint.broadcast("alerts", "alerts_updated", [])
```
joshlarson, line 204: This is cool!

test/dotcom_web/live/schedule_finder_live_test.exs
```diff
@@ -0,0 +1,344 @@
...
+  describe "Daily Departures (non-subway)" do
+    test "loads & shows individual departures", %{conn: conn} do
+      departures =
+        2
+        |> Faker.random_between(20)
+        |> Factories.ScheduleFinder.build_list(:daily_departure)
+
+      expect(Dotcom.ScheduleFinder.Mock, :daily_departures, 2, fn _, _, _, _ ->
+        {:ok, departures}
+      end)
```
joshlarson, line 276: **Issue:** Especially for daily schedules and _espcially_ for upcoming departures, I think we gain most of our value by asserting on a pipeline that goes all the way through `Schedules` --> `ScheduleFinder.daily_schedules` --> `HTML`. Mocking the logicky bits feels to me like we lose a huge amount of the value that we otherwise could get here, and also prevents us from changing the `DailySchedules` (or `UpcomingDepartures`) interface.

test/support/factories/routes/route.ex
```diff
@@ -69,8 +69,8 @@ defmodule Test.Support.Factories.Routes.Route do
       long_name: Faker.Company.catch_phrase(),
       color: Faker.Color.rgb_hex(),
       direction_destinations: %{
-        0 => Faker.Address.street_address(),
-        1 => Faker.Address.street_address()
+        0 => Faker.Address.state(),
+        1 => Faker.Address.state()
```
joshlarson, line 73: `Green Line C Branch Westbound to Oregon`

Image

joshlarson (changes requested):
I haven't finished an in-depth review here, but had a few first impressions that could be worth discussion.

I'm marking request changes, but this definitely falls into the _request_ changes category, rather than _demand_ changes; I'm open to being persuaded that this approach is good, either for now or for the long-term.

test/dotcom_web/live/schedule_finder_live_test.exs
thecristen: Added a commit to improve on this! [62af67f](https://github.com/mbta/dotcom/pull/3126/commits/62af67ffc462c84b8cc3ef21f34f147217c49fd3)

test/dotcom_web/live/schedule_finder_live_test.exs
thecristen: I'm going to leave this as is for now and we can revisit it if desired.

joshlarson (approved):

PR #3130, opened 2026-04-23
Resolve two search regions

lvachon1:
## Scope

**Asana Ticket:** [♿ Resolve two search regions](https://app.asana.com/1/15492006741476/project/555089885850811/task/1211600360082756?focus=true)

## Implementation

Added some javascript code to add "aria-hidden" attributes to the main nav search containers and update them accordingly as the page resizes and mobile or desktop view switches out.

These components already had a width of 0 and display:none, however that doesn't seem to be enough to hide them.  This code sets the ariaHidden property of the two search containers so that only one of them is visible at a time (depending on window width [aka mobile/desktop])

### DOM for desktop mode 
DOMDesktop

### A11y view for desktop mode
Screenshot 2026-04-23 at 1 14 15 PM

### DOM for mobile mode
DOMMobile

### A11y view for mobile mode
Screenshot 2026-04-23 at 1 14 37 PM

## How to test

Visit any url with the main navigation in mobile and desktop mode.  Observe that in each mode only one search region is visible.

http://localhost:4001/

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -35,6 +35,7 @@ import translateAnalytics from "./translate-analytics.js";
 import previousEventsButton from "./view-previous-events";
 
 import mobileAppBanner from "./mobile-app-banner.js";
+import mobileSearchA11y from "./mobile-search-a11y.js";
 
 // Establish Phoenix Socket and LiveView configuration.
 import { Socket } from "phoenix";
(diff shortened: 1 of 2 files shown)


jlucytan (changes requested):
This appears to have some linting issues (can run `npm run ci:lint:js` locally). I see the checks skipped the linting for whatever reason but I do think this should be linted before it's merged in.

lvachon1:
Linting applied

jlucytan (approved):
thanks for linting!

PR #3132, opened 2026-04-23
Improve screenreader interaction with headers with links

lilybarrett:
## Scope

There is a screen reader issue on the homepage causing some headings to be read as multiple levels of headings, i.e., VoiceOver will say "heading level 2, level 1" (noisy!) instead of just "heading level 2."

Acceptance Criteria
* Update headings on the homepage so that each one is only identified as a single level of heading, not multiple ones (ex. both level 3 and level 2)

Steps to reproduce:
* Open [mbta.com](https://mbta.com/) and turn on a screen reader (ex. Voice Over)
* Navigate by heading (in Voice Over: ctrl + option + cmd + h)
* Observe that some headers are identified as both level 2 and level 3, including:
"Bus", "Subway", etc. in the Alerts tab

**Asana Ticket:** https://app.asana.com/1/15492006741476/project/555089885850811/task/1211600360082773?focus=true

## Implementation
The issue with some of these headings is that they have nested links within them, whereas the screenreader needs links to wrap the headers, not the other way around.

Note that I'm new to Elixir, so if there's a cleaner way to handle this, let me know. Right now, I've gone through and made one-off changes to sections/pages that were flagged in the Asana ticket. I also did a quick check for other headings with nested elements in the dotcom app, but a more thorough audit _could_ be necessary.

## How to test
* Open [mbta.com](https://mbta.com/) and turn on a screen reader (ex. Voice Over)
* Navigate by heading (in Voice Over: ctrl + option + cmd + h)
* Headers should be identified as just having one level, including:
"Bus", "Subway", etc. in the Alerts tab

### DIFF
--- a/assets/css/_homepage.scss
+++ b/assets/css/_homepage.scss
@@ -210,6 +210,69 @@
     }
   }
 
+  .heading-link {
+    h2,
+    h3,
+    h4 {
+      color: $gray-dark;
+      display: inline-flex;
+      font-size: inherit;
+      line-height: inherit;
+
+      &:hover {
+        text-decoration: inherit;
+      }
+
+      > .header-text {
+        margin-left: .5em;
+        text-decoration: underline;
+
+        @include media-breakpoint-down(xs) {
+          font-size: $font-size-base;
+          margin-bottom: $base-spacing;
+        }
+      }
+    }
+
+    h2 {
+      @include media-breakpoint-down(sm) {
+        font-size: 1.125rem;
+        line-height: 1.5rem;
+      }
+
+      @include media-breakpoint-only(md) {
+        font-size: 1.31rem;
+        line-height: 1.68rem;
+      }
+
+      @include media-breakpoint-up(lg) {
+        font-size: 2rem;
+        line-height: 2.5rem;
+      }
+    }
+
+    h3 {
+      @include media-breakpoint-down(sm) {
+        font-size: .875rem;
+        line-height: 1rem;
+      }
+
+      @include media-breakpoint-only(md) {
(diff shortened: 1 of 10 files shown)


joshlarson (changes requested):
I'm not an expert here, so it's possible that I'm getting this wrong, but I think that this is not quite the right solution. From a UX perspective, I'm noticing that for the links under the alerts section of the homepage, in prod, it says that it's a `link`, but on this branch, the fact that it's a `link` is hidden.

👇 Prod 👇 
Image

👇 This branch 👇 
Image

(Screenshots captured by taking a screen _recording_ and then taking screenshots from the resulting video... I definitely know how to use a computer! 😅)

---

My hunch is that `aria-label` is not the right solution here - that the root of the problem is that for some reason, the link itself is a "level 2 link" (whatever that means), and that to truly solve this, we need to make the link itself not be level 2. Then the readout would be `heading level 3, visited, link, Subway`, which I think is closer to what we actually want.

lilybarrett:
@joshlarson Great catch! I investigated and you're totally right that we should be broadcasting links. There are a number of headers wrapping links in the code, when it should be the other way around for voiceover to work properly. I've made changes accordingly.

assets/css/_homepage.scss
```diff
@@ -210,6 +210,69 @@
     }
   }
 
+  .heading-link {
```
lilybarrett, line 213: Added this style here so the links still maintain the header styling while behaving like links.

joshlarson (approved):
~Looks~ _Sounds_ great!

PR #3143, opened 2026-04-29
feat: Add app store redirects for mticket

EmmaSimon:
## Scope

**Asana Ticket:** No ticket, [slack thread](https://mbta.slack.com/archives/C8THJ1NJY/p1777477613853979)

## Implementation

CEX is asking for an mticket redirect that works the same way as the existing MBTA Go one. This just takes that code and makes the app store and desktop URLs params.

## How to test

Open https://dev-green.mbtace.com/mTicketapp on iOS, Android and desktop, the URLs are properly redirected to either app store or https://dev-green.mbtace.com/mbta-endorsed-apps


thecristen (approved):
👍🏼 Thanks!

PR #3144, opened 2026-04-29
feat(ScheduleFinderLive): use visibilitychange (closes #3125)

thecristen:
## Scope

Builds off prior art:
- #3125 

## Implementation

- Basically rewrites the JS logic into a [Phoenix client hook](https://hexdocs.pm/phoenix_live_view/js-interop.html#client-hooks-via-phx-hook). 
  - On component mount, assigns the `visibilitychange` event listener. It pushes the document `visibilityState` to the server, which can be handled in the parent LiveView via `handle_event`.
  - On mount also immediately sends the current `visibilityState`
  - Reconnections also trigger immediately sending the current `visibilityState`
  - On component `destroyed()` we remove the event listener.

Then I edited the Schedule Finder to respond to the changed value:
- Toggles a new assign, `:should_refresh?`, which if `true` will schedule (ahem) the next upcoming departure update.

## How to test

I put an `IO.inspect()` in `assign_upcoming_departures/1` to see how frequently it's being called after changing the tab visibility state! I'll note it's not a perfect/immediate transition - for example, if the next upcoming departure update is already queued when the tab changes to a hidden state, that next update will still happen.

### DIFF
--- a/lib/dotcom_web/live/schedule_finder_live.ex
+++ b/lib/dotcom_web/live/schedule_finder_live.ex
@@ -73,6 +73,7 @@ defmodule DotcomWeb.ScheduleFinderLive do
          |> assign_new(:loaded_trips, fn -> %{} end)
          |> assign_new(:selected_service_name, fn -> Map.get(selected_service, :label, "") end)
          |> assign_new(:daily_schedule_date, fn -> service_date() end)
+         |> assign_new(:should_refresh?, fn -> true end)
          |> assign_alerts()
          |> assign_departures()
          |> assign_upcoming_departures()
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/live/schedule_finder_live.ex
```diff
@@ -292,7 +305,7 @@ defmodule DotcomWeb.ScheduleFinderLive do
             stop_id: stop_id
           })
 
-        schedule_refresh_upcoming_departures(parent_pid)
+        _ = if should_refresh?, do: schedule_refresh_upcoming_departures(parent_pid)
```
lvachon1, line 308: Just curious, why assign the value of this condition if it's being disregarded anyway?  I assume it's to silence a warning?

lvachon1 (approved):
This is great!  Thanks for taking my half-baked idea over the line.

lib/dotcom_web/live/schedule_finder_live.ex
thecristen: You're right!! `mix dialyzer` complained about an unmatched return. :/

PR #3145, opened 2026-04-29
feat(Dotcom.Endpoint): use Bandit adapter

thecristen:
## Scope

Newly-created Phoenix projects as of version 1.7 use the [Bandit](https://github.com/mtrudel/bandit) webserver via the [Bandit.PhoenixAdapter](https://hexdocs.pm/bandit/1.8.0/Bandit.PhoenixAdapter.html) adapter. It's been reported to be faster and more performant than the prior industry standard (Cowboy). I figure we should try it!

## Implementation

While I switched the configured `:adapter` I also decided to clean up the files around configuring our `DotcomWeb.Endpoint` 🧹 

I also updated Bandit (it was already a dependency of something else) to the latest patch version.

Finally, I adjusted another thing, the endpoint's `check_origin` value, because I noticed this was throwing lots of errors in prod 🙃 

## Screenshots

No change

## How to test

Going to deploy and see that the prod configuration works (all dev environments use `MIX_ENV=prod`).

Special request: I'd also like to verify this doesn't break anything when running the application locally using Docker.


lvachon1:
I can verify that local docker stuff works just fine on this branch.  It took a while to compile (as expected), but started right up on the first try.

joshlarson (approved):
🤠 🩹 🇮🇹

PR #3149, opened 2026-04-29
perf: Load just the predictions for the station in question

joshlarson:
Draft PR for performance testing


joshlarson:
Testing is all done - turns out this was an anti-tradeoff! It made API calls worse _and_ made performance worse!

Screenshot 2026-04-29 at 9 55 19 PM

(The place where the lines switch sides in ☝️ chart is where I swapped the dev-green versus dev-blue labels, to see if the performance differences had anything to do with specifics of the environments themselves.... they didn't 😵‍💫)

PR #3150, opened 2026-04-29
fix: Boston Stadium trains button should link to /schedules/bostonstadium

lvachon1:
## Scope

**Asana Ticket:** [fix: Boston Stadium trains button should link to /schedules/bostonstadium](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214400724705466?focus=true)

## Implementation

Added code to filter out CR-Foxboro
Added a hand coded route for Boston Stadium Trains and logic to render buttons for it in the Commuter Rail context
This also fixes the issue of Boston Stadium Trains showing up in recently visited

### Schedules
Screenshot 2026-04-29 at 3 17 33 PM

### Schedules/Commuter-Rail
Screenshot 2026-04-29 at 3 17 47 PM

## How to test

http://localhost:4001/schedules

http://localhost:4001/schedules/commuter-rail

Confirm that the Foxboro Event Service link is gone and has been replaced with the Boston Stadium Trains link
Confirm that it goes to the World Cup / Boston Stadium page

### DIFF
--- a/lib/dotcom_web/templates/mode/_grid_buttons.html.heex
+++ b/lib/dotcom_web/templates/mode/_grid_buttons.html.heex
@@ -24,16 +24,20 @@
       name: "Boston Stadium Trains",
       sort_order: 20014,
       type: 2
-    } %>
-  {render("_grid_button.html",
-    conn: @conn,
-    route: wc_route,
-    show_icon?: @show_icons?,
-    link_path: ~p"/schedules/Bostonstadium",
-    has_alert?: has_alert?(wc_route, @alerts, @date_time),
-    id_prefix: assigns[:id_prefix],
-    index: Enum.count(@routes)
-  )}
+    }
+
+  test_route = @routes |> Enum.at(0) %>
+  {if !is_atom(test_route) and test_route.type == 2,
+    do:
+      render("_grid_button.html",
+        conn: @conn,
+        route: wc_route,
+        show_icon?: @show_icons?,
+        link_path: ~p"/schedules/bostonstadium",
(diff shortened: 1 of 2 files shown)


lib/dotcom_web/templates/mode/_grid_buttons.html.heex
```diff
@@ -10,6 +10,25 @@
...
+      render("_grid_button.html",
+        conn: @conn,
+        route: wc_route,
+        show_icon?: @show_icons?,
+        link_path: ~p"/schedules/bostonstadium",
+        has_alert?: false,
+        id_prefix: assigns[:id_prefix],
+        index: Enum.count(@routes)
+      )}
   <%= if assigns[:sub_routes] do %>
```
thecristen, line 32: **suggestion(not blocking)**: I am a bit curious about whether you could leverage the current `:sub_routes` mechanism for showing the Boston Stadium route. It appears to be used [here](https://github.com/mbta/dotcom/blob/f1c960480c1afb30a1b784944b0a7f8bb5d8395c/lib/dotcom_web/templates/mode/hub.html.heex#L22-L28) to render extra buttons for the Green Line pages on the subway mode page.

thecristen (approved):
Works! ⚽

lib/dotcom_web/templates/mode/_grid_buttons.html.heex
lvachon1: I gave that a shot and it didn't work out.  Subroutes use the condensed_grid_button layout which doesn't have room for anything but an icon.

PR #3155, opened 2026-05-01
Prevent decorative SVGs from making screenreader noise

lvachon1:
## Scope

**Asana Ticket:** [♿ Prevent decorative SVGs from making screenreader noise](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214318354497078?focus=true)

## Implementation

I initially attempted to suppress voice over with aria-hidden, display:none, etc... but that did not seem to work.
I then looked into stripping out the title tag at the component level, leaving the files alone.  Unfortunately there are several different icon implementations of mixed languages and repos, but they all used the same set of SVGs so...
I decided to strip all the of title tags from the SVGs and add back in aria-labels in the places we depended on them.
This has brought forward some nice features, such as the shuttle bus icon just had a title of "Shuttle" whereas now the aria-label can be "Massport Shuttle 33" or "Green Line Shuttle" instead.

The hard part of this approach is going through all the icons and making sure they either have A) Accompanying text B) An aria-label or C) are decorative and have aria-hidden or equivalent.  I'll need some help.

<hr/>

- Wrote a script to delete title tags from all SVGs in the specified path
- Ran that script
- Added aria-label support to the icon typescript component, and added labels for the parking and accessibility icons
- Added a name for the colored lines route object so their metro icons would have a proper label in the stop page header
- Removed the default "Marker" alt text from the leaflet map Marker, Marker, <sup>Marker<sup>Marker...</sup></sup> component
- Added aria-labels to shuttle bus icons
- Enhanced existing labels of stop page header mode icons
- Added aria-labels to feature icons of the autocomplete search results
- Added aria-labels to connection icons on line diagram page

TODO: Find out any more places that were relying on svg title tags and use aria-labels or alt text instead. 

## Screenshots

A/V Recordings upcoming

## How to test

This basically affects the whole website, so I'm probably going to put it on blue or green and ask people to play with it for a while to make sure I haven't missed anything.

Use voiceover on _everything_, notice how much less repetition there is, and ensure all icons that need labels have them.


jlucytan:
probably should add an aria-label for the icons in the alerts tab; old behavior did read out the title in the SVG and there isn't any text indicating what the icons are otherwise

image

joshlarson:
Similar to Lucy's comment, it looks like subway and commuter rail connections are just showing to screen readers as "image".

### `main`
Screenshot 2026-05-08 at 4 38 33 PM
Screenshot 2026-05-08 at 4 42 01 PM

### This branch
Screenshot 2026-05-08 at 4 38 11 PM
Screenshot 2026-05-08 at 4 41 47 PM

jlucytan:
fwiw the icons in @joshlarson 's screenshots are React components, not Elixir ones so you'll have to go in there to mess with them

joshlarson:
I think I found a few more icons that should be read out that aren't on this branch. On the [subway stop hub page](http://localhost:4001/stops/subway), there are two places where icons are read out in prod, and I'm pretty sure they still should be:

Screenshot 2026-05-14 at 4 56 57 PM
Screenshot 2026-05-14 at 4 59 01 PM

lib/dotcom_web/views/stop_view.ex
```diff
@@ -20,7 +20,24 @@ defmodule DotcomWeb.StopView do
...
+      aria_hidden?: false
+    })
+    |> add_aria_label(feature)
+  end
+
+  defp add_aria_label({:safe, [_, _, attributes, _, _, _, _, _, _] = html}, feature) do
+    attributes =
+      attributes
+      |> Enum.concat([
+        "aria-label = \"#{feature |> Atom.to_string() |> String.replace("_", " ")}\""
```
joshlarson, line 35: **Issue (non-blocking):** This is behaving in two kind of funky ways. 

One is that for ♿, it's now reading out `access` rather than `accessible`

👇 This branch 👇 
Image

👇 Main 👇 
Image

---

The other is that on this branch now these icons are being read out as groups that you can navigate into, so for instance, "access, group -> image -> end of, access, group" (at least this is how VoiceOver is reading it).

---

Marking this as non-blocking because while I do think it makes the output a bit more annoying, it doesn't hide anything, and can reasonably be treated as a follow-up. Totally up to you whether to fix this in this PR or punt it to a separate ticket that we can refine for later!

joshlarson (approved):
Found one additional issue, but I don't think it needs to block my approval!

As we will soon no longer say, let's "ship it Ship It"!

lib/dotcom_web/views/stop_view.ex
lvachon1: Let's refine later, this PR is big enough already 😆

PR #3174, opened 2026-05-11
WC[feat]: Update Foxboro Event Service from CR status summary

lvachon1:
## Scope

**Asana Ticket:** [⚽️⚠️ Remove Foxboro Event Service from CR status summary](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214574388398226?focus=true)

## Implementation

Added a function that returns if its a match day or not.  Used that function as part of the logic for CR-Foxboro status summary overhaul for Boston Stadium Trains

## Screenshots
Screenshot 2026-05-11 at 3 18 16 PM

## How to test

Unit tests!

### DIFF
--- a/lib/dotcom_web/components/system_status/commuter_rail_status.ex
+++ b/lib/dotcom_web/components/system_status/commuter_rail_status.ex
@@ -233,7 +233,7 @@ defmodule DotcomWeb.Components.SystemStatus.CommuterRailStatus do
             if is_map(assigns.status.status) do
               assigns.status.rows
             else
-              [%{label: "Normal Service", icon_atom: :normal}]
+              [%{label: ~t"Normal Service", icon_atom: :normal}]
             end
           else
             [%{label: "No Scheduled Service", icon_atom: :no_scheduled_service}]
--- a/lib/dotcom_web/components/system_status/commuter_rail_status.ex
+++ b/lib/dotcom_web/components/system_status/commuter_rail_status.ex
@@ -7,6 +7,7 @@ defmodule DotcomWeb.Components.SystemStatus.CommuterRailStatus do
 
   import DotcomWeb.Components, only: [bordered_container: 1]
   import DotcomWeb.Components.SystemStatus.StatusIcon, only: [status_icon: 1]
+  import DotcomWeb.WorldCupTimetableLive, only: [match_day?: 0]
 
   alias Dotcom.SystemStatus.CommuterRail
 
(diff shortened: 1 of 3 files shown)


test/dotcom_web/components/system_status/commuter_rail_status_test.exs
```diff
@@ -614,6 +614,78 @@ defmodule DotcomWeb.SystemStatus.CommuterRailStatusTest do
     end
   end
 
+  test "Shows no service for CR-Foxboro on non-match days" do
+    expect(Dotcom.Utils.DateTime.Mock, :now, 2, fn ->
+      ~D[2027-01-01]
```
thecristen: **issue(blocking)**: This function actually normally returns a `DateTime` and not a `Date`! So we should test with that.

lib/dotcom_web/live/world_cup_timetable_live.ex
```diff
@@ -69,6 +70,11 @@ defmodule DotcomWeb.WorldCupTimetableLive do
      ]}
   ]
 
+  def match_day?() do
+    date = @date_time_module.now()
```
thecristen: **suggestion**: this is a `Datetime` and I'm skeptical it'll successfully match against the `Date` that is being compared against. Luckily we can convert it to a `Date` pretty easily (and even a service date in the suggestion here)

```suggestion
    date = Dotcom.Utils.ServiceDateTime.service_date(@date_time_module.now())
```

lib/dotcom_web/components/system_status/commuter_rail_status.ex
```diff
@@ -212,7 +213,45 @@ defmodule DotcomWeb.Components.SystemStatus.CommuterRailStatus do
...
+          end,
+        rows:
+          if match_day?() do
+            if is_map(assigns.status.status) do
+              assigns.status.rows
+            else
+              [%{label: "Normal Service", icon_atom: :normal}]
+            end
+          else
+            [%{label: "No Scheduled Service", icon_atom: :no_scheduled_service}]
```
thecristen:
```suggestion
            [%{label: ~t"No Scheduled Service", icon_atom: :no_scheduled_service}]
```

lib/dotcom_web/components/system_status/commuter_rail_status.ex
```diff
@@ -212,7 +213,45 @@ defmodule DotcomWeb.Components.SystemStatus.CommuterRailStatus do
...
+            end
+          else
+            :no_scheduled_service
+          end,
+        rows:
+          if match_day?() do
+            if is_map(assigns.status.status) do
+              assigns.status.rows
+            else
+              [%{label: "Normal Service", icon_atom: :normal}]
```
thecristen:
```suggestion
              [%{label: ~t"Normal Service", icon_atom: :normal}]
```

thecristen (changes requested):
Super cool! I thought about suggesting you put the bulk of your logic in `Dotcom.SystemStatus.CommuterRail.commuter_rail_status/0` but then I realized the rows are computed separately... so I think where you have this change is the right choice.

test/dotcom_web/components/system_status/commuter_rail_status_test.exs
lvachon1: I've updated the tests to use DateTime sigils.  I'm not sure why the dates seemed to work on my machine, they certainly failed during CI.

thecristen (approved):

PR #3180, opened 2026-05-13
WC: fix description list items flowing into each other

jlucytan:
## Scope

**Asana Ticket:** [⚽️ Fix tablet-sized layout on "What You Can Bring on Boston Stadium Trains" page](https://app.asana.com/1/15492006741476/project/385363666817452/task/1214777883134932?focus=true)

## Implementation
simply forces the sides of the `dl--item` to be clear. I don't think there are any cases where we wouldn't want this to happen, so I've simply removed the breakpoint for this style entirely.

## Screenshots
Before (Window width 624px) 
image

After (same width)
image

## How to test
Visit the [What You Can Bring on Boston Stadium Trains page](http://localhost:4001/security/what-you-can-bring-boston-stadium-trains) on this branch and check the appearance of the description list at different breakpoints. 

Would also be good to check other pages with description lists to make sure their behavior hasn't regressed; of the ones I'm aware of: 
- Fares: [local](http://localhost:4001/fares) | [prod](https://www.mbta.com/fares)
- Main menu, in the fares section on mobile and the right side at larger breakpoints, "Most Popular Fares" 
image
image

### DIFF
--- a/assets/css/_description-list.scss
+++ b/assets/css/_description-list.scss
@@ -6,9 +6,7 @@
 
 .dl--item {
   // added this because at larger breakpoints items float into different rows
-  @include media-breakpoint-up(md) {
-    clear: both;
-  }
+  clear: both;
 }
 
 .dt {


lvachon1 (approved):
Looks great, thanks

joshlarson (approved):

PR #3182, opened 2026-05-13
refactor: Move `UpcomingDepartures` module to a separate folder

joshlarson:
## Scope

No ticket, but I'm planning on creating a bunch of new modules and thus files related to upcoming departures, so I figured it made sense to create a separate folder dedicated to those in particular.

I imagine that we'll probably want to do something similar for trip details, although I haven't thought that in-depth about that. My thought was to do that later, once we start doing this work for trip details, but I would be open to doing it now if that would be preferred.

## How to test

This should have no functional change, so [departures pages](http://localhost:4001/departures/?route_id=Green&direction_id=0&stop_id=place-pktrm) should still Just Work™

### DIFF
--- a/lib/dotcom/upcoming_departures/upcoming_trip_details.ex
+++ b/lib/dotcom/upcoming_departures/upcoming_trip_details.ex
@@ -4,6 +4,8 @@ defmodule Dotcom.UpcomingDepartures.UpcomingTripDetails do
   the stop specified, along with arrival times (or departure times when relevant).
   """
 
+  alias Dotcom.ScheduleFinder.TripDetails
+
   defstruct [
     :stop,
     :stops_after,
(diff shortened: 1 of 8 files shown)


thecristen (dismissed):
This is just reminding me that I hate that we include the folder name in the module name even when they're the same name 😂

lib/dotcom/upcoming_departures/upcoming_trip_details.ex
```diff
@@ -0,0 +1,20 @@
...
+
+  defstruct [
+    :stop,
+    :stops_after,
+    :stops_before,
+    :vehicle_info
+  ]
+
+  @type t :: %__MODULE__{
+          stop: nil | TripDetails.TripStop.t(),
```
thecristen, line 17: Dialyzer is flagging this because you haven't clarified where `TripDetails` comes from!

thecristen (approved):
Looks good, save for the failing type check.

lib/dotcom/upcoming_departures/upcoming_trip_details.ex
joshlarson: [Fixed](https://github.com/mbta/dotcom/pull/3182/commits/6ecbaa2d23e3e6249c9f3bd07c2f4159a7e8fb71) and [fixed](https://github.com/mbta/dotcom/pull/3182/commits/244767375400f94e7ae4964097c65cb9fc83ec4c)!

PR #3186, opened 2026-05-14
WC[feat]: Add alerts to Boston Stadium Trains timetable

lvachon1:
## Scope

**Asana Ticket:** [⚽️⚠️ Add alerts to Boston Stadium Trains timetable](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214574388398220?focus=true)

## Implementation

Added function to get alerts for a route (now that I think about it, this may be wrong)
Added alerts for CR-Foxboro to the World Cup Timetable Page
Added some CSS so the alerts fit better on the World Cup Timetable Page
Updated test to mock out repos used by this new feature

### Wide
Screenshot 2026-05-14 at 1 45 22 PM

### Narrow
Screenshot 2026-05-14 at 1 47 53 PM

## How to test

http://localhost:4001/schedules/bostonstadium

Confirm that the test alert shows as expected.

### DIFF
--- a/assets/css/_alerts.scss
+++ b/assets/css/_alerts.scss
@@ -359,8 +359,3 @@
     top: calc(#{$base-spacing} / 1.75);
   }
 }
-
-//World Cup alerts style tweak for wider pages.
-.wc-alerts {
-  width: 100%
-}
--- a/assets/css/_alerts.scss
+++ b/assets/css/_alerts.scss
@@ -360,3 +360,7 @@
   }
 }
 
+//World Cup alerts style tweak for wider pages.
+.wc-alerts {
+  width: 100%
+}
--- a/lib/dotcom_web/live/world_cup_timetable_live.ex
+++ b/lib/dotcom_web/live/world_cup_timetable_live.ex
@@ -9,7 +9,10 @@ defmodule DotcomWeb.WorldCupTimetableLive do
   import DotcomWeb.WorldCupTimetable.MatchLink, only: [match_link: 1, selected_match_banner: 1]
 
   on_mount {DotcomWeb.Hooks.Breadcrumbs, :world_cup_timetable}
+  @alerts_repo Application.compile_env!(:dotcom, :repo_modules)[:alerts]
 
+  @date_time Application.compile_env!(:dotcom, :date_time_module)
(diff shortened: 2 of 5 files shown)


assets/css/_alerts.scss
```diff
@@ -360,3 +360,7 @@
   }
 }
 
+//World Cup alerts style tweak for wider pages.
+.wc-alerts {
+  width: 100%
+}
```
joshlarson: **Suggestion (non-blocking):** Instead of adding this CSS class, would it work to just add the tailwind `w-full` class to the element?

lib/dotcom_web/live/world_cup_timetable_live.ex
```diff
@@ -71,10 +74,16 @@ defmodule DotcomWeb.WorldCupTimetableLive do
 
   @impl true
   def mount(_params, _session, socket) do
+    route = @routes_repo.get("CR-Foxboro")
+    alerts = Dotcom.Alerts.route_alerts(route.id)
```
joshlarson:
```suggestion
    now = @date_time.now()
    alerts = @alerts_repo.by_route_ids(["CR-Foxboro"], now)
```

☝️ Would save you from needing to add the new `route_alerts/1` function, and is how at least some other route-specific touchpoints get their alerts.

joshlarson (changes requested):
This looks great! Especially when I'm looking at the page with the right branch checked out 😅!

I do think it's worth using the existing `@alerts_repo` function, rather than adding a new one. Happy to discuss that though - if it turns out to be more effort than it's worth, or if you don't agree, let's talk!

assets/css/_alerts.scss
lvachon1: Good idea

lib/dotcom_web/live/world_cup_timetable_live.ex
lvachon1: Another good idea, I knew there had to be a function already written for this.

joshlarson (approved):
I snuck an additional commit on, because I noticed that when there are no alerts, it renders `<div class="w-full"></div>`, which (in the context of the flex that contains it) just adds some awkward space between the world cup callout and the match picker.

PR #3190, opened 2026-05-18
Ferry[feat]: Handle changes to ferry fares for F10 and F2H

lvachon1:
## Scope

**Asana Ticket:** [⛴️ 💸 Handle changes to ferry fares for F10 and F2H](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214773101169719?focus=true)

## Implementation

- Added calculate_ferry overrides for the Hinghum/Hull <-> Logan fare ($2.40)
- Added override for the new loop ferry (Boat-F10) fare ($2.40)  
- Fixed a bug I noticed with the East Boston ferry where it was charging $9.75 for a journey that should be $2.40 (Long Wharf to Logan).  I might want to make a better solution for that if it happens on other route combos.  

Unable to test the Boat-F10 stuff yet locally, will do so on dev green.

### Wrong fare on dev for E. Boston Ferry to Logan
Screenshot 2026-05-18 at 1 43 40 PM

### Fixed on branch
Screenshot 2026-05-18 at 1 43 13 PM

### $2.40 Fare for Hinghum/Hull to Logan
Screenshot 2026-05-18 at 1 43 01 PM

### $2.40 Fare for Loop Ferry Trips
...Coming Soon

## How to test

https://dev-green.mbtace.com/trip-planner?plan=h8QVX3VudXNlZF9kYXRldGltZV90eXBlxADEEl91bnVzZWRfd2hlZWxjaGFpcsQAxAhkYXRldGltZcQZMjAyNi0wNi0xMlQxNTo1MDowMC0wNDowMMQNZGF0ZXRpbWVfdHlwZcQIbGVhdmVfYXTEBGZyb22ExAhsYXRpdHVkZcQJNDIuMzY1NTc3xAlsb25naXR1ZGXECS03MS4wNjEyOcQEbmFtZcQNTm9ydGggU3RhdGlvbsQHc3RvcF9pZMQLcGxhY2Utbm9ydGjEBW1vZGVzhsQDQlVTxAVmYWxzZcQFRkVSUlnEBHRydWXEBFJBSUzEBWZhbHNlxAZTVUJXQVnEBWZhbHNlxA5fcGVyc2lzdGVudF9pZMQBMMQNX3VudXNlZF9GRVJSWcQAxAJ0b4TECGxhdGl0dWRlxAk0Mi4zNTk3ODnECWxvbmdpdHVkZcQJLTcxLjAyNzM0xARuYW1lxBxMb2dhbiBBaXJwb3J0IEZlcnJ5IFRlcm1pbmFsxAdzdG9wX2lkxApCb2F0LUxvZ2Fu

Look for Loop Ferry Fares
Look for East Boston Fares
Look for Hingham/Hull Fares

### DIFF
--- a/lib/dotcom/trip_plan/fares.ex
+++ b/lib/dotcom/trip_plan/fares.ex
@@ -109,8 +109,8 @@ defmodule Dotcom.TripPlan.Fares do
     origin_id = mbta_id(from.stop)
     destination_id = mbta_id(to.stop)
 
-    between_ids = between |> Enum.map(fn stop -> stop.name end)
-    [name: Fares.calculate_ferry(origin_id, destination_id, between_ids)]
+    between_names = between |> Enum.map(fn stop -> stop.name end)
+    [name: Fares.calculate_ferry(origin_id, destination_id, between_names)]
   end
 
   defp fare_filter_for_route(route, _, _, _) when mbta_shuttle?(route) do
--- a/lib/dotcom/trip_plan/fares.ex
+++ b/lib/dotcom/trip_plan/fares.ex
@@ -76,9 +76,10 @@ defmodule Dotcom.TripPlan.Fares do
   # All other Logan Express buses are $9.00
   def cents_for_leg(leg) when agency_name?(leg, "Logan Express"), do: 900
 
-  def cents_for_leg(%Leg{from: from, route: route, to: to}) when agency_name?(route, "MBTA") do
+  def cents_for_leg(%Leg{from: from, route: route, to: to, intermediate_stops: between} = leg)
+      when agency_name?(route, "MBTA") do
     route
-    |> fare_filter_for_route(from, to)
+    |> fare_filter_for_route(from, to, between)
     |> Keyword.put_new(:duration, :single_trip)
     |> Keyword.put_new(:reduced, nil)
     |> Fares.Repo.all()
@@ -89,7 +90,7 @@ defmodule Dotcom.TripPlan.Fares do
   # Non-transit legs don't have a fare
   def cents_for_leg(_), do: 0
 
-  defp fare_filter_for_route(route, from, to) when route.type == 2 do
+  defp fare_filter_for_route(route, from, to, _) when route.type == 2 do
     if mbta_id(route) == "CR-Foxboro" do
       [name: :foxboro, duration: :round_trip]
     else
@@ -104,17 +105,19 @@ defmodule Dotcom.TripPlan.Fares do
     end
   end
 
-  defp fare_filter_for_route(route, from, to) when route.type == 4 do
+  defp fare_filter_for_route(route, from, to, between) when route.type == 4 do
     origin_id = mbta_id(from.stop)
     destination_id = mbta_id(to.stop)
-    [name: Fares.calculate_ferry(origin_id, destination_id)]
+
+    between_ids = between |> Enum.map(fn stop -> stop.name end)
+    [name: Fares.calculate_ferry(origin_id, destination_id, between_ids)]
   end
 
-  defp fare_filter_for_route(route, _, _) when mbta_shuttle?(route) do
+  defp fare_filter_for_route(route, _, _, _) when mbta_shuttle?(route) do
     [name: :free_fare]
   end
 
-  defp fare_filter_for_route(route, from, _) when route.type == 3 do
+  defp fare_filter_for_route(route, from, _, _) when route.type == 3 do
     route_id = mbta_id(route)
     origin_id = mbta_id(from.stop)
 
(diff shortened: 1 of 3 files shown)


test/fares/fares_test.exs
```diff
@@ -90,7 +90,7 @@ defmodule FaresTest do
           end
 
         assert Fares.fare_for_stops(:ferry, origin_id, destination_id) == {:ok, expected_name},
-               "Unexpected result for #{origin_id} to #{destination_id}"
+               "Unexpected result for #{origin_id} to #{destination_id}, expected #{expected_name}"
```
joshlarson, line 93: Love this ✨

lib/fares/fares.ex
```diff
@@ -102,20 +104,31 @@ defmodule Fares do
...
-    :ferry_cross_harbor
+    :ferry_east_boston
   end
 
   def calculate_ferry(origin, destination)
       when "Boat-Long" in [origin, destination] and "Boat-Lewis" in [origin, destination] do
     :ferry_east_boston
   end
 
+  def calculate_ferry(origin, destination)
```
joshlarson: **Suggestion/Question (non-blocking):** I'm somewhat new to this code, but would it simplify this if we made a list of inner harbor stops, something like:
```elixir
@inner_harbor_stops [
  "Boat-Lewis",
  "Boat-Logan",
  "Boat-Long",
  # etc...
]
```

and then compared like

```elixir
def calculate_ferry(origin, destination)
    when origin in @inner_harbor_stops
    and destination in @inner_harbor_stops do
  :ferry_east_boston
end
```

---

I'm not sure that would work quite as simply as I just said, because 👇 trip starts and ends in the inner harbor, but goes out to Quincy in between, which does justify the $6.50 cost.... but I do think there's gotta be an easier way than trying to list out all of the possible origin/destination pairs.

Image

Maybe if we checked the origin and destination _and_ all intermediate stops? I dunno <-- and that's part of why this comment is non-blocking! 😅

joshlarson:
Hmmmm, related to my comment, would [this trip (first result)](http://dev-green.mbtace.com/trip-planner?plan=h8QVX3VudXNlZF9kYXRldGltZV90eXBlxADEEl91bnVzZWRfd2hlZWxjaGFpcsQAxAhkYXRldGltZcQZMjAyNi0wNi0xMlQxNTo1MDowMC0wNDowMMQNZGF0ZXRpbWVfdHlwZcQIbGVhdmVfYXTEBGZyb22ExAhsYXRpdHVkZctARS4c3l0YCcQJbG9uZ2l0dWRly8BRwyCWeHzpxARuYW1lxDBMb25nIFdoYXJmLCA2NiBMb25nIFdIUkYsIEJvc3RvbiwgTUEsIDAyMTEwLCBVU0HEB3N0b3BfaWTEAMQFbW9kZXOGxANCVVPEBWZhbHNlxAVGRVJSWcQEdHJ1ZcQEUkFJTMQFZmFsc2XEBlNVQldBWcQFZmFsc2XEDl9wZXJzaXN0ZW50X2lkxAEwxA1fdW51c2VkX0ZFUlJZxADEAnRvhMQIbGF0aXR1ZGXECTQyLjM1OTc4OcQJbG9uZ2l0dWRlxAktNzEuMDI3MzTEBG5hbWXEHExvZ2FuIEFpcnBvcnQgRmVycnkgVGVybWluYWzEB3N0b3BfaWTECkJvYXQtTG9nYW4=) actually only cost $2.40, or would it cost $9.75 (or twice that?) because it does a full round-trip through Hingham, Hull, and back?

Screenshot 2026-05-21 at 5 39 36 PM

joshlarson (commented):
Oops - I left a comment that alluded to one of my detailed comments here, and I thought that would pull my other comments out of the `Pending` state. Alas... this comment should do that though!

lib/fares/fares.ex
lvachon1: I'm not sure if there's a simple way to calculate ferry fares based on intermediate stops given the current code structure.  It seems only `to` `from` and `route` are the only things used in the fare calculation when no transfers are involved.

lvachon1: I have found an `intermediate_stops` field in the trip leg data that is a list of stop names in between from and to.  Using this I was able to exclude any long-way-around trips when the origin and destination are in the inner harbor.

lib/dotcom/trip_plan/fares.ex
```diff
@@ -104,17 +105,19 @@ defmodule Dotcom.TripPlan.Fares do
...
   end
 
-  defp fare_filter_for_route(route, from, to) when route.type == 4 do
+  defp fare_filter_for_route(route, from, to, between) when route.type == 4 do
     origin_id = mbta_id(from.stop)
     destination_id = mbta_id(to.stop)
-    [name: Fares.calculate_ferry(origin_id, destination_id)]
+
+    between_ids = between |> Enum.map(fn stop -> stop.name end)
+    [name: Fares.calculate_ferry(origin_id, destination_id, between_ids)]
```
joshlarson: **Suggestion (non-blocking):** These are stop names, rather than stop ID's.

```suggestion
    between_names = between |> Enum.map(fn stop -> stop.name end)
    [name: Fares.calculate_ferry(origin_id, destination_id, between_names)]
```

Ideally, these would actually be the stop names, but OTP client doesn't provide those (yet!).

joshlarson (approved):
Nice!

Hat-tip for the A+ assert message, and the function name `long_way_around_trip_logic`!

PR #3191, opened 2026-05-18
Ferry[fix]: Add George's Island to the list of F2H stops

lvachon1:
## Scope

https://mbta.slack.com/archives/C076FBBC21K/p1779129911512189

## Implementation

Service is back to Geroge's Island!  Put Boat-George back in the list of Boat-F2H stops on the timetable controller.

## Screenshots
Screenshot 2026-05-18 at 2 58 17 PM

## How to test

**Set environment to point at dev-green**

http://localhost:4001/schedules/Boat-F2H/timetable?date=2026-05-24&schedule_direction%5Bdirection_id%5D=0#direction-filter

Confirm that George's Island is in the timetable in the right location during the weekend, but does not show on days when there is no service.

### DIFF
--- a/lib/dotcom_web/controllers/schedule/timetable_controller.ex
+++ b/lib/dotcom_web/controllers/schedule/timetable_controller.ex
@@ -429,6 +429,7 @@ defmodule DotcomWeb.ScheduleController.TimetableController do
     "Boat-F2H" => [
       "Boat-Hingham",
       "Boat-Hull",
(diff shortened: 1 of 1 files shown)


jlucytan (approved):
I can see it on the inbound but can't on the outbound (which has no data for me, so no table exists), but don't see any reason to block this PR since it's probably unrelated.

PR #3192, opened 2026-05-18
fix: Subway status handling for multiple skipped stop alerts

joshlarson:
## Scope

**Asana Ticket:** [In status, handle when there's separate alerts for inbound/outbound stop closures at same time](https://app.asana.com/1/15492006741476/project/385363666817452/task/1214905054024584?focus=true)

This fixes two bugs:
- When a station closure and a unidirectional bypass are active at the same time, they should be combined on the homepage to show that the whole station is closed - e.g. `Stop Skipped: Tappan Street`, but they're instead being combined to show the unidirectional one - `Stop Skipped: Tappen Street (Westbound)`.
  - This only affects the Green line because [we add direction ID's to alerts differently for light rail versus heavy rail](https://github.com/mbta/dotcom/blob/dd08be281e7095548eb29863fcc6e9c380e21bec/lib/alerts/alert.ex#L157-L172).
- When there are bypasses and/or station closures affecting different stops on the same route, all affected directions are attached to all affected stops, rather than having each stop get its own affected direction.

Screenshot 2026-05-18 at 4 39 20 PM

This (dev-blue [alerts page](https://dev-blue.mbtace.com/alerts/subway) on the left; [homepage](https://dev-blue.mbtace.com/) on the right) shows both bugs in action. For the two alerts on the Green Line, the bidirectional alert at Coolidge Corner gets combined with the unidirectional alert at Tappan Street into just the Westbound direction, which is what shows on homepage status (right). For the two Orange Line alerts, it's the same phenomenon, except that the bidirectional and unidirectional alerts combine into "Both Directions" for _both_ stations (which then gets hidden, because "Both Directions" only shows when there are also unidirectional alerts).

## Implementation

- In `affected_stops`, instead of pulling all direction ID's at the beginning, pull stop / direction ID combos per-alert, and then combine them at the very end.

### dev-blue
Screenshot 2026-05-18 at 9 49 42 PM
The above scenario on dev-blue (middle), and this branch (right), along with the expanded display on the [dev-blue subway alerts page](https://dev-blue.mbtace.com/alerts/subway), which shows how closely the summary matches the actual alerts.

### prod
Screenshot 2026-05-18 at 9 57 34 PM
Current real-world data.

## How to test
Look at this branch against the prod API, and observe that Englewood is shown as `Westbound`, while Tappan Street is shown as `Both Directions`... in prod, right now both are shown implicitly as `Both Directions`.

Look at this branch against the dev-blue API, and compare against the screenshots above.

### DIFF
--- a/lib/dotcom/alerts/affected_stops.ex
+++ b/lib/dotcom/alerts/affected_stops.ex
@@ -20,25 +20,34 @@ defmodule Dotcom.Alerts.AffectedStops do
 
   @impl Behaviour
   def affected_stops(alerts, route_ids) do
-    stop_ids = alerts |> get_all_entities(:stop)
+    alerts
+    |> Stream.flat_map(fn alert -> affected_stops_for_alert(alert, route_ids) end)
+    |> Enum.group_by(& &1.stop.id)
+    |> Enum.map(fn {_stop_id, affected_stops} -> affected_stops |> combine_directions() end)
+  end
 
-    direction_ids = alerts |> get_all_direction_ids()
+  defp affected_stops_for_alert(alert, route_ids) do
+    affected_stop_ids = alert |> Alert.get_entity(:stop)
 
+    alert
+    |> Alert.get_entity(:direction_id)
+    |> Enum.flat_map(&convert_nil_to_all/1)
+    |> Enum.uniq()
+    |> outer_product(route_ids)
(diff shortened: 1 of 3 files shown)


lvachon1 (approved):
Looks all good to me

PR #3193, opened 2026-05-19
fix(tooltips): tooltip hover triangle 

jlucytan:
## Scope

**Asana Ticket:** [♿ Update tooltips to remain open if cursor is moved onto tooltip display](https://app.asana.com/1/15492006741476/project/555089885850811/task/1211709578469472)

## Implementation
Updates bootstrap tooltips to be able to be hovered over (currently disappears when you try), and have an invisible triangle to allow you to move your mouse more naturally into the tooltip. 

## Screenshots
Green triangle for illustration purposes only; this PR does not actually color the area
image

the tooltip for the parking icon is slightly off center but it was off-center before this PR as well
image

there should be no visual change to the tooltips
image

## How to test
Go to a line diagram page with connections, hover over the icons and observe that the tooltip appears and does not disappear when  you try to hover over the tooltip itself. 

[Orange line on local](http://localhost:4001/schedules/Orange/line) for convenience

### DIFF
--- a/assets/css/_bootstrap.scss
+++ b/assets/css/_bootstrap.scss
@@ -68,3 +68,19 @@
     max-width: 70vw;
   }
 }
+
+// Persistent tooltips for line diagrams
+.m-schedule-diagram [data-toggle='tooltip'] {
+  &:has(.c-icon__bus-pill--small)::before,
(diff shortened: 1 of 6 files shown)


lvachon1 (approved):
Works great!

PR #3195, opened 2026-05-19
fix(stops): Restore tooltips to stop page

jlucytan:
## Scope

**Asana Ticket:** [Restore tooltips on Stations page](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214903096193366)

## Implementation
Turns `show_tooltip?` on for `stop_feature_icon`, which, based on a quick search of the codebase, seems to be only used on the [stops index page](https://www.mbta.com/stops/subway)

## Screenshots
Tooltips are back!
image

## How to test
Go to [Stops/Stations page](http://localhost:4001/stops/subway) and verify that the tooltips are back


lvachon1 (approved):
Nice and simple, yay!

PR #3206, opened 2026-05-22
feat: Load ferry timetables from data

joshlarson:
## Scope

**Asana Ticket:** [🏖️⛴️ Update Quincy/Winthrop timetables to be data-based again](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214330808616829?focus=true)

This is also a soft blocker for the Harbor Loop Ferry work, since this will allow us to not hard-code data for that timetable.

## Implementation

- See comments in the code for more implementation notes.

## Screenshots

Screenshot 2026-05-22 at 6 34 15 PM

## How to test

Look at the [Quincy](http://localhost:4001/schedules/Boat-F7/timetable) and [Winthrop](http://localhost:4001/schedules/Boat-F6/timetable) timetables, and compare them against their production [Quincy](https://www.mbta.com/schedules/Boat-F7/timetable) and [Winthrop](https://www.mbta.com/schedules/Boat-F6/timetable) counterparts.


lvachon1 (approved):
This is some cool code, I like how generalized it is.  I bet we'll find an exception to these elegant rules at some point, but I can see this being useful for more than just these loops.  I think this kind of "merge schedules with different stops into a timetable" function would've made the weird WC CR shuttles trivial.

PR #3214, opened 2026-05-29
feat(ScheduleFinderLive): show LiveView error

thecristen:
## Scope

The other day @joshlarson and I were discussing what happens to the UI when we use a separate process to populate Upcoming Departures and that process happens to stop sending updates to the LiveView. Turns out the current behavior is super underwhelming and even problematic -- the display essentially stops updating, and the data eventually becomes wildly outdated.

This PR is not technically about that. But I did notice that LiveView itself adds a CSS class to the parent container, `.phx-error`, when the connection to the server is lost or fails. This PR adjusts Upcoming Departures to display its `"There was a problem loading upcoming departures"` state when this happens.

## Implementation

- Added a [Tailwind variant](https://v3.tailwindcss.com/docs/plugins#adding-variants) that's active when the element has a parent element with the `.phx-error` class.
- Use it to, when there's a LiveView error, both hide the potentially outdated UI and show the `"There was a problem loading upcoming departures"` message.

## How to test

image

It's actually very easy to play with in the browser. Open up the JS console, and toggle the error view by typing `liveSocket.disconnect()` or `liveSocket.connect()`!

### DIFF
--- a/assets/tailwind.config.js
+++ b/assets/tailwind.config.js
@@ -103,7 +103,8 @@ module.exports = {
       addVariant("phx-change-loading", [
         ".phx-change-loading&",
         ".phx-change-loading &"
-      ])
+      ]),
+      addVariant("phx-error", [".phx-error&", ".phx-error &"])
     ]),
     // Base styling for HTML elements
     plugin(({ addBase }) =>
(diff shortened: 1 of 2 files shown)


lvachon1 (approved):
Good catch, works great!

PR #3219, opened 2026-05-29
chore: extract Upcoming Departures into a LiveView

thecristen:
## Scope

I got tired of navigating the gigantic codebase of `DotcomWeb.ScheduleFinderLive`, and I think Upcoming Departures has enough of its own thing going on that it could be extracted. You can render child LiveViews within parent LiveViews. So now here is a new LiveView, dedicated to Upcoming Departures.

## Implementation

Most of the code was migrated as-is*. I moved a few smaller functional components into separate files, so that other files could easily import them. I added some more documentation in various spots.

*I changed the computation of the last trip time, slightly. Just felt like we could use a simpler query.
*I also slightly adjusted it to use `start_async/3` and `handle_async/3` instead of `assign_async/3`, felt easier to reason about and try to capture errors for

## Screenshots

No change! This should result in no visible changes from the current implementation.

## How to test

No idea, but there are a few new tests covering some of the upcoming departures functionality now!

### DIFF
--- a/lib/dotcom_web/components/components.ex
+++ b/lib/dotcom_web/components/components.ex
@@ -11,6 +11,7 @@ defmodule DotcomWeb.Components do
     endpoint: DotcomWeb.Endpoint,
     router: DotcomWeb.Router
 
+  import DotcomWeb.ViewHelpers, only: [mode_name: 1]
   import MbtaMetro.Components.Badge, only: [badge: 1]
   import MbtaMetro.Components.Button, only: [button: 1]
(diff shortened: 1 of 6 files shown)


joshlarson (approved):
Love it - `ScheduleFinderLive` was getting pretty unwieldy - this'll make it a lot easier to wrap our minds around it.

At first, I was a bit concerned that this would cause each SF page to open multiple sockets, but it looks like LiveView is smart enough to send updates for different LiveViews through the same socket!

PR #3220, opened 2026-05-29
fix(timetable): use date param when getting timetable for franklin shuttles

rudiejd:
## Scope

**Asana Ticket:** ad-hoc

## Implementation

Problem: 
The Franklin line World Cup gameday function head was using a hard-coded date of 6/13 instead of using the passed in from parameters

Solution:
Use date passed in from HTTP request parameters when rendering timetable for the Franklin line on World Cup game days.


thecristen (approved):
🎉

PR #3221, opened 2026-05-29
fix: Don't crash the timetable page when the shuttle route isn't available

joshlarson:
## Scope

This should fix the fact that the [Providence timetable on dev](https://dev.mbtace.com/schedules/CR-Providence/timetable) is unable to load right now.

## Implementation

- Replace `@routes_repo.get(shuttle_route_id)` with a skeleton `%Route{id: shuttle_route_id, type: 3}`. We don't need any info about the shuttle route other than its route ID - the schedule query happens elsewhere (and gracefully handles a missing route).

## Screenshots

Screenshot 2026-05-29 at 3 43 00 PM

## How to test

- Clear your cache (`Dotcom.Cache.Multilevel.flush` in an IEx shell), and then look at the [Providence timetable](http://localhost:4001/schedules/CR-Providence/timetable) using dev or prod data.
- On this branch, it will load exactly what you'd expect
- On `main`, it'll crash

### DIFF
--- a/lib/dotcom_web/controllers/schedule/timetable_controller.ex
+++ b/lib/dotcom_web/controllers/schedule/timetable_controller.ex
@@ -250,7 +250,7 @@ defmodule DotcomWeb.ScheduleController.TimetableController do
         } = conn
       )
       when route.id == "CR-Providence" do
-    shuttle_route = @routes_repo.get("Shuttle-CantonJunctionStoughton")
+    shuttle_route = %Route{id: "Shuttle-CantonJunctionStoughton", type: 3}
 
     shuttle_schedules =
       timetable_schedules(%{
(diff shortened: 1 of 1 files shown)


lib/dotcom_web/controllers/schedule/timetable_controller.ex
```diff
@@ -251,7 +250,7 @@ defmodule DotcomWeb.ScheduleController.TimetableController do
         } = conn
       )
       when route.id == "CR-Providence" do
-    shuttle_route = @routes_repo.get("Shuttle-CantonJunctionStoughton")
+    shuttle_route = %Route{id: "Shuttle-CantonJunctionStoughton", type: 3}
```
thecristen, line 251: Seems same should apply to the call for `route.id == "CR-Franklin"`

joshlarson: Yeah, fair. Turns out that that shuttle route does reliably exist in all environments (for now), which is why I only aimed for this one.

I'll fix it too - might as well not leave this as a surprise when transit data removes the shuttle route 😬

joshlarson: [Done!](https://github.com/mbta/dotcom/pull/3221/commits/9f098361fc1d01951152a98ecadc97e82f1520a1)

thecristen (approved):

PR #3223, opened 2026-06-01
fix(Dotcom.ServicePatterns): extra_service date handling

thecristen:
## Scope

[Slack thread](https://mbta.slack.com/archives/C076FBBC21K/p1780325860708129) - came up while testing the upcoming rating and noticing Jun 13 not being clear

## Implementation

It turns out services using the `:extra_service` typicality  might not use `added_dates` to describe its single date of service, but we were relying on parsing that to show special dates.

### Before - extended event service without date (or missing entirely)
image

image

### After - extended event service now annotated with Jun 13
image
image

## How to test

Using dev-blue data, you can visit schedule finders for most routes using [this service](https://api-dev-blue.mbtace.com/services/RTL20262-hmb26sp6-Saturday-01) on June 13.

### DIFF
--- a/lib/dotcom/service_patterns.ex
+++ b/lib/dotcom/service_patterns.ex
@@ -61,10 +61,8 @@ defmodule Dotcom.ServicePatterns do
     |> to_service_pattern()
   end
 
-  defp unwrap_multiple_holidays(
-         %{typicality: :holiday_service, added_dates: added_dates} = service
-       )
-       when length(added_dates) > 1 do
+  defp unwrap_multiple_holidays(%{typicality: extra, added_dates: added_dates} = service)
+       when extra in [:extra_service, :holiday_service] and length(added_dates) > 1 do
     for added_date <- added_dates do
       %{
         service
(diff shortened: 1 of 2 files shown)


lib/dotcom/service_patterns.ex
```diff
@@ -61,10 +61,8 @@ defmodule Dotcom.ServicePatterns do
     |> to_service_pattern()
   end
 
-  defp unwrap_multiple_holidays(
-         %{typicality: :holiday_service, added_dates: added_dates} = service
-       )
-       when length(added_dates) > 1 do
+  defp unwrap_multiple_holidays(%{typicality: extra, added_dates: added_dates} = service)
+       when extra in [:extra_service, :holiday_service] and length(added_dates) > 1 do
```
joshlarson: **Suggestion (non-blocking):** 👇 

```suggestion
  defp unwrap_multiple_holidays(%{typicality: typicality, added_dates: added_dates} = service)
       when typicality in [:extra_service, :holiday_service] and length(added_dates) > 1 do
```

joshlarson (approved):

PR #3228, opened 2026-06-03
refactor(UpcomingDeparturesLive): subscribe to data updates from a GenServer

thecristen:
## Scope

**Asana Ticket:** [📅🔎🎭 Extract upcoming departures calculation from LiveView into a separate module](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214587664555338?focus=true) (**update**: and [📅🔎🎭 De-duplicate Upcoming Departures GenServers](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214587664555339?focus=true) with commit f46c55a839402db49063f8e9d3074e5e2529dba1)

The linked ticket describes the background and prior art nicely!

## Implementation

> [!WARNING]
> There are a few other changes bundled into here, so I'll describe it commit-by-commit.
> 
> These changes felt needed after I ran into many surprises testing this PR. Because this PR makes `UpcomingDeparturesLive` receive data from another process, and tests run in their own process, and LiveView renders in a separate process, _and_ test mocks (via Mox's `expect/4` or `stub/3` calls) individually run in other processes.... I quickly found myself in a brittle mess of race conditions. The first two commits helped me get through that mess.

### 4722e989b1683c731d9cced1df401751f4f59a01 refactor(UpcomingDeparturesLive): use behaviour

> [!TIP]
>  This makes `upcoming_departures/1` able to be mocked in tests.

[This section of the Mox documentation](https://mox.hexdocs.pm/Mox.html#module-blocking-on-expectations) describes a technique to prevent the test from finishing executing before it has a chance to call the mock and meet the expectations. It basically entails sending a message from the `expect/4` function to the test, and asserting on receiving said message in the test.

The `upcoming_departures/1` function internally makes several different calls to the MBTA V3 API, which we can dutifully mock in the tests. But it proved hard to use this test technique with so many different mocks!

I went back and forth on whether to make `upcoming_departures/1` itself mockable or not. For tests which call code which calls `upcoming_departures/1`:
- If we don't mock it, those tests can use the same predictions/schedules/`now()` mocks to produce the same output even if we change the `upcoming_departures/1` internals
- If we do mock it, our tests can help enforce changes to the shape of the `upcoming_departures/1` output (because of the addition of the Behaviour), and we are free to change the internals without adjusting other mocks. It doesn't capture how particular predictions/schedules/`now()` states create certain outputs (though that's covered directly in the function's own tests)

I'm getting rambly. I decided to mock it, so that I'd only have one thing to mock for the associated LiveView's tests, and one place to fix Mox's treatment.

### 81826c6df8a4687d32c96e2fc27894e6c39f22a1 refactor(UpcomingDepartures.Processor) inline date

> [!TIP]
>  This means we don't have to mock `Dotcom.Utils.DateTime.now()` _every_ single time we deal with calling `upcoming_departures/1` in tests.

We were always calling `now/0` right before calling `upcoming_departures/1` anyways, so I figure it doesn't make any particular performance impact to make that call from inside of `upcoming_departures/1`  🤷🏼‍♀️ 

### 27c46de206613a0b38c9b971ce426ea3183400fd & 6da3ee7c10220a87323b776c5d4dc35d89f5e6b1 feat(UpcomingDepartures.Server): add worker process + fix(UpcomingDeparturesLive): unbreak all the tests

> [!TIP]
>  **This is the main part!** Instead of calling `upcoming_departures/1` in a LiveView, move that to a separate process.

#### About `Dotcom.UpcomingDepartures.Server`

It's a `GenServer` that's configured to be a long-lived process handling sending upcoming departures to other processes! 

- The actual subscribing/unsubscribing/broadcasting-to-subscribers pieces are actually handled ultimately by `Phoenix.PubSub` (via the `DotcomWeb.Endpoint` `subscribe/1`, `unsubscribe/1`, and `broadcast/3` functions) . 
- Each route/direction/stop is associated with a topic. e.g. the Orange line inbound from Back Bay is associated with the topic name `"departures:Orange:1:place-bbsta"`. (`Phoenix.PubSub` requires the topic to be a string)
- A single `Dotcom.UpcomingDepartures.Server` process is started for each unique topic.
  - On process startup, it stores the function we'll call to get upcoming departures, the topic name, and a list of unique subscribers (initialized to empty).
  - It handles periodically calling the stored function and broadcasting its output to the topic's subscribers
  - When someone new subscribes, it replies immediately with the current upcoming departures
  - While it does internally keep an updated list of current subscribers to a topic, it's only used for checking when we have no subscribers left. When that happens, we stop the process.

My theory was that because subscribing/unsubscribing to a topic is decoupled from the `Dotcom.UpcomingDepartures.Server` process, if `Dotcom.UpcomingDepartures.Server` happens to crash/ and get restarted, it could quickly resume broadcasting to subscribers since `Phoenix.Pubsub` will still know who's subscribed to which topic. 

However that'd mess up our subscriber count tracking! 

So I also added some logic to the `Dotcom.UpcomingDepartures.Server` `terminate/2` callback, to send _another_ broadcast to the topic indicating it's been terminated. In `UpcomingDeparturesLive` we can detect that event and clear the results from the screen before the data gets stale (here it should switch to "There was a problem loading upcoming departures").
 
## Screenshots

There should be no visual changes from this! Compare side-to-side with a prod or staging environment of your choice.

## How to test

I added some logging so we could see what's going on. We could remove this before going live if we'd like, as it'd get quite noisy :)
image

I hadn't quite gotten so far in my own testing, but I'd love to verify that terminating the server produces the expected output, I might probably come back and add that tomorrow.

### DIFF
--- a/lib/dotcom/application.ex
+++ b/lib/dotcom/application.ex
@@ -53,6 +53,7 @@ defmodule Dotcom.Application do
           Alerts.CacheSupervisor,
           {DynamicSupervisor, name: Dotcom.UpcomingDepartures.Supervisor},
           {Phoenix.PubSub, name: Dotcom.PubSub},
+          DotcomWeb.Presence,
           DotcomWeb.Endpoint
         ] ++
         if Application.get_env(:dotcom, :env) != :test do
--- a/lib/dotcom/application.ex
+++ b/lib/dotcom/application.ex
@@ -51,7 +51,7 @@ defmodule Dotcom.Application do
           Predictions.Supervisor,
           Alerts.BusStopChangeSupervisor,
           Alerts.CacheSupervisor,
-          {DynamicSupervisor, name: UpcomingDeparturesSupervisor},
+          {DynamicSupervisor, name: Dotcom.UpcomingDepartures.Supervisor},
           {Phoenix.PubSub, name: Dotcom.PubSub},
           DotcomWeb.Endpoint
         ] ++
--- a/lib/dotcom/application.ex
+++ b/lib/dotcom/application.ex
@@ -51,6 +51,7 @@ defmodule Dotcom.Application do
           Predictions.Supervisor,
           Alerts.BusStopChangeSupervisor,
           Alerts.CacheSupervisor,
+          {DynamicSupervisor, name: UpcomingDeparturesSupervisor},
           {Phoenix.PubSub, name: Dotcom.PubSub},
           DotcomWeb.Endpoint
         ] ++
--- a/lib/dotcom/upcoming_departures.ex
+++ b/lib/dotcom/upcoming_departures.ex
@@ -12,7 +12,7 @@ defmodule Dotcom.UpcomingDepartures do
 
   @spec subscribe(map()) :: :ok
   def subscribe(params) do
-    {:ok, topic} = topic_name(params)
+    topic = topic_name(params)
     :ok = DotcomWeb.Endpoint.subscribe(topic)
 
     topic
@@ -22,7 +22,7 @@ defmodule Dotcom.UpcomingDepartures do
 
   @spec unsubscribe(map()) :: :ok
   def unsubscribe(params) do
-    {:ok, topic} = topic_name(params)
+    topic = topic_name(params)
     :ok = DotcomWeb.Endpoint.unsubscribe(topic)
 
     worker = get_worker(topic)
@@ -33,7 +33,7 @@ defmodule Dotcom.UpcomingDepartures do
   end
 
   defp topic_name(%{route_id: route_id, direction_id: direction_id, stop_id: stop_id}) do
-    {:ok, "departures:#{route_id}:#{direction_id}:#{stop_id}"}
+    "departures:#{route_id}:#{direction_id}:#{stop_id}"
   end
 
   defp get_or_start_worker(topic) do
--- a/lib/dotcom/upcoming_departures.ex
+++ b/lib/dotcom/upcoming_departures.ex
@@ -12,10 +12,9 @@ defmodule Dotcom.UpcomingDepartures do
 
   @spec subscribe(map()) :: :ok
   def subscribe(params) do
-    :ok =
-      params
-      |> topic_name()
-      |> DotcomWeb.Endpoint.subscribe()
+    topic = topic_name(params)
+    :ok = DotcomWeb.Endpoint.subscribe(topic)
+    _ = DotcomWeb.Presence.track(self(), topic, "upcoming_departures", %{})
 
     params
     |> get_or_start_worker()
@@ -28,12 +27,6 @@ defmodule Dotcom.UpcomingDepartures do
       params
       |> topic_name()
       |> DotcomWeb.Endpoint.unsubscribe()
-
-    worker = get_worker(params)
-
-    if worker do
-      GenServer.cast(worker, {:unsubscribe, self()})
-    end
   end
 
   def topic_name(%{route_id: route_id, direction_id: direction_id, stop_id: stop_id}) do
--- a/lib/dotcom/upcoming_departures.ex
+++ b/lib/dotcom/upcoming_departures.ex
@@ -12,40 +12,44 @@ defmodule Dotcom.UpcomingDepartures do
 
   @spec subscribe(map()) :: :ok
   def subscribe(params) do
-    topic = topic_name(params)
-    :ok = DotcomWeb.Endpoint.subscribe(topic)
+    :ok =
+      params
+      |> topic_name()
+      |> DotcomWeb.Endpoint.subscribe()
 
-    topic
+    params
     |> get_or_start_worker()
     |> GenServer.cast({:subscribe, self()})
   end
 
   @spec unsubscribe(map()) :: :ok
   def unsubscribe(params) do
-    topic = topic_name(params)
-    :ok = DotcomWeb.Endpoint.unsubscribe(topic)
+    :ok =
+      params
+      |> topic_name()
+      |> DotcomWeb.Endpoint.unsubscribe()
 
-    worker = get_worker(topic)
+    worker = get_worker(params)
 
     if worker do
       GenServer.cast(worker, {:unsubscribe, self()})
     end
   end
 
-  defp topic_name(%{route_id: route_id, direction_id: direction_id, stop_id: stop_id}) do
+  def topic_name(%{route_id: route_id, direction_id: direction_id, stop_id: stop_id}) do
     "departures:#{route_id}:#{direction_id}:#{stop_id}"
   end
 
-  defp get_or_start_worker(topic) do
-    get_worker(topic) || start_worker(topic)
+  defp get_or_start_worker(params) do
+    get_worker(params) || start_worker(params)
   end
 
-  defp get_worker(topic) do
-    GenServer.whereis({:global, topic})
+  defp get_worker(params) do
+    GenServer.whereis({:global, params})
   end
 
-  defp start_worker(topic) do
-    case DynamicSupervisor.start_child(Dotcom.UpcomingDepartures.Supervisor, {Server, topic}) do
+  defp start_worker(params) do
+    case DynamicSupervisor.start_child(Dotcom.UpcomingDepartures.Supervisor, {Server, params}) do
       {:ok, pid} -> pid
       {:error, {:already_started, pid}} -> pid
     end
--- a/lib/dotcom/upcoming_departures.ex
+++ b/lib/dotcom/upcoming_departures.ex
@@ -45,7 +45,7 @@ defmodule Dotcom.UpcomingDepartures do
   end
 
   defp start_worker(topic) do
-    case DynamicSupervisor.start_child(UpcomingDeparturesSupervisor, {Server, topic}) do
+    case DynamicSupervisor.start_child(Dotcom.UpcomingDepartures.Supervisor, {Server, topic}) do
       {:ok, pid} -> pid
       {:error, {:already_started, pid}} -> pid
     end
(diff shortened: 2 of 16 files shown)


lib/dotcom/application.ex
```diff
@@ -51,6 +51,7 @@ defmodule Dotcom.Application do
           Predictions.Supervisor,
           Alerts.BusStopChangeSupervisor,
           Alerts.CacheSupervisor,
+          {DynamicSupervisor, name: UpcomingDeparturesSupervisor},
```
joshlarson: **Suggestion (non-blocking):** I know it doesn't really matter (hence non-blocking), but it was confusing me a bit that the supervisor wasn't namespaced where I was expecting `Dotcom.UpcomingDepartures` stuff to be.

```suggestion
          {DynamicSupervisor, name: Dotcom.UpcomingDepartures.Supervisor},
```

---

Also it kind of blows my mind that you can basically invent a module like this on the fly 🤯.

lib/dotcom/upcoming_departures/server.ex
```diff
@@ -0,0 +1,82 @@
...
+      @upcoming_departures_module.upcoming_departures(%{
+        direction_id: direction_id,
+        route: route,
+        stop_id: stop_id
+      })
+    end
+
+    send(self(), :refresh)
+
+    {:ok, %{departures_fn: departures_fn, topic: topic, subscribers: MapSet.new([])}}
```
joshlarson: Oooh I like `departures_fn`! It feels a lot more elegant than having `route`, `stop_id`, etc each tracked as part of state.

lib/dotcom/upcoming_departures.ex
```diff
@@ -4,11 +4,50 @@ defmodule Dotcom.UpcomingDepartures do
...
+
+    worker = get_worker(topic)
+
+    if worker do
+      GenServer.cast(worker, {:unsubscribe, self()})
+    end
+  end
+
+  defp topic_name(%{route_id: route_id, direction_id: direction_id, stop_id: stop_id}) do
+    {:ok, "departures:#{route_id}:#{direction_id}:#{stop_id}"}
```
joshlarson: **Suggestion:** Since this always returns `{:ok, String.t()}`, and all of its consumers know that and only handle that case, can we get rid of the `:ok` part, and just return the string?

lib/dotcom/upcoming_departures/server.ex
```diff
@@ -0,0 +1,82 @@
...
+  @routes_repo Application.compile_env!(:dotcom, :repo_modules)[:routes]
+  @upcoming_departures_module Application.compile_env!(:dotcom, :upcoming_departures_module)
+
+  def start_link(topic) do
+    GenServer.start_link(__MODULE__, topic, name: {:global, topic})
+  end
+
+  @impl GenServer
+  def init(topic) do
+    [_departures, route_id, direction_id, stop_id] = String.split(topic, ":")
```
joshlarson: **Suggestion (non-blocking):** Instead of constructing the topic string in `UpcomingDepartures.subscribe/1` and then having to parse that string here, what if we passed the params in here, and constructed the topic by calling `UpcomingDepartures.topic/1` from here.

That way, we wouldn't have to parse a string that we just constructed, and could instead just rely on the data that's already structured.

---

**Suggestion 2/ Question:** Given that the `Server` already knows who all of its subscribers are, do we actually need a PubSub, or could the Server just send messages to each of its subscribers directly?

If we got rid of the PubSub, then we wouldn't need to worry about topic strings, since the Server would just send messages directly to the relevant PID's, and my suggestion above wouldn't be relevant anymore.

(Another benefit of removing PubSub here is that then the subscriber only needs to handle one type of message, whereas now it needs to handle both PubSub-style messages and the plain-old `:subscribed` ones.)

test/dotcom/upcoming_departures/server_test.exs
```diff
@@ -0,0 +1,213 @@
...
+  end
+
+  test "gets restarted if it crashes", %{topic: topic} do
+    Phoenix.PubSub.subscribe(Dotcom.PubSub, topic)
+
+    {:ok, pid} = start_supervised({Server, topic})
+    assert ^pid = GenServer.whereis({:global, topic})
+    Process.exit(pid, :kill)
+    # need some time for it to restart
+    Process.sleep(100)
```
joshlarson: **Question (non-blocking):** Is there a way to do some kind of exponential backoff to check this at, say, 1ms, 10ms, and 100ms? I only ask because the `Process.sleep` introduces noticeable lag when running this test file.

It's not a huge deal for now - if it's not trivial, then I'd advocate for a follow-up investigation task, rather than blocking this PR on it.

lib/dotcom/upcoming_departures/processor.ex
```diff
@@ -19,34 +19,24 @@ defmodule Dotcom.UpcomingDepartures.Processor do
...
-          | {:before_service, UpcomingDeparture.t()}
-          | {:no_realtime, [UpcomingDeparture.t()]}
+  @impl Dotcom.UpcomingDepartures.Behaviour
   def upcoming_departures(%{
         direction_id: direction_id,
-        now: now,
         route: route,
         stop_id: stop_id
       }) do
+    now = @date_time.now()
```
joshlarson, line 39: **Question/Suggestion (non-blocking):** I don't oppose this on principle, but I thought we were migrating `UpcomingDepartures.Processor` towards being purely functional (e.g. a future task is passing `predicted_schedules` in as an argument, rather than calculating it in `upcoming_departures/1`). So I'm wondering if moving `now` into this function is something that we're going to wind up undoing in the next PR?

joshlarson (changes requested):
Love this change! And poking around with it running, I tried to get it to break, and couldn't, so this feels very promisigin! I left a few code-style suggestions and questions, but none of them are blocking.

The thing that _is_ blocking approval for me is the fact that something about this change caused half of the tests `ScheduleFinderLiveTest` to become super flaky - its pass rate is under 50%! I would be okay with just tagging the flaky tests as `@flaky` for now, or digging in a bit deeper to see what's actually going on with them.

Weirdly, the new tests are totally fine.

lib/dotcom/upcoming_departures/server.ex
```diff
@@ -0,0 +1,82 @@
...
+    {:noreply, state}
+  end
+
+  def handle_info(_, state), do: {:noreply, state}
+
+  @impl GenServer
+  def handle_cast({:subscribe, caller_pid}, state) do
+    Logger.notice("subscribing #{inspect(caller_pid)} to #{state.topic}")
+    new_state = add_subscriber(caller_pid, state)
+    send(caller_pid, {:subscribed, state.departures_fn.()})
```
joshlarson: **Suggestion:** The LiveView that receives this update doesn't have to do anything special due to this being a `:subscribed` message, but they do need to know that it's upcoming departures. 

(Imagine if a LiveView subscribed to more than one thing, but just got messages called `:subscribe` from each of them 😅)

```suggestion
    send(caller_pid, {:upcoming_departures, state.departures_fn.()})
```

lib/dotcom/upcoming_departures/processor.ex
thecristen: I'm not going to undo it in the next PR, but maybe we might in the PR after that?

test/dotcom/upcoming_departures/server_test.exs
thecristen: Adding a test helper function for this!  https://github.com/mbta/dotcom/pull/3228/commits/88dc7a0cbfbff93dd2e4fb623f7a818d7e7cf91d

lib/dotcom/upcoming_departures/server.ex
thecristen: Renamed in https://github.com/mbta/dotcom/pull/3228/commits/0a7670979c19ba00d1dbb0648fc315de48d6b665

lib/dotcom/upcoming_departures.ex
thecristen: Fair enough! Changed in https://github.com/mbta/dotcom/pull/3228/commits/a245ad8ee094b7ba9ad8bbc6f8a8d78a812be408

lib/dotcom/application.ex
thecristen: Updated in https://github.com/mbta/dotcom/pull/3228/commits/197c9b76083353d956aaa491517490f3d0f305a3

lib/dotcom/upcoming_departures/server.ex
thecristen: 1) I think that works fine! updated in https://github.com/mbta/dotcom/pull/3228/commits/fa4da5205081d9954f4119b026591557c15f6add

thecristen: 2. It does work without the PubSub (one of my many iterations did just that)!  Some aspects to mull over:
- (This one's the main one I consider important here) If we stored subscribers in the `Server` state, we'd lose them in the event of a `Server` restart (unless we did our own effort to save them somewhere else and restore them or something)
- As far as I understand PubSub's pretty dang optimized and performant for what we're using it for here, as opposed to having to loop through however many subscribers in a given GenServer process (I know the non-PubSub version seems to perform okay enough, though)
- PubSub handles clearing dead processes from its subscribers without any work on our part
- If we ever dared create a setup where our nodes could talk to each other, PubSub would _just work_ across them

thecristen: @joshlarson I added a few more commits to update this to go all-in on PubSub and use Phoenix.Presence to track subscribers and help us detect when there are no more subscribers. I'd love for your feedback there!

joshlarson (approved):
My only blocking comment has been resolved 🎉 !

lib/dotcom/upcoming_departures/processor.ex
joshlarson: Fair enough! If we do decide to move it back, we'll move it back.

PR #3229, opened 2026-06-04
refactor: Extract timetable template to its own component

joshlarson:
## Scope

The end goal of this work is to have a timetable component whose invocation will be (almost) as simple as `<.timetable timetable={@timetable} />`. That will make it easier to render the Harbor Loop Ferry with two timetables on the same page, and give us more flexibility in how we render different types of timetable pages in the future.

This PR goes about halfway - it splits out a separate component for the timetable (actually two... more on that in a moment), and adds more structure to the `%Timetable{}` struct, introduced in https://github.com/mbta/dotcom/pull/3206.  The remaining refactor work is to pull the last bits of non-user-interaction data (i.e. alerts) into the `Timetables` module. Once we do that, rendering a different layout for the Harbor Loop Ferry should be almost trivial.

### Two timetables?

We already have [two different timetable implementations](https://github.com/mbta/dotcom/blob/66909432d2e5def198cd8eca3eb05fa22b815fb9/lib/dotcom_web/templates/schedule/_timetable.html.heex#L217-L253), although they're currently kind of muddled together, using the somewhat opaquely-named `use_pdf_schedules?` flag.
- When `use_pdf_schedules?` is true, then the timetable template expects `@timetable_schedules`, `@header_stops`, and `@header_schedules` to be set.
- When `use_pdf_schedules?` is false, then the timetable still expects `@header_stops` and `@header_schedules` to be set, but it gets its cell data from `@trip_schedules` and `@trip_messages` instead (along with data from a few other fields, like `@track_changes`.

A core part of this PR makes is to explicitly separate those two implementations into `<.linear_timetable />` (so named because it can't support loops or trips that visit the same stops in different orders, and `<.timetable />`, which has no such constraint, but has fewer features (e.g. it doesn't support track changes... for now!). Most of the rest of this PR is cleanup around the edges of the contract between `<.timetable />` and the template that calls it, as well as the `%Timetable{}` struct that gets passed in.

**Asana Ticket:** [⛴️ Make timetable its own component(s)](https://app.asana.com/1/15492006741476/task/1215424202403217?focus=true)

## Implementation

> [!WARNING]
> This PR is quite large, so I'll walk through it commit-by-commit.

- [refactor: Extract timetable rendering to a component](https://github.com/mbta/dotcom/pull/3229/changes/d32132374a09171bad6b02d807e2ac1466dfa9f6)
- [refactor: Convert `_timetable_schedule` template into `<.timetable_cell />` component](https://github.com/mbta/dotcom/pull/3229/changes/8350af3bbd19b741516ce0b42cb906546371f40a)
- [refactor: Change `use_pdf_timetable?` to `linear_timetable?` and swap the sense](https://github.com/mbta/dotcom/pull/3229/changes/495c2acbcd897350d2d1dff37454ad0754c08730)

- [refactor: Replace `content_tag` invocations with regular HEEx code](https://github.com/mbta/dotcom/pull/3229/changes/0930f9e8ed6c7ef9350eb97523a0bebb05544eab)
- [refactor: Extract `<.timetable_cell_wrapper />` and remove unnecessary variables](https://github.com/mbta/dotcom/pull/3229/changes/3e6fc8ea5843d8e59e1df5c0f4407703081d648b)
- [refactor: Use `@value` syntax instead of `assigns[:value]`](https://github.com/mbta/dotcom/pull/3229/changes/429fa70397412e187532333f05f03912d12ee2fb)
  - The `assigns[:value]` invocations were necessary when this code was in the template, but the way I pulled them into the component meant they were guaranteed to always be assigned (even if they were assigned as `nil`), meaning that `@value` should work just fine now.
- [fix: Map lookup syntax error](https://github.com/mbta/dotcom/pull/3229/changes/81dd350139f7d9c5593b6f3e221dfb3418cc699b)
  - Well, not exactly a syntax error, but the `|| %{}` was inside of the lookup `[brackets]` - i.e. `map[key || %{}]`. I'm guessing it was originally intended to be `map[key] || %{}`, but the fact that everything already worked (and that `key` in this context was guaranteed never to be falsey), means that it was easier to just get rid of the `|| %{}` entirely.

- [refactor: Split `<.timetable />` and `<.linear_timetable />` into separate components](https://github.com/mbta/dotcom/pull/3229/changes/3814f4d6a8ebed484dddc5c1fe2cac44e2135a0a)
- [refactor: Remove portions of timetable components that aren't needed](https://github.com/mbta/dotcom/pull/3229/changes/1fd723dde68ea1873e33d7e52b8a57fed6806d12)
- [refactor: Compute `cell_background` in a function and remove a for loop](https://github.com/mbta/dotcom/pull/3229/changes/e743e18d720aa5197fb1922624fd2b137675bc57)
- [refactor: Extract `<.stop_header_cell />`](https://github.com/mbta/dotcom/pull/3229/changes/d8e4bbd4ad7aad52b8f853ed9068a71a93c0f7ed)
- [refactor: Iterate over `@timetable_schedules` instead of `@header_stops`](https://github.com/mbta/dotcom/pull/3229/changes/548007965032a6e0631ec340b4a1e4698676575b)

- [refactor: Assign `timetable` directly to the `conn` instead of `timetable_schedules`](https://github.com/mbta/dotcom/pull/3229/changes/523880669ab17510d5f4cdced42dc92f2ae8cac3)
- [cleanup: Remove non-ferry section, and add a warning comment](https://github.com/mbta/dotcom/pull/3229/changes/523b661f77c7b4cdfe0293b9ae9b08b3668a5f69)
  - We will eventually want to add this (or something like it) back, but I didn't want to deal with speculative code right now.
- [refactor: Use `header_trips` instead of `header_schedules` in the timetable](https://github.com/mbta/dotcom/pull/3229/changes/f2e26c61ff4b81de56296ec44f08c66644cd59ad)
- [refactor: Make timetable rows their own struct](https://github.com/mbta/dotcom/pull/3229/changes/3498e3dd2d0fff96e72af0ac158081b9b4a82dbe)
- [refactor: Move stop header calculation to timetable backend](https://github.com/mbta/dotcom/pull/3229/changes/606399525b664f6c696db8d6a5e1356739c7a5c6)
- [refactor: Make `@header_schedules` not required in timetable template](https://github.com/mbta/dotcom/pull/3229/changes/6be4277d9b745a9628982f8d07a73c41421ba152)
- [cleanup: Remove `stop_id` field from `%Timetable.Cell{}`](https://github.com/mbta/dotcom/pull/3229/changes/20901a92870e7f91d1438afd50baf2d415ef8929)
- [refactor: Move trip header calculation to timetable backend](https://github.com/mbta/dotcom/pull/3229/changes/7a14937333c1a7a1dfec88a54370608f6ce11650)
- [refactor: Rename variables to avoid calling everything trips](https://github.com/mbta/dotcom/pull/3229/changes/6073b44609fc6d4a14ef38e9ea0165fdb54e9527)
  - Purely a code-cosmetic change, but when the `Timetables` module was first introduced, it didn't deal with `%Schedules.Trip{}` structs at all, so `trips` was a relatively non-confusing name for a list of `%Schedule{}`-lists (the schedules for a trip). This PR added handling for actual trip structs, so finding a new name for the variable formerly known as `trips` became much more virtuous.

## Screenshots

No visual changes.

## How to test

This should have no actual changes, so take a look at some [timetables](http://localhost:4001/schedules/CR-Providence/timetable) [for](http://localhost:4001/schedules/CR-Needham/timetable) [commuter](http://localhost:4001/schedules/CR-Franklin/timetable) [rail](http://localhost:4001/schedules/CR-Needham/timetable) and [timetables](http://localhost:4001/schedules/Boat-F2H) [for](http://localhost:4001/schedules/Boat-F4) [ferry](http://localhost:4001/schedules/Boat-Lynn) (don't forget the [loop](http://localhost:4001/schedules/Boat-F6) [ferry](http://localhost:4001/schedules/Boat-F7) [routes](http://localhost:4001/schedules/Boat-F10/timetable)) and compare them to their prod equivalents.

> [!NOTE]
> The Harbor Loop Ferry doesn't start until [June 29](http://localhost:4001/schedules/Boat-F10/timetable?date=2026-06-29#date-filter), and doesn't run on [weekends](http://localhost:4001/schedules/Boat-F10/timetable?date=2026-07-11#date-filter), so the fact that it doesn't a show a timetable for today is a feature, not a bug.

---
  - https://app.asana.com/0/0/1215424202403217


thecristen (changes requested):
There are three broken tests from the old timetable view test file, please fix! 
I think the rest looks and works great. 🚀

thecristen (approved):
✨

PR #3230, opened 2026-06-04
SF2.0[fix]: Skipped Stops disappear when the schedule drifts into the past, even if the vehicle is still upstream

lvachon1:
## Scope

**Asana Ticket:** [📅🔎 Skipped Stops disappear when the schedule drifts into the past, even if the vehicle is still upstream](https://app.asana.com/1/15492006741476/project/555089885850811/task/1213813725381825?focus=true)

## Implementation

Removed time filtering from predicted schedules on trip details so that we continue to show skipped stops that have predictions in the past, but the vehicle hasn't skipped yet on this trip

### Local ------------------------- Dev
Screenshot 2026-06-04 at 1 40 21 PM

## How to test

Find a route with skipped stops, preferably a bus.  Find a stop right before the skipped stops, watch a vehicle approach and confirm that the skipped stops do not disappear while the vehicle is still on it's way to your stop.

If there's a Red Sox home game, the 8, 19, 60, and 65 will be skipping stops around Fenway and you can use this link and vary the route between those to find a bus that is coming soon.  I got lucky, these seem to be rather infrequent busses.

https://dev.mbtace.com/departures/?route_id=65&direction_id=1&stop_id=1806

### DIFF
--- a/lib/dotcom/upcoming_departures/processor.ex
+++ b/lib/dotcom/upcoming_departures/processor.ex
@@ -296,7 +296,6 @@ defmodule Dotcom.UpcomingDepartures.Processor do
 
     predicted_schedules =
       PredictedSchedule.group(predictions, schedules)
-      |> Enum.reject(&past_schedule?(&1, now))
 
     vehicle =
       predicted_schedules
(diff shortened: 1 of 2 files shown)


joshlarson (dismissed):
I do think this change deserves a test in `UpcomingDeparturesTest`.

thecristen (approved):
And thanks for the test addition!

PR #3236, opened 2026-06-08
[Alerts] Fix alerts with URLs which have query parameters

lvachon1:
## Scope

**Asana Ticket:** [⚠️ Fix alerts with URLs which have query parameters](https://app.asana.com/1/15492006741476/project/555089885850811/task/1213810651947031?focus=true)

## Implementation
Updated the url detecting regex to look for query param characters(`?` and `&`), url_escaped characters(ie `%20`), and `#`
Added a test for the new functionality

**Note:**  I'm not going to expand the scope of this ticket, but our URL regex pattern is a bit conservative in what it allows.  As long as we don't need to link to any [exotic domains](https://en.wikipedia.org/wiki/List_of_English-language_generic_Internet_top-level_domains) it'll be fine.  We should take a look at this sometime.

### Branch
Screenshot 2026-06-08 at 1 51 28 PM

### Dev-Blue
Screenshot 2026-06-08 at 1 51 36 PM

## How to test

View an alert with a URL that has any combination of query parameters(`?foo=bar`), url escaped characters(`hello%20world`), and/or an anchor (`index.html#part3`)

I currently have one setup on dev here:

http://localhost:4001/schedules/CR-NewBedford/alerts

Confirm that the URL is fully parsed (correctly) and leads to the correct URL when clicked.

### DIFF
--- a/lib/alerts/url_parsing_helpers.ex
+++ b/lib/alerts/url_parsing_helpers.ex
@@ -1,7 +1,7 @@
 defmodule Alerts.URLParsingHelpers do
   @moduledoc "Helpers to parse out the url from a string and create an html link"
 
-  @url_regex ~r/(https?:\/\/)?([\da-z\.-]+)\.([a-z]{2,6})([\/\w\.-]*)*\/?/i
+  @url_regex ~r/(https?:\/\/)?([\da-z\.-]+)\.([a-z]{2,6})([\/\w\.-\?&%#]*)*\/?/i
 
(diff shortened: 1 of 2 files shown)


thecristen (changes requested):
Can you please add a test case or two to `Alerts.URLParsingHelpersTest` to cover the scenario from the ticket?

lvachon1:
Test added!

thecristen (approved):

PR #3239, opened 2026-06-08
cleanup: Remove unnecessary `clipPath` from realtime icon SVG file

joshlarson:
## Scope

While reviewing https://github.com/mbta/dotcom/pull/3228, I noticed that running `mix test test/dotcom_web/live/upcoming_departures_live_test.exs` generated a bunch of warnings like 👇
```heex
warning: Duplicate id found while testing LiveView: clip0_472_18139

    <clipPath id="clip0_472_18139"><rect width="12" height="12" fill="white"></rect></clipPath>
    <clipPath id="clip0_472_18139"><rect width="12" height="12" fill="white"></rect></clipPath>
    <clipPath id="clip0_472_18139"><rect width="12" height="12" fill="white"></rect></clipPath>
    <clipPath id="clip0_472_18139"><rect width="12" height="12" fill="white"></rect></clipPath>
    <clipPath id="clip0_472_18139"><rect width="12" height="12" fill="white"></rect></clipPath>


LiveView requires that all elements have unique ids, duplicate IDs will cause
undefined behavior at runtime, as DOM patching will not be able to target the correct
elements.
```

## Implementation

- Turns out, that clip path was never necessary - it just cropped the 12x12 image into a 12x12 rectangle!
- I also got rid of the hard-coded height and width in the SVG definition - both places where the icon is used (stop page and departures page) already set the size themselves (incidentally to 12px at default font size)

## Screenshots

Screenshot 2026-06-08 at 2 53 22 PM

No real change, and no real need for a screenshot, but this is what I was looking at while making sure it worked, so why not?

## How to test

Look at a [departures page](http://localhost:4001/departures/?route_id=39&direction_id=1&stop_id=11131) and a [stop page](http://localhost:4001/stops/11131), and make sure the icon still looks right.


jlucytan (approved):
looks good! Thanks for removing it, the warnings can be distracting in the console for sure.

PR #3249, opened 2026-06-11
feat(timetable): Upcoming Changes for commuter rail

jlucytan:
## Scope

**Asana Ticket:** [🟣⚽️ Upcoming Changes widget for Commuter Rail](https://app.asana.com/1/15492006741476/project/555089885850811/task/1215166490913220)

## Screenshots
image

Ferry still has banner alerts 
image

## How to test
Changes are on dev-blue at the moment. Alerts are from Alerts Manager staging. 

[Newburyport timetable](https://dev-blue.mbtace.com/schedules/CR-Newburyport/timetable) on dev-blue for convenience.

Please verify the following: 
- Alerts that are only today do NOT appear in Upcoming Changes 
- Alerts that include today, but end in the future DO appear in Upcoming Changes
- Alerts that begin in the next 7 days appear in Upcoming Changes
- Alerts that begin after the next 7 days do NOT appear in Upcoming Changes
  -  if such alerts exist, "# later change(s) >" appears as a link at the bottom of the changes list
- Alerts are sorted in ascending order by start time, then end time
- Widgets stack for mobile widths

On ferry page (which also uses the timetable controller): 
- banner alerts still appear

### DIFF
--- a/lib/util/util.ex
+++ b/lib/util/util.ex
@@ -23,6 +23,26 @@ defmodule Util do
     # to_local_time(utc_now_fn.())
   end
 
+  @doc "Allows you to compare DateTime to nil values.
+  By default, nil is an indefinite future."
+  @spec safe_time_compare(DateTime.t(), DateTime.t(), boolean()) :: :lt | :eq | :gt
+  def safe_time_compare(a, b, nil_is_greater? \\ true) do
+    cond do
+      a == nil ->
+        if b == nil do
+          :eq
+        else
+          if nil_is_greater?, do: :gt, else: :lt
+        end
+
+      b == nil ->
+        if nil_is_greater?, do: :lt, else: :gt
+
+      true ->
+        DateTime.compare(a, b)
+    end
+  end
+
   @spec time_is_greater_or_equal?(
           DateTime.t() | NaiveDateTime.t(),
           DateTime.t() | NaiveDateTime.t()
--- a/lib/util/util.ex
+++ b/lib/util/util.ex
@@ -23,7 +23,7 @@ defmodule Util do
     # to_local_time(utc_now_fn.())
   end
 
-  @doc "Allows you to compare DateTime to nil values.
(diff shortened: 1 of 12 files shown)


lib/util/util.ex
```diff
@@ -23,6 +23,25 @@ defmodule Util do
     # to_local_time(utc_now_fn.())
   end
 
+  @doc "Handles comparison of DateTime to nil values.
+  By default, nil is an indefinite future."
+  @spec safe_time_compare(DateTime.t(), DateTime.t(), boolean()) :: :lt | :eq | :gt
```
thecristen: Oh I like that! The typespec needs to be updated to reflect that the inputs might be `nil` here.
```suggestion
  @spec safe_time_compare(DateTime.t() | nil, DateTime.t() | nil, boolean()) :: :lt | :eq | :gt
```

test/dotcom_web/components/system_status/commuter_rail_upcoming_changes_test.exs
```diff
@@ -0,0 +1,143 @@
...
+      later = DateTime.shift(today, day: 9)
+
+      alert1 = Factories.Alerts.Alert.build(:alert, active_period: [{later, nil}])
+      alert2 = Factories.Alerts.Alert.build(:alert, active_period: [{later, nil}])
+
+      # EXERCISE
+      rendered_component = render_upcoming_changes([alert1, alert2])
+
+      # VERIFY
+      assert String.contains?(
```
thecristen: **nit(non-blocking)**: I'd recommend using a `data-test` attribute on the `<.link />` element and verifying its existence that way, rather than checking for the text.

lib/util/util.ex
thecristen: Elixir is... not a strict enough language to make anything fail thanks to that 😆

thecristen (approved):
I love how many different existing things you've been able to reuse here, in addition to your new contributions! 🎉

PR #3251, opened 2026-06-11
fix: Allow `PredictedSchedule.Collection` to correctly handle Prediction ID changes

joshlarson:
Prediction ID changes are often represented as an ADD followed by a REMOVE. In the previous implementation, the fact that the REMOVE came last caused the prediction to fully get deleted. This now allows the Collection to store multiple predictions per {trip_id, stop_sequence}, and only removes the one that was intended to be removed.

Prediction ID changes are unusual... except at certain stations, like Forest Hills, where they happen all the time when a trip is assigned to a particular track!

## Scope

**Asana Ticket:** [📅🔎 Correctly handle when a Prediction's ID changes](https://app.asana.com/1/15492006741476/project/1210709545912056/task/1215634938817006?focus=true)

## Implementation

The fix is in `PredictedSchedule.Collection` - when a Prediction's ID changes, the API handles that by issuing an ADD event for the new ID, and then a REMOVE event for the old one. In our old implementation, we were treating predictions as unique per `{trip_id, stop_sequence}`, which meant that that second REMOVE was deleting the prediction that we'd just ADDed.

This fixes that by changing the underlying data structure in the collection per `{trip_id, stop_sequence}` to be something like 👇

```elixir
%{
  schedule: schedule,
  predictions: %{
    prediction_id_1 => prediction_1,
    prediction_id_2 => prediction_2
  }
}
```

That way, during that brief moment when a new prediction is ADDed with the new ID, it keeps track of both so that it can REMOVE the right one.

## Screenshots

Screenshot 2026-06-11 at 12 26 29 PM

## How to test

Watch [the Forest Hills Orange Line departures](http://localhost:4001/departures/?route_id=Orange&direction_id=1&stop_id=place-forhl). If you hang out on that page for long enough, you'll see trips start to disappear, typically a few minutes before departure. Well, you'll see that issue in [dev](https://dev.mbtace.com/departures/?route_id=Orange&direction_id=1&stop_id=place-forhl), but not [prod](https://www.mbta.com/departures/?route_id=Orange&direction_id=1&stop_id=place-forhl) (which doesn't have the latest streaming code yet), and you hopefully _won't_ see it when running this branch.

### DIFF
--- a/lib/predicted_schedule/collection.ex
+++ b/lib/predicted_schedule/collection.ex
@@ -74,7 +74,7 @@ defmodule PredictedSchedule.Collection do
 
   @type key_t() :: {Schedules.Trip.id_t(), non_neg_integer()}
   @type entry_t() :: %{
-          schedule: Schedules.Trip.t(),
+          schedule: Schedules.Schedule.t(),
           predictions: %{Predictions.Prediction.id_t() => Predictions.Prediction.t()}
         }
   @type t() :: %{
--- a/lib/predicted_schedule/collection.ex
+++ b/lib/predicted_schedule/collection.ex
@@ -186,9 +186,6 @@ defmodule PredictedSchedule.Collection do
             %PredictedSchedule{schedule: schedule, prediction: prediction}
           end)
       end
-
-      # dbg(schedule)
-      # dbg(predictions)
     end)
   end
 
--- a/lib/predicted_schedule/collection.ex
+++ b/lib/predicted_schedule/collection.ex
@@ -73,9 +73,13 @@ defmodule PredictedSchedule.Collection do
   """
 
   @type key_t() :: {Schedules.Trip.id_t(), non_neg_integer()}
+  @type entry_t() :: %{
+          schedule: Schedules.Trip.t(),
+          predictions: %{Predictions.Prediction.id_t() => Predictions.Prediction.t()}
+        }
   @type t() :: %{
-          base: %{key_t() => PredictedSchedule.t()},
-          populated: %{key_t() => PredictedSchedule.t()}
+          base: %{key_t() => entry_t()},
+          populated: %{key_t() => entry_t()}
         }
 
   @spec new([Schedules.Schedule.t()]) :: t()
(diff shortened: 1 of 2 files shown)


lib/predicted_schedule/collection.ex
```diff
@@ -73,9 +73,13 @@ defmodule PredictedSchedule.Collection do
   """
 
   @type key_t() :: {Schedules.Trip.id_t(), non_neg_integer()}
+  @type entry_t() :: %{
+          schedule: Schedules.Trip.t(),
```
thecristen: Wrong type, this is `Schedules.Schedule.t()`

lib/predicted_schedule/collection.ex
```diff
@@ -120,15 +124,15 @@ defmodule PredictedSchedule.Collection do
   def delete_prediction(%{populated: populated} = collection, prediction) do
     key = key(prediction)
 
-    new_map =
-      populated
-      |> Map.get(key)
-      |> case do
-        %PredictedSchedule{schedule: schedule} = ps when schedule != nil ->
-          populated |> Map.put(key, %PredictedSchedule{ps | prediction: nil})
+    entry = populated |> Map.get(key)
```
thecristen: question: Should we consider a possibility where `entry` resolves to `nil`?

thecristen (approved):
Awesome! A small comment/question and then you can fix CI

lib/predicted_schedule/collection.ex
joshlarson: Huh - yeah we should.

lib/predicted_schedule/collection.ex
joshlarson: Woops - kinda bummed the type checker didn't notice that....

joshlarson: [Fixed](https://github.com/mbta/dotcom/pull/3251/commits/d2ab99ec8b288f05921fe4cab5548ce555392a70)

lib/predicted_schedule/collection.ex
joshlarson: [Done](https://github.com/mbta/dotcom/pull/3251/commits/a7ec85d4d74c0daa836227209ba52e899ebc5f0f)

PR #3252, opened 2026-06-11
fix(Departures): Handle the case when a prediction has a nil trip

joshlarson:
Sometimes (especially for added trips), a prediction will become available before its associated trip is. We handle that reasonably gracefully - we create the prediction like `%Prediction{trip: nil}`, and then pass it thru to the rest of the system.

The problem is that some of its consumers depend on the trip information, like its headsign, so we do actually need to get ahold of the right trip eventually. For predictions partway through a trip, they often jitter by a second or two, so we get events that cause us to re-parse the prediction, and re-fetch the trip, but for predictions near the beginning, they can be stable for a long time before jittering.

This PR addresses that by having `UpcomingDepartures.Processor` fetch the trip itself, and use that instead of the prediction's trip for things like headsign. That way, we'll get trip info as soon (on potentially a five second delay due to `UpcomingDepartures.Server`'s refresh) as its available, rather than having to wait for the prediction to jitter itself into another update.

## Scope

**Asana Ticket:** [📅🔎 Show departures for predictions with nil trips if their trip ID corresponds to a real trip](https://app.asana.com/1/15492006741476/project/1210709545912056/task/1215623459165604?focus=true)

## Implementation

- This adds the `trip_id` field directly to the `%Prediction{}` struct so that if the `trip` itself isn't available, the `trip_id` can still be used.
- It makes little touches to a zillion different files (okay, 13) in order to parse and correctly handle the presence of that field.

## Screenshots

I haven't been able to figure out how to get good screenshots of this, especially because right now there are some (I think intentional) data differences between dev and prod.

## How to test

[Green Line Westbound predictions at Government Center](http://localhost:4001/departures/?route_id=Green&direction_id=0&stop_id=place-gover) are a decent place to see this, because trips get added there pretty frequently (see also [GL-B](http://localhost:4001/departures/?route_id=Green-B&direction_id=0&stop_id=place-gover) and [GL-C](http://localhost:4001/departures/?route_id=Green-C&direction_id=0&stop_id=place-gover) there, since the added trips are usually B or C branch, seeing as they start at Gov't Ctr).

### DIFF
--- a/lib/predictions/prediction.ex
+++ b/lib/predictions/prediction.ex
@@ -7,6 +7,7 @@ defmodule Predictions.Prediction do
 
   defstruct id: nil,
             trip: nil,
+            trip_id: nil,
             stop: nil,
             platform_stop_id: nil,
             route: nil,
@@ -31,6 +32,7 @@ defmodule Predictions.Prediction do
   @type t :: %__MODULE__{
           id: id_t,
           trip: Schedules.Trip.t() | nil,
+          trip_id: Schedules.Trip.id_t() | nil,
           stop: stop,
           platform_stop_id: Stops.Stop.id_t() | nil,
           route: Routes.Route.t(),
--- a/lib/dotcom/predictions/event_broadcaster.ex
+++ b/lib/dotcom/predictions/event_broadcaster.ex
@@ -4,7 +4,7 @@ defmodule Dotcom.Predictions.EventBroadcaster do
   """
   use GenStage
 
-  alias Predictions.{Prediction, StreamParser}
+  alias Predictions.StreamParser
   alias ServerSentEventStage.Event
 
   def start_link(opts) do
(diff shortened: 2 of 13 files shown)


lib/predictions/prediction.ex
```diff
@@ -31,6 +32,7 @@ defmodule Predictions.Prediction do
   @type t :: %__MODULE__{
           id: id_t,
           trip: Schedules.Trip.t() | nil,
+          trip_id: Schedules.Trip.id_t() | nil,
```
thecristen, line 35: This shouldn't ever be `nil`, I think?

joshlarson: We don't assume that in either [repo parsing](https://github.com/mbta/dotcom/blob/6c60d04469b0e8dd9894d5743ef37b88cac524eb/lib/predictions/repo.ex#L249-L251) or [stream parsing](https://github.com/mbta/dotcom/blob/6c60d04469b0e8dd9894d5743ef37b88cac524eb/lib/predictions/stream_parser.ex#L90-L93), so I didn't want to risk borking something.

thecristen (approved):

PR #3253, opened 2026-06-11
fix: resolve memory leak under load

paulswartz:
The application experiences Out of Memory (OOM) crashes under high request volume due to unbounded memory accumulation in the metrics-collecting agents DotcomWeb.Stats and Req.Stats.

Specifically:
- These agents listen to Phoenix route dispatch and Req/Finch HTTP client telemetry events.
- For each unique route/status, they stored a list of all request durations.
- This caused memory footprint to grow linearly with the number of requests between dispatch flushes.
- The use of ++ list append is an O(N) operation that copies the list, generating excessive garbage collection overhead.

To fix this:
- Refactored both agents to aggregate metrics in constant memory using `{count, sum}` tuples instead of lists.
- On incoming events, the state is updated to `{count + 1, sum + duration}`.
- During dispatch, the average is calculated using `sum / count`.
- This ensures constant memory usage regardless of traffic volume.

This does not fix the underlying issue that this code is substantially duplicated.

### DIFF
--- a/lib/dotcom_web/stats.ex
+++ b/lib/dotcom_web/stats.ex
@@ -28,13 +28,22 @@ defmodule DotcomWeb.Stats do
     status = metadata.conn.status
     duration = measurement[:duration]
 
-    Agent.update(__MODULE__, fn state ->
-      if Kernel.get_in(state, [method, path, status]) do
-        Kernel.update_in(state, [method, path, status], &(&1 ++ [duration]))
-      else
-        Kernel.put_in(state, [Access.key(method, %{}), Access.key(path, %{}), status], [duration])
-      end
-    end)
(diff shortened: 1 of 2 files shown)


thecristen (approved):
🎉 Thanks for finding

thecristen:
@paulswartz Can you fix the credo error? I'll set this to merge once that's resolved.

```
┃ [F] → Function body is nested too deep (max depth is 2, was 3).
┃       lib/req/stats.ex:28:55 #(Req.Stats.handle_event)
┃ [F] → Function body is nested too deep (max depth is 2, was 3).
┃       lib/dotcom_web/stats.ex:33:57 #(DotcomWeb.Stats.handle_event)
```

PR #3256, opened 2026-06-12
further improvements to stats logging

paulswartz:
- centralize implementation (`Telemetry.Stats`)
- use built-in `Path.dirname/1` instead of custom Regex
- use ETS table for concurrent metric logging, instead of having all requests go through a single Agent (metric calculation still goes through the agent)

When all of the metric logging goes through a single Agent, we lose the concurrency we'd otherwise have. Each request needs to wait for previous requests to make their log, creating a single bottleneck. By using an ETS table with `{:write_concurrency, true}` each process can update the metrics concurrently without blocking each other.

### DIFF
--- a/lib/dotcom_web/stats.ex
+++ b/lib/dotcom_web/stats.ex
@@ -1,14 +1,28 @@
 defmodule DotcomWeb.Stats do
   @moduledoc """
-  This Agent attaches to telemetry events emitted by Phoenix and aggregates them.
+  This GenServer attaches to telemetry events emitted by Phoenix and aggregates them by path and status.
   """
 
-  use Agent
+  use GenServer
 
   @doc """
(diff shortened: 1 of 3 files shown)


lib/dotcom_web/stats.ex
```diff
@@ -3,84 +3,55 @@ defmodule DotcomWeb.Stats do
   This Agent attaches to telemetry events emitted by Phoenix and aggregates them.
```
jlucytan: nit(non-blocking): Should this doc be updated from "Agent" to "GenServer"?

```suggestion
  This GenServer attaches to telemetry events emitted by Phoenix and aggregates them.
```

jlucytan (approved):

lib/dotcom_web/stats.ex
paulswartz: Yep! There were a couple other places I missed this as well.

PR #3258, opened 2026-06-12
fix: resolve presence tracking process and stream leak in departures

paulswartz:
Makes sure we also untrack the Presence when we unsubscribe a Manager.


thecristen (approved):
Ah, that tracks... I think I'd somehow had a notion of PubSub's handling (where it removes dead processes from its subscribers) to somehow apply analogously to Presence. Anyways, thanks so much for this!

paulswartz:
I believe it does, but if we're ever unsubscribing without the process terminating (such as transitioning between pages) then the Presence tracking gets out of sync.

PR #3259, opened 2026-06-12
fix: configure gc_interval for Nebulex local cache

paulswartz:
https://nebulex.hexdocs.pm/declarative-caching.html#memory-management

Without the `gc_interval` I don't think we're actually evicting anything from the cache.


thecristen (approved):
👍🏼

PR #3261, opened 2026-06-12
fix(alert): More permissive URL regex

jlucytan:
## Scope

**Asana Ticket:** [⚠️ Fix alerts with URLs with weird chars](https://app.asana.com/1/15492006741476/project/385363666817452/task/1215629073500541)

## Implementation
Adds a bunch of characters to the url regex to be allowed. 

## Screenshots
Before: link shown is split
image

After: link is together
image

## How to test
Find some live alerts in local that are broken (need to be pointed to `api-dev`) and verify they're no longer broken.

http://localhost:4001/schedules/CR-Providence/timetable 
Providence currently has one

### DIFF
--- a/lib/alerts/url_parsing_helpers.ex
+++ b/lib/alerts/url_parsing_helpers.ex
@@ -1,7 +1,7 @@
 defmodule Alerts.URLParsingHelpers do
   @moduledoc "Helpers to parse out the url from a string and create an html link"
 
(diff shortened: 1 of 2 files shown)


lvachon1 (approved):
I didn't realize character ranges in regex expressions (eg, a-z or 0-9) can be between any two ascii code points, neat!

PR #3262, opened 2026-06-15
[Feat]: Separate date and time pickers in Trip Planner

lvachon1:
## Scope

**Asana Ticket:** [🗺️ Separate date and time pickers in Trip Planner](https://app.asana.com/1/15492006741476/project/555089885850811/task/1210941646065572?focus=true)

## Implementation

- Moved the DatePicker component out of Metro and into Dotcom
- Removed the time part of the datepicker component
- Created a separate TimePicker component
- Combined the two together to update the same variable in trip_planner_live
- Updated error handling to unify DatePicker and TimePicker
- Added logic to select the current time (rounded up) when no time is previously specified
- Made it look nice

## Screenshots
Screenshot 2026-06-15 at 1 10 27 PM

Screenshot 2026-06-15 at 1 25 08 PM

## How to test

- Confirm that without a `?plan` that when the date and time pickers are shown they default to "now-ish".
- Fill out the form and confirm that the selected time is properly passed into OTP
- Change the time and confirm the results update
- Refresh the page now that a new plan has been generated and confirm that the same results are returned
- Select a date+time in the past and confirm that the error state renders properly
- Select a proper date+time, confirm the error state clears the results update


lib/dotcom_web/components/timepicker.ex
```diff
@@ -0,0 +1,126 @@
...
+        <option
+          :for={hour <- 1..12}
+          value={hour}
+          selected={option_selected?(@form, hour, :timepicker_hour)}
+        >
+          {hour}
+        </option>
+      </select>
+      <select
+        class="c-select trip-plan-time"
```
thecristen: **observation**(non-blocking): The rest of this component is structured and named pretty generically, and then there's this `trip-plan-time` CSS class. Similarly, its companion datepicker component is also pretty generic, save for the use of `phx-hook="TripPlannerDatePicker"`. And that hook itself is something that could potentially be reused outside of a trip planner context. What I'm getting at here is... could these new date/time pickers be truly general and not intertwined with trip planning concerns?

lib/dotcom_web/components/timepicker.ex
```diff
@@ -0,0 +1,126 @@
...
+  """
+  def render(assigns) do
+    assigns =
+      assigns
+      |> assign(:errors, format_changeset_errors(assigns.errors))
+
+    ~H"""
+    <div class={error_class?(@errors)}>
+      <select
+        class="c-select trip-plan-time  .mbta-input"
```
thecristen: typo here (` .mbta-input`)

lib/dotcom_web/components/timepicker.ex
```diff
@@ -0,0 +1,126 @@
+defmodule DotcomWeb.Components.TimePicker do
+  @moduledoc """
+  A time picker live component, 12h hh:mmAM format, every 5 minutes
+
+  assigns:
+  :form - the form this component is in
+  :errors - the errors to render
+
+  """
+  use Phoenix.LiveComponent
```
thecristen: **suggestion**: Given that this module has no `handle_event` (or `handle_` anything) callbacks here, and the fact that this component isn't doing anything to change its own state, I think this can be converted into a regular function component rather than a `LiveComponent`

lib/dotcom_web/components/timepicker.ex
```diff
@@ -0,0 +1,126 @@
...
+        id="timepicker_minute"
+        name={@form[:timepicker_minute].name}
+      >
+        <option
+          :for={minute <- 0..55//5}
+          value={minute |> Integer.to_string() |> String.pad_leading(2, "0")}
+          selected={
+            option_selected?(
+              @form,
+              minute |> Integer.to_string() |> String.pad_leading(2, "0"),
```
thecristen: **suggestion**: Might want to extract this as a function. Or compute the list of formatted numbers once at the start before returning the markup.

lib/dotcom_web/live/trip_planner_live.ex
```diff
@@ -383,6 +384,53 @@ defmodule DotcomWeb.TripPlannerLive do
...
+         } = params
+       ) do
+    # Get the date from datepicker (datetime)
+    {:ok, old_datetime} =
+      if is_binary(datetime) do
+        datetime |> DateTime.from_iso8601() |> elem(1)
+      else
+        datetime
+      end
+      |> DateTime.shift_zone("America/New_York")
```
thecristen: Just for consistency/consolidation's sake, let's get the timezone from our configuration. A few other files do it like so
```elixir
@timezone Application.compile_env!(:dotcom, :timezone)
```
Same goes for line 425.

thecristen (changes requested):
This works really nicely!! Mostly minor comments, here.

lib/dotcom_web/components/timepicker.ex
lvachon1: I've made the class name more generic, I'm not sure where to put the rule though.  It seems silly to make a new file just for one a few lines, is there a better place that already exists I can put it?

lvachon1: To answer your larger question, I do think that these components can be used independently of the trip planner, nothing in their logic requires the trip planner form itself.  Anything that would use them would have to emulate some aspects of the trip planner form rendering and handling, more thought is needed if that is the right way to go.  For now, I kept datepicker.js in the trip-planner folder because there are already two other datepicker-ish  files in the parent folder (and /time-controls!).  

Hopefully when we replace all the react pages we can delete many of these confusing/redundant files.

thecristen (approved):

PR #3265, opened 2026-06-16
feat(dependabot): limit PRs for root package.json

thecristen:
## Scope

We have two `package.json` files! It turns out the root level one was using the default Dependabot configuration (which is set to 5 open PRs)

## Implementation

- Adds configuration for NPM packages in the `/` directory
- Adjusts the total number of open PRs across all 3 configurations to 5 total


lvachon1 (approved):
Thank you!

PR #3270, opened 2026-06-16
fix(UpcomingChanges): gettext later changes link

jlucytan:
## Scope

No ticket, just realized that there was no localization on this text.

## Screenshots
No visual changes. There currently aren't any translation strings for this text so nothing will change when toggling to the Spanish site in dev at this time.

## How to test
Verify that there are no changes to the text appearance when "x later changes" appears in the Upcoming Changes widget on the commuter rail timetable


thecristen (approved):
🫡

PR #3271, opened 2026-06-17
fix(a11y): Don't read out icons in fare cards

jlucytan:
## Scope

**Asana Ticket:** [♿️🎧 Mode Page: Decorative mode icon is read out as "image" in Fares section of /subway, /commuter-rail, etc](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214647592812619)

## Implementation
Added `aria-hidden="true"` to the fare card templates so that screen readers won't read them 

## Screenshots
No visual changes

## How to test
When using VoiceOver on the local pages utilizing the following templates with this branch checked out, observe that you cannot navigate to the icons using VO and VO does not read them out: 
- individual fare cards : [example page - fares page](http://localhost:4001/fares)
- grouped fare card : [example page - subway fares page](http://localhost:4001/fares/subway-fares)

All content pages should be using these fare cards, but doing a spot check on other fare pages may be valuable just in case.

### DIFF
--- a/lib/dotcom_web/templates/cms/paragraph/_fare_card.html.heex
+++ b/lib/dotcom_web/templates/cms/paragraph/_fare_card.html.heex
@@ -101,7 +101,7 @@ card_title = [content_tag(:span, name, class: "c-fare-card__mode"), duration] %>
 
 <%= content_tag :div, class: "grid-container" do %>
   <div class="c-fares-grid">
-    <div aria-hidden= "true">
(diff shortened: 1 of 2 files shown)


lvachon1 (approved):
Sounds good to me regarding the icons.  An out-of-scope but related issue I noticed:

Image

This reading seems a bit off: "space" and "3 items" doesn't seem right.  Something for a future ticket, looks like it's probably a content issue.

PR #3276, opened 2026-06-22
fix(PageController): don't copy the whole conn

paulswartz:
If we use `conn` in the async closure, the whole thing (including any existing data) is copied. By extracting only the items we need, we limit the amount of extra memory used.


joshlarson (approved):
Thanks for catching this - hopefully it makes a difference!

paulswartz:
@joshlarson are the dev-blue deploy failures due to an issue with the code?

PR #3281, opened 2026-06-23
[feat] Stale old alerts on line diagram banner

lvachon1:
## Scope

**Asana Ticket:** [🐬 Stale Symphony alert on line diagram banner on GL and GL-E pages](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214740102739529?focus=true)

## Implementation

- Added logic to alerts.ex to identify and mark alerts as stale when the current active period has started more than five weeks ago
- Added filter to remove stale alerts from the line diagram page.
- Added tests

Interestingly the Symphony alert has been downgraded in priority and no longer shows on the line diagram anyway.

### Dev-Blue 
Screenshot 2026-06-29 at 3 13 26 PM

### Local
Screenshot 2026-06-29 at 3 13 19 PM

## How to test

http://localhost:4001/schedules/1/line

https://dev-blue.mbtace.com/schedules/1/line

On dev-green/blue, Bus 1 has a perpetual delay that started in May.  Confirm it shows up on dev-blue/green, but not on this branch (when pointed at dev-blue/green)

### DIFF
--- a/lib/alerts/alert.ex
+++ b/lib/alerts/alert.ex
@@ -137,7 +137,6 @@ defmodule Alerts.Alert do
     |> set_priority()
     |> set_direction_ids()
     |> ensure_entity_set()
-    |> check_freshness()
   end
 
   @spec update(t(), Keyword.t()) :: t()
@@ -181,17 +180,14 @@ defmodule Alerts.Alert do
     %__MODULE__{alert | priority: Priority.priority(alert)}
   end
 
-  defp check_freshness(%__MODULE__{} = alert) do
+  def stale?(%__MODULE__{} = alert) do
     now = Timex.now()
     five_weeks_ago = DateTime.add(now, -5 * 7, :day)
 
-    stale? =
-      case Dotcom.Alerts.StartTime.next_active_time(alert, now) do
-        {:current, datetime} -> datetime |> DateTime.before?(five_weeks_ago)
-        _ -> false
-      end
-
-    %__MODULE__{alert | stale?: stale?}
+    case Dotcom.Alerts.StartTime.next_active_time(alert, now) do
+      {:current, datetime} -> datetime |> DateTime.before?(five_weeks_ago)
+      _ -> false
+    end
   end
 
   @spec build_struct(Keyword.t()) :: t()
--- a/lib/alerts/alert.ex
+++ b/lib/alerts/alert.ex
@@ -190,7 +190,7 @@ defmodule Alerts.Alert do
       if is_nil(current_active_period) do
         false
       else
-        current_active_period |> elem(0) |> DateTime.before?(five_weeks_ago)
+        (current_active_period |> elem(0) || now) |> DateTime.before?(five_weeks_ago)
       end
 
     %__MODULE__{alert | stale?: stale?}
@@ -200,24 +200,12 @@ defmodule Alerts.Alert do
   def current_active_period(%__MODULE__{} = alert, now) do
     alert.active_period
     |> Enum.find(fn {start, stop} ->
-      cond do
-        is_nil(start) and DateTime.after?(stop, now) ->
-          true
-
-        is_nil(start) ->
-          false
-
-        is_nil(stop) and DateTime.before?(start, now) ->
-          true
-
-        is_nil(stop) ->
-          false
-
-        DateTime.before?(start, now) and DateTime.after?(stop, now) ->
-          true
-
-        true ->
-          false
+      case {start, stop} do
+        # nil start should never happen, but some tests include it
+        {nil, stop} -> DateTime.after?(stop, now)
+        {start, nil} -> DateTime.before?(start, now)
+        {start, stop} -> DateTime.before?(start, now) and DateTime.after?(stop, now)
+        _ -> false
       end
     end)
   end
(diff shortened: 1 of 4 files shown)


lib/dotcom_web/templates/schedule/_line.html.heex
```diff
@@ -31,7 +31,11 @@
 
 <div class="page-section m-schedule-line">
   {DotcomWeb.AlertView.group(
-    alerts: @alerts,
+    alerts:
+      @alerts
+      |> Enum.filter(fn %{created_at: created_at} ->
+        DateTime.after?(created_at, DateTime.add(@date_time, -5 * 7, :day))
+      end),
```
joshlarson: **Suggestion (blocking):** The concept of `stale` is going to be applicable beyond just this page. I would recommend defining `stale?` in [`alert.ex`](https://github.com/mbta/dotcom/blob/fc23cc2de937c8076f4b78b34caf77d148a6c129/lib/alerts/alert.ex) or somewhere else reusable.

---

That would also make it easier to test.

lib/dotcom_web/templates/schedule/_line.html.heex
```diff
@@ -31,7 +31,11 @@
 
 <div class="page-section m-schedule-line">
   {DotcomWeb.AlertView.group(
-    alerts: @alerts,
+    alerts:
+      @alerts
+      |> Enum.filter(fn %{created_at: created_at} ->
```
joshlarson: **Issue (blocking):** The field we want to be looking at is when the alert first becomes active, not when it was first entered into alerts manager.

(There's no `active_at` field on an alert - you'll need to look through the active periods in the alert, finding the one that's currently active, and see when that active period began.)

joshlarson (changes requested):
Left a few comments - one is stylistic, the other is functional, but they're both blocking me from approving.

Also, I know the ticket mentions GL and GL-E, but this solution applies to any line diagram page (which I think is good), so I would also update the PR title (and thus the commit message when it gets merged) to not be GL/GL-E specific.

lib/dotcom_web/templates/schedule/_line.html.heex
lvachon1: Good idea, done!

lib/dotcom_web/templates/schedule/_line.html.heex
lvachon1: Thanks for catching this

lvachon1:
I figured there already had to be a way to determine the current active period.  Cristen found two others, so I'll be updating this to use the best version instead of my own.

lib/alerts/alert.ex
```diff
@@ -178,6 +181,19 @@ defmodule Alerts.Alert do
     %__MODULE__{alert | priority: Priority.priority(alert)}
   end
 
+  defp check_freshness(%__MODULE__{} = alert) do
+    now = Timex.now()
+    five_weeks_ago = DateTime.add(now, -5 * 7, :day)
```
joshlarson:
```suggestion
    five_weeks_ago = DateTime.shift(now, week: -5)
```

lib/alerts/alert.ex
```diff
@@ -178,6 +181,19 @@ defmodule Alerts.Alert do
     %__MODULE__{alert | priority: Priority.priority(alert)}
   end
 
+  defp check_freshness(%__MODULE__{} = alert) do
+    now = Timex.now()
```
joshlarson: **Suggestion:** We've been slowly migrating most of our code to use [`Dotcom.Utils.DateTime.now/0`](https://github.com/mbta/dotcom/blob/f9a5b4cdd4cd4e0e38a46da850d3d2347b70379b/lib/dotcom/utils/date_time.ex#L29-L30) instead of `Timex` (pulling the `Dotcom.Utils.DateTime` module in as a configurable attribute so that we can mock it in tests... [example from `UpcomingDepartures`](https://github.com/mbta/dotcom/blob/f9a5b4cdd4cd4e0e38a46da850d3d2347b70379b/lib/dotcom/upcoming_departures/processor.ex#L24))

lib/dotcom_web/templates/schedule/_line.html.heex
```diff
@@ -31,7 +31,9 @@
 
 <div class="page-section m-schedule-line">
   {DotcomWeb.AlertView.group(
-    alerts: @alerts,
+    alerts:
+      @alerts
+      |> Enum.filter(fn alert -> !Alerts.Alert.stale?(alert) end),
```
joshlarson:
```suggestion
      |> Enum.reject(&Alerts.Alert.stale?/1),
```

lib/alerts/alert.ex
```diff
@@ -58,6 +58,7 @@ defmodule Alerts.Alert do
             lifecycle: :unknown,
             priority: :low,
             severity: 5,
+            stale?: false,
```
joshlarson:
```suggestion
```

Leftovers 🙂

lib/alerts/alert.ex
```diff
@@ -116,6 +117,7 @@ defmodule Alerts.Alert do
           lifecycle: lifecycle,
           priority: Priority.priority_level(),
           severity: severity,
+          stale?: boolean(),
```
joshlarson:
```suggestion
```

Leftovers 🙂

joshlarson (approved):
I added a few style comments, and this otherwise looks and works great!

Thanks for going on what wound up being a journey with this PR!

PR #3282, opened 2026-06-23
Gate flags path behind basic auth

lilybarrett:
## Scope

While it's unlikely that users outside the MBTA organization will visit the feature-flagging route, we should gate it behind `basic_auth`. 

## How to test
* Navigate to `/_flags`
* Input credentials

### DIFF
--- a/lib/dotcom_web/router.ex
+++ b/lib/dotcom_web/router.ex
@@ -76,7 +76,7 @@ defmodule DotcomWeb.Router do
   end
 
   scope "/_flags", Laboratory do
-    pipe_through([:browser, :browser_live])
+    pipe_through([:basic_auth])
     forward("/", Router)
   end
 


arkadyan:
I don't think we want to do this. We shouldn't use basic auth ever for anything. But also, it's fine for the flags endpoint to be exposed. People need to know where to look for it, and there's no harm if they go to it.

lilybarrett:
> I don't think we want to do this. We shouldn't use basic auth ever for anything. But also, it's fine for the flags endpoint to be exposed. People need to know where to look for it, and there's no harm if they go to it.

@arkadyan Got it! My concern is that since the route is exposed in prod, curious riders or non-MBTA users could wind up toggling translations on before they're ready for the public -- which could possibly impact UX/incur unwanted costs. Thoughts?

lib/dotcom_web/router.ex
```diff
@@ -76,7 +76,7 @@ defmodule DotcomWeb.Router do
   end
 
   scope "/_flags", Laboratory do
-    pipe_through([:browser, :browser_live])
+    pipe_through([:basic_auth])
```
lvachon1, line 79: I think you want to use `:basic_auth_readonly` here to use the same methods and credentials as `/preview`.  Currently this page 500s for me locally.  

The credentials for these accounts can be found on onepassword under Dotcom Basic Auth Readonly or something like that.  Locally, they are configured in your `.env` file

arkadyan: @lvachon1 We shouldn't be using any form of basic auth on the site. Cristen is aware of this, and I believe there should be tickets for the team to remove any existing usage.

arkadyan:
@lilybarrett There are no on-the-fly translations for the website, so there's no cost implication. If any visitor, internal or external, toggles an in-development feature, they should understand there could be UI or functionality impacts, so that's on them.

PR #3289, opened 2026-06-25
Lb/feature flag with new language picker

lilybarrett:
## Scope

The current Languages picker is a styled skin for the Google Translate plugin. This PR provides an initial iteration (with styling/UX subject to future change once designs are confirmed) of a new language picker set behind the feature flag for Smartling translations.

We have a .po file for Spanish in the DotCom repo, which I believe is why we're getting the Spanish translations when I select them. The other languages besides English & Spanish don't work yet since we don't have the Smartling translations/files for them in the repo yet.

Coming in future PRs:
* Refined/confirmed designs and UX

**Asana Ticket:** 
https://app.asana.com/1/15492006741476/project/1213994587940653/task/1214847351822159?focus=true

## How to test
* Run the app
* Navigate to `/_flags` and enable the Smartling translations flag
* Select Spanish from the languages menu in either the header or footer -- the website should be translated to Spanish.
* Other languages are also selectable from the menu, but the website content won't change since we don't yet have `.po` files for them.

https://github.com/user-attachments/assets/14c2c34e-5b64-48f4-a138-4792373353f4


lib/dotcom_web/templates/layout/_footer.html.heex
```diff
@@ -97,10 +97,13 @@
         </div>
         <div class="container">
           <div class="lower-footer">
-            <%= unless assigns[:exclude_google_translate] do %>
+            <%= unless assigns[:exclude_google_translate] || Laboratory.enabled?(@conn, :use_smartling_translations) do %>
               {render("_google_translate_plugin.html", conn: @conn)}
+              <div class="m-menu__language"></div>
```
lilybarrett, line 102: TODO: Add new language picker to the footer

lvachon1 (approved):
This works great, I'm excited to see this whole translation project get properly underway here on MBTA.com

lib/dotcom_web/templates/layout/root.html.heex
```diff
@@ -89,7 +89,7 @@
       <!-- End Google Tag Manager (noscript) -->
     <% end %>
     <div class="body-wrapper" id="body-wrapper">
-      <%= if !Application.get_env(:dotcom, :is_prod_env?) do %>
+      <%= if !Application.get_env(:dotcom, :is_prod_env?) && !Laboratory.enabled?(@conn, :use_smartling_translations) do %>
```
joshlarson, line 92: **Question/Request (non-blocking):** Would it be possible to keep that banner at the top when Smartling is enabled? Or was the wacky "menu location is actually hard-coded, and gets messed up by the banner" issue too annoying?

lib/dotcom_web/templates/layout/_footer.html.heex
```diff
@@ -95,12 +95,28 @@
             )}
           </nav>
         </div>
-        <div class="container">
+        <div class="container m-footer-links">
```
joshlarson, line 98: **Issue (non-blocking):** I'm not sure if it's `m-footer-links` or some other CSS thing (or just something fiddly in the new DOM), but there's a funky line-wrap in the footer icons at browser widths from 1088 to 1343 px.

Image

(FWIW, the line-wrapping behavior is already wrong and weird in prod - it's just that the language stuff in the footer is making that always happen at certain common breakpoints, whereas in prod, it only happens at very narrow screen widths)

lib/dotcom/locales.ex
```diff
@@ -7,6 +7,22 @@ defmodule Dotcom.Locales do
 
   @default_locale %Locale{code: "en", endonym: "English"}
   @development_locale %Locale{code: "es", endonym: "Español"}
+
+  @development_additional_locales [
+    @default_locale,
+    @development_locale,
```
joshlarson, line 13: **Question (non-blocking):** Is `@development_locale` still special, or would it make sense to lump all of the locales together (probably still keeping `@default_locale` separate)?

joshlarson (commented):
I see that @lvachon1 approved (thank you Luc!). I was snooping a bit, and had a few non-blocking questions and thoughts, which I may as well still leave here.

Some of them may need to be addressed before this becomes fully productionalized, but I don't think any of them need to block further development on this feature (and possibly some or all of the issues/questions are really more core-dotcom responsibilities anyway, and don't need to take up your time! 😅)

---

Thanks for this work!

lib/dotcom_web/templates/layout/root.html.heex
lilybarrett: I can bring it back in the next iteration!

lib/dotcom_web/templates/layout/_footer.html.heex
lilybarrett: Thank you for this callout! This may change completely based on designs, but I'll make a note.

lib/dotcom/locales.ex
lilybarrett: Yeah, good question. I kept `@development_locale` intact so functionality could remain unchanged when the feature flag was turned on, but we could def merge these down the road.

PR #3302, opened 2026-06-30
Lev/feat/stale alert on subway status

lvachon1:
## Scope

**Asana Ticket:** [🐬 Stale Symphony alert in subway current status](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214740102739531?focus=true)

## Implementation

Added a filter to remove `stale?` alerts from the subway status indicator on the homepage.

This builds upon the work already done in https://github.com/mbta/dotcom/pull/3281

### Dev-blue showing stale alert on green line
Screenshot 2026-06-30 at 1 29 16 PM
Screenshot 2026-06-30 at 1 35 23 PM

### Local only showing the fresh blue line alert
Screenshot 2026-06-30 at 1 29 10 PM
Screenshot 2026-06-30 at 1 35 35 PM

## How to test

This is hard.  We cannot create alerts with start times in the past, so you have to find one somewhere in the wild.  Tappan street seems to be the best test case at the moment.


thecristen (commented):
I'm curious if, philosophically, this new piece of logic is better off in `Dotcom.SystemStatus.status_alert?/2` -- that function is currently used to filter alerts before passing them into the function this PR modified.

lvachon1:
> I'm curious if, philosophically, this new piece of logic is better off in `Dotcom.SystemStatus.status_alert?/2` -- that function is currently used to filter alerts before passing them into the function this PR modified.

I tried that out and it should work, but looks like it'll also affect `commuter_rail_alert_groups()` in `alerts.ex`.

thecristen (approved):

PR #3307, opened 2026-07-01
Changes to Dotcom translation workflow

lilybarrett:
## Scope

We realized that Dotcom's custom translation task is making it so a single string, `All clear, captain!` was the only thing getting translated from English. This changes the code so we can properly update our `.po` files for translations going forward.

**Asana Ticket:** [Implement: Fix workflows/Add CI checks/Update ReadMe](https://app.asana.com/1/15492006741476/project/1213994587940653/task/1214847351822167?focus=true)

## Implementation
* Remove `translator.yml` and `translate.ex`
* Remove `terminologies`
* Update `gettext` pluralization logic
* Add more information to the ReadMe about handling internationalization
* Add a new command, `mix localize`, that should be run manually to update `.po` files before making PRs
* Add a CI task to ensure `.po` files have been properly updated
* Commit an updated `.po` file for English phrases

## How to test
* Pull down this branch
* Run `mix localize` locally to update your `.po` files
* Navigate to `/_flags` and turn on the Smartling flag
* Navigate to the homepage and play around with picking different languages


mix.exs
```diff
@@ -41,6 +42,18 @@ defmodule DotCom.Mixfile do
...
 
+  # Aliases are shortcuts or tasks specific to the current project.
+  # For example, to install project dependencies and perform other setup tasks, run:
+  #
+  #     $ mix setup
+  #
+  # See the documentation for `Mix` for more info on aliases.
+  defp aliases do
+    [
+      localize: ["gettext.extract", "gettext.merge priv/gettext --no-fuzzy"]
```
joshlarson: **Praise / Minor Suggestion:** I really like this! 

How would you feel about adding another one, like 👇 

```elixir
check_localized: ["gettext.extract --check-up-to-date"]
```

(This would largely be a convenience for me, since I have an easy way of running individual commands with no flags in my editor... this is the least blocking of any suggestions lol)

config/config.exs
```diff
@@ -18,7 +18,7 @@ config :dotcom, :httpoison, HTTPoison
 
 config :dotcom,
   default_locale_code: "en",
-  locale_codes: ["en", "es", "ht", "pt", "vi", "zh"]
+  locale_codes: ["en", "es", "ht", "pt-BR", "vi", "zh-CN", "zh-TW", "fr-FR"]
```
joshlarson, line 21: **Issue:** I think by virtue of adding `fr-FR`, you've caused another `.po` file to com into existence on `mix compile`. Would it make sense to commit that file as well?

README.md
```diff
@@ -131,6 +131,56 @@ Then, start the server with `iex -S mix phx.server`
...
+
+This is done via [the `Gettext` Elixir library](https://hex.pm/packages/gettext), which uses
+[the GNU `gettext` tool](https://www.gnu.org/software/gettext) under the hood,
+[managing `.po` files for message strings](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html).
+
+To internationalize a string in an interface, do the following:
+
+- Wrap strings to be translated in the `gettext` function, i.e. `gettext("your string here")`.
+  Note that you may need to wrap the function call in `{...}` for attribute strings and
+  `<%= ... %>` for text on the page.
```
joshlarson: **Nit:** HEEx works with `{}` for text on the page as well, so 👇 are equivalent.

```heex
<span><%= foo(@bar) %></span>
```

```heex
<span>{foo(@bar)}</span>
```

All that to say, you can drop the "`<%= ... %>` for text on the page" part of this note.

README.md
```diff
@@ -131,6 +131,56 @@ Then, start the server with `iex -S mix phx.server`
...
+
+#### Rules Of Gettext
+
+1. No newlines inside a gettext call. Newlines are a new gettext call.
+2. Don't start or end with whitespace. If you break a sentence up, leave the whitespace in the HTML untranslated.
+3. If you break apart a sentence across multiple gettext calls, use `pgettext` and add context including the full sentence to help translators. Grep the app for `pgettext` for examples.
```
joshlarson, line 153: Is this the sort of thing that would apply to links... e.g. in a sentence like "transfer to the [Orange Line](https://www.mbta.com/schedules/Orange/line) at [Downtown Crossing](https://www.mbta.com/stops/place-dwnxg)"?

config/config.exs
joshlarson: Gotta be honest - I was rushing when I typed this - it's not a `.po` file 🤦‍♂️ - it's a JSON file - `priv/cldr/locales/fr.json`. 

I stand by my question/suggestion/request to commit it though!

joshlarson (approved):
This seems quite reasonable, and the workflow you described in the README is quite cool!

Thanks for setting this up.

(I snuck a commit on to add that new file 😇)

config/config.exs
lilybarrett: Yes! From what I see, CLDR (and these files) are for additional on-the-fly rules for translation. They're not necessary for Smartling to work properly, but still good to commit.

mix.exs
lilybarrett: Done!

README.md
lilybarrett: Yeah! Here's an example from the MyCharlie repo:

```
<p>
      {pgettext(
        "To access discounted fares on The RIDE, you'll need to"
      )}
      <LiveComponents.link
        href="https://www.mbta.com/accessibility/the-ride/ride-fares-and-adding-funds#reduced"
        text={
          pgettext(
            "link your reduced fare card to your RIDE ID"
          )
        }
      />.
</p>
```

README.md
lilybarrett: Fixed!

lib/dotcom/gettext/plural.ex
```diff
@@ -1,5 +1,33 @@
...
+  Plural forms are documented in the gettext docs here:
+  https://www.gnu.org/software/gettext/manual/gettext.html#Plural-forms
+
+  There's a list of plural forms for most languages here:
+  https://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html
+  """
+  @behaviour Gettext.Plural
+
+  # Haitian Creole is not supported by default by Gettext.Plural (which uses Expo.PluralForms)
+  def nplurals("ht"), do: 2
```
arkadyan, line 14: suggestion: It looks like this comes from MyCharlie, but I'm not sure this is correct. (Same for `plural` below.) Based on [sources](https://preply.com/en/blog/haitian-creole-nouns/#id-45ce40431dad84b7a7105032cb91d84792a2eaed) and experimenting in Google Translate, it looks like Haitian Creole only has a single form for any number of items. I'm not an expert, though, so it would probably be good to confirm with someone who is. Maybe the Fares team could ask the translators they work with?

Not for this PR, but as a follow-up, it would be great to contribute details about `ht` to the `expo` library so that we don't need custom patches.

lilybarrett: @arkadyan Thank you for looking into that! I'll comment out the Haitian Creole-specific lines and add TODOs for human review to confirm those assumptions -- Let me know if that sounds like a good move?

Totally agree re: the follow-up.

lilybarrett: Actually @arkadyan on a second look, the compiler fails when I remove those lines:

```
** (Gettext.Plural.UnknownLocaleError) unknown locale "ht". If this is a locale you need to handle,
consider using a custom pluralizer module instead of the default
Gettext.Plural. You can read more about this on the Gettext docs at
https://hexdocs.pm/gettext/Gettext.Plural.html
```

lilybarrett: For posterity: We have decided to keep this as-is for now but will remember to follow up with human review.

arkadyan: @lilybarrett Could you please make a ticket for this so we don't forget?

PR #3311, opened 2026-07-01
refactor(Alerts.Sort): more performant comparison

thecristen:
## Scope

While tackling #3310 I noticed that `Alerts.Sort.sort/2` was featuring _prominently_ in my flame charts, and I wanted to see if I could do anything about that.

## Implementation

**_First of all_**. `/stops/<mode>` pages don't even show alerts. So the first commit moves the call which assigns alerts to `conn` to the specific method behind rendering the individual stop pages, instead of being invoked for every page handled by `DotcomWeb.StopController`.

Then when debugging `Alerts.Sort.sort/2`, I noticed the sorting logic included assessing the next active period start time. And I remembered we'd later implemented a different function for computing that, over at `Dotcom.Alerts.StartTime.next_active_time/2`! So I wanted to see if reusing that was more performant (and looks like it is!).

These charts were generated from loading South Station page with an already warmed cache.

| Before (6.5MB) | After (3.3MB) |
|--------|--------|
| flame_on (alert before) | flame_on (alert after)  | 

## How to test

We could verify if the subway **Planned Work** section still sorts those alerts in the same way as before.

### DIFF
--- a/lib/dotcom/alerts/start_time.ex
+++ b/lib/dotcom/alerts/start_time.ex
@@ -41,6 +41,9 @@ defmodule Dotcom.Alerts.StartTime do
         time
       ) do
     cond do
+      is_nil(start_time) ->
+        next_active_period_active_time(rest_of_active_periods, time)
+
       ends_before?(end_time, time) ->
         next_active_period_active_time(rest_of_active_periods, time)
 
--- a/lib/dotcom/alerts/start_time.ex
+++ b/lib/dotcom/alerts/start_time.ex
@@ -41,11 +41,13 @@ defmodule Dotcom.Alerts.StartTime do
         time
       ) do
     cond do
-      is_nil(start_time) ->
-        next_active_period_active_time(rest_of_active_periods, time)
-
+      # period already ended -- keep iterating
       ends_before?(end_time, time) ->
         next_active_period_active_time(rest_of_active_periods, time)
+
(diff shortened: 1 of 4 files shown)


lib/dotcom/alerts/start_time.ex
```diff
@@ -41,9 +41,14 @@ defmodule Dotcom.Alerts.StartTime do
         time
       ) do
     cond do
+      # period already ended -- keep iterating
       ends_before?(end_time, time) ->
         next_active_period_active_time(rest_of_active_periods, time)
 
+      # period hasn't ended yet, but no defined start time -- use the current time
+      is_nil(start_time) ->
```
thecristen, line 49: Some of our tests have alerts involving active periods with a `nil` start time, so I added this clause to handle those. I'm not sure if this'll ever be encountered in real alerts from Alerts Manager, though.

lvachon1: I'm glad I saw this PR, I have one open with yet a 3rd way of computing the active time period.  I'm going to put mine into draft for now.  I saw this exact same behavior regarding only our tests having nil start times, thanks for following up with the transit data team about if that's even possible.

lvachon1 (approved):
I only found one place where the sorting of alerts wasn't quite the same between prod and this branch.  I don't think it's relevant though, the alerts seem to be for the same timeframe.

Image

PR #3318, opened 2026-07-06
Add general accessibility note to Trip Planner

lvachon1:
## Scope

**Asana Ticket:** [[By 7/13] 🗺️ Add general accessibility note to Trip Planner](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216114279196738?focus=true)

## Implementation

Created generic CTA/callout component with attrs for `link`, `icon`, `icon-type`, `arrow`,  and CSS `classes`.  Used it for the new a11y notification on the trip planner

## Screenshots

Screenshot 2026-07-06 at 3 30 15 PM

## How to test

Check the a11y checkbox on and off and confirm the notice appears and disappears accordingly


thecristen (commented):
**Question**: Was there a particular reason you didn't use/extend the existing `callout` and/or `callout_link` components? It seems to have a lot of overlap with this one, design-wise (similar styling, includes the arrow).

lvachon1:
> **Question**: Was there a particular reason you didn't use/extend the existing `callout` and/or `callout_link` components? It seems to have a lot of overlap with this one, design-wise (similar styling, includes the arrow).

There was no particular reason, I suppose I just didn't see much substance to those components and didn't think they were a good place to start.  Mostly I just looked at the callout/cta components and picked the one that looked closest to the design in the ticket to use as a base for a generic version.   In my mind I guess it was easier to take a component with everything on it, and add switches to toggle stuff.  Either way I'm only making the 'too many similar components' problem worse here, and it would be wise to consolidate some of these components either by replacing them wholesale or having each one call cta with params and leaving the consumers alone.

lvachon1:
I tried to clean things up a bit here, callout and mbta_go_cta now call cta under the hood.  The rest of the callouts seem to be WC specific and not long for this world, so I didn't think it was worth it to refactor those as well.

lib/dotcom_web/components/components.ex
```diff
@@ -475,20 +479,42 @@ defmodule DotcomWeb.Components do
...
+  attr :classes, :string, required: false, default: nil
+  attr :icon, :string, required: false, default: nil
+  attr :icon_type, :string, required: false, default: "icon-svg"
+  attr :rest, :global
+
+  def cta(assigns) do
+    ~H"""
+    <a
+      id="cta-banner"
+      href={@link}
```
thecristen, line 488: When `@link` is nil, such as in the trip planner accessibility note, this ends up creating a `<a>` with no `href`, and also causes some fun side effect of presenting link-like behavior such as underlining on hover:

Image

**suggestion**: I think in this case the CTA shouldn't be a link element at all. Maybe this function can render different tags depending on whether `@link` is populated or not. Phoenix actually has a neat function that can help here ([dynamic_tag](https://phoenix-live-view.hexdocs.pm/Phoenix.Component.html#dynamic_tag/1) doc).

lib/dotcom_web/components/components.ex
```diff
@@ -442,9 +442,13 @@ defmodule DotcomWeb.Components do
   """
   def callout(assigns) do
     ~H"""
-    <div class="callout font-bold text-center" {@rest}>
+    <.cta
+      classes="callout font-bold text-center"
+      style="background-color: var(--colors-charcoal-90)"
```
thecristen:
```suggestion
      classes="callout font-bold text-center bg-charcoal-90"
```

**nit**: This should work too!

assets/css/_callout.scss
```diff
@@ -78,3 +78,15 @@
...
 }
+
+.cta-a {
+  margin-bottom: var(--spacing-5);
+  color: black;
+  text-decoration: none;
+  padding: var(--spacing-3);
+  display: flex;
+  align-items: center;
+  background-color: #e2ecf8;
```
thecristen: **nit**: I believe this can use `var(--colors-cobalt-90)`

thecristen: Though the mockup looks a little darker to me, making me wonder if it's `var(--colors-cobalt-80)`... however I cannot find the source at the moment 🙃

thecristen (changes requested):
Cool! Just requesting changes for one thing: the emergent fake link behavior is awkward. While it's probably not technically inaccessible, IMO this'd be more semantically pleasing without it.

assets/css/_callout.scss
thecristen: [confirmation](https://www.figma.com/design/fXXxB7zL0OsxOItwfWtAxQ/branch/5dRtncIa83jjLbvop1NjfN/MBTA.com?m=auto&node-id=10448-3115&t=imTpV8MNOlhhLsSa-1), it is 80

I think the rest might be already right to the spec

lib/dotcom_web/components/components.ex
lvachon1: Ooooh, `dynamic_tag` is great, thank you!

lib/dotcom_web/components/components.ex
```diff
@@ -504,7 +477,14 @@ defmodule DotcomWeb.Components do
...
-    <a
+    <.dynamic_tag
+      tag_name={
+        if @link do
+          "a"
+        else
+          "div"
+        end
+      }
       id="cta-banner"
```
thecristen: suggestion(blocking): Let's drop the hardcoded `id` attribute here so that we don't inadvertently give all CTAs the same ID

thecristen (changes requested):
...sorry for missing this one in my prior reviews (!)

lib/dotcom_web/components/components.ex
lvachon1: Good catch

thecristen (approved):
Let's roll

thecristen (approved):
Image Hm, that's a new one... you'll have to fix this to get it to merge.

PR #3319, opened 2026-07-07
Add CI check for updating translations

lilybarrett:
* Adds CI check for ensuring translations are up to date (Tip: Be sure to run `mix localize` locally before pushing up your PR!)
* Removes line numbers from `.pot` files to prevent how much the `.po`/`.pot` files have to be updated. Right now, there will be updates whenever content moves to a different line in the source code (even if the content of those translated messages itself does not change). My Charlie removed line numbers for this reason, and I figured it made sense to do the same here.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -108,6 +108,7 @@ jobs:
       - run: npm install --prefix assets
       - run: npm --prefix assets run webpack:build
       - run: mix compile --all-warnings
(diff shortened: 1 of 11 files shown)


joshlarson (approved):
Love it! I had originally thought that the line numbers would be helpful for finding the actual instances of `"text"`, but they generate so much diff noise, and just having the file names is enough.

PR #3327, opened 2026-07-10
fix(CMS.Repo): avoid caching errors as []

thecristen:
## Scope

Prompted by investigating the missing homepage content reported in [this thread](https://mbta.slack.com/archives/C022MNWN2CU/p1783693246668319) this morning!

## Implementation

The missing content on the page is associated with calls to `CMS.Repo.whats_happening()`. I noticed that its current implementation would cause errors from this call to get cached as empty lists.

### How?

To dissect the initial code:

```elixir
@decorate cacheable(..omitted for brevity)
def whats_happening do
  case @cms_api.view("/cms/whats-happening", []) do
    {:ok, api_data} -> Enum.map(api_data, &WhatsHappeningItem.from_api/1)
    _ -> []
  end
end
```

This `cacheable` decorator function has a default value set for its `:match` function as described [here](https://nebulex.hexdocs.pm/2.6.6/Nebulex.Caching.Decorators.html#module-shared-options), TLDR the following results are *not* cached: `:error`, `nil`, and `{:error, _}` tuples. Which means the above setup obscures errors from the `@cms_api.view/2` call by turning them into `[]`, which in turn **_will_** get cached.

To prevent this from happening, I adjusted this to follow a pattern we use in many places throughout Dotcom:

```elixir
def function_thats_called do
  case other_function_thats_cached do
    {:ok, value} -> (whatever)
    _ -> (whatever we want to obscure the error with)
  end
end

@decorate cacheable(...etc)
def other_function_thats_cached do
  @cms_api.view(...) # something that can actually return errors!
end
```

I took a cursory look for similar things happening elsewhere in the homepage, and adjusted those too.

### DIFF
--- a/priv/gettext/en/LC_MESSAGES/dotcom.po
+++ b/priv/gettext/en/LC_MESSAGES/dotcom.po
@@ -3861,11 +3861,6 @@ msgstr ""
 msgid "Service Inquiry"
 msgstr ""
 
-#: lib/dotcom_web/components/timetable.ex
-#, elixir-autogen, elixir-format
-msgid "Service alert or delay"
-msgstr ""
-
 #: lib/dotcom_web/components/components.ex
 #, elixir-autogen, elixir-format
 msgid "Service changes are in effect June 6 – July 12. Check each day of your line’s %{timetable_link} for the most up-to-date schedule."
--- a/lib/cms/repo.ex
+++ b/lib/cms/repo.ex
@@ -55,8 +55,8 @@ defmodule CMS.Repo do
   @spec news_entry_by(Keyword.t()) :: NewsEntry.t() | :not_found
   def news_entry_by(opts) do
     case do_news_entry_by(opts) do
-      [record | _] -> record
-      [] -> :not_found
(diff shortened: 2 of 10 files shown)


priv/gettext/en/LC_MESSAGES/dotcom.po
```diff
@@ -3861,11 +3861,6 @@ msgstr ""
 msgid "Service Inquiry"
 msgstr ""
 
-#: lib/dotcom_web/components/timetable.ex
-#, elixir-autogen, elixir-format
-msgid "Service alert or delay"
-msgstr ""
```
jlucytan, line 3867: question(non-blocking): why did these translation files end up updated here?

jlucytan (approved):
lgtm! just had a curiosity question if you happen to know the answer

priv/gettext/en/LC_MESSAGES/dotcom.po
thecristen: I don't know! 😅

PR #3329, opened 2026-07-10
fix(fares): Allow grouped fare cards to ingest Summarys

jlucytan:
## Scope

**Asana Ticket:** [🐞 500 error when visiting the latest (preview) version of the ferry fares page](https://app.asana.com/1/15492006741476/project/1213039586590742/task/1215975015154292)

## Implementation
Changes a few group_fare_card calls to handle the case that the fare returned from `fare_from_token` is a Summary instead of a Fare.

Additionally trims whitespace from tokens so `ferry_winthrop ` (note the space) doesn't result in an `invalid` fare

## Screenshots
Before
before fixing grouped fare card:
image

after fixing grouped fare card, before trimming single tokens: 
image

After
image

## How to test
Verify that this ferry fare [preview page](http://localhost:4001/fares/ferry-fares?preview=&vid=92067&nid=3643) loads and properly displays the grouped fare and the Winthrop fare.

### DIFF
--- a/test/dotcom/content_rewriters/liquid_objects/fare_test.exs
+++ b/test/dotcom/content_rewriters/liquid_objects/fare_test.exs
@@ -366,8 +366,17 @@ defmodule Dotcom.ContentRewriters.LiquidObjects.FareTest do
   end
 
   describe "fare_object_request" do
-    test "it handles fare requests with trailing spaces" do
-      assert fare_request("subway ") == {:ok, "$2.40"}
+    test "it handles fare object requests with trailing spaces" do
+      result =
+        Repo.all(
+          name: :ferry_winthrop,
+          includes_media: :mticket,
+          reduced: nil,
+          duration: :single_trip
+        )
+        |> List.first()
+
+      assert fare_object_request("ferry_winthrop ") == result
     end
 
     test "it handles fare requests for a given mode name and media" do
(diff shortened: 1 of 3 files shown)


test/dotcom/content_rewriters/liquid_objects/fare_test.exs
```diff
@@ -362,6 +366,10 @@ defmodule Dotcom.ContentRewriters.LiquidObjects.FareTest do
   end
 
   describe "fare_object_request" do
+    test "it handles fare requests with trailing spaces" do
+      assert fare_request("subway ") == {:ok, "$2.40"}
```
joshlarson: **Issue:** This test duplicates the test introduced above, and doesn't test the function in the `describe` invocation - I'm guessing you meant to test `fare_object_request/1` here?

joshlarson (changes requested):
Looks good! Well... I imagine content/design will want to update how exactly this renders, but at least for now, this is a huge improvement over the page 500-ing.

The one blocker for me is the test issue - once that's fixed, then I think this is good to go!

test/dotcom/content_rewriters/liquid_objects/fare_test.exs
jlucytan: Good catch, thanks!

joshlarson (approved):

PR #3335, opened 2026-07-13
Lb/update languages menu ux behind flag

lilybarrett:
## Scope

We want to redesign the header with a two-tier menu that, among other things, reimagines what the Languages button/menu looks like. This adds that redesign behind the feature flag that also enables Smartling translations .

**Asana Ticket:**
https://app.asana.com/1/15492006741476/project/1213994587940653/task/1216081563934501?focus=true

## Implementation
* Please note that this adds some changes to the Typescript code that handles interactions w/ the headers/menus on Dotcom -- namely to allow Dotcom to handle multiple tiers of headers rather than just one.
* Additional CSS classes and design updates behind the feature flag -- You may notice some code duplication here.
* Some minor code refactors

## Screenshots

Screenshot 2026-07-14 at 4 35 32 PM

## How to test
- Pull down this branch or navigate to the deployment branch
- Ensure nothing in the header UX/functionality looks broken (BEFORE turning on the feature flag)
- Navigate to `/_flags` and turn on the feature flag
- Interact with the new two-tiered header -- ensuring that buttons/links on both menu tiers still work as expected, changing the screen size, etc.
- Turn on Voiceover and tab around, ensuring that the Languages button adheres to accessibility expectations. (It should behave similarly to the existing buttons on the menu).

Note: Changes to the footer will come in future work, once designs are complete.


assets/css/_global-navigation.scss
```diff
@@ -56,6 +77,16 @@ nav.m-menu--desktop {
   }
 }
 
+nav.m-top-tier-menu--mobile {
```
lilybarrett, line 64: Tried to follow naming conventions with the new styling classes I added, but I'm sure I'm missing context/it's possible I'm not using them correctly. Please let me know!

assets/css/_header.scss
```diff
@@ -73,6 +73,75 @@
   }
 }
 
+.top-tier-nav-container {
```
lilybarrett, line 77: Tried to follow naming conventions with the new styling classes I added, but I'm sure I'm missing context/it's possible I'm not using them correctly. Please let me know!

assets/ts/app/global-navigation.ts
```diff
@@ -71,35 +69,73 @@ export function setup(rootElement: HTMLElement): void {
   clearAllBodyScrollLocks();
   if (!rootElement) return;
 
-  const header: HTMLElement = rootElement.querySelector("header")!;
-  if (!header) return;
+  // SUPPORT MULTIPLE HEADERS
```
lilybarrett, line 72: I know there's been some performance improvement work recently -- this code for handling/looping through multiple headers could benefit from additional eyes on it!

assets/css/_global-navigation.scss
heybenji: In the Rider Design System, we refer to this as the ["MBTA.com Header"](https://www.figma.com/design/IupfoRXJNItGDD8ZaZCxDz/MBTA-Rider-Design-System-Components?node-id=863-793&t=aSedBa1m9cPmLB3S-4) (catchy, I know) and the second tier as the ["MBTA.com Navbar"](https://www.figma.com/design/IupfoRXJNItGDD8ZaZCxDz/MBTA-Rider-Design-System-Components?node-id=900-527&t=aSedBa1m9cPmLB3S-4) component.

assets/css/_header.scss
```diff
@@ -73,6 +73,75 @@
...
+  position: relative;
+}
+
+.top-tier-header-navbar.new {
+  height: $header-navbar-height-new;
+
+  @include media-breakpoint-only(xs) {
+    height: $header-navbar-height-xs;
+  }
```
thecristen: **suggestion**: to better match the design, let's remove the defined `height` here and add vertical padding! Here's the CSS variable we have which corresponds to the `--spacing-small` variable used in Figma

```suggestion
  padding: var(--spacing-sm);
```

(and both mobile and desktop designs happen to use the same padding)

lib/dotcom_web/templates/layout/_top_tier_nav_desktop.html.heex
```diff
@@ -0,0 +1,47 @@
+<header class="top-tier-header new">
+  <div class="container top-tier-nav-container">
+    <nav id="languages-menu" class="top-tier-header-navbar new m-menu--desktop">
```
thecristen: Image

(desktop only) It looks like there's some extra padding on the right of the language menu that is making it misaligned with the search bar below it.

lib/dotcom_web/templates/layout/_top_tier_nav_mobile.html.heex
```diff
@@ -0,0 +1,46 @@
+<header class="top-tier-header new top-tier-nav">
+  <div class="container top-tier-nav-container new">
+    <nav id="languages-menu" class="top-tier-header-navbar new m-top-tier-menu--mobile">
```
thecristen: Image

(mobile only)

**question**: Is darkening the bottom navbar when the top nav menu is expanded expected here?

assets/css/_global-navigation.scss
```diff
@@ -23,6 +23,28 @@ header {
...
+    text {
+      fill: $white;
+    }
+  }
+
+  background-color: $brand-primary-darkest;
+  color: $white;
+}
+
+
```
thecristen, line 38: I'm not sure this section is needed, because I think the existing CSS will achieve the same as-is?

lib/dotcom_web/templates/layout/_new_header.html.heex
```diff
@@ -33,17 +39,33 @@
...
+            type="button"
+            aria-expanded="false"
+            aria-controls="search-header-mobile__container"
+            aria-label={~t(Search MBTA.com)}
+          >
+            <i class="fa fa-fw fa-search" aria-hidden="true"></i>
+            <i class="fa fa-fw fa-close" aria-hidden="true"></i>
+          </button>
+        </div>
+      <% end %>
```
thecristen, line 69: I'm a bit confused why this diff is needed! Could you explain what's happening here?

lib/dotcom_web/templates/layout/_new_nav_desktop.html.heex
```diff
@@ -12,7 +12,7 @@
       <span class="c-indicator--angle"></span>
     </a>
 
-    <div class="m-menu--desktop__menu">
+    <div class={"#{if Laboratory.enabled?(@conn, :use_smartling_translations), do: "m-menu--multiple_tiered_desktop__menu", else: "m-menu--desktop__menu"}"}>
```
thecristen: I think you might be able to get away without toggling the CSS class here (and I don't think you need the `.m-menu--multiple_tiered_desktop__menu` class either)

lib/dotcom_web/templates/layout/_top_tier_nav_mobile.html.heex
```diff
@@ -0,0 +1,46 @@
+<header class="top-tier-header new top-tier-nav">
```
thecristen: **question**: Why have a separate template for the mobile version of the top tier nav? There's almost no designed content difference between the desktop/mobile designs, only the MBTA logo change that could be handled in CSS. I think you don't need this file (and the associated CSS duplicated for it).

thecristen (changes requested):
Thank you for this addition and light refactor - you're a great pattern-matcher!

In addition to the suggestions here, I noticed that the top tier nav completely disappears when you're between certain desktop and mobile breakpoints 😅 (shrink the page width and you'll see!).

assets/css/_global-navigation.scss
lilybarrett: @thecristen There are subtle differences: the logos need to be white, and the background color needs to be darker! I agree there's a lot of repetition here, though. Thoughts on a better way to handle it?

lib/dotcom_web/templates/layout/_top_tier_nav_mobile.html.heex
lilybarrett: There are _some_ different classes applied in either case -- but I'm still getting familiar with Elixir templates, so I imagine there's a better way to handle even that. Any suggestions?

lib/dotcom_web/templates/layout/_new_nav_desktop.html.heex
lilybarrett: The only important difference included in the new class is this:

```
top: $header-navbar-height-new * 2;
```

I'll try to find a better way to account for that without changing the original class!

lib/dotcom_web/templates/layout/_new_header.html.heex
lilybarrett: It's a very subtle difference in styling (with the addition of the new classes) that moves the logo in the original header. In the original header, on mobile, the logo is between the menu and the search bar. In the new design, the logo is always left-aligned.

I'm definitely not saying this is the best way to implement that, though -- especially being an Elixir newbie. Let me know your thoughts!

lib/dotcom_web/templates/layout/_top_tier_nav_mobile.html.heex
lilybarrett: That's a good question...I'll follow up with Chase!

assets/css/_global-navigation.scss
lilybarrett: For posterity, we were able to remove a lot of the new class code but kept the background and foreground colors.

lib/dotcom_web/templates/layout/_new_header.html.heex
lilybarrett: For posterity: We were able to remove one of the new classes!

lib/dotcom_web/templates/layout/_new_nav_desktop.html.heex
lilybarrett: For posterity, the plan: Don't add the other styles to the new class, just the top value. Toggle that class on/off based on the flag, let it work with the previous class.

lib/dotcom_web/templates/layout/_top_tier_nav_mobile.html.heex
lilybarrett: For posterity: The logo change is valid, but the toggle class only adds a margin-right value that may not be doing anything particularly useful

lib/dotcom_web/templates/layout/_new_nav_desktop.html.heex
lilybarrett: For posterity: Applied `position: relative` to the bottom header and the height adjusts automatically!

lib/dotcom_web/templates/layout/_top_tier_nav_desktop.html.heex
lilybarrett: Fixed!

lilybarrett: Fixed!

assets/css/_global-navigation.scss
lilybarrett: It seems like the navbar uses some `header`-prefixed classes, so this may involve some work to standardize the naming. @thecristen Thoughts on whether a new ticket should be created to standardize the naming for the two-tiered header in the future?

lib/dotcom_web/templates/layout/_new_header.html.heex
```diff
@@ -1,5 +1,11 @@
-<aside class="c-modal__cover m-menu--cover" aria-hidden="true" data-nav="veil"></aside>
-<header class="header--new">
+<%= if Laboratory.enabled?(@conn, :use_smartling_translations) do %>
+  {render("_top_tier_nav_mobile.html", assigns)}
+  {render("_top_tier_nav_desktop.html", assigns)}
```
thecristen: Having both of these in the DOM is causing `<nav id="languages-menu">` to be duplicated (ideally `id` values should be unique per-document).

I don't think you need both of these. If we reveal the current`_top_tier_nav_mobile.html` on the desktop breakpoint... it looks (and seems to work) the exact same as the `_top_tier_nav_desktop.html` output. Can we remove one?

Image

thecristen (changes requested):

lib/dotcom_web/templates/layout/_new_header.html.heex
lilybarrett: @thecristen On *actual* mobile, there's a difference in the logo/logo size that's used, but I think that's actually already being handled with CSS. I'll remove one of these.

lilybarrett: @thecristen Fixed here: https://github.com/mbta/dotcom/pull/3335/changes/00319d2c8dd0991171468c058877bbb660bed72a

thecristen (approved):
🎉 Thanks!!

PR #3337, opened 2026-07-14
chore: Add a mix task for deploying with the right tag

joshlarson:
## Scope

We've had a few instances recently of manually typing out the wrong tag number, and this is the sort of thing that computers can do quite well! So I stole a bit of [my work](https://github.com/mbta/skate/pull/2538) from my days on Skate, and this is the result!

## How to test

Run `mix deploy.prod` and see it open a browser window with the right tag selected.

You can also fiddle with the `today_formatted()` function if you want to pretend to be running the script on a different day, in order to test out the `same_day?` logic.

(Don't worry - the mix task won't actually execute a deploy - it just opens the [New Release page](https://github.com/mbta/dotcom/releases/new) with the right tag set - you'll still have to publish the release and approve the Github Action before anything goes to prod!)

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -4,12 +4,12 @@
 
 This can be done on GitHub Actions via the following:
 
-- Create a tag following this format
-  - Pull the latest main locally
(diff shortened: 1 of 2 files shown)


thecristen (approved):
I just used it to create today's release! 😆 Thanks

PR #3338, opened 2026-07-15
refactor & test homepage alerts

thecristen:
## Scope

**Asana Ticket:** [📈 Speed up alerts call on the homepage](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216174323767592?focus=true)

## Implementation

Showing alerts on the homepage means we have to fetch all alerts, group them by applicable mode and accessibility effect, and fetch all the relating routes and stops so we could render the right route pills or stop links. When starting from an empty cache it's a _lot_ of data fetching!

This PR does a couple things:

1. The call to fetch all alerts was doing a little bit more than it needed to -- notably, `Alerts.Repo.all/1` calls a function that calls `Alerts.Sort.sort/1`. For the homepage, we don't need any alert sorting - so I added a function that fetches all current alerts, without the sorting.
2. Extracted the two data pipelines into two functions that I can add tests for. And rewrote them! The most impactful change there was probably using `Task.async_stream/3` to manage fetching many routes or stops in parallel. Another improvement over the original code is that it doesn't break if the `Routes.Repo` and `Stops.Repo` calls return `nil` (which is a valid case).

## How to test

I used `Util.log_duration/3` to measure how long it takes to render the alerts tab HTML content given a list of alerts. All numbers are in `ms`

- this branch, no cache - 883.581
- this, everything cached - 3.682
- `main` branch, no cache - 8338.736
- `main`, everything cached - 6.147

### DIFF
--- a/lib/dotcom/alerts.ex
+++ b/lib/dotcom/alerts.ex
@@ -262,4 +262,84 @@ defmodule Dotcom.Alerts do
       %{} -> false
     end)
   end
+
+  def routes_with_high_priority_alerts_by_mode(alerts) do
+    modes = [:subway, :bus, :commuter_rail, :ferry]
+    empty_by_mode = Map.new(modes, fn mode -> {mode, MapSet.new()} end)
+
+    route_ids_by_mode =
+      alerts
+      |> Enum.filter(&(Alerts.Priority.priority(&1) == :high))
+      |> Enum.reduce(empty_by_mode, fn alert, acc ->
+        route_ids = Alert.get_entity(alert, :route) |> MapSet.delete(nil)
+
+        alert
+        |> alert_route_type()
+        |> Enum.map(&Route.type_atom/1)
+        |> Enum.reduce(acc, fn mode, acc2 ->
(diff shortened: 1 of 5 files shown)


lib/dotcom/alerts.ex
```diff
@@ -262,4 +262,84 @@ defmodule Dotcom.Alerts do
       %{} -> false
     end)
   end
+
+  def routes_with_high_priority_alerts_by_mode(alerts) do
```
joshlarson, line 266: **Question:** Would it improve performance even more if we cached this? Either with a `@cacheable` decorator with a TTL of 1min, or with a GenServer, similar to SubwayStatus?

joshlarson (approved):

lib/dotcom/alerts.ex
thecristen: I'd lean toward a `GenServer` so that we could keep it up to date with the underlying alerts. Maybe as a follow-on task?

thecristen: Made a followup ticket, thanks for the suggestion!

PR #3341, opened 2026-07-17
chore(Routes.Route): remove param override

thecristen:
## Scope

**Asana Ticket:** [🐞 GL search results aren't working for branches](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214902494254017?focus=true)

## Implementation

We used the `Phoenix.Param` [protocol](https://phoenix.hexdocs.pm/Phoenix.Param.html) to convert the Route struct into URL parameters, but I don't think we wish to convert all Green Line branch links to direct to the combined Green Line page anymore.

So I deleted this.

## How to test

`Util.site_path(:schedule_path, [:show, Routes.Repo.get("Green-B")])` outputs `"/schedules/Green-B"` instead of `"/schedules/Green"`

### DIFF
--- a/lib/routes/route.ex
+++ b/lib/routes/route.ex
@@ -307,12 +307,6 @@ defmodule Routes.Route do
   end
 end
 
-defimpl Phoenix.Param, for: Routes.Route do
-  alias Routes.Route
-  def to_param(%Route{id: "Green" <> _rest}), do: "Green"
-  def to_param(%Route{id: id}), do: id
-end
-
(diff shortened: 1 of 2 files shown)


joshlarson (approved):
Nice catch - this feels like one of those things that could easily have driven someone (me) up a tree, trying to figure out why `~p"/schedules/#{%Route{id: "Green-E"}}"` was returning `/schedules/Green`... I wouldn't even have known to look here!

PR #3343, opened 2026-07-17
feat(Dotcom.Telemetry): more memory metrics

thecristen:
1. Enables metrics on the LiveDashboard, which isn't pretty but does let us see metrics go up and down locally
2. Sends more detailed metrics about memory.

On dev-green you can see the additional metrics (bottom chart)
image


joshlarson (approved):
📈📉📈📉

PR #3344, opened 2026-07-17
chore: hibernate_after in many places

thecristen:
From debugging our application locally over the past week, I found that
- most of our memory usage is tied up in **processes** (this is almost true for prod metrics too, but there we have even more memory attributed to our usage of ETS)
- many processes maintain a pretty large heap size.

From the GenServer documentation:

> Hibernating a [GenServer](https://elixir.hexdocs.pm/1.19.5/GenServer.html) causes garbage collection and leaves a continuous heap that minimises the memory used by the process.
>
> Hibernating should not be used aggressively as too much time could be spent garbage collecting, which would delay the processing of incoming messages. Normally it should only be used when you are not expecting new messages to immediately arrive and minimising the memory of the process is shown to be beneficial.

So in this branch I explored using this strategically.
- the alerts-fetching processes were already using `:hibernate` (except after initializing, which I've added here), so I figured it's ok to have the `SystemStatus` caches, which rely on alerts, to use these too
- `ServiceDateRollover` is a ~once a day thing, so I added `:hibernate` after that message,
- for a few other places: `:hibernate_after` will hibernate after a given amount of milliseconds without messages. This felt appropriate in most cases, because most of our GenServers have sporadic activity

Some slightly more involved tweaks here:
- fixup(SystemStatus.SubwayCache): remove state - since we already have the status saved in the ETS table, it didn't make sense to also save it in the GenServer state
- fixup(Vehicles.Repo): remove table from state - similar story here, but kept a reference to the table name

### Results 

Alas, what do the `:observer` load charts say. The lower left charts depict memory usage pretty soon after application startup.

| Before | After |
|--------|--------|
| image | image | 

There's a noticeable reduction in the baseline memory usage here.

### DIFF
--- a/lib/alerts/cache/store.ex
+++ b/lib/alerts/cache/store.ex
@@ -149,7 +149,7 @@ defmodule Alerts.Cache.Store do
     # no cover
     _ = :ets.new(:alert_banner, [:set, :protected, :named_table, read_concurrency: true])
 
-    {:ok, []}
+    {:ok, [], :hibernate}
   end
 
   @impl true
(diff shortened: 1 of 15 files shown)


lvachon1 (approved):
Everything looks good and works well on my end.

PR #3354, opened 2026-07-21
QF | [WC Cleanup]: Add redirect on Boston Stadium Train timetable page

lvachon1:
## Scope

**Asana Ticket:** [QF | ⚽️❌ Add redirect on Boston Stadium Train timetable page](https://app.asana.com/1/15492006741476/project/555089885850811/task/1214427583552363?focus=true)

## Implementation

Removed WorldCupTimetableLive scope entry in router,  added redirect away from that URL to /schedules/commuter-rail

## Screenshots

N/A

## How to test

http://localhost:4001/schedules/bostonstadium
Confirm that link redirects to the main commuter rail schedules page


joshlarson (approved):
I might be wrong, but could we also remove `WorldCupTimetableLive` and its associated tests, either as part of this or as a follow-up?

PR #3360, opened 2026-07-22
chore(Application): stop running unused stream

thecristen:
This feature was added back when we were developing the current iteration of the stop page in 
- #1585 

As explained in the linked PR, this was added because the existing data emitted was already transformed into map markers for usage on the existing Leaflet map, which didn't lend itself well for consuming in other ways. But we never ended up adding live vehicles to the stop page, so that new vehicle channel was never consumed by anything! 

I decided to just comment it out, since I think we'd like to use it soon enough.

## How to test

Can verify this hasn't impacted showing live vehicles on the line diagram pages.


lvachon1 (approved):
Bus, Subway, and Rail live vehicle positions on the line diagram map look good to me.  Are there any other places that we have them?

PR #3361, opened 2026-07-24
feat: Have new ferry timetables scroll to the right trip

joshlarson:
## Scope

~No ticket, but I noticed that this hasn't worked since [I introduced the new timetable for loopy ferries](https://github.com/mbta/dotcom/pull/3206)... I just completely forgot to do the offset thing!~

I was wrong! There was a ticket!

**Asana Ticket**: [⛴️🐞 New timetables don't auto-scroll to the right trip](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216392923703216?focus=true)

## Implementation

- Change the timetable to deal with `DateTime`'s instead of `String`'s. 
- Have the `Timetables` module calculate the offset based on a `date_time` argument.

The commits in this PR do somewhat show a journey, since I changed my approach partway through, but the easiest way to review it is probably to look at [the first commit](https://github.com/mbta/dotcom/pull/3361/changes/16b74de072b014f8e6c24a1f747bb0393fd831d9), which changes the times stored in the `Timetable` struct from strings to proper `DateTime`'s, and then to look at the whole PR (most importantly, the `Dotcom.Timetables.first_unfinished_trip_index/2` function).

## Screenshots

Screenshot 2026-07-24 at 2 38 00 PM

## How to test

Check out timetables for the [Quincy](http://localhost:4001/schedules/Boat-F7/timetable), [Winthrop](http://localhost:4001/schedules/Boat-F6/timetable), and [Harbor Loop](http://localhost:4001/schedules/Boat-F10/timetable) ferries. Note that they scroll to the right trip.

You can fake the definition of `now` in order to try out other times (like before the first trip starts, or after the last trip ends) by changing [this line](https://github.com/mbta/dotcom/blob/63c29d0a1166b1b9a29c865c996d58756d469ff6/lib/dotcom/utils/date_time.ex#L30) to something like 👇 
```elixir
def now(), do: ~N[2026-07-24T14:43:00] |> Timex.to_datetime("America/New_York")
```

### DIFF
--- a/lib/dotcom/timetables/timetable.ex
+++ b/lib/dotcom/timetables/timetable.ex
@@ -25,7 +25,7 @@ defmodule Dotcom.Timetables.Timetable do
     @moduledoc """
     A struct representing timetable cells. See `Dotcom.Timetables` for more information.
     """
-    defstruct [:time, :trip, :stop_id]
+    defstruct [:time, :trip]
 
     @type t() :: %__MODULE__{
             time: String.t(),
--- a/lib/dotcom/timetables/timetable.ex
+++ b/lib/dotcom/timetables/timetable.ex
@@ -2,11 +2,12 @@ defmodule Dotcom.Timetables.Timetable do
   @moduledoc """
   A struct representing timetables. See `Dotcom.Timetables` for more information.
   """
-  defstruct [:rows, :trips]
+  defstruct [:rows, :trips, :offset]
 
   @type t() :: %__MODULE__{
           rows: [__MODULE__.Row.t()],
-          trips: [Schedules.Trip.t()]
+          trips: [Schedules.Trip.t()],
+          offset: non_neg_integer()
         }
 
   defmodule Row do
--- a/lib/dotcom/timetables/timetable.ex
+++ b/lib/dotcom/timetables/timetable.ex
@@ -2,12 +2,11 @@ defmodule Dotcom.Timetables.Timetable do
   @moduledoc """
   A struct representing timetables. See `Dotcom.Timetables` for more information.
   """
-  defstruct [:rows, :trips, :offset]
+  defstruct [:rows, :trips]
 
   @type t() :: %__MODULE__{
           rows: [__MODULE__.Row.t()],
-          trips: [Schedules.Trip.t()],
-          offset: non_neg_integer()
+          trips: [Schedules.Trip.t()]
         }
 
   defmodule Row do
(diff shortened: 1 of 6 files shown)


joshlarson:
I'm actually a bit on the fence about whether this actually should be calculated in the timetable. I could very easily be persuaded that `offset` is actually a frontend concern, and that the timetable shouldn't have to think about it (and shouldn't have to think about the current time either). The existence of `offset` also forces the timetable to have to take an argument other than the schedules, which makes it less _pure_ (it also means that we can't cache it, or at least can't cache it for as long).

I did it this way because it was more convenient - it uses `schedule_lists_for_trips` to figure out the offset, and `schedule_lists_for_trips` is an intermediate variable that doesn't exist (in its current form) on the finished timetable.

I'm interested in what others think!

joshlarson:
I've become convinced by my argument above that `offset` shouldn't be part of the timetable itself. I'm gonna refactor this a bit before putting it up for actual review.

lib/dotcom/timetables/timetable.ex
```diff
@@ -25,7 +25,7 @@ defmodule Dotcom.Timetables.Timetable do
     @moduledoc """
     A struct representing timetable cells. See `Dotcom.Timetables` for more information.
     """
-    defstruct [:time, :trip, :stop_id]
```
joshlarson, line 28: Not strictly related to this PR, but this was wrong, so I figured I'd fix it.

lib/dotcom_web/components/timetable.ex
```diff
@@ -153,6 +155,10 @@ defmodule DotcomWeb.Components.Timetable do
   # `:if={!ferry?(@route)}`, which means that that section will need
   # to be added here before we can use this for non-ferry routes.
   def timetable(assigns) do
+    assigns =
+      assigns
+      |> assign(:offset, Timetables.first_unfinished_trip_index(assigns.timetable, assigns.now))
```
joshlarson, line 160: I'm on the fence about whether the call to `first_unfinished_trip_index` should be here, versus in the controller - I sort of see an argument to be made in favor of having the field be something like `initial_scroll_position`, so that the component is doing less thinking and more rendering. That was what I first tried, but it felt silly to do computation on the timetable right before passing that very same timetable into this component, so here we are.

I'm gonna leave it as-is for now, unless anyone objects 🙂

thecristen: I like what you ended up deciding here!

thecristen (approved):
This is really cool!

PR #3363, opened 2026-07-27
Experiment:  Preserve customer support form info between refreshes

lvachon1:
## Scope

**Asana Ticket:** [🙋 Investigate customer support form reloading issues](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216855140054434?focus=true)

## Implementation

As part of the above investigation I have written a stop-gap/band-aid fix that will make the experience less frustrating for users that experience page refreshes.

This JS code saves the state of the form in localStorage when any field is changed.  If the page is refreshed/loaded and the cached form data is <24h old, it will fill out the form and put the user back where they were.  

Unfortunately it is not allowed to set file inputs from JS, the user will have to manually re-select them. 

The Captcha will/should not be affected either.  Those clicks actually expire after a few minutes.

## Screenshots

https://github.com/user-attachments/assets/61533828-5c0e-4a4c-a005-fab75de26304

## How to test

http://localhost:4001/customer-support

Play with the form, put it into various states, refresh and see if it behaves as expected.  Drop-downs, text fields, hidden/shown sections and checkboxes should all resume their previous state upon refresh.  If a user is typing when the form crashes/refreshes they should lose at most 1 character.

## Moving Forward

It appears that our DotcomWeb controllers import Phoenix.LiveView and that establishes a websocket connection and hearbeats.  However, on this page at least, no other data is being sent that way.  I think we should see if we can remove that requirement and see if the lack of a websocket connection will help.

### DIFF
--- a/assets/js/support-form.js
+++ b/assets/js/support-form.js
@@ -21,6 +21,10 @@ export default function($ = window.jQuery) {
         handleSubmitClick($, toUpload);
         handleModeChangeSelection($);
         handleSubjectChange($);
+        loadCache();
+        document.getElementById("support").addEventListener("change", saveCache)
+        document.getElementById("support").addEventListener("submit", clearCache)
+
       });
     },
     { passive: true }
@@ -551,6 +555,7 @@ export function handleSubmitClick($, toUpload) {
         $(".support-confirmation--error").addClass("hidden-xs-up");
         reactivateSubmitButton($);
         $("#support").remove();
+        clearCache();
       },
       error: (xhr, errorString, errorObject) => {
         console.error(errorString);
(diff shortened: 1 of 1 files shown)


assets/js/support-form.js
```diff
@@ -576,3 +582,34 @@ export function handleSubjectChange($) {
...
+  localStorage.setItem("support-form-cache", cacheString)
+
+}
+
+export const loadCache = ()=>{
+  const cacheString = localStorage.getItem("support-form-cache")
+
+  if(!cacheString || cacheString.length<10){return;}//No data, bail
+  const cacheData = JSON.parse(cacheString)
+  if(cacheData.time < Date.now()-86400000){return;}//Data is too old, bail
```
joshlarson: **Suggestion (non-blocking):** Some quick back-of-the-envelop math tells me that this is one day. How would you feel about pulling this out to a constant, like so, to make it super explicit?
```javascript
ONE_DAY_MS = 1000 * 60 * 60 * 24
```

assets/js/support-form.js
```diff
@@ -576,3 +582,34 @@ export function handleSubjectChange($) {
...
+  const cacheString = JSON.stringify({time: Date.now(), data: Array.from(elements).map(elem => [elem.id, elem.checked || elem.value])})
+  localStorage.setItem("support-form-cache", cacheString)
+
+}
+
+export const loadCache = ()=>{
+  const cacheString = localStorage.getItem("support-form-cache")
+
+  if(!cacheString || cacheString.length<10){return;}//No data, bail
+  const cacheData = JSON.parse(cacheString)
```
joshlarson: **Issue (blocking):** If `cacheString` somehow winds up becoming an invalid JSON, then that causes the javascript to crash, which prevents future attempts to save state (and I think it also prevents form submission). This error is unrecoverable unless the user clears their browser data.

**Another issue (non-blocking):** If the `time` field somehow goes missing, but the JSON is otherwise valid, then `cacheData.time` will become `undefined`, and in its infinite wisdom, Javascript has decided that all numerical comparisons with `undefined` should evaluate to `false`, which means that we'll keep our `undefined`-time old submission around forever. I'm thinking if `cacheData.time` is `undefined`, we probably also want to bail.

joshlarson (changes requested):
This is a great change! I left a few comments/suggestions. The one thing that's holding me back from approving is the invalid-JSON thing... I know it's unlikely, but rule 1 about getting data from users is to assume that it's probably invalid or wrong in some way 😅

assets/js/support-form.js
lvachon1: I've turned the magic number into a more readable constant

assets/js/support-form.js
lvachon1: I've put a try/catch block around the JSON parsing and everything that depends on it succeeding, and I've added checks to make sure our data is valid before we do any work on it.

joshlarson (approved):

PR #3364, opened 2026-07-27
Update timetable date picker to support selecting the current date

lvachon1:
## Scope

**Asana Ticket:** [Update timetable date picker to support selecting the current date](https://app.asana.com/1/15492006741476/project/385363666817452/task/1215567127775712?focus=true)

## Implementation

Removed the special case that handled "today" and returned a nil URL.  Now "today" is treated like any other day.

## Screenshots

Screenshot 2026-07-27 at 1 21 06 PM

## How to test

Open any timetable (http://localhost:4001/schedules/CR-NewBedford), open the datepicker and select Today (while already viewing Today's timetable).  Confirm that clicking on Today behaves the same as clicking on any other valid date.

### DIFF
--- a/lib/build_calendar.ex
+++ b/lib/build_calendar.ex
@@ -196,10 +196,6 @@ defmodule BuildCalendar do
   end
 
   @spec build_url(url_fn, Date.t(), Date.t()) :: String.t()
-  defp build_url(url_fn, today, today) do
-    url_fn.(date: nil, date_select: nil, shift: nil)
-  end
-
   defp build_url(url_fn, date, _) do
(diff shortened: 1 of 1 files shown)


joshlarson (changes requested):
I misunderstood this too in exactly the same way when I first read through the ticket, but the issue isn't how we were handling today (I actually think that the special-casing is good, and we should keep it) - it's what happens if you select the date that's currently selected. The problem is that if you're look at the timetable for a particular date, and then you select that date in the calendar picker, it keeps the calendar picker open and doesn't do anything.

lvachon1:
Closing since I solved the wrong problem, oops!

PR #3365, opened 2026-07-27
[Fix]: Search tool breaks when you hit "enter" too soon

lvachon1:
## Scope

**Asana Ticket:** [🐞🔎 Search tool breaks when you hit "enter" too soon](https://app.asana.com/1/15492006741476/project/385363666817452/task/1213972909607100?focus=true)

## Implementation

Added event handler to handle the case where the user presses enter before the debounce period expires

## Screenshots

https://github.com/user-attachments/assets/364e345f-a36e-4ac6-8445-b21d73429052

## How to test

http://localhost:4001/search

Enter a search term and press enter very quickly (<300ms after the last keystroke) and confirm that results now show.

### DIFF
--- a/lib/dotcom_web/live/search_page_live.ex
+++ b/lib/dotcom_web/live/search_page_live.ex
@@ -59,7 +59,27 @@ defmodule DotcomWeb.SearchPageLive do
     end
   end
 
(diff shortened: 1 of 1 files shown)


joshlarson (approved):
🔎

PR #3366, opened 2026-07-27
feat: Add a new timetable implementation behind a feature flag

joshlarson:
## Scope

One of the blockers to the new Harbor Loop ferry layout (which puts both timetables on the same page) is that the original timetable implementation has controls that assume that there's only one timetable on the page (if there are multiple, then the `Earlier Boats`/`Later Boats` buttons cause all timetables to scroll). I spent some time digging through the existing Javascript code, and decided to try my hand at rebuilding some of it in a more "Phoenix-forward" way, using Phoenix hooks, and relying more on tailwind, rather than having styling attached to parts of the timetable on load. At the time, I figured that would be easier than trying to fiddle with the existing JS code / timetable DOM and trying to guess what each bit was doing.

This was the result.

**Asana Ticket:** [⛴️ Tie "Earlier Trips"/"Later Trips" controls directly to the timetable component](https://app.asana.com/1/15492006741476/project/555089885850811/task/1215456475351494?focus=true)

## Implementation

- This PR adds a new feature flag: `new_timetables`, and uses it to decide what timetable logic to run.
  - When the flag is set to true, then it uses the new timetable logic for all timetables (this is obviously not production-ready, since the new timetable code lacks features that exist on CR timetables, and produces a slightly wrong timetable for more complicated ferry routes like [the Hingham/Hull](http://localhost:4001/schedules/Boat-F2H/timetable), but it's useful for comparing).
  - When the flag is false, then we continue as we previously did, using the new timetable backend logic only for the Quincy, Winthrop, and Harbor Loop ferry routes, and using the old timetable components for all routes.
 - The old timetable relies on a few Javascript files to work.
   - [`scroll-to.js`](https://github.com/mbta/dotcom/blob/f4cae04aba4fd68c99469227a8ca6e8a9c8e73d8/assets/js/scroll-to.js): Used to make the timetables scroll to the right trip automatically on page-load.
     - Unchanged.
   - [`timetable-scroll.js`](https://github.com/mbta/dotcom/blob/f4cae04aba4fd68c99469227a8ca6e8a9c8e73d8/assets/js/timetable-scroll.js): This is the code that runs when the `Earlier Boats`/`Later Boats` buttons are clicked.
     - Transplanted to a Phoenix hook.
   - [`timetable-stick.js`](https://github.com/mbta/dotcom/blob/f4cae04aba4fd68c99469227a8ca6e8a9c8e73d8/assets/js/timetable-stick.js): This is what causes the trip-name headers to look like they stick to the top of the screen on CR timetables.
     - Unchanged for now (this feature doesn't work on the new timetables yet, and isn't necessary for ferry timetables); will likely be transplanted to a Phoenix hook when the time comes.
   - [`timetable-style.js`](https://github.com/mbta/dotcom/blob/f4cae04aba4fd68c99469227a8ca6e8a9c8e73d8/assets/js/timetable-style.js): This does a bunch of things, including setting some of the styling that makes the stop header column sticky.
     - Mostly replaced with Tailwind/CSS. The rest will likely be transplanted to a Phoenix hook.

## Screenshots

Screenshot 2026-07-24 at 6 47 14 PM

☝️ is what the new timetable looks like (with a little disclaimer banner that indicates the new-ness of the new timetable).

This timetable should mostly look the same as the existing one, but it does resolve a few visual quirks. The screenshots below are from the old timetable, showing visual bugs that I wasn't able to produce on this branch:

---

👇  While we don't officially support a no-JS version of the site, the Javascript on this page sometimes takes long enough to load that this flashes on the screen... or even lingers (when I took this screenshot, it was like this for about ten seconds). 👇 
Screenshot 2026-07-27 at 3 40 05 PM

---

👇  Some exciting quirks at certain screen widths where the shadow extends a bit too far up or down, making it look randomly darker in a few spots. 👇 
Screenshot 2026-07-24 at 4 27 09 PM
Screenshot 2026-07-27 at 5 36 52 PM

---

👇  At some screen widths, the above-the-headers fake-table isn't aligned exactly with the table below, leading to an awkward jog 👇
Screenshot 2026-07-27 at 5 36 09 PM

---

👇 This wasn't a quirk exactly, but I thought it would be nice to keep the earlier/later train/boat buttons at narrow screen widths, so instead of fully hiding those buttons at narrow screen widths, I decided to just hide the words.

Screenshot 2026-07-27 at 5 44 17 PM

## How to test

Go to [the flags dashboard](http://localhost:4001/_flags) and set `New timetables` to true. Then navigate to any timetable page you're interested in, and have fun!

## Notes

This is not a complete feature. I think this component is mostly, but not entirely, ready to go to production for the loopy ferry routes, but it still needs a few things. If we decide to roll forward with this PR, then I'll make follow-up tickets for those:
- We need to make sure that this has the same screen reader readout as the existing timetable, including the aria-label for the table itself (which miraculously doesn't overwrite the whole table in screen-reader-land!)
- Stop rows don't show parking and accessibility info
- Earlier/Later buttons aren't disabled at the left-most and right-most parts of their journeys.
- Unsure about this, but cell padding may need to be increased - the cells currently look cramped compared to the production timetables, although it's possible that that's because the aforementioned parking and accessibility info forces the row headers (and thus the whole rows) taller.

Once ☝️ issues are resolved, then we can roll this out by having `Boat-F6`, `Boat-F7`, and `Boat-F10` use the new timetable component whether or not the feature flag is enabled, and continue to serve the old component for other routes when the flag is disabled.

... and there are other issues that are CR-specific that need to be resolved before this component can be used for CR timetables:
- Currently no sticky headers - once you scroll past the trip names, they aren't visible anymore.
- No track-change, flag stop, early departure icons in the cells.
- No bikes-allowed/disallowed icons on the trip headers.
(to be fair, the [old "new timetable" component](https://github.com/mbta/dotcom/blob/f4cae04aba4fd68c99469227a8ca6e8a9c8e73d8/lib/dotcom_web/components/timetable.ex#L149-L233) also doesn't support the various icons, although it does support sticky headers)

As well as a few issues around the timetable algorithm and data (both issues are only visible with `New timetables` enabled - the timetables render correctly in production):
- The [Needham Line](http://localhost:4001/schedules/CR-Needham/timetable?date=2026-09-01&schedule_direction[direction_id]=0#date-filter) gets a few trips from the Providence Line, which causes the new timetable algorithm to show the whole Providence Line.
- The new timetable algorithm is greedy, which doesn't work when the first few trips have stops in a particular order. The [Hingham/Hull timetable](http://localhost:4001/schedules/Boat-F2H/timetable?date=2026-09-01#date-filter) demonstrates this clearly, with a few duplicated rows.

All of these are known, and fortunately don't affect the Harbor Loop Ferry, which is the one that needs this component the most!


assets/js/scroll-to.js
```diff
@@ -4,10 +4,11 @@ export default () => {
 
 const scrollTo = () => {
   window.requestAnimationFrame(() => {
-    const initialElToScrollTo = document.querySelector("[data-scroll-to]");
-    if (initialElToScrollTo) {
-      doScrollTo(initialElToScrollTo);
-    }
+    document.querySelectorAll("[data-scroll-to]").forEach((elToScrollTo) => {
+      if (elToScrollTo) {
```
thecristen, line 8: **nit**: I can't imagine `elToScrollTo` would ever be `undefined` or `null` in this situation??

assets/ts/.eslintrc.js
```diff
@@ -34,7 +34,16 @@ const config = {
     "@typescript-eslint/no-var-requires": "off",
     // note you must disable the base rule as it can report incorrect errors
     "no-shadow": "off",
-    "@typescript-eslint/no-shadow": ["error"]
+    "@typescript-eslint/no-shadow": ["error"],
+    // Allows us to mark variables as intentionally unused with a leading _underscore
```
thecristen, line 38: neat!

lib/dotcom_web/components/new_timetable.ex
```diff
@@ -0,0 +1,143 @@
...
+  defp header_column_classes() do
+    "w-36 sm:w-44 md:w-52"
+  end
+
+  defp ferry?(route), do: Routes.Route.type_atom(route) == :ferry
+
+  defp format!(nil), do: ""
+  defp format!(time), do: Dotcom.Utils.Time.format!(time, :hour_12_minutes)
+
+  defp vehicle_name(route), do: Routes.Route.vehicle_name(route) <> "s"
```
thecristen, line 142: 😭 at this holdover lol

lib/dotcom_web/components/new_timetable.ex
```diff
@@ -0,0 +1,143 @@
...
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </div>
+    """
+  end
+
+  attr :id, :string, required: true
+  attr :scroll_direction, :integer, required: true
```
thecristen: **suggestion**:  add `values: [-1, 1]` here!

thecristen (approved):
Thanks for all the detailed notes, including about the followup work! Great progress here!

assets/js/scroll-to.js
joshlarson: I did this because Typescript was yelling at me - maybe there's a better way??

lib/dotcom_web/components/new_timetable.ex
joshlarson: Ooooh I actually have a follow-up PR queued up for right after this one 😁😁😁

lib/dotcom_web/components/new_timetable.ex
joshlarson: [Done!](https://github.com/mbta/dotcom/pull/3366/changes/3ba2fc33828adf11ff26728d728a337ed087ddb4)

I wish LiveView would warn if you provide a value that isn't one of the `values` specified, but at least it's useful for documentation.

assets/js/scroll-to.js
joshlarson: Talked offline, and agreed that while there may be a better way, neither of us know it!

PR #3367, opened 2026-07-28
chore: Make translation-checking its own job

joshlarson:
## Scope

No ticket, but I noticed that [this build](https://github.com/mbta/dotcom/actions/runs/30377690155/job/90337453924) had failed, and it took me a bit of digging to figure out why.

## Screenshots

Screenshot 2026-07-28 at 12 47 05 PM

From [the build](https://github.com/mbta/dotcom/actions/runs/30379824341) for this very PR!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -194,7 +194,7 @@ jobs:
       - run: npm run ci:format:ts
 
   check_translations:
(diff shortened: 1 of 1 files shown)


thecristen (approved):
Looks like that should do the trick!

PR #3368, opened 2026-07-29
[Fix]: Update timetable date picker to support selecting the current date

lvachon1:
## Scope

**Asana Ticket:** [Update timetable date picker to support selecting the current date](https://app.asana.com/1/15492006741476/project/555089885850811/task/1215567127775712?focus=true)

## Implementation

It appears that our problem is that the URL is identical to the one w…e're on, and so the browser treats it as an anchor link (which it is) and scrolls to that anchor (which it's already on).  I added a random parameter to the URL to disable this behavior.  It works, and is not the first time I've used this trick, but I'm not sure it is the cleanest solution.

## Screenshots

https://github.com/user-attachments/assets/e53d5e6a-c55c-4095-9ddf-d44fe5e557ad

## How to test

http://localhost:4001/schedules/CR-NewBedford/timetable

Confirm that re-selecting the currently selected date closes the calendar and refreshes the page.

### DIFF
--- a/lib/build_calendar.ex
+++ b/lib/build_calendar.ex
@@ -201,7 +201,7 @@ defmodule BuildCalendar do
   end
 
   defp build_url(url_fn, date, _) do
-    url_fn.(date: Date.to_iso8601(date), date_select: nil, shift: nil)
+    url_fn.(date: Date.to_iso8601(date), date_select: nil, shift: nil, random: :rand.uniform())
   end
 
   @spec month_relation(Date.t(), Date.t(), Date.t()) :: __MODULE__.Day.month_relation()
--- a/lib/build_calendar.ex
+++ b/lib/build_calendar.ex
@@ -169,7 +169,7 @@ defmodule BuildCalendar do
     for date <- Date.range(first_day(shifted), last_day(shifted)) do
       %BuildCalendar.Day{
         date: date,
-        url: build_url(url_fn, date, today),
+        url: build_url(url_fn, date, today, selected),
         month_relation: month_relation(date, last_day_of_previous_month, last_day_of_this_month),
         selected?: date == selected,
         holiday?: MapSet.member?(holiday_set, date),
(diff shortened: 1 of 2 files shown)


thecristen (commented):
😵‍💫 Creative _and_ makes me uncomfortable.

I noticed in `BuildCalendar.Day.td/1` that it renders the link with the `#date-filter` anchor, and that removing this seems to make the link behave the way we want (at least when I tried it in Vivaldi just now). I wonder if there's something we could do with that, like removing that anchor if we happen to be looking at the current date, or something like that.

joshlarson (commented):
Another possible idea is to add the `date-picker-toggle` class to the date cell when `day.selected?` is true ([this line](https://github.com/mbta/dotcom/blob/fb6307facfa4f00b866e0b63ee6df47c647fd847/lib/build_calendar.ex#L90)). That makes the currently-selected day act as a toggle, just like the `Schedule for:` button that expands, and the `Cancel` button that collapses, the date picker.

([This](http://github.com/mbta/dotcom/blob/fb6307facfa4f00b866e0b63ee6df47c647fd847/assets/js/date-picker.js#L51) is the JS that already does that for the `Schedule for:` and `Cancel` buttons)

lvachon1:
I took Josh's idea and re-used the CSS class that already triggers the behavior we want via JS.  Thanks!

lib/build_calendar.ex
```diff
@@ -89,7 +89,8 @@ defmodule BuildCalendar do
           {day.holiday?, "schedule-holiday"},
           {day.selected? && day.month_relation == :current, "schedule-selected"},
           {day.month_relation == :next, "schedule-next-month"},
-          {day.today?, "schedule-today"}
+          {day.today?, "schedule-today"},
+          {day.selected?, "date-picker-toggle"}
```
joshlarson, line 93: **Suggestion (non-blocking):** I wonder if this could be combined with the other `day.selected?` condition above (the fact that this works flawlessly makes me think that the `month_relation` check on the other `day.selected?` check isn't necessary).

Non-blocking because this does work as-is, and because I don't have 100% confidence that you actually can combine them, and I don't want to force you to block this on that 🙂

lib/build_calendar.ex
```diff
@@ -169,7 +170,7 @@ defmodule BuildCalendar do
     for date <- Date.range(first_day(shifted), last_day(shifted)) do
       %BuildCalendar.Day{
         date: date,
-        url: build_url(url_fn, date, today),
+        url: build_url(url_fn, date, today, selected),
```
joshlarson: **Suggestion:** I don't think you need this fourth `selected` parameter anymore. All of the `build_url/4` definitions below ignore it as `_`.

joshlarson (approved):
Nice! Left a few little style comments, but none are approval-blocking. Appreciate you going on what wound up being a surprisingly convoluted journey for this bug 😵‍💫

lib/build_calendar.ex
lvachon1: I've removed the now redundant fourth argument

PR #3371, opened 2026-07-29
feat: Nicer `/_flags` page

joshlarson:
## Scope

No ticket, but I noticed that the helpful banner that tells you what environment you're in was missing from the `/_flags` page. I figured... while we're here, why not make it look pretty?

Blocked by:
- https://github.com/mbta/laboratory/pull/3

## Screenshots

Screenshot 2026-07-29 at 3 28 28 PM

## How to test

Go to [the flags page](http://localhost:4001/_flags), and toggle the flags as you see fit!


lib/dotcom_web/controllers/flags_controller.ex
```diff
@@ -0,0 +1,26 @@
+defmodule DotcomWeb.FlagsController do
+  use DotcomWeb, :controller
+
+  use Phoenix.VerifiedRoutes,
+    endpoint: DotcomWeb.Endpoint,
+    router: DotcomWeb.Router,
+    statics: DotcomWeb.static_paths()
```
thecristen: FYI if you wanted to not need to include this, you can add this to the form

```heex
<input type="hidden" name="redirect" value={~p"/_flags"} />
```

then in the controller can read the redirect param and do

```elixir
|> redirect(to: params["redirect"])
```

joshlarson: Whoa - [cool](https://github.com/mbta/dotcom/pull/3371/commits/628fcb277c3631fd8c0a503238335f3a4c6913c4)!

thecristen (approved):

PR #3377, opened 2026-07-30
refactor(SystemStatus.CommuterRail): use service check

thecristen:
## Scope

This PR covers a few things I noticed while analyzing the homepage. 

### Checking if a commuter rail route has service today

It's not directly related to the homepage, but as the system became overwhelmed and various things fell apart, one call that'd sometimes fail is this which feeds into the Commuter Rail Status widget:
```
  # Returns a boolean indicating whether or not the route has a schedule
  # for today. This is used to determine if the route is running service today.
  @spec service_today?(String.t()) :: boolean()
  defp service_today?(id) do
    [id]
    |> @schedules_condensed_repo.by_route_ids()
    |> Enum.any?(fn %{time: time} -> Dotcom.Utils.ServiceDateTime.service_today?(time) end)
  end
```

This PR replaces this with the existing `Dotcom.ServicePatterns.has_service?/1` function, which defaults to checking the current service date and uses a much simpler check (against V3 API `/service` result rather than checking times in each result from `/schedules`).

### Getting alerts for each status widget

Both `subway_alert_groups()` and `commuter_rail_alert_groups()` currently fetch alerts by fetching **all alerts** then filtering the result down to the relevant routes. I realized we could start with fewer alerts by using the existing `Alerts.Repo.by_route_types/2` function. So that's what the final commit does.

## Screenshots

No change!

## How to test

Compare with the current commuter rail status -- should work exactly the same way.

### DIFF
--- a/lib/dotcom/alerts.ex
+++ b/lib/dotcom/alerts.ex
@@ -127,7 +127,7 @@ defmodule Dotcom.Alerts do
   @spec subway_alert_groups() :: [{Route.t(), [Alert.t()]}]
   def subway_alert_groups() do
     alerts =
-      @alerts_repo_module.all(@date_time_module.now())
+      @alerts_repo_module.by_route_types([0, 1], @date_time_module.now())
       |> Enum.reject(&service_impacting_alert?/1)
 
     non_banner_alerts = excluding_banner(@alerts_repo_module.banner(), alerts)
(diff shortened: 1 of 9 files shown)


joshlarson (approved):
Iterating over all of the schedules for a route on a given day... 😅 

Nice catch!

PR #3380, opened 2026-07-31
refactor(Dotcom.PageController): avoid fetching and iterating all alerts ever

thecristen:
## Scope

The homepage's "Alerts" tab shows routes and stops which have important alerts.

My local load testing on the homepage has found that fetching and processing alerts is by far the most computationally intensive part of the homepage. This PR tries to make that more performant.


thecristen:
This turned out to not help as much as I'd hoped, maybe will revisit later!

PR #3381, opened 2026-07-31
QF | UndefinedFunctionError: function :the_ride.id/0

lvachon1:
## Scope

**Asana Ticket:** [QF | UndefinedFunctionError: function :the_ride.id/0](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217075266456934?focus=true)

## Implementation

Added a special case for `:the_ride` since it is just an atom, and not a full route with an id.

## Screenshots

N/A

## How to test

http://localhost:4001/schedules/

Confirm the page doesn't crash


thecristen (approved):
Thank you 😅

PR #3385, opened 2026-07-31
fix(Plugs.Cookies): avoid using other paths

thecristen:
## Scope

Just happened to discover a few other "routes" stored in the cookie. :( 

## Implementation

It was thanks to poor path parsing. We have, for example, `/schedules/map_api` which one could hit. This would store "map_api" as a "route", which would eventually be looked up in `Routes.Repo` when rendering "Recently Visited" routes. Not great, and a waste of resources.

The other aspect of this I wanted to raise is that, if you happen to have a nonsensical route in your cookie, this is going to be requested from the V3 API over and over and [over](https://api-dev.mbtace.com/routes/nonsense). This is because we don't cache error responses. Normally that's a good thing, but in light of recognizing our website gets hammered with excess requests.... this creates a suboptimal situation. So I added an extra change in `Routes.Repo` which caches errors, but only for 30 seconds. Curious what folks think of that.

## Screenshots

No change!

## How to test

The final test I added in the test file fails on `main` but passes here!

### DIFF
--- a/lib/routes/repo.ex
+++ b/lib/routes/repo.ex
@@ -83,13 +83,15 @@ defmodule Routes.Repo do
 
   defp update_direction_names(route), do: route
 
-  @decorate cacheable(cache: @cache, on_error: :nothing, opts: [ttl: @ttl])
-
+  @decorate cacheable(cache: @cache, on_error: :nothing, match: &match_route/1)
   defp cached_get(id, opts) do
     with %{data: [route]} <- MBTA.Api.Routes.get(id, opts) do
       {:ok, parse_route(route)}
     end
   end
+
+  defp match_route({:error, _}), do: {true, nil, ttl: :timer.seconds(30)}
+  defp match_route(route), do: {true, route, ttl: @ttl}
 
   @impl Routes.Repo.Behaviour
   def get_shapes(route_id, opts) do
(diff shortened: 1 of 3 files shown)


lib/routes/repo.ex
```diff
@@ -83,14 +83,16 @@ defmodule Routes.Repo do
...
-  @decorate cacheable(cache: @cache, on_error: :nothing, opts: [ttl: @ttl])
-
+  @decorate cacheable(cache: @cache, on_error: :nothing, match: &match_route/1)
   defp cached_get(id, opts) do
     with %{data: [route]} <- MBTA.Api.Routes.get(id, opts) do
       {:ok, parse_route(route)}
     end
   end
 
+  defp match_route({:error, _}), do: {true, nil, ttl: :timer.seconds(30)}
```
joshlarson, line 93: **Issue (non-blocking):** We should probably distinguish between `:something`, `:nothing`, and `:error`. If I'm understanding this right, our previous (and current) behavior is to interpret `:error` as `:nothing` (because a 404 error does indicate that the queried route doesn't exist), which means that outside of that, we can't tell the difference between `:nothing` and `:error`.

...but we really should be caching `:something` and `:nothing`, and not caching `:error`.

---

I marked this issue as non-blocking because while I do stand by ☝️ analysis, I'm not convinced that it's worth blocking this fix on doing this Right™, and also your approach of caching `:nothing` and `:error` for 30 seconds does seem like a reasonable compromise for now. I may drop something onto the tech debt backlog about it though.

joshlarson (approved):
Pretty neat!

I left a non-blocking comment that I think we should address at some point, whether before merge or not.

jlucytan (approved):

lib/routes/repo.ex
thecristen: I'll admit I'm not exactly sure what "nothing" would mean for some queries, like this one. But I agree we could give it a more careful look!

PR #3386, opened 2026-07-31
fix(StopController): alerts by_stop_id/1

thecristen:
I just realized we have an existing way to fetch alerts by stop ID and a place which wasn't using it.
Running the tests reveals that the stop pages (for individual stops) were fetching all stops for a stop ID twice - the second time while getting banner alerts. So I ended up refactoring it a little to avoid that duplication.

Should be a no-op.


joshlarson (approved):
Neat!

PR #3390, opened 2026-08-03
hotfix(ContentSecurityPolicy): unblock Google Translate script

thecristen:
## Scope

While in a meeting today we noticed the language picker in the navigation bar has stopped working. 😩 

<figure>
image
<figcaption>Something is missing between the buttons and the search bar...</figcaption>
</figure>

## Implementation

It was getting blocked by our CSP. The fix is to add the (perhaps new) URLs to our Content Security Policy's `script-src` directive.

## Screenshots

image

### DIFF
--- a/lib/dotcom_web/plugs/content_security_policy.ex
+++ b/lib/dotcom_web/plugs/content_security_policy.ex
@@ -77,7 +77,9 @@ defmodule DotcomWeb.Plugs.ContentSecurityPolicy do
         snap.licdn.com
         translate.google.com/translate_a/element.js
         translate-pa.googleapis.com
+        translate.googleapis.com
         www.instagram.com
+        www.google.com/sorry/index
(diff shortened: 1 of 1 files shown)


lib/dotcom_web/plugs/content_security_policy.ex
```diff
@@ -109,8 +111,6 @@ defmodule DotcomWeb.Plugs.ContentSecurityPolicy do
     websocket_url = "#{Keyword.get(endpoint_config, :url, [])[:host]}"
 
     [
-      {:connect_src, "ws://#{websocket_url}"},
-      {:connect_src, "wss://#{websocket_url}"},
```
thecristen, line 113: I removed these after seeing errors in the console about these not resolving to valid URLs.

lvachon1 (approved):

PR #3391, opened 2026-08-03
QF [Fix]:  FunctionClauseError: no function clause matching in Date.compare/2

lvachon1:
## Scope

**Asana Ticket:** [QF | FunctionClauseError: no function clause matching in Date.compare/2](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217022094152012?focus=true)

[Sentry Log](https://mbtace.sentry.io/issues/trace/aadfa8d3ce85490ab38ae6eb66d9facf/?groupId=7641114837&node=error-aadfa8d3ce85490ab38ae6eb66d9facf&pageEnd=2026-07-30T10%3A47%3A03.457&pageStart=2026-07-29T10%3A47%3A03.457&referrer=slack&source=issue_details&timestamp=1785365223.452)

## Implementation

Saw a nil passed into this function via error logs, added cases to handle if end_date or start_date (or both) are nil.

Is this the right logic to use for these edge cases?  The error condition returns true, so it seems we prefer false positives over false negatives.  I used that logic to give the partial data the best chance to pass a test that makes sense.

## Screenshots

N/A

## How to test

Unit tests!

### DIFF
--- a/lib/dotcom_web/plugs/date_in_rating.ex
+++ b/lib/dotcom_web/plugs/date_in_rating.ex
@@ -17,8 +17,17 @@ defmodule DotcomWeb.Plugs.DateInRating do
   def call(%Conn{assigns: %{date: date}, query_params: %{"date" => _}} = conn, dates_fn: dates_fn) do
     in_rating? =
       case dates_fn.() do
+        %{start_date: nil, end_date: end_date} ->
+          Date.before?(date, end_date)
+
+        %{start_date: start_date, end_date: nil} ->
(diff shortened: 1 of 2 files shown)


thecristen (changes requested):
I think that's correct - preferring false positives over false negatives. And thanks for the extra tests!!

Once case I'm not sure this handles is if the current date happens to **equal** the start or end date, can you please add a case for that?

thecristen (approved):
🎉

PR #3392, opened 2026-08-03
refactor(ScheduleController): remove unused predictions

thecristen:
One could argue they were used for the Green line's vehicle tooltips, except there's another code path it goes through which also populates the vehicle tooltip. For the vehicle maps, the result seems pretty much the same.

This should be a no-op.

## Screenshots

(dev on left, this branch on right) (it's identical)

image

image


thecristen:
The controller I deleted was introduced in this PR which was for an older version of the schedule page:

- https://github.com/mbta/dotcom-legacy/pull/541

lvachon1 (approved):
🧹 Thanks for cleaning up!

PR #3402, opened 2026-08-04
feat: Github Action to kick off Artillery test

thecristen:
## Scope

**Asana Ticket:** [(Implementation!) Follow up on "📈 Investigate Distributed Load Testing in AWS"](https://app.asana.com/1/15492006741476/project/1215975015154295/task/1216630210027189)

## Implementation

In search of a simple approach, we came to identify running our load tests from a Github Actions runner as a viable option. The referenced [action](https://www.artillery.io/docs/cicd/github-actions#setting-up-a-github-actions-workflow) will set up Artillery's [Docker](https://www.artillery.io/docs/docker#running-tests-scripts-inside-a-docker-container) container and run its `run` command, which will execute the script.

It's minimal by design, and doesn't use Artillery's distributed load testing features. This action is configured to be run manually against either dev-green or dev-blue environments.

### DIFF
--- a/integration/load_tests/homepage-heavy-load.yml
+++ b/integration/load_tests/homepage-heavy-load.yml
@@ -9,9 +9,9 @@ config:
       arrivalRate: 100
   environments:
     dev-green:
-      target: "http://dev-green.mbtace.com"
+      target: "https://dev-green.mbtace.com"
     dev-blue:
-      target: "http://dev-blue.mbtace.com"
+      target: "https://dev-blue.mbtace.com"
 scenarios:
   - name: "Stress test single page"
     flow:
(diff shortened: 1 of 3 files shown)


integration/load_tests/homepage-heavy-load.yml
```diff
@@ -8,9 +8,9 @@ config:
       duration: 300
       arrivalRate: 100
   environments:
-    green:
+    dev-green:
```
joshlarson, line 11: In order to work properly, the `http`'s below will need to be `https`'es

thecristen: Thanks, updated

joshlarson (approved):
Let's try it!

PR #3408, opened 2026-08-07
performance: Move API calls out of render and into subway_status

jlucytan:
## Scope

**Asana Ticket:** [📈 🛠️ [Subway Status] Move API calls / stop-loading out of component-rendering](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216359457265957?focus=true&auto_reloaded=true)

## Implementation
Moves `affected_stops` and `endpoint_stops` out of `status_row_heading` and into `subway_status` so there are fewer API calls at render time on the home page, and the data is cached. 

The alerts page and schedule finder still make calls to populate the subheading for the Planned Work section, however.

Although commuter_rail_upcoming_changes also utilizes the `status_row_heading` component, I don't believe it currently shows any subheadings, so no data is added there.

## Screenshots
Should have no visual changes. 

Old flamegraph:
flame-main-content-tag-blue

New flamegraph:
flame-branch-content-tag-blue

## How to test
[Homepage](http://localhost:4001), [alerts page](http://localhost:4001/alerts/subway), schedule finder alert banners, commuter rail upcoming changes appearances should not change between this branch and prod.


lib/dotcom_web/components/system_status/status_row_heading.ex
```diff
@@ -131,10 +134,8 @@ defmodule DotcomWeb.Components.SystemStatus.StatusRowHeading do
     }
   end
 
-  defp decorations(%{status: status, alerts: alerts, route_ids: route_ids})
+  defp decorations(%{status: status, subheading_data: {:endpoint_stops, endpoints}})
        when status in [:service_change, :shuttle, :single_tracking, :suspension] do
```
joshlarson, line 138: **Suggestion (non-blocking):** I think we can (and should) remove the `when status in` check here - now that you've moved the "which statuses get endpoints?" check into the backend, the frontend can just be like "endpoints? okay I'll draw them like this"

lib/dotcom_web/components/system_status/subway_status.ex
```diff
@@ -101,6 +104,7 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatus do
       prefix={@row.status_entry.prefix}
       plural={@row.status_entry.plural}
       future={@row.status_entry.future}
+      subheading_data={@subheading_data}
```
joshlarson, line 107: **Suggestion (non-blocking):** If you add `subheading_data: nil` to [the "normal status" clause of rows_for_status_entry/3](https://github.com/mbta/dotcom/blob/4648a6413766edd724afb12a06983eba7db3b93d/lib/dotcom_web/components/system_status/subway_status.ex#L238-L254), then every status_entry map at this point in the code will have a `subheading_data` field defined, and then you can replace this with 👇, similar to all the other fields here, and get rid of the assigns-fiddling you're doing above.
```heex
<.status_row_heading
  # ...
  subheading_data={@row.status_entry.subheading_data}
/>
```

joshlarson (changes requested):
I like what I see so far, and having these calls not run per-component-load is going to be a huge win!

However, I think something's gone a bit wrong with combining subheadings. Compare dev-green on the left to this branch on the right.

Image

It looks like status on the [alerts page](http://localhost:4001/alerts/subway) is combining subheadings from different rows even when the rows themselves aren't being combined, but on that page, each row should only show the affected stops or endpoints from the alert associated with that row.

lib/dotcom_web/components/system_status/subway_status.ex
```diff
@@ -68,7 +69,23 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatus do
...
 
   defp alerts_page_row(%{row: %{alerts: [alert]}} = assigns) do
-    assigns = assigns |> assign(:alert, alert)
+    # Override the pseudo-combined subheading_data with the specific data for this alert
+    subheading_data =
+      subheading_data(
+        status: alert.effect,
+        alerts: [alert],
+        route_ids: [assigns.row.route_info.route_id]
+      )
```
jlucytan, line 78: This is only used on the alerts page, where we currently can take the performance hit (well, it's not a regression at any rate). Not ideal, would be nice to not have to call the API again (see also similar calls in `planned_disruptions.ex` and `schedule_finder_live.ex`) but it's the quickest fix without having to do wider structural changes.

test/dotcom_web/components/system_status/subway_status_test.exs
```diff
@@ -774,10 +774,15 @@ defmodule DotcomWeb.Components.SystemStatus.SubwayStatusTest do
 
       stops = Factories.Stops.Stop.build_list(2, :stop)
 
-      expect(Dotcom.Alerts.AffectedStops.Mock, :affected_stops, fn
-        [_], _ ->
+      [affected_branch1 | affected_branch2] = affected_branches
```
joshlarson: **Issue (style/naming):** `affected_branch2` will be a list, because the `|` splits off the head and tail of a linked list. 

```elixir
iex> list = [0, 2, 1, 1, 6]
iex> [foo | bar] = list

iex> foo
0

iex> bar
[2, 1, 1, 6]
```

If you meant to split a 2-item list into two equivalent branches, then I'm guessing you meant to do `[affected_branch1, affected_branch2] = affected_branches`

```elixir
iex> list = [0, 2]
iex> [foo, bar] = list

iex> foo
0

iex> bar
2
```

---

That would also allow you to treat `^affected_branch1` and `^affected_branch2` the same in the `expect` statement below.

lib/dotcom_web/live/schedule_finder_live.ex
```diff
@@ -275,6 +276,13 @@ defmodule DotcomWeb.ScheduleFinderLive do
       |> Enum.filter(fn %{informed_entity: %{direction_id: direction_id}} ->
         Enum.any?([nil, direction], &(&1 in direction_id))
       end)
+      |> Enum.map(fn alert ->
+        Map.put(
+          alert,
+          :subheading_data,
+          subheading_data(status: alert.effect, alerts: [alert], route_ids: [route.id])
+        )
+      end)
```
joshlarson, line 285: **Question / Issue:** Is this doing anything? Or is it possibly leftover from an earlier attempt?

lib/dotcom/system_status/subway.ex
```diff
@@ -135,6 +149,39 @@ defmodule Dotcom.SystemStatus.Subway do
...
   end
 
+  # Additional information required to compute the subheading for a status entry.
+  # - {:affected_stops, [affected_stop_t()]} if the status is a station closure
+  # - {:endpoint_stops, [{endpoint_t(), endpoint_t()}]} if the status is a service change
+  # - {:delay} if the status is a delay
+  # - nil if there is no data
+  @spec subheading_data(status: atom(), alerts: [Alert.t()], route_ids: [Routes.Route.id_t()]) ::
+          subheading_data_t()
+  def subheading_data(status: :station_closure, alerts: alerts, route_ids: route_ids) do
```
joshlarson: **Issue / Idea (non-blocking):** Since this is called by a number of non-system-status modules (planned disruptions, and eventually the departures page, for starters), it may make sense to move this to the `Dotcom.Alerts` module, or maybe even its own namespace.

---

FWIW - this is just about the lowest priority idea imaginable, so if you don't agree, or it would be non-trivial effort, then feel free to ignore this!

lib/dotcom/system_status/subway.ex
```diff
@@ -7,15 +7,29 @@ defmodule Dotcom.SystemStatus.Subway do
...
+  @type endpoint_t() :: Dotcom.Alerts.EndpointStops.Behaviour.endpoint_t()
+
   @type status_time() :: :current | {:future, DateTime.t()}
 
   @type status_t() :: :normal | Dotcom.Alerts.service_effect_t()
 
+  @type subheading_data_t() ::
+          {:affected_stops, [affected_stop_t()]}
+          | {:endpoint_stops, [{endpoint_t(), endpoint_t()}]}
+          | {:delay}
```
joshlarson: **Question / Issue (non-blocking):** What does `:delay` represent in this context? Does it represent both the `Due to Single Tracking` and `Up to 20min` states? 

If so, I would recommend having separate subheading states for those, since they do render differently.

---

That said, I'm totally on board with keeping this as-is for now, and refactoring it later, hence the "non-blocking" portion of this.

joshlarson (approved):
This looks great, and when running locally and looking at flamegraphs, I was seeing a consistent improvement of about 70ms (rendering going from ~268ms to ~198ms), which is HUGE. Plus, _hand-wavy-guess_ it wouldn't shock me if this was also one of the more memory-consuming parts of individual page loads, which makes this even more impactful.

I left some stylistic feedback, but none of it is blocking, and can all be addressed as follow-ups, if at all.

lib/dotcom_web/live/schedule_finder_live.ex
jlucytan: The schedule finder alerts banners actually also leverage `status_row_heading`, so a `subheading_data` field has to be provided. As it turns out, the way this information is piped through is through this alerts assignment elsewhere in the file (`<.alerts_status_group alerts={@alerts}>`. In the `alerts_status_group` component, I've assigned `subheading_data={alert.subheading_data}`). Open to less hacky ways to do this though. 

Here's an example of a schedule finder alert banner with a subheading; the subheading wouldn't appear without subheading data added.
image

lib/dotcom/system_status/subway.ex
jlucytan: I agree it would make sense to be tied to alerts and have moved it there in [commit 614d66f653776084794e8668d130385d948b86ba](https://github.com/mbta/dotcom/pull/3408/changes/614d66f653776084794e8668d130385d948b86ba)

lib/dotcom/system_status/subway.ex
jlucytan: I don't think subheading does "up to x minutes"? It provides additional context ("Due to single tracking") if the delay is due to single tracking, otherwise it's empty. `decorations` in `status_row_heading` determines if it should say "Due to single tracking" or not. If there's an alert type I've overlooked that has a different behavior for the subheading, let me know.

image

Now that I think about it, maybe it's not worth having `{:delay}` in it at all? Since `decorations` actually only needs `status: :delay` to decide if it needs to check all the alerts passed to it for single tracking. Or it could make sense to move the check into `subheading_data` and change the subheading data to `{:due_to_single_tracking}` or something and have `decorations` use that result instead of iterating through the alerts at render time.

test/dotcom_web/components/system_status/subway_status_test.exs
jlucytan: Knew there had to be a better way to do this, thanks!

lib/dotcom/system_status/subway.ex
joshlarson: Right! I had misremembered exactly how this worked, so I thought it was

**Delays**
_Up to 20 min_

but it's actually

**Delays up to 20 min**
(possibly with _Due to Single Tracking_)

---

I think the "purest" answer here would be to move the "Due to Single Tracking" check into `subheading_data` like you said, so that way _everything_ that renders in the smaller font next to (or under) the main status comes from `subheading_data`, and the frontend doesn't have to think about it at all.

...but we're already in an "impure" intermediate state with this, and calculating this on render doesn't pose the same kind of performance problem/s that endpoints and affected stops do, so I think this is fine for now.

lib/dotcom_web/live/schedule_finder_live.ex
joshlarson: Ahhh I understand!

I agree that this is hacky (especially because it creates a world where now we have an `%Alert{}` struct that's being passed around, but has this extra `subheading_data` key, which isn't found on the alert itself.

...and like the other comment above, I think this is a worthwhile tradeoff to make for now.

PR #3409, opened 2026-08-08
feat: Enable/Disable the Earlier/Later buttons on scroll/resize

joshlarson:
## Scope

**Asana Ticket:** [⛴️ Disable Earlier/Later Boats buttons at the right scroll positions](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216965272523309?focus=true)

## Implementation

Yet another Phoenix hook.

> [!NOTE]
> This is my first real foray into Phoenix hooks, and I'm not sure that I put them in the right place. We've now got two different hooks attached to different parts of the timetable... maybe that's the way to do it? Maybe not?

## Screenshots

https://github.com/user-attachments/assets/649932dd-31f8-41be-8945-2d90faa31faa

## How to test

[Enable new timetables](http://localhost:4001/_flags), and then visit [any](http://localhost:4001/schedules/Boat-F10/timetable) [timetable](http://localhost:4001/schedules/CR-Greenbush/timetable) [page](http://localhost:4001/schedules/Boat-F6/timetable) and play around with the buttons. Try shrinking the window when scrolled all the way to the right - that nudges some of the trips off the edge of the screen, which re-enables the `Later Trips` button.

Given special attention to [the Lynn ferry page](http://localhost:4001/schedules/Boat-F6/timetable) (which has few enough trips that it doesn't need the buttons at typical desktop browser widths). Try resizing the window so that it's too narrow for its five trips and note that the buttons reappear.

Try other things I didn't think of - have fun!

### DIFF
--- a/assets/ts/phoenix-hooks/index.ts
+++ b/assets/ts/phoenix-hooks/index.ts
@@ -3,6 +3,7 @@ import MBTAGoCTABanner from "./mbta-go-cta-banner";
 import PageVisibility from "./page-visibility";
 import ScrollIntoView from "./scroll-into-view";
 import TimetableScroll from "./timetable-scroll";
+import TimetableScrollBar from "./timetable-scroll-bar";
 import TripPlannerForm from "./trip-planner-form";
 import TripPlannerMap from "./trip-planner-map";
 import TripPlannerDatePicker from "../../js/trip-planner/datepicker";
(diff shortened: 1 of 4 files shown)


lvachon1 (approved):
Looks great,  I tried breaking it with window and font shenanigans, but it wasn't fooled.

PR #3410, opened 2026-08-08
feat: Add accessibility and parking indicators to new timetable

joshlarson:
## Scope

**Asana Ticket:** [⛴️ Add parking and accessibility to new timetable stop row headers](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216965272523307?focus=true)

## Implementation

- Literally just copy-pasted the parking and accessibility indicator stuff from [the existing timetable component](https://github.com/mbta/dotcom/blob/47fe6d8058ce90b43e4799237d7307fa9701665b/lib/dotcom_web/components/timetable.ex#L252-L279).

## Screenshots

Direct current-main-versus-new comparison to show how much better the world is with these icons there.
Screenshot 2026-08-07 at 9 28 56 PM

Comparison of the [old timetable component](https://github.com/mbta/dotcom/blob/47fe6d8058ce90b43e4799237d7307fa9701665b/lib/dotcom_web/components/timetable.ex) (with the feature flag disabled) with the results of this PR - to show that we're getting closer and closer to feature parity.
Screenshot 2026-08-07 at 9 29 32 PM

## How to test

Check out the [Winthrop](http://localhost:4001/schedules/Boat-F6/timetable) and [Quincy](http://localhost:4001/schedules/Boat-F7/timetable) timetables (those are good ferry choices because they're the only ones with docks that have parking). [Framingham/Worcester](http://localhost:4001/schedules/CR-Worcester/timetable) is a good commuter rail timetable to look at because it has all four permutations of accessible/inaccessible and parking/no-parking (Newtonville has neither icon, for instance).

### DIFF
--- a/lib/dotcom_web/components/new_timetable.ex
+++ b/lib/dotcom_web/components/new_timetable.ex
@@ -67,8 +67,38 @@ defmodule DotcomWeb.Components.NewTimetable do
               class="sticky left-0"
               style="background-color: inherit;"
             >
-              <div class={"#{header_column_classes()} px-2 py-1 font-medium #{font_size_classes()} border-r-xs border-gray-lighter"}>
-                <.link navigate={~p"/stops/#{row.stop.id}"}>{row.stop.name}</.link>
+              <div class={"#{header_column_classes()} px-2 py-1 font-medium #{font_size_classes()} border-r-xs border-gray-lighter flex gap-2"}>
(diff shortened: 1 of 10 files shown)


jlucytan (approved):
question(non-blocking): is there value in making these into separate components? I assume that the other timetable component will eventually go away and this part won't be duplicated anymore, but just checking.

PR #3412, opened 2026-08-10
[Fix]: FunctionClauseError: no function clause matching in Calendar.ISO.parse_date/1

lvachon1:
## Scope

**Asana Ticket:** [QF | FunctionClauseError: no function clause matching in Calendar.ISO.parse_date/1](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216997800645302?focus=true)

## Implementation

Logs were showing that this function was being passed a map (or other invalid arguments), which was not handled, and the app crashed.  I added some extra guards to handle bad arguments.

## Screenshots

N/A?

## How to test

Unit tests

### DIFF
--- a/lib/util/util.ex
+++ b/lib/util/util.ex
@@ -486,6 +486,10 @@ defmodule Util do
     end
   end
 
+  def parse_valid_daate(_) do
+    {:error, :invalid_date}
+  end
+
   @doc """
 
   Parses a string into a valid datetime, or returns an error.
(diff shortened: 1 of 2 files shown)


thecristen (approved):
Technically fixes the 500 error!

I did see your note about what's getting passed into this function and decided to look into it further -- the linked Sentry issue indicates this happening at `/events?date[month]=7&date[year]=2023`, which indeed makes that date param a map. So I think my earlier PR for handling the date param must have entirely broke that functionality... dang.

With this PR that page loads, but it redirects to the current date. I'll add a followup ticket for restoring the prior functionality for navigating to other month's events.

PR #3414, opened 2026-08-12
feat(Util): get or save `:persistent_term`

thecristen:
## Scope

[Slack thread](https://mbta.slack.com/archives/C06TP48CJ9W/p1786547373071049)

From a linked page from the linked thread:

> Erlang's [persistent_term](https://www.erlang.org/doc/apps/erts/persistent_term.html) is a great option if you want to keep a large state in memory and you modify it rarely or ideally, never. It solves a problem that you'll run into if you use :ets or a [GenServer](https://hexdocs.pm/elixir/GenServer.html) to store a large state in memory: whenever a process fetches your large state, the BEAM will copy the entire state and return it to the calling process. Additionally, the state is copied again and again whenever the process uses it. If your state is large, this will seriously impact your processing speed and memory usage.

> That's where persistent_term comes in. Instead of returning a copy of your state, it returns a reference to it. That way, only one version of the state stays in memory, but processes can still look up data and pass it around without problems.

I think we have an opportunity to use this tool, so long as we're careful to avoid updates or deletes to the things we store in `:persistent_term`.

Everything gets reset on redeploy, so we don't have to worry about unexpected stale states.

## Implementation

- Added a function which either retrieves an item from `:persistent_term`, or computes and returns the result (while storing it for future use).
- Uses it in two places which I noticed while profiling the homepage:
  - Generating our Content Security Policy, which is identical for every response
  - Formatting module names for use in cache keys, which will be consistent for each module name

## Screenshots

No change!

## How to test

Not much to test here.

### DIFF
--- a/lib/dotcom/cache/key_generator.ex
+++ b/lib/dotcom/cache/key_generator.ex
@@ -11,7 +11,10 @@ defmodule Dotcom.Cache.KeyGenerator do
   def generate(mod, fun, args) do
     unique_id = args |> inspect() |> Base.encode64()
 
-    "#{clean_mod(mod)}|#{fun}|#{unique_id}"
+    module_name =
+      Util.get_or_save_persistent_term({:key_generator, mod}, fn -> clean_mod(mod) end)
+
+    "#{module_name}|#{fun}|#{unique_id}"
   end
 
   defp clean_mod(mod) do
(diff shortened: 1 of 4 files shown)


lvachon1 (approved):
This is a very cool feature, I look forward to seeing how we can apply it further.

PR #3416, opened 2026-08-13
[Fix]: FunctionClauseError: no function clause matching in DotcomWeb.ScheduleController.FinderApi.load_from_repos/2

lvachon1:
## Scope

**Asana Ticket:** [QF | FunctionClauseError: no function clause matching in DotcomWeb.ScheduleController.FinderApi.load_from_repos/2](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216994482750470?focus=true)

## Implementation

Digging around the Sentry logs it appears that when/if invalid or incomplete URL params are supplied to `schedules/finder_api/journeys` the app crashes.  

Added checks to ensure code only runs when supplied with proper parameters.  Added an error message to handle invalid params instead of crashing.

## Screenshots

N/A

### Invalid arguments:
http://localhost:4001/schedules/finder_api/journeys?date=2026-08-13&stop=place-sstat&is_current=true&id=1

### Valid arguments (empty result)
http://localhost:4001/schedules/finder_api/journeys?date=2026-08-13&stop=place-sstat&is_current=true&id=1&direction=1

### DIFF
--- a/lib/dotcom_web/controllers/schedule/finder_api.ex
+++ b/lib/dotcom_web/controllers/schedule/finder_api.ex
@@ -78,7 +78,7 @@ defmodule DotcomWeb.ScheduleController.FinderApi do
   end
 
   def journeys(conn, _) do
-    DotcomWeb.ControllerHelpers.return_invalid_arguments_error(conn)
+    return_invalid_arguments_error(conn)
   end
 
   # Use alternative JourneyList constructor to only return trips with predictions
@@ -184,7 +184,7 @@ defmodule DotcomWeb.ScheduleController.FinderApi do
   end
 
   def trip(conn, _) do
-    ControllerHelpers.return_invalid_arguments_error(conn)
+    return_invalid_arguments_error(conn)
   end
 
   # Use internal API to generate list of relevant schedules and predictions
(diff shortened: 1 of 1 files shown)


lib/dotcom_web/controllers/schedule/finder_api.ex
```diff
@@ -68,6 +77,10 @@ defmodule DotcomWeb.ScheduleController.FinderApi do
     json(conn, journeys)
   end
 
+  def journeys(conn, _) do
+    DotcomWeb.ControllerHelpers.return_invalid_arguments_error(conn)
```
jlucytan: nit(non-blocking): Can all calls to `return_invalid_arguments_error` be made to match? There's one at L103 and another at L174

lib/dotcom_web/controllers/schedule/finder_api.ex
```diff
@@ -87,7 +100,7 @@ defmodule DotcomWeb.ScheduleController.FinderApi do
   end
 
   def departures(conn, _) do
-    ControllerHelpers.return_invalid_arguments_error(conn)
+    return_invalid_arguments_error(conn)
```
jlucytan, line 103: question: I don't see `return_invalid_arguments_error` as a direct import -- should this still call `ControllerHelpers`?

lvachon1: I've cleaned up the other calls

jlucytan (approved):

PR #3417, opened 2026-08-13
[Fix]: Add websocket connect_src lines to CSP for older IOS devices

lvachon1:
## Scope

**Asana Ticket:** [[🐞 Line diagram + departures page are down for iOS 15 users](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217464268376067?focus=true)](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217464268376067?focus=true)

## Implementation

Restored ws:// and wss:// lines to the CSP.  The original solution used the :host of the :url config field, which is only defined for prod. This solution uses the :host of the :static_url config field which is the same value by default and defined for all environments.

## How to test

I was able to see the new CSP entries in the response headers.  A proper Apple device test would be nice though.


thecristen (approved):
Thank you 🙏🏼

thecristen:
> The original solution used the :host of the :url config field, which is only defined for prod. 
@lvachon1 I think you can use the original solution? Prod in this Mix context means `MIX_ENV`=prod, which applies to all of our deployed environment.

PR #3418, opened 2026-08-13
[Fix]: Line diagram + departures page are down for iOS 15 users

lvachon1:
## Scope

**Asana Ticket:**[🐞 Line diagram + departures page are down for iOS 15 users](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217464268376067?focus=true)

## Implementation

Restored original solution


thecristen (approved):
🚀 Let's give it a whirl

PR #3419, opened 2026-08-13
fix(events): Add an empty state for months with no events

joshlarson:
## Scope

**Asana Ticket:** [🗓️ Events page should still show months without events](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217464268376075?focus=true)

## Implementation

- Instead of relying exclusively on `Enum.group_by`, which simply doesn't include entries for months that have no events, we find the earliest and latest relevant months and return a list (possibly empty) of event teasers for each month in between.

## Screenshots

Screenshot 2026-08-13 at 2 53 29 PM

## How to test

Look at the [events page](http://localhost:4001/events) and notice how it doesn't look like we've just forgotten about August.

### DIFF
--- a/lib/dotcom_web/views/event_view.ex
+++ b/lib/dotcom_web/views/event_view.ex
@@ -34,6 +34,8 @@ defmodule DotcomWeb.EventView do
   @doc "Returns a list of event teasers, grouped/sorted by month"
   @spec grouped_by_month([Teaser.t()], number) :: [{number, [Teaser.t()]}]
   def grouped_by_month(events, year) do
+    {earliest_year, latest_year} = events |> Stream.map(& &1.date.year) |> Enum.min_max()
+
     by_month =
       events
       |> Enum.filter(&(&1.date.year == year))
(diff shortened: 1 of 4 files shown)


lib/dotcom_web/views/event_view.ex
```diff
@@ -34,10 +34,21 @@ defmodule DotcomWeb.EventView do
   @doc "Returns a list of event teasers, grouped/sorted by month"
   @spec grouped_by_month([Teaser.t()], number) :: [{number, [Teaser.t()]}]
   def grouped_by_month(events, year) do
-    events
-    |> Enum.filter(&(&1.date.year == year))
-    |> Enum.group_by(& &1.date.month)
-    |> Enum.sort_by(&elem(&1, 0))
+    {earliest_year, latest_year} = events |> Stream.map(& &1.date.year) |> Enum.min_max()
```
lvachon1, line 39: Elixir has such fun Enum functions.  `min_max()` is a new one for me. That could save some boilerplate code I've typed countless times in my life.

lvachon1 (approved):

PR #3421, opened 2026-08-14
feat(TripPlannerLive): show leg fares via Fares V2

thecristen:
## Scope

**Asana Ticket:** [🗺️ 💸 Spike: Try turning on GTFS fares V2 for OTP](https://app.asana.com/1/15492006741476/project/1204352509486278/task/1216769442870973?focus=true)

## Implementation

> [!IMPORTANT]
> While updating the backend to ingest the new data applies to everyone, showing the new fare information for each leg is behind a new feature flag! 

## Screenshots

If you turn on the feature flag, **and** your `OPEN_TRIP_PLANNER_URL` environment variable points to an application which was built with the Fares V2 feature activated (like in this [otp-deploy branch](https://github.com/mbta/otp-deploy/pull/50)), you'll see more elaborate fare summaries. Each column corresponds to a transit leg; there are many fare products for each leg because there are so many different ways to pay your fare at the T!

image

## How to test

Nothing to see here yet.

> [!NOTE]
> The existing calculated total fare is kept as-is for now.


config/config.exs
```diff
@@ -133,7 +133,8 @@ config :laboratory,
     {:new_timetables, "New timetables",
      "Uses the new timetable component and logic for all timetables"},
     {:use_smartling_translations, "Smartling translations",
-     "Uses Smartling's translation workflows"}
+     "Uses Smartling's translation workflows"},
+    {:fares_v2, "Fares v2", "Exposes information from GTFS Fares V2 in trip plans"}
```
joshlarson, line 137: **Suggestion (non-blocking):** Should we put the feature flags in alphabetical order?

lib/dotcom_web/components/trip_planner/results.ex
```diff
@@ -19,6 +19,7 @@ defmodule DotcomWeb.Components.TripPlanner.Results do
   attr :results, :any
   attr :class, :string, default: ""
   attr :accessible_grouping?, :boolean
+  attr :fares_v2, :any, default: nil
```
joshlarson: **Suggestion (non-blocking):** Could this be a boolean?

```suggestion
  attr :fares_v2, :boolean, default: false
```

mix.exs
```diff
@@ -128,7 +128,7 @@ defmodule DotCom.Mixfile do
       {:ex_machina, "2.8.0", only: [:dev, :test]},
       {:ex_unit_summary, "0.2.0", only: [:dev, :test]},
       {:excoveralls, "0.18.5", only: :test},
-      {:faker, "0.19.0-alpha.1"},
+      {:faker, "0.19.0-alpha.1", override: true},
```
joshlarson: **Suggestion (non-blocking):** If, when we upgrade OTP Client, we also upgrade faker to `0.19.0`, then the override won't be necessary.

joshlarson (commented):
Pretty neat! I'm pretty much ready to approve once the [OTP client PR is merged](https://github.com/mbta/open_trip_planner_client/pull/29), and we update the OTP client version in `mix.exs` accordingly.

mix.exs
thecristen: I forgot to upgrade its Faker version!

config/config.exs
thecristen: heh, not sure how meaningful that is. One advantage of appending new features is it leaves the corresponding buttons more consistently placed (e.g. the 1st button can stay directing to new timetables for however long that feature exists).

joshlarson: Fair enough!

FWIW, I didn't do that for the new timetables flag (I put it before the Smartling one, which preceded it), and buttons won't be reliably consistently placed anyway (since once the new timetables one is deleted, everything below it will shift up), and I figured if the list gets long enough, it'll be easier to find feature flags by alphabetical-order-of-name, rather than trying to remember when a particular flag got added (although maybe that's just a sign that we should avoid having too many flags?)

But also this isn't a huge deal, and we can reorder them later if we so choose, so I'm gonna resolve this comment for now.

mix.exs
joshlarson: [I upgraded it in an earlier PR](https://github.com/mbta/open_trip_planner_client/pull/28/changes#diff-dfa6f4ed74c90e5d4fda283d547d366586e690387289bcfd473e3fa5f9ace308R73), so the latest OTP Client release has faker `0.19.0`!

So we can set that version here now and lose the override.

joshlarson (approved):

PR #3426, opened 2026-08-18
feat: save rendered static markup for faster retrieval

thecristen:
## Scope

**Asana Ticket:** a prerequisite to [📈 Optimize nav-bar rendering](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217274403544453?focus=true)

I have two goals

1. Easier, simpler rendering - I want rendering to be less work on our server. So to support that, I'm writing markup directly instead of invoking functions, where it makes sense.
2. Moving some of the rendered markup into `:persistent_term`, as described in 
- #3414 

## Implementation

I intend to follow up with similar PRs for the header navigation, footer, and parts of the homepage. But this PR only impacts the root template.

I tackled two pieces, which will never, ever* change from between individual requests:

### The hidden icons
Background: We currently render a bunch of icons and use JS to copy them and render them with search results in the navigation, here: 
image

### Top-tier nav
image

Now, the language menu actually does change slightly, depending which language is selected. 
image

So unlike the prior case, we must save a version rendered for each supported language. And the locale has to be part of the persistent term key, so we can retrieve the correct markup for the selected language.

Saving per-language can work by doing this in an isolated task:

```elixir
Task.async(fn ->
  Dotcom.Locales.set_locale("fr") # only sets the language for this fleeting process

  # trigger rendering and saving the translated markup
end)
```

Finally, this is invoked on application startup, so every actual request should retrieve the already-saved markup.

## Screenshots

No changes!

## How to test

The [before](https://github.com/user-attachments/assets/412b6a09-fdfe-455b-8bc7-8d80f8cdb8cb) and [after](https://github.com/user-attachments/assets/7e0440fd-d0f5-4c0c-ab2b-67258d85f41b) flamegraphs are kind of dramatic because the root.html rendering seems to get represented different in the "after" version - almost like it's more broken up into smaller pieces. Overall removing a bunch of HTML escaping work seems to save us ~9ms. The hidden icons template goes from 8.6ms to.. I can't even find it anymore.

### DIFF
--- a/lib/dotcom/application.ex
+++ b/lib/dotcom/application.ex
@@ -71,6 +71,8 @@ defmodule Dotcom.Application do
 
     :ok = Logster.attach_phoenix_logger()
 
+    _ = Dotcom.Cache.PersistentTermWarmer.warm_homepage()
+
     opts = [strategy: :one_for_one, name: Dotcom.Supervisor]
 
     Supervisor.start_link(children, opts)
--- a/lib/dotcom/cache/persistent_term_warmer.ex
+++ b/lib/dotcom/cache/persistent_term_warmer.ex
@@ -1,0 +1,15 @@
+defmodule Dotcom.Cache.PersistentTermWarmer do
+  @moduledoc """
+  Functions to trigger at application startup to warm caches that are stored in
+  :persistent_term.
+
+  By starting these processes at application startup, we can ensure that the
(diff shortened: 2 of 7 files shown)


lvachon1 (approved):
This is very cool!  I've always been a fan of using plain HTML when possible, to see that it's noticeably faster is very satisfying.

PR #3428, opened 2026-08-18
[Symphony]: Hide long-term alerts in departures pages

lvachon1:
## Scope

**Asana Ticket:** [🐬 Hide long-term alerts in departures pages](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216640941991690?focus=true)

## Implementation

Added function to filter out stale alerts that don't target this stop specifically

### Dev at Hynes
Screenshot 2026-08-18 at 1 16 58 PM

### Local at Hynes
Screenshot 2026-08-18 at 1 17 15 PM

### Local at Symphony
Screenshot 2026-08-18 at 1 17 22 PM

## How to test

http://localhost:4001/departures?route_id=Green&direction_id=1&stop_id=place-hymnl - Confirm that the Symphony closure alert is gone for other stops on the route

http://localhost:4001/departures?route_id=Green&direction_id=1&stop_id=place-symcl - Confirm that for Symphony itself the alert still shows up


joshlarson (commented):
The feature looks good (I think), but I think it would be worth it to add a few tests to [the alerts section of the schedule finder live tests](https://github.com/mbta/dotcom/blob/78e76fde8a90c38dd0f87e0fb4a04beac73324f4/test/dotcom_web/live/schedule_finder_live_test.exs#L207-L236) validating that this behaves the way we want.

test/dotcom_web/live/schedule_finder_live_test.exs
```diff
@@ -388,11 +388,13 @@ defmodule DotcomWeb.ScheduleFinderLiveTest do
     end
   end
 
-  defp visit_with_valid_params(conn, route_types \\ [0, 1, 2, 3, 4]) do
-    route_id = FactoryHelpers.build(:id)
-    direction_id = FactoryHelpers.build(:direction_id)
-    stop_id = FactoryHelpers.build(:id)
-
+  defp visit_with_set_params(
```
joshlarson, line 432: **Question:** Is this necessary? It looks like this is only being called by `visit_with_valid_params/2`.

And/or was this created with the intent to add test coverage for the boolean logic here (in which case I would say that we should in fact add that test coverage 🙂).

joshlarson (changes requested):
I know I'm a broken record about this kind of thing, but I do think we should have a test that validates that the stale alert doesn't show up typically (and a test that validates that it _does_ show up on the Symphony page, and that other station closures do show up, etc).

test/dotcom_web/live/schedule_finder_live_test.exs
lvachon1: This was created to support the test I wanted to write that needed to specify specific params not just randomly valid ones.  I have now added the tests that actually use this function.

test/dotcom_web/live/schedule_finder_live_test.exs
```diff
@@ -235,6 +235,39 @@ defmodule DotcomWeb.ScheduleFinderLiveTest do
...
+        informed_entity: %{stop: alert_stop_id},
+        effect: Faker.Util.pick([:station_closure, :stop_closure])
+      )
+    end)
+
+    {:ok, view, _html} = visit_with_set_params(conn, route_id, direction_id, stop_id)
+    assert !has_element?(view, "section.c-alert-group")
+  end
+
+  test "show stale alerts (on proper stop)", %{conn: conn} do
```
joshlarson, line 260: **Issue (blocking):** There appears to be a race condition in this test (which annoyingly can't be reproduced with just a `seed` value). About half the time I run this, it fails with 👇 

```elixir
1) test show stale alerts (on proper stop) (DotcomWeb.ScheduleFinderLiveTest)
   test/dotcom_web/live/schedule_finder_live_test.exs:255
   Expected truthy, got false
   code: assert has_element?(view, "section.c-alert-group")
   arguments:

       # 1
       #Phoenix.LiveViewTest.View<
         id: "phx-GM8Okuco13lHZADk",
         module: DotcomWeb.ScheduleFinderLive,
         pid: #PID<0.648.0>,
         endpoint: DotcomWeb.Endpoint,
         ...
       >

       # 2
       "section.c-alert-group"

   stacktrace:
     test/dotcom_web/live/schedule_finder_live_test.exs:268: (test)
```

At a glance, I'm guessing that this is because the page renders in a separate process, and when the assertion is made, the page hasn't loaded the alerts yet.

joshlarson (changes requested):

test/dotcom_web/live/schedule_finder_live_test.exs
```diff
@@ -235,6 +235,39 @@ defmodule DotcomWeb.ScheduleFinderLiveTest do
...
+    route_id = FactoryHelpers.build(:id)
+    direction_id = FactoryHelpers.build(:direction_id)
+    stop_id = "place-symcl"
+
+    expect(Dotcom.ScheduleFinder.Mock, :current_alerts, 2, fn _, _ ->
+      Factories.Alerts.Alert.build_list(1, :alert_for_informed_entity,
+        informed_entity: %{stop: stop_id},
+        effect: Faker.Util.pick([:station_closure, :stop_closure])
+      )
```
thecristen, line 270: If being active now is important, call `Test.Support.Factories.Alerts.Alert.active_now(alert)` on the factory-generated alert to help ensure it's active!

test/dotcom_web/live/schedule_finder_live_test.exs
lvachon1: It turns out the issue was the lack of a specified direction_id.  Schedule Finder discards all alerts that aren't for it's direction, so my test alerts were never making it through.

test/dotcom_web/live/schedule_finder_live_test.exs
lvachon1: Thanks!  This at least helped me confirm active period wasn't the problem, heh.  Turns out I wasn't specifying a direction_id for the informed entities so the factory was randomly picking one and Schedule Finder filters out alerts for the wrong direction.  Oops!

lvachon1: It is important for the opposite test case though, just to make sure that `stale?` is the only thing hiding the alert.

joshlarson (approved):

PR #3429, opened 2026-08-18
feat(load-testing): Add phases to the load testing script

joshlarson:
## Summary

This adds phases to the load testing script, so that we can measure our code's performance against 100, 200, 500, 1k, 2k, 5k, and 10k requests per second. 

## Scope

**Asana Ticket:** [📈 Set up the GHA Artillery script to increase load in stages](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217421262176610?focus=true)

## Screenshots

Screenshot 2026-08-18 at 4 49 07 PM

## How to test

Once this is merged, we can re-run the [load test Github Action](https://github.com/mbta/dotcom/actions/workflows/load-test.yml) and see how much load dev-green and dev-blue can handle. 

For extra fun, we could deploy an old release to one of those environments and see it perform worse (hopefully).

### DIFF
--- a/integration/load_tests/homepage-heavy-load.yml
+++ b/integration/load_tests/homepage-heavy-load.yml
@@ -22,16 +22,6 @@ config:
     - name: "1k req / sec"
       duration: 120
       arrivalRate: 1000
-    - name: "Rest"
-      pause: 60
-    - name: "2k req / sec"
-      duration: 120
-      arrivalRate: 2000
-    - name: "Rest"
-      pause: 60
-    - name: "10k req / sec"
-      duration: 120
-      arrivalRate: 10000
   environments:
     dev-green:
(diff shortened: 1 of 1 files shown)


integration/load_tests/homepage-heavy-load.yml
```diff
@@ -15,8 +40,5 @@ config:
 scenarios:
   - name: "Stress test single page"
     flow:
-      - loop:
-          - get:
-              url: "/"
-          - think: 1
-        count: 100
```
joshlarson, line 22: I'm open to adding this back in, but IMO it was just confusing me, because in this particular case, it was just sort of acting as a slightly delayed multiplier, making it hard to tell just by looking at the script what load it was generating.

lvachon1 (approved):

PR #3431, opened 2026-08-18
fix: Alignment bug in new timetable

joshlarson:
## Scope

No ticket, but I just noticed that at certain browser widths (at a guess... even number of pixels? 1000px for instance), if you drag the new timetable scroll bar almost but not quite to the "Latest trips" end of the screen and then click `Later <Vehicles>`, it goes one pixel too far, and nudges the header with the `Earlier`/`Later` buttons out of alignment.

## Implementation

- Prevent the buttons from nudging the scroll bar any further than the actual max scroll. 

## Screenshots

Screenshot 2026-08-18 at 6 52 28 PM

## How to test

Navigate to [the Harbor Loop Ferry Timetable page](http://localhost:4001/schedules/Boat-F10/timetable?schedule_direction[direction_id]=1#direction-filter) with a browser width of 1000px - on `main`, you should see the top bar get nudged a little - on this branch, you should see it stay put.

Try other routes and other browser widths!

### DIFF
--- a/assets/ts/phoenix-hooks/timetable-scroll.ts
+++ b/assets/ts/phoenix-hooks/timetable-scroll.ts
@@ -2,6 +2,14 @@ import { ViewHook } from "phoenix_live_view";
 
 let animating = false;
 
+const min = (a: number, b: number) => {
+  if (a < b) {
+    return a;
+  } else {
+    return b;
+  }
+};
+
 const TimetableScroll: Partial<ViewHook> = {
   mounted() {
     if (this.el) {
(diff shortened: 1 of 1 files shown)


joshlarson:
This doesn't fully solve the issue - it's still possible to get the misalignment by resizing the browser window and positioning the scroll bar in a specific place. 

But this at least solves the problem for the vast majority of cases that someone is likely to actually do 😅

assets/ts/phoenix-hooks/timetable-scroll.ts
```diff
@@ -2,6 +2,14 @@ import { ViewHook } from "phoenix_live_view";
 
 let animating = false;
 
+const min = (a: number, b: number): number => {
```
lvachon1: `Math.min` is built in and should work here?

joshlarson: I was thinking "there's _got_ to be a built-in min function", and then just completely forgot to follow up with that thought - thanks for catching! [Fixed now](https://github.com/mbta/dotcom/pull/3431/commits/47e395d6c701beaa2d3f1c80128f6c9f9684b55f)

lvachon1 (approved):

PR #3432, opened 2026-08-18
feat: New bespoke Harbor Loop Ferry timetable layout

joshlarson:
## Scope

**Asana Ticket:** [⛴️ Render F10 schedule using the new bespoke F10 layout](https://app.asana.com/1/15492006741476/project/555089885850811/task/1215284488469002?focus=true)

> [!IMPORTANT]
> This feature is _not_ behind a feature flag, so if this PR is merged, then on the next deploy, this will go live to riders. I think that's okay, but since most of the other timetable work I've done has been behind the `new_timetables?` feature flag, I wanted to make sure that this PR's non-flagged nature was clear.

## Implementation

> [!NOTE]
> I used 🤖 Copilot to generate the first commit of this PR. I've inspected the code and tests, and my tweaks and changes are the remaining commits.

- Defines a couple of special assigns so that the Harbor Loop special case can load both timetables onto the page.
- It uses the `NewTimetable.timetable/1` component rather than `Timetable.timetable/1` because the buttons on `Timetable.timetable/1` can only work if there's a single timetable on the page (this is the entire reason that `NewTimetable` exists).

## Screenshots

Screenshot 2026-08-18 at 7 24 47 PM

Screenshot 2026-08-18 at 7 33 36 PM

## How to test

Visit [the Harbor Loop Ferry timetable page](http://localhost:4001/schedules/Boat-F10/timetable?schedule_direction[direction_id]=1#direction-filter).

Try it out for different dates and browser widths.

A thing that I tried, just to validate a corner case that I admit would surprise me, was to change `evening_schedules` or `morning_schedules` to `[]` in the `Boat-F10` `assign_trip_schedules/1` clause. This is the result (Copilot's initial implementation was much less nice):

Screenshot 2026-08-27 at 8 43 43 AM

(I did this before the refactor in [this commit](https://github.com/mbta/dotcom/pull/3432/commits/02ea127a3f15626d78bd87a18309faf5d9c9499f), which removes the explicit `morning/evening_schedules` variables.)

Try out whatever else you think of!

### DIFF
--- a/lib/dotcom_web/templates/schedule/_timetable.html.heex
+++ b/lib/dotcom_web/templates/schedule/_timetable.html.heex
@@ -69,7 +69,12 @@
             timetable={@morning_timetable}
           />
         <% else %>
-          <p class="text-gray-500">{~t"No trips scheduled for this direction."}</p>
+          {render("_empty.html",
+            date: @date,
+            direction: Routes.Route.direction_name(@route, 0),
+            conn: @conn,
+            error: assigns[:schedule_error]
+          )}
         <% end %>
       </div>
       <div>
@@ -85,7 +90,12 @@
             timetable={@evening_timetable}
           />
         <% else %>
-          <p class="text-gray-500">{~t"No trips scheduled for this direction."}</p>
+          {render("_empty.html",
+            date: @date,
+            direction: Routes.Route.direction_name(@route, 1),
+            conn: @conn,
+            error: assigns[:schedule_error]
+          )}
         <% end %>
       </div>
     <% @new_timetables? -> %>
--- a/lib/dotcom_web/templates/schedule/_timetable.html.heex
+++ b/lib/dotcom_web/templates/schedule/_timetable.html.heex
@@ -58,7 +58,7 @@
     <% assigns[:dual_direction_timetable?] -> %>
       <div class="mb-5">
         <div class="container">
-          <h2 class="mb-1">{~t"Morning"}</h2>
+          <h2 class="my-1">{~t"Morning"}</h2>
           <p class="mb-3">{~t"Counterclockwise Loop"}</p>
         </div>
         <%= if @morning_trip_count > 0 do %>
@@ -81,7 +81,7 @@
       </div>
       <div>
         <div class="container">
-          <h2 class="mb-1">{~t"Evening"}</h2>
+          <h2 class="my-1">{~t"Evening"}</h2>
           <p class="mb-3">{~t"Clockwise Loop"}</p>
         </div>
         <%= if @evening_trip_count > 0 do %>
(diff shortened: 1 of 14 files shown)


thecristen (commented):
Image 

This key feels wildly out of place in this layout and I think we should do something about that. Though we will need that last line about the bikes.. (or _do_ we? 🤔 )

lib/dotcom_web/templates/schedule/_trip_view_filters.html.heex
```diff
@@ -15,13 +15,15 @@
...
     <% end %>
   </div>
-  {render("_direction_select.html",
-    conn: @conn,
-    direction_id: @direction_id,
-    route: @route,
-    show_date_select?: @show_date_select?,
-    column_width: 8
-  )}
+  <%= unless @route.id == "Boat-F10" do %>
```
thecristen: **suggestion**: It might make more sense to tie this to the `::dual_direction_timetable?` assign rather than hardcoding it to this route ID.

lib/dotcom_web/templates/schedule/_timetable.html.heex
```diff
@@ -51,10 +51,57 @@
...
         date: @date,
-        direction: Routes.Route.direction_name(@route, @direction_id),
+        direction: @direction_id && Routes.Route.direction_name(@route, @direction_id),
         conn: @conn,
         error: assigns[:schedule_error]
       )}
+    <% assigns[:dual_direction_timetable?] -> %>
+      <div class="mb-5">
+        <div class="container">
+          <h3 class="text-xl font-bold mb-1">{~t"Morning"}</h3>
```
thecristen: **suggestion(blocking)**: Given the structure of the rest of the page, this and the Evening heading should be `<h2>` (which I also think looks fine without the extra text sizing and bolding CSS)

thecristen (changes requested):
Just some tiny comments, it looks great!

lib/dotcom_web/templates/schedule/_timetable.html.heex
joshlarson: [Done](https://github.com/mbta/dotcom/pull/3432/commits/93bc3ba37379076c534ce68565ac013a3177deb8) - thanks!

lib/dotcom_web/templates/schedule/_trip_view_filters.html.heex
joshlarson: Yes, [yes it does](https://github.com/mbta/dotcom/pull/3432/commits/e6b125e430c90d1c9c6bca7e3a181c676ab582bb).

joshlarson:
> This key feels wildly out of place in this layout and I think we should do something about that. Though we will need that last line about the bikes.. (or do we? 🤔 )

Yeah, I agree... although I actually think it feels out of place on _all_ of the ferry timetables, since those icons are only relevant for Commuter Rail. Personally, I'd rather address that as a separate task, although I'm open to whatever.

thecristen (approved):
⛵

PR #3433, opened 2026-08-19
[Symphony]:  Symphony logic final (final) version

lvachon1:
## Scope

**Asana Ticket:** [🐬 Symphony logic final (final) version](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217611059875181?focus=true)

## Implementation

Changed the business logic for stale? alerts, now they're only for the symphony work.  Updated alerts_test tests to suit the new logic.  

### Symphony alert hidden on Subway Status
Screenshot 2026-08-19 at 11 44 37 AM 
Screenshot 2026-08-19 at 11 46 54 AM

### Symphony alert hidden on Green-E line
Screenshot 2026-08-19 at 11 44 12 AM

### Symphony alert shown in Green-E alerts
Screenshot 2026-08-19 at 11 43 58 AM

### Symphony alert hidden in Green line
Screenshot 2026-08-19 at 11 43 27 AM

### Symphony alert shown in Green alerts
Screenshot 2026-08-19 at 11 43 40 AM

## How to test

http://localhost:4001/ & http://localhost:4001/alerts/subway - Confirm the symphony alert isn't shown in Subway Status
http://localhost:4001/schedules/Green-E/line  & http://localhost:4001/schedules/Green/line - Confirm the alert doesn't show on the line diagram but does on the Alerts tab

Depending on the order in which these PRs are merged you may want to test the departures page as well.

http://localhost:4001/departures?route_id=Green&direction_id=1&stop_id=place-symcl - Confirm the alert shows
http://localhost:4001/departures?route_id=Green&direction_id=1&stop_id=place-hymnl - Confirm the alert doesn't show

### DIFF
--- a/lib/alerts/alert.ex
+++ b/lib/alerts/alert.ex
@@ -179,13 +179,11 @@ defmodule Alerts.Alert do
   end
 
   def stale?(%__MODULE__{} = alert) do
-    now = Dotcom.Utils.DateTime.now()
-    five_weeks_ago = DateTime.shift(now, week: -5)
+    affected_stops = alert.informed_entity.stop
+    symphony? = affected_stops |> Enum.any?(fn stop -> stop == "place-symcl" end)
+    closed? = alert.effect == :station_closure or alert.effect == :stop_closure
 
-    case Dotcom.Alerts.StartTime.next_active_time(alert, now) do
-      {:current, datetime} -> datetime |> DateTime.before?(five_weeks_ago)
-      _ -> false
-    end
+    symphony? and closed?
   end
 
   @spec build_struct(Keyword.t()) :: t()
--- a/lib/alerts/alert.ex
+++ b/lib/alerts/alert.ex
@@ -44,6 +44,8 @@ defmodule Alerts.Alert do
 
   @lifecycles [:new, :ongoing, :ongoing_upcoming, :unknown, :upcoming]
 
+  @symphony_stop_id "place-symcl"
+
   defstruct id: "",
             active_period: [],
             banner: "",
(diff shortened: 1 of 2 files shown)


lib/alerts/alert.ex
```diff
@@ -179,13 +179,11 @@ defmodule Alerts.Alert do
   end
 
   def stale?(%__MODULE__{} = alert) do
-    now = Dotcom.Utils.DateTime.now()
-    five_weeks_ago = DateTime.shift(now, week: -5)
+    affected_stops = alert.informed_entity.stop
+    symphony? = affected_stops |> Enum.any?(fn stop -> stop == "place-symcl" end)
```
joshlarson: **Suggestions (non-blocking):**

- Pull `place-symcl` out to a module attribute (basically a constant)
- Rename `affected_stops` to `affected_stop_ids`, since `alert.informed_entity.stop` actually returns stop ID's, not stops.
- `Enum.any?/2` combined with an equality check is the same as `Enum.member?/2`.

```elixir
@symphony_stop_id "place-symcl"

# ...

affected_stop_ids = alert.informed_entity.stop
symphony? = affected_stop_ids |> Enum.member?(@symphony_stop_id)
```

joshlarson (approved):
Looks good! I appreciate that this changes the definition of `stale?` without changing anything about its interaction with the outside world, which means that it's fully, 100% independent of [your other stale-related PR](https://github.com/mbta/dotcom/pull/3428) 🎉 

---

I feel compelled to flag that technically, this would match an alert that was created to close Symphony _and_ another stop, so if, say, a single alert closed both Symphony and Prudential, then that alert would also get marked as `stale`, and wouldn't show up.

I still consider this non-blocking, both because I think the scenario above is exceedingly unlikely, and also surprisingly difficult to solve for (since individual platforms are typically informed entities during station closures as well, so it's not as simple as saying "are _all_ of the informed stop entities `place-symcl`?"), but I also think it is worth noting... data can do surprising things!

lib/alerts/alert.ex
lvachon1: I've made these changes, thanks

PR #3436, opened 2026-08-19
hotfix: revert root template adjustment in GTM script

thecristen:
The string interpolation won't work in the inlined JavaScript. This isn't verifiable in the other environments because prod's the only one with the needed environment variables, but this is verifiable locally by hardcoding the `gtm_id`/`gtm_preview`/`gtm_auth` variables and seeing the values not get properly propagated. Honestly, how rude...


joshlarson (approved):
How rude indeed.

PR #3437, opened 2026-08-20
feat(header): save rendered static markup for faster retrieval

thecristen:
Same scope/implementation as this PR, just focused on the header!

- https://github.com/mbta/dotcom/pull/3426

This required some light refactoring, because I can't save the `Laboratory.enabled?(@conn, :use_smartling_translations)` calls in `:persistent_term`.

This one was a lot harder to isolate in the flamegraphs, sorry. Best I can offer is an attempt from rendering the pages (which obviously includes many other things beyond the navigation).

| Before | After |
|--------|--------|
| 142,486µs | 116,482µs | 
|  image | image |


lvachon1 (approved):
Everything looks good on my end, thanks!

PR #3446, opened 2026-08-21
feat(ServicePatterns): separate (no school) services

thecristen:
## Scope

**Asana Ticket:** [🛜 🐞 Daily Schedules sometimes doesn't show school trips](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217345168848931?focus=true)

## Implementation

The current code merges together typical **services** to show a clean dropdown menu.

image

The typical services can encompass service across multiple route patterns, some of those which aren't typical. One example is [37-B-1](https://api-dev.mbtace.com/route_patterns/37-B-1), which powers our very occasional school trips on the 37 (typicality of 3). This also has not concept of which stops are actually being served or not - services relate to routes, and route patterns further describe how stops along the route are (or are not) served at a given time.

Anyway, our code was combining `Monday - Thursday schedules (no school)` into `Monday - Thursday schedules` but the underlying datepicker picks the next valid service date, which across the combined services is today (well, yesterday), which isn't actually served at our school-trip-only stops (yet!). Altogether creating a confusing experience.

Because of its behavior in selecting dates which don't have a school trip, it can't show said trip.
The quickest way to resolve this is to stop letting the `"(no school)"` services get merged into the otherwise identical school version of the service, and show them both separately to avoid ambiguity. 

## Screenshots

Before | After

image

| Before | After |
|--------|--------|
| image  |  image |
| Can't see school Fridays! | image | 

## How to test

Try it on your favorite routes!

Note - unfortunately, there's still bugs with respect to showing the entire school trip!  The aforementioned example doesn't show stops from Forest Hills to Avenue Louis Pasteur. This'll hopefully be resolved in a future PR.

image

### DIFF
--- a/lib/dotcom/service_patterns.ex
+++ b/lib/dotcom/service_patterns.ex
@@ -284,6 +284,7 @@ defmodule Dotcom.ServicePatterns do
          service: %Service{typicality: :typical_service} = service
        }) do
     typical_groups = [
+      {:no_school, nil, fn s -> s.description =~ "(no school)" end},
       {:monday_thursday, ~t"Monday - Thursday schedules",
        fn s ->
          s.type == :weekday &&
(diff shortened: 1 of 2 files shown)


lvachon1 (approved):
🚌!

PR #3447, opened 2026-08-21
chore(load-tests): Length "rest" phases between tests, and add lighter phases

joshlarson:
## Scope

No ticket. This fell out of a few discoveries when running the first batch of load tests:
- The first phase - `100 req / sec` - was enough to knock over an older build, so I wanted to add a few lighter phases that would give us useful <100% graphs.
- Having a rest phase of 1 minute wasn't enough to allow the different phases to show up as different on the graphs. This increases the rest phase to ~2 minutes (should I bump it up to 3??)~ _3 minutes_.


lvachon1 (approved):
📈

PR #3448, opened 2026-08-24
[Perf]: Optimize `DotcomWeb.PageView._tabbed_nav.html`

lvachon1:
## Scope

**Asana Ticket:** [📈 Optimize `DotcomWeb.PageView._tabbed_nav.html`](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217663241793923?focus=true)

## Implementation

Used our new static cache to cache our static HTML for the tab list.
Replaced `~p` with a plain string in subway status `add_url` which is much faster
Used our dynamic cache to cache the alerts HTML with locale in the key and a TTL of 1 minute. 
Went ahead and added a dynamic cache to subway_status as well 

Old render time: >7000 µs
New render time <500 µs

### Main (>7000µs)
flame_on (4)

### This branch (<1000µs)
flame_on (3)

## How to test

http://localhost:4001/ - Confirm everything looks the same and works the same across all languages.

### DIFF
--- a/lib/dotcom_web/views/page_view.ex
+++ b/lib/dotcom_web/views/page_view.ex
@@ -24,6 +24,7 @@ defmodule DotcomWeb.PageView do
   @spec alerts(%{alerts: [Alerts.Alert.t()], locale: String.t()}) :: Phoenix.HTML.Safe.t()
   def alerts(%{alerts: alerts, locale: locale}) do
     _ = locale
+
     [routes, stops] =
       [
         &Dotcom.Alerts.routes_with_high_priority_alerts_by_mode/1,
--- a/lib/dotcom_web/views/page_view.ex
+++ b/lib/dotcom_web/views/page_view.ex
@@ -249,4 +249,10 @@ defmodule DotcomWeb.PageView do
        |> Enum.map(&Phoenix.View.render_to_iodata(DotcomWeb.PageView, &1, %{}))}
     end)
   end
+
+  def tab_list(locale) do
+    Util.get_or_save_persistent_term({__MODULE__, :tab_list, locale}, fn ->
+      {:safe, Phoenix.View.render_to_iodata(DotcomWeb.PageView, "_tab_list.html", %{})}
+    end)
+  end
 end
--- a/lib/dotcom_web/views/page_view.ex
+++ b/lib/dotcom_web/views/page_view.ex
@@ -21,7 +21,7 @@ defmodule DotcomWeb.PageView do
               opts: [ttl: :timer.minutes(1)],
               key: {"homepage|alerts-tab", locale}
             )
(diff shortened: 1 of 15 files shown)


lib/dotcom_web/views/page_view.ex
```diff
@@ -15,8 +15,14 @@ defmodule DotcomWeb.PageView do
   alias CMS.Partial.Banner
   alias DotcomWeb.PartialView
 
-  @spec alerts([Alerts.Alert.t()]) :: Phoenix.HTML.Safe.t()
-  def alerts(alerts) do
+  @decorate cacheable(
+              cache: @cache,
+              on_error: :raise,
+              opts: [ttl: :timer.minutes(1)],
+              key: {"homepage|alerts-tab", locale}
```
thecristen, line 22: Great find!

lib/dotcom_web/views/page_view.ex
```diff
@@ -15,8 +15,14 @@ defmodule DotcomWeb.PageView do
...
+            )
+  @spec alerts(%{alerts: [Alerts.Alert.t()], locale: String.t()}) :: Phoenix.HTML.Safe.t()
+  def alerts(%{alerts: alerts, locale: locale}) do
```
thecristen, line 25: nit: Elixir seems to report the `locale` variable as being unused when we're using it in this particular way.
maybe we can silence it with something equally silly:

```suggestion
  def alerts(%{alerts: alerts, locale: locale}) do
    _ = locale
```

lib/dotcom_web/views/page_view.ex
```diff
@@ -249,4 +255,10 @@ defmodule DotcomWeb.PageView do
        |> Enum.map(&Phoenix.View.render_to_iodata(DotcomWeb.PageView, &1, %{}))}
     end)
   end
+
+  def tab_list(locale) do
```
thecristen, line 261: Add a call to this in `Dotcom.Cache.PersistentTermWarmer` so this can be cached right at application startup!

thecristen (changes requested):
This is so cool! Longer term, I'm pretty sure we'd want to handle the alerts a bit more judiciously (by say, updating it immediately when the underlying alerts update, a la subway status), but in the meantime this seems to work great! Just a couple small comments:

lib/dotcom_web/views/page_view.ex
lvachon1: ✅ Done, thanks for the reminder.

lvachon1:
I agree that some smarter cache handling could reduce load even further.  If we already have a process subscribed to alert updates that's doing cache things, it shouldn't be too hard to put these cache writes there.  Then the homepage could either be read-only or have a higher TTL (just in case).  I'll make a follow-up ticket in the tech debt board.

thecristen (approved):

PR #3449, opened 2026-08-25
[Perf]: Optimize what's happening section rendering

lvachon1:
## Scope

**Asana Ticket:** [📈 Optimize what's happening section rendering](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217663241793919?focus=true)

## Implementation

Added a cache to `page_controller.whats_happening_items/0` since the URL encoding inside was a bit slow.  I first tried only escaping the content parameter and just concatenating the rest, but that didn't make much difference.  So I'm back to using my favorite new hammer: function caches.  

There was a cache already in the CMS repo for the data that supplies this function, but I don't like the idea of stacked caches. I can imagine a race condition that makes us wait `2*@TTL` for an update, so I removed the Repo cache for `whats_happening()`.  This way the cache will contain the final output instead of the data to calculate it.

Summary: Execution time went from ~6000us to ~180us in exchange for about 3k of memory (probably less, since I got rid of the Repo cache).

Let me know if I'm using my shiny new cache hammers too often on all these performance nails I see.

### Main Whats Happening
Main_WhatsHappening

### Branch Whats Happening (aka a cache hit)
Branch_WH

## How to test

http://localhost:4001/?locale=en - Confirm that the What's Happening section still shows up, no change should be visible except speed.  You (probably) have to be on VPN for the CMS to work, at least that's how my setup is configured.

### DIFF
--- a/lib/cms/repo.ex
+++ b/lib/cms/repo.ex
@@ -110,12 +110,6 @@ defmodule CMS.Repo do
     end
   end
 
-  @decorate cacheable(
-              cache: @cache,
-              key: "cms.repo|whats-happening",
-              on_error: :nothing,
-              opts: [ttl: @ttl]
-            )
   def do_whats_happening do
     @cms_api.view("/cms/whats-happening", [])
   end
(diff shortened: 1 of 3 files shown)


joshlarson (approved):
Looks good! Every millisecond counts!

Once we figure out how CMS stuff is localized, we'll need to make sure that this cache either stores everything that's needed for localized rendering, or include the locale in the cache key. But that's a problem that existed in the "Before" picture too.

PR #3456, opened 2026-08-31
Line Page | Set up a feature flagged "Schedules & Maps (new)" tab and page

lvachon1:
## Scope

**Asana Ticket:** [🚩💈Line Page | Set up a feature flagged "Schedules & Maps (new)" tab and page](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217398531030279?focus=true)

## Implementation

Added a new feature flag
Added a new tab to the /schedules page when that flag is enabled
Created a new live view with a few prepared assigns
Added that live view to the router

### Feature Flag
Screenshot 2026-08-31 at 1 46 44 PM

### New Tab
Screenshot 2026-08-31 at 1 46 13 PM

### New tab is gone when flag is off
Screenshot 2026-08-31 at 1 46 54 PM

### New (blank) live view
Screenshot 2026-08-31 at 1 54 46 PM

## How to test

http://localhost:4001/_flags - Confirm the new flag is there 
http://localhost:4001/schedules/Red/line - Confirm the new tab is visible and links to the new (mostly empty) live view

### DIFF
--- a/lib/dotcom_web/live/line_diagram_live.ex
+++ b/lib/dotcom_web/live/line_diagram_live.ex
@@ -23,7 +23,7 @@ defmodule DotcomWeb.LineDiagramLive do
 
     ~H"""
     <marquee style="font-size:2cm" scrollamount="16" scrolldelay="60">
-      Hello World
+      🚧 Under Construction 🚧
     </marquee>
     """
   end
--- a/lib/dotcom_web/live/line_diagram_live.ex
+++ b/lib/dotcom_web/live/line_diagram_live.ex
@@ -7,7 +7,7 @@ defmodule DotcomWeb.LineDiagramLive do
   @route_patterns_repo Application.compile_env!(:dotcom, :repo_modules)[:route_patterns]
 
   def mount(params, _session, socket) do
-    route_id = params |> Map.get("route_id", "Red")
+    route_id = params |> Map.get("route_id")
     direction_id = params |> Map.get("direction_id", "1")
     route_patterns = @route_patterns_repo.by_route_id(route_id)
 
--- a/lib/dotcom_web/live/line_diagram_live.ex
+++ b/lib/dotcom_web/live/line_diagram_live.ex
@@ -1,0 +1,28 @@
+defmodule DotcomWeb.LineDiagramLive do
+  @moduledoc """
+  The primary view for looking up stops, maps, and schedules for a particular line
+  """
+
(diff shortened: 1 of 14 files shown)


lib/dotcom_web/live/line_diagram_live.ex
```diff
@@ -0,0 +1,30 @@
...
+     |> assign(:direction_id, direction_id)
+     |> assign(:route_patterns, route_patterns)
+     |> assign(:route_id, route_id)}
+  end
+
+  def render(assigns) do
+    dbg(assigns)
+
+    ~H"""
+    <marquee style="font-size:2cm" scrollamount="16" scrolldelay="60">
```
joshlarson, line 23: When I was looking at this, and the marquee floated in from the right in its giant font, I actually busted out laughing in the middle of the office 🤣

lib/dotcom_web/live/line_diagram_live.ex
```diff
@@ -0,0 +1,28 @@
+defmodule DotcomWeb.LineDiagramLive do
+  @moduledoc """
+  The primary view for looking up stops, maps, and schedules for a particular line
+  """
+
+  use DotcomWeb, :live_view
+  @route_patterns_repo Application.compile_env!(:dotcom, :repo_modules)[:route_patterns]
+
+  def mount(params, _session, socket) do
+    route_id = params |> Map.get("route_id", "Red")
```
joshlarson: **Nit (non-blocking):** While we do need to have defaults for `direction_id` and `route_patterns` (and tickets for doing those defaults correctly), the `route_id` will come through the path, and will always be present. I don't think we need a default here

(I don't even think we need error handling for this, since I'm pretty sure there's no way to get to this code without a `route_id`, and if we do, then the default error behavior of a regular `500` plus a Sentry alert should be sufficient)

joshlarson (approved):
It's happening!

lib/dotcom_web/live/line_diagram_live.ex
lvachon1: I think it's hilarious that tag still works, I'm glad you found it as amusing as I do.

lib/dotcom_web/live/line_diagram_live.ex
lvachon1: Makes sense.  I've removed my bias for the Red line 😝

PR #3457, opened 2026-08-31
feat: 503 and 504 outage pages

joshlarson:
## Scope

> [!NOTE]
> I'm not entirely sure that this is the right way to version this, since these pages won't actually be served from a running Dotcom instance. I figure this is as good a place as any to stash them and iterate on them though 🤷 

**Asana Ticket:** [📈 🛠️ [#3378] Implement custom error pages for MBTA.com when app is down](https://app.asana.com/1/15492006741476/project/555089885850811/task/1210443882503716?focus=true)

## Implementation

> [!NOTE]
> 🤖 I used Copilot to build the first draft of this, and then iterated on it both with further AI prompts and by hand. Each commit whose prefix includes `(robot)` was generated by AI - the others were hand-written by yours truly.

## Screenshots

Screenshot 2026-08-31 at 2 39 52 PM
Screenshot 2026-08-31 at 2 39 38 PM

## How to test

Open the files. The buses blink!


lib/dotcom_web/controllers/schedule/timetable_controller.ex
```diff
@@ -37,6 +37,7 @@ defmodule DotcomWeb.ScheduleController.TimetableController do
     as: :assign_direction_id
 
   def show(conn, _) do
+    raise "nope"
```
jlucytan: This will have to be removed if this branch is to be merged into main

jlucytan (changes requested):
If this is just meant to be an archive of the current state of the pages, perhaps we can keep this as a draft branch to be used for reference for perpetuity? Although the visibility isn't the greatest, so I'm not opposed to merging it in either, but I think it's a bit confusing to have the pages there while they're not being served, and future devs will just have to know that updating them does nothing.

lib/dotcom_web/controllers/schedule/timetable_controller.ex
joshlarson: Omfg I was so diligent about not committing that... apparently [until](https://github.com/mbta/dotcom/pull/3457/changes/b42b56753629f6dbce1d32b9550c44226f3413b5) `fix(robot): 0504 attempt 2`.

joshlarson:
> [...] I think it's a bit confusing to have the pages there while they're not being served, and future devs will just have to know that updating them does nothing.

Yeah that's a good point. I wonder if we should put them in a clearer place, alongside a README that explains what they're for and how they actually get "deployed" (which .. uh... currently I don't even know)

jlucytan (approved):
lgtm, if you find a good place to move them and/or can add a comment somewhere in the repo (maybe the README would be good enough?) about their non-served status, I think that would be useful

thecristen (changes requested):
Lighthouse is flagging the SVGs for not having an an accessible text alternative. I think the big bus picture is decorative and should use `aria-hidden="true"`, and probably doesn't need `role="img"`. And maybe the same for the app store ones, but the links might need better labels (`"Download on the App Store"` doesn't say _which_ app store 😅)

joshlarson:
@thecristen and @jlucytan I just re-requested review from both of you - I think I've addressed your comments (and made the VO experience for the 503 and 504 pages a lot nicer).

I'm also hoping to add a 502 page, since we see that error too sometimes, but I'm planning on having that page be 100% identical to the others, but with a 502 blinky bus and the appropriate text changed.

thecristen (approved):
Cool!

jlucytan (approved):

PR #3459, opened 2026-08-31
chore: Add AI note to Github Pull Request Template

joshlarson:
As we use AI more, we would like to extend each other the courtesy of informing when our PR's are all or partially AI authored. So here is a small addition to the PR template to make it easy to do so!

> [!NOTE]
> :robot: If you used AI to write some or all of this PR, please
> uncomment this block and use it to explain how AI was used.
>
> Remember that you are responsible for all work that you PR, whether
> it's hand-written or AI-generated.

No ticket.


thecristen (approved):
I am a human and I approve this PR...

PR #3464, opened 2026-09-01
[LineDiagram] : Inline Metro Map into Dotcom

lvachon1:
## Scope

**Asana Ticket:** [🛠️💈 Inline Metro Map into Dotcom](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217383866704961?focus=true)

## Implementation

Moved the live component over
Moved the JS hook over
Moved the CSS over
Moved the maplibre-gl dependency over

[Removed the map from metro](https://github.com/mbta/mbta_metro/pull/39)

Rational:  Anticipating that we will need to update and enhance this map, it was decided that it would be easier to work on as a Dotcom component than as a Metro component.  

Thoughts:  Should we update our maplibre-gl version?

## Screenshots

No visual difference!

## How to test

http://localhost:4001/trip-planner - Confirm that the map still works (markers, lines, zoom, pan, rotate)
http://localhost:4001/preview/stop-map - This too why not

### DIFF
--- a/assets/css/map.css
+++ b/assets/css/map.css
@@ -1,24 +1,24 @@
 @import "../node_modules/maplibre-gl/dist/maplibre-gl.css";
 
-.mbta-metro-map-wrapper {
+.mbta-map-wrapper {
   height: 100%;
 }
 
-.mbta-metro-map svg {
+.mbta-map svg {
   fill: var(--colors-cobalt-30);
 }
 
-.mbta-metro-map-pin {
+.mbta-map-pin {
   width: 4rem;
   height: 4rem;
 }
 
-.mbta-metro-map-point {
+.mbta-map-point {
   width: 1rem;
   height: 1rem;
 }
 
-.mbta-metro-map-icon {
+.mbta-map-icon {
   width: 2rem;
   height: 2rem;
 }
--- a/assets/css/map.css
+++ b/assets/css/map.css
@@ -1,0 +1,24 @@
+@import "../node_modules/maplibre-gl/dist/maplibre-gl.css";
+
+.mbta-metro-map-wrapper {
+  height: 100%;
+}
+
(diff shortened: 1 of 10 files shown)


assets/css/map.css
```diff
@@ -0,0 +1,24 @@
+@import "../node_modules/maplibre-gl/dist/maplibre-gl.css";
+
+.mbta-metro-map-wrapper {
```
joshlarson: **Suggestion (non-blocking):** Can we rename these to not be metro?

... we might even be able to get rid of them entirely, since now that the map component is in Dotcom, we can use Tailwind.

(For historical context, while I don't remember the exact issue, using Tailwind styles in Metro made it difficult or impossible to import the library into Dotcom without releasing a new version - so Cristen replaced all of the Tailwind in Metro with plain old CSS. That constraint no longer exists for this new and improved map!)

lib/dotcom_web/components/map.ex
```diff
@@ -0,0 +1,143 @@
+defmodule DotcomWeb.LiveMap do
```
joshlarson: **Suggestion (non-blocking):** I'm going to vote that we call this `Map`, rather than `LiveMap`... but if we do call it `LiveMap`, can we rename the file to `live_map.ex`?

joshlarson (approved):
Woop!

assets/css/map.css
lvachon1: I've removed "metro" from classes, but kept the CSS.  Seems cleaner to me.

lib/dotcom_web/components/map.ex
lvachon1: `Map` works for me, and I figure I should probably follow the convention of the rest of the components and put it in the `DotcomWeb.Components` namespace.

PR #3454, opened 2026-08-28
chore: Replace `Mock` with `Mox`

joshlarson:
## Scope

**Asana Ticket:** [🛠️ Remove Mock library, replace with Mox](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217234747139794?focus=true)

## Implementation

> [!NOTE]
> This PR was written mostly by AI, using this prompt:
> 
> > We have a number of tests in this repository that use Mox for stubbing and expectations, and some that use Mock. Convert all tests that use Mock to using Mox, and remove `:mock` from our dependencies.
>
> This PR will be easiest to review commit-by-commit.
> - Commits starting with `chore(robot)` are from ☝️'s initial effort.
> - Commits starting with `cleanup(robot)` are follow-up enhancements generated with further AI prompting. The prompts are in the commit messages.
> - Commits that don't include `(robot)` were done by hand.

## Screenshots

Screenshot 2026-08-28 at 1 00 04 PM

Screenshot 2026-08-28 at 12 59 47 PM

## How to test

`mix test`

PR #3465, opened 2026-09-01
fix: Remove `with_mock` call in `ScheduleControllerTest`

joshlarson:
## Scope

No ticket, but CI is currently failing on `main`:

```elixir
error: undefined function with_mock/4 (expected DotcomWeb.ScheduleControllerTest to define such a function or for it to be imported, but none are available)
 14      │
 15  454 │       with_mock(Schedules.Repo, [:passthrough],
 16      │       ^^^^^^^^^
 17      │
 18      └─ test/dotcom_web/controllers/schedule_controller_test.exs:454:7: DotcomWeb.ScheduleControllerTest."test schedules_for_stop/2 doesn't crash when a nil route is encountered"/1
 19
 20
 21 == Compilation error in file test/dotcom_web/controllers/schedule_controller_test.exs ==
 22 ** (CompileError) test/dotcom_web/controllers/schedule_controller_test.exs: cannot compile module DotcomWeb.ScheduleControllerTest (errors have been logged)
```

Turns out [this PR](https://github.com/mbta/dotcom/pull/3453) added a new call to `with_mock` at the same time that [this PR](https://github.com/mbta/dotcom/pull/3454) wiped `:mock` (and with it `with_mock`) out of our dependency list.

## Implementation

> [!NOTE]
> :robot: I used AI to convert the `with_mock` call to `Mox` syntax. The prompt I used was:
> > Look at test/dotcom_web/controllers/schedule_controller_test.exs. There's one test in there that uses `with_mock`. Please update it to use Mox instead

AI note notwithstanding, this test does not run. The whole test is excluded from running because of [its `external` moduletag](https://github.com/mbta/dotcom/blob/218f14cf6cc8746b4055f661f9d4f8f1bdc19c51/test/dotcom_web/controllers/schedule_controller_test.exs#L13), which means that its only impact on the test suite is its current failure to compile causing the whole build to crash.

## How to test

`mix test`

### DIFF
--- a/test/dotcom_web/controllers/schedule_controller_test.exs
+++ b/test/dotcom_web/controllers/schedule_controller_test.exs
@@ -451,23 +451,22 @@ defmodule DotcomWeb.ScheduleControllerTest do
     end
 
     test "doesn't crash when a nil route is encountered", %{conn: conn} do
-      with_mock(Schedules.Repo, [:passthrough],
-        schedules_for_stop: fn
-          "TEST 1234", _ ->
-            [
-              %Schedules.Schedule{
-                route: nil,
-                stop: %Stops.Stop{id: "TEST 1234"},
-                departure_time: ~U[2219-05-18 22:25:06.098765Z]
-              }
-            ]
-        end
-      ) do
-        conn = ScheduleController.schedules_for_stop(conn, %{"stop_id" => "TEST 1234"})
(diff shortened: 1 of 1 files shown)


lvachon1 (approved):
Thanks!  I should've caught this sooner, oops.

PR #3467, opened 2026-09-02
fixup(playwright): adjust CSS selector for map pin

thecristen:
Just noticed a [test failing on the dev deploy](https://github.com/mbta/dotcom/actions/runs/33557171651/job/100024458551). Easy to miss without a deploy, this should fix it.


joshlarson (approved):
Woop - thanks for catching that

PR #3473, opened 2026-09-03
[Line Diagram]: Set up new line page LiveView to use the same header & tabs layout as the existing one

lvachon1:
## Scope

**Asana Ticket:** [💈Set up new line page LiveView to use the same header & tabs layout as the existing one](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217398531030287?focus=true)

## Implementation

Copied over the header from the existing line diagram.
The tab code depended on `conn` when we're already on to `socket` so I had to reimplement some parts of that.
Added in a basic layout skeleton to separate main and right-rail content

TODO:  Make the URL generating code smarter, we don't technically need it, but for general web usability it would be nice

## Screenshots

Screenshot 2026-09-03 at 2 02 33 PM

We need some actual content soon, otherwise I'm might just add a view counter, web ring links, and 
Best Viewed in: netscape_now

## How to test

http://localhost:4001/schedules/Red/line_new - Confirm the header and tabs look right and behave appropriately.

### DIFF
--- a/lib/dotcom_web/live/line_diagram_live.ex
+++ b/lib/dotcom_web/live/line_diagram_live.ex
@@ -111,10 +111,26 @@ defmodule DotcomWeb.LineDiagramLive do
         <div class="schedule__header-tabs">{header_tabs(assigns)}</div>
       </div>
     </div>
-
-    <marquee style="font-size:2cm" scrollamount="16" scrolldelay="60">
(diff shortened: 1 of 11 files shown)


joshlarson (approved):
Just in time for the Spring, 1997 rating! Great stuff!

PR #3475, opened 2026-09-03
dev: Have webpack clean output directory on startup

jlucytan:
Resolves `Error: EMFILE: too many open files, watch 'dotcom/priv/static/css/stop.c6f8a32fa3a07730bd2d.hot-update.js.map'`

## Scope

**Asana Ticket:** 

No ticket, but discussed in [Slack](https://mbta.slack.com/archives/C06TP48CJ9W/p1788382746672769?thread_ts=1788380327.595099&cid=C06TP48CJ9W)

## Implementation
Keeps local webpack output folder clean so it doesn't get overwhelmed by hot reload artifacts. Not sure what this incurs in startup costs, if at all (since I assume server rebuilds the initial chunks anyway, emptying out the folder beforehand surely wouldn't add too much time? But if it's building off old chunks, then maybe it's more expensive? Unsure.), but it's probably not a noticeable amount. 

## Screenshots
N/A

## How to test
Observe that `priv/static/css` has many hotreload files. With setting on, observe that starting server cleans out the directory and starts with fresh chunks.

### DIFF
--- a/assets/webpack.config.dev.js
+++ b/assets/webpack.config.dev.js
@@ -13,7 +13,8 @@ module.exports = env =>
 
     output: {
       publicPath: `http://localhost:${port}/`,
-      path: path.resolve(__dirname, "../priv/static/css")
+      path: path.resolve(__dirname, "../priv/static/css"),
+      clean: true
     },
 
     devServer: {


lvachon1 (approved):
My css directory had around 90k files in it, and in that state startup was extremely slow.  I was able to watch webpack slowly delete each file one by one, and at that rate I'd be waiting for hours.  After manually cleaning out the css folder from everything `*.hot-update.*` startup went back to its usual few seconds.  This should help keep our filesystems happy, thanks!

PR #3478, opened 2026-09-04
feat: Breadcrumbs for new line diagram page

joshlarson:
## Scope

**Asana Ticket:** [💈 Add breadcrumbs to new line diagram page](https://app.asana.com/1/15492006741476/project/1210709545912056/task/1218191070934369?focus=true)

## Implementation

I chose to add an `AssignRoute` hook, rather than just querying for the route in the `Breadcrumbs` hook, because I didn't want to create a world where we were re-querying the same route ID over and over again. So the state of the world now is that we have an `AssignRoute` hook that assigns the route to the socket, and the `Breadcrumbs` hook (if called in the context of the `:schedule_page`) assumes that `:route` is already assigned.

> [!NOTE]
> :robot: This was mostly human-written, but [the commit that adds proper not-found handling](https://github.com/mbta/dotcom/pull/3478/commits/6cba70d003582a49416842399824af134b6b1184) was written by Copilot (and then [refined with a human touch](https://github.com/mbta/dotcom/pull/3478/commits/9d6fd82bcce9519009650d21c0d4b91e23ad8129))

## Screenshots

Screenshot 2026-09-04 at 6 02 36 PM

Why yes, I did make sure to get a screenshot where the "Watch Your Step" message had a visible step in it.

## How to test

Look at the [new line diagram page](http://localhost:4001/schedules/Boat-F7/line_new) and see that the breadcrumbs are the same as they are on any of the other tabs.

### DIFF
--- a/lib/dotcom_web/hooks/assign_route.ex
+++ b/lib/dotcom_web/hooks/assign_route.ex
@@ -1,0 +1,14 @@
+defmodule DotcomWeb.Hooks.AssignRoute do
+  @moduledoc """
+  Assign the route, before both disconnected and connected mounts.
+  """
+  import Phoenix.Component, only: [assign: 3]
+
+  @routes_repo Application.compile_env!(:dotcom, :repo_modules)[:routes]
+
+  def on_mount(:default, %{"route_id" => route_id}, _session, socket) do
+    route = @routes_repo.get(route_id)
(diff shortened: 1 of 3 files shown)


lvachon1 (approved):
🍞

PR #3479, opened 2026-09-08
feat(ScheduleFinderLive): CSS transition schedules

thecristen:
Switching schedules in the departures page can be very jarring. This adds a CSS transition which hopefully makes it less jarring in most cases.

# Before

https://github.com/user-attachments/assets/1ae4d303-fe1c-47b3-8b9b-2020f4009b57

# After

https://github.com/user-attachments/assets/c38bd876-f79c-4dae-acb8-50119d1cc4ab


lvachon1 (approved):
Very cinematic, I like it!

PR #3481, opened 2026-09-08
QF | Line Diagram PDF schedules

lvachon1:
## Scope

**Asana Ticket:** [QF | 💈 PDF schedules](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216638931645034?focus=true)

## Implementation

Added Route PDF Schedules and Maps to the Right Rail of the new Line Diagram

Updated the styles to reflect the new design

## Screenshots
Screenshot 2026-09-09 at 1 06 59 PM

## How to test

http://localhost:4001/schedules/CR-NewBedford/line_new - (or any route with PDFs) Confirm that the PDF link shows up and goes to the correct PDF

### DIFF
--- a/lib/dotcom_web/live/line_diagram_live.ex
+++ b/lib/dotcom_web/live/line_diagram_live.ex
@@ -154,7 +154,7 @@ defmodule DotcomWeb.LineDiagramLive do
         pdfs when is_list(pdfs) -> pdfs
       end
 
-    assigns |> assign(:pdfs, pdfs)
+    assigns = assigns |> assign(:pdfs, pdfs)
 
     unless Enum.empty?(pdfs) do
       ~H"""
--- a/lib/dotcom_web/live/line_diagram_live.ex
+++ b/lib/dotcom_web/live/line_diagram_live.ex
@@ -156,19 +156,12 @@ defmodule DotcomWeb.LineDiagramLive do
 
     unless Enum.empty?(pdfs) do
       ~H"""
-      <.unstyled_accordion
-        summary_class="flex items-center border-t-xs border-gray-lightest bg-brand-primary-lightest cursor-pointer group/row pb-3 pl-2 justify-between"
-        chevron_class="fill-brand-primary mr-4 flex items-center pt-[0.5rem] transform scale-[1.8]"
-      >
-        <:heading>
-          <h3>{~t(PDF Schedules and Maps)}</h3>
-        </:heading>
-        <:content>
-          <div class="p-3 border-xs flex items-center border-gray-lightest pb-1">
-            {route_pdf_link(pdfs, @route, @date)}
-          </div>
-        </:content>
-      </.unstyled_accordion>
+      <div>
+        <h2 class="text-xl">{~t(Printable Schedules)}</h2>
+        <div class="p-1 flex items-center pb-1">
+          {route_pdf_link(pdfs, @route, @date)}
+        </div>
+      </div>
       """
     end
   end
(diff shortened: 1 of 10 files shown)


lib/dotcom_web/live/line_diagram_live.ex
```diff
@@ -33,6 +34,13 @@ defmodule DotcomWeb.LineDiagramLive do
       params |> Map.get("schedule_direction", %{direction_id: 0}) |> Map.get("direction_id")
 
     tab_params = %{"schedule_direction[direction_id]": direction_id}
+    date = Map.get(socket.assigns, :date, @date_time_module.now())
+
+    pdfs =
+      Dotcom.RoutePdfs.fetch_and_choose_pdfs(
+        route_id,
+        date
+      )
```
joshlarson, line 43: **Suggestion (non-blocking):** You could restructure this whole thing to first assign a date if missing, using [`assign_new`](https://phoenix-live-view.hexdocs.pm/Phoenix.Component.html#assign_new/3), and then add a separate `assign_pdfs` function. Like so:

```elixir
def mount(...) do
  {:ok,
   socket
   |> assign(:route_id, route_id)
   # ... more assigns
   |> assign(:tab_params, tab_params)
   |> assign_new(:date, &@date_time_module.now/0)
   |> assign_pdfs()}
end

# ...

defp assign_pdfs(%{assigns: %{date: date}} = socket) do
  pdfs =
    Dotcom.RoutePdfs.fetch_and_choose_pdfs(
     route_id,
     date
   )

  assign(socket, :route_pdfs, pdfs)
end
```

A few things to note about ☝️:

1. `assign_new/3` takes a 0-arity function, not a value, hence `&@date_time_module.now/0` rather than `@date_time_module.now()`
2. Given that `assign_pdfs` is called after `assign_new(:date, ...)`, you wouldn't need to worry about handling the case where the socket doesn't have a `date` assigned. 

---

I recognize that this is kind of an aesthetic thing - I find it easier to follow `mount/3` when it's defined as a simple list of `socket |> assign_this(...) |> do_that(...) |> assign_frobnicator(...)`. Hence why the comment is non-blocking 🙂

lib/dotcom_web/live/line_diagram_live.ex
```diff
@@ -131,8 +141,30 @@ defmodule DotcomWeb.LineDiagramLive do
...
     </div>
     """
   end
+
+  defp route_pdf_sidebar_content(assigns) do
+    pdfs =
+      case Map.get(assigns, :route_pdfs, []) do
+        {:error, _} -> []
+        pdfs when is_list(pdfs) -> pdfs
+      end
```
joshlarson, line 155: **Suggestion (non-blocking):** Would it make sense to move this to the `assign`ed `route_pdfs` field, rather than doing this logic in the component?

If you follow the suggestion in my earlier comment about adding an `assign_pdfs/1` function, then I think that function could be a good place to transform `{:error, _}` into `[]` as well.

joshlarson (approved):
Noice! Left two code-style suggestions - take 'em or leave 'em.

lib/dotcom_web/live/line_diagram_live.ex
lvachon1: Makes sense to me, done

lib/dotcom_web/live/line_diagram_live.ex
lvachon1: ...and done.  Thanks.

PR #3482, opened 2026-09-08
feat: Show wall-clock time for ferry trips on stop pages

joshlarson:
## Scope

**Asana Ticket:** [⛴️ Ferry stop pages follow realtime consistency guidelines](https://app.asana.com/1/15492006741476/project/385363666817452/task/1218185546770966?focus=true)

## Implementation

> [!NOTE]
> :robot: There are four commits here, each generated by Copilot.
>
> 1. Actually does the work by changing `isCR` to be true for commuter rail _and_ ferry trips.
> 2. Updates the name `isCR` to `renderAbsoluteTime`, since it doesn't really make sense to have a variable called `isCR` that's true for trips that aren't commuter rail trips.
> 3. I noticed that `renderAbsoluteTime` was being used to determine track number rendering, so I asked Copilot to restore the previous `isCR`, and use it for that and only that.
> 4. A name like `renderAbsoluteTime` makes more sense in Elixir, where you can name a function like `render_absolute_time?`. In Javascript, it's more customary to name functions like `shouldRenderAbsoluteTime`, and so that's what this commit does.

## Screenshots

Screenshot 2026-09-08 at 6 00 40 PM

Screenshot 2026-09-08 at 5 59 17 PM

## How to test

Go to [a ferry dock page](http://localhost:4001/stops/Boat-Charlestown), and admire the wall-clock display. Click around inside the departures box - there's a lot to see in there!


lvachon1 (approved):
⏰

PR #3485, opened 2026-09-09
Line Diagram | Guide Links

lvachon1:
## Scope

**Asana Ticket:** [💈 Guide links](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216638931645036?focus=true)

## Implementation

Created a look-up-table of access and beginner guides.  
Created a component that renders the correct guides given a route.

We could update this in the future with vanity URLs for these images, but considering the URLs imply they haven't changed since 2018 I don't think it's a pressing matter.

### Commuter Rail Guides
Screenshot 2026-09-09 at 2 31 56 PM

### Subway Guides
Screenshot 2026-09-09 at 2 32 44 PM

### Bus Guides
Screenshot 2026-09-09 at 2 32 56 PM

### Silver Line Guides == Bus Guides
Screenshot 2026-09-09 at 2 34 53 PM

### Ferry Guides
Screenshot 2026-09-09 at 2 33 16 PM

## How to test

[Silver Line 1](http://localhost:4001/schedules/751/line_new)
[71 Bus](http://localhost:4001/schedules/71/line_new)
[Red Line](http://localhost:4001/schedules/Red/line_new)
[Fall River/New Bedford CR](http://localhost:4001/schedules/CR-NewBedford/line_new)
[Loop Ferry](http://localhost:4001/schedules/Boat-F10/line_new)

Confirm that the correct guides show up, have the correct images, and link to the correct pages.

### DIFF
--- a/lib/dotcom_web/live/line_diagram_live.ex
+++ b/lib/dotcom_web/live/line_diagram_live.ex
@@ -8,6 +8,65 @@ defmodule DotcomWeb.LineDiagramLive do
   @alerts_repo Application.compile_env!(:dotcom, :repo_modules)[:alerts]
   @date_time_module Application.compile_env!(:dotcom, :date_time_module)
 
+  @guides [
+    %{
+      title: ~t(Subway Beginner's Guide),
+      image:
+        "/sites/default/files/styles/max_2600x2600/public/media/2018-12/Subway-Wordless-for-homepage-revised-2018-12-11.png",
+      link: "/guides/subway-guide",
+      modes: [0, 1]
+    },
+    %{
+      title: ~t(Bus Beginner's Guide),
+      image:
+        "/sites/default/files/styles/max_2600x2600/public/media/2018-12/Guides-Bus-Singleword-revised-2018-12-11.png",
+      link: "/guides/bus-guide",
+      modes: [3]
+    },
+    %{
+      title: ~t(Commuter Rail Beginner's Guide),
+      image:
+        "/sites/default/files/styles/max_2600x2600/public/media/2018-11/Guides-Commuter-HomepageWordless.png",
+      link: "/guides/commuter-rail-guide",
+      modes: [2]
+    },
+    %{
+      title: ~t(Ferry Beginner's Guide),
+      image:
+        "/sites/default/files/styles/max_2600x2600/public/media/2019-09/ferry-v2-wordless-for-homepage.png",
+      link: "/guides/ferry-guide",
+      modes: [4]
+    },
+    %{
+      title: ~t(Subway Access Guide),
+      image:
+        "sites/default/files/styles/max_2600x2600/public/Accessibility/govt-center-press-access-button.jpg",
+      link: "/accessibility/subway-guide",
+      modes: [0, 1]
+    },
+    %{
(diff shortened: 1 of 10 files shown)


lib/dotcom_web/live/line_diagram_live.ex
```diff
@@ -41,7 +100,8 @@ defmodule DotcomWeb.LineDiagramLive do
      |> assign(:route_id, route_id)
      |> assign(:route, route)
      |> assign(:tab, "new_line")
-     |> assign(:tab_params, tab_params)}
+     |> assign(:tab_params, tab_params)
+     |> assign(:guides, @guides)}
```
thecristen: **suggestion**: Since you already have the route and route type here, I'd actually recommend doing the filtering to select the appropriate guides here. Then in the template you already have the right `@guides` to show. I also believe this'll be more performant than doing the filtering in the `<.guides />` component each render.

lib/dotcom_web/live/line_diagram_live.ex
```diff
@@ -122,16 +182,30 @@ defmodule DotcomWeb.LineDiagramLive do
...
   end
+
+  def guides(assigns) do
+    ~H"""
+    <a
+      :for={guide <- @guides |> Enum.filter(fn guide -> assigns.route.type in guide.modes end)}
+      href={guide.link}
+      class="text-black text-lg font-bold"
+    >
+      <img src={guide.image} class="w-[380px] rounded-sm mb-[8px]" />
```
thecristen: **suggestion**: Add `alt=""` so assistive tech can dismiss these images as decorative, now that we've got sensible accompanying text for the link

thecristen (changes requested):

lib/dotcom_web/live/line_diagram_live.ex
lvachon1: Good call, added an empty alt tag so screen readers ignore the image.

lib/dotcom_web/live/line_diagram_live.ex
lvachon1: Good point, `on_mount` should only run once but the render calls will happen far more often.  I've moved the logic up the chain, thanks!

lib/dotcom_web/live/line_diagram_live.ex
```diff
@@ -122,16 +185,30 @@ defmodule DotcomWeb.LineDiagramLive do
...
-      <div class="col-md-5">
+      <div class="col-md-5 gap-[32px] flex flex-col">
         <marquee
           style="font-size:1cm;filter: drop-shadow(2px 4px 6px orange);"
           scrollamount="16"
           scrolldelay="60"
         >
           ⚠️ Watch Your Step ⚠️
         </marquee>
+        <.guides route={@route} guides={@guides} />
```
thecristen, line 196: nit: You don't need to pass in `@route` anymore!

thecristen (approved):

lib/dotcom_web/live/line_diagram_live.ex
lvachon1: Oops, thanks

PR #3486, opened 2026-09-09
chore(deps): Upgrade Elixir to 1.20.4, and upgrade some deps whose old versions are now incompatible.

joshlarson:
## Scope

**Asana Ticket:** [🛠️ (2026/09) Upgrade Elixir and Erlang](https://app.asana.com/1/15492006741476/project/555089885850811/task/1218291482484756?focus=true)

## Implementation

> [!NOTE]
> :robot: While I did the original Elixir upgrade by hand, I used Copilot to clean up most of the resulting compile-time warnings. As usual, commits with `robot` in the prefix (e.g. `fix(warnings/robot)`) were AI-generated. The others were human-generated.

Elixir version 1.20 introduced some nifty new type checking and warnings, some of which we ran afoul of, generating warnings that I wanted to get cleaned up before issuing this PR. Those new type checks were:
- [It now warns on unused `require`s](https://elixir.hexdocs.pm/changelog.html#1-enhancements-2)
- [It's a lot smarter about type inference now](https://elixir.hexdocs.pm/changelog.html#1-enhancements-2). Many of the commits remove or clean up assertions/refutations that _can't_ fail based on the types.
- Specifically, [it's a lot better at detecting unnecessary pattern-matched clauses](https://elixir.hexdocs.pm/changelog.html#typing-across-clauses), sometimes based on patterns specified elsewhere in the code. A bunch of the cleanup commits remove unnecessary pattern-matches that we hadn't removed before now.
- It also is a little too excitable about macros. See `live_isolated_component` below.

Commits 2-10 of this PR address ☝️ warnings within our codebase that I (well, Copilot) was actually able to clean up.

[Commit 11](https://github.com/mbta/dotcom/pull/3486/commits/7fcb5fd44eade1ce5d83b5c21069d4007713bbab) upgrades `ex_cldr_units` and `ex_cldr_numbers`, because those two libraries are run at compile-time, and the old versions had type issues that Elixir 1.20 warned about.

Commits [12](https://github.com/mbta/dotcom/pull/3486/commits/3aa7671a45d19d03625a836843ee47dcae8620f0) and [13](https://github.com/mbta/dotcom/pull/3486/commits/af28dc5e6dffbc3bda234f8b73f1a1a5f182aaa5), well... see `live_isolated_component` section below.

### Unnecessary pattern-matched clauses

I wanted to highlight this example because it's pretty cool!

The `DotcomWeb.ScheduleController.TripInfo` module defines a [private `current_trip` function](https://github.com/mbta/dotcom/blob/1841cc18f0f1a1337187cc9a4d2a847da74ab3f0/lib/dotcom_web/controllers/schedule/trip_info.ex#L122-L127), which handles empty and non-empty journey lists like so:
```elixir
  defp current_trip([%Journey{} | _] = times, now) do
    do_current_trip(times, now)
  end

  defp current_trip([], _now), do: nil
```

However, `current_trip` is only called in two places ([one](http://github.com/mbta/dotcom/blob/1841cc18f0f1a1337187cc9a4d2a847da74ab3f0/lib/dotcom_web/controllers/schedule/trip_info.ex#L61-L75), and [two](https://github.com/mbta/dotcom/blob/1841cc18f0f1a1337187cc9a4d2a847da74ab3f0/lib/dotcom_web/controllers/schedule/trip_info.ex#L77-L82)), both of which explicitly have a pattern-match preventing `journeys` from being an empty list:
```elixir
  defp trip_id(...)
       when journeys != [] do
    current_trip(journeys, ...)
  end
```

So Elixir's new type checker detected that `journeys` can never be `[]`, and that [the empty-list-handling clause](https://github.com/mbta/dotcom/blob/1841cc18f0f1a1337187cc9a4d2a847da74ab3f0/lib/dotcom_web/controllers/schedule/trip_info.ex#L127) of `current_trip/2` was unnecessary.

Pretty neat! 

### `live_isolated_component`

The flip-side of this is that the new type-checking gets a bit too excited about macros. `live_isolated_component` itself is defined as a macro, like so:
```elixir
defmacro live_isolated_component(component, opts) do
  do_live_isolated_component(
    component,
    quote(do: if(is_map(unquote(opts)), do: [assigns: unquote(opts)], else: unquote(opts)))
  )
end
```

In [the tests where we use it](https://github.com/mbta/dotcom/blob/1841cc18f0f1a1337187cc9a4d2a847da74ab3f0/test/dotcom_web/components/search_results_live_test.exs#L43-L71), we invoke `live_isolated_component` like so:
```elixir
live_isolated_component(SearchResultsLive, create_assigns())
```
And `create_assigns()` is known at compile-time to be a `%{map}`. 

So we end up with the macro expanding to something like
```elixir
  do_live_isolated_component(
    component,
    if(is_map(%{map}), do: [assigns: %{map}], else: %{map}))
  )
```

And of course `%{map}` is a map, so we get a warning.

Copilot suggested the hack of wrapping the `create_assigns()` call in `Function.identity/1`, so that Elixir would forget what its type was.

(Arguably less important: The reason there are two commits was because I wanted to show [the AI's first attempt](https://github.com/mbta/dotcom/pull/3486/commits/3aa7671a45d19d03625a836843ee47dcae8620f0), which was to wrap each `create_assigns()` call in `Function.identity/1`, as well as [my proposed cleanup ((also actually implemented by Copilot))](https://github.com/mbta/dotcom/pull/3486/commits/af28dc5e6dffbc3bda234f8b73f1a1a5f182aaa5) of moving the `Function.identity/1` call to the definition of `create_assigns/0`) 

## How to test

We should somewhat thoroughly test this out in a deployed environment to make sure that everything still works.


mix.exs
```diff
@@ -123,14 +123,14 @@ defmodule DotCom.Mixfile do
...
-      {:ex_cldr_units, "3.20.1"},
+      {:ex_cldr_units, "3.20.5"},
       {:ex_doc, "0.40.1", only: :dev},
       {:ex_machina, "2.8.0", only: [:dev, :test]},
       {:ex_unit_summary, "0.2.0", only: [:dev, :test]},
       {:excoveralls, "0.18.5", only: :test},
       {:faker, "0.19.0"},
       {:floki, "0.38.0"},
-      {:flame_on, "~> 1.0", only: [:dev]},
+      {:flame_on, github: "DockYard/flame_on", branch: "main", only: [:dev]},
```
joshlarson, line 133: `v1.0.0` of `flame_on` pins `meck` at version `0.9.2`, which is incompatible with OTP 29*. There is [a commit on `main`](https://github.com/DockYard/flame_on/commit/9a0cb1da756a09f0fd356c949021a717fda7ccb1) that relaxes that pin, and allows `flame_on` to work with the latest OTP, but there's no release with that commit on it yet. [I reported an issue](https://github.com/DockYard/flame_on/issues/67), so hopefully that will get released soon, and we'll be able to go back to the simpler `flame_on` import.

(Incidentally, that same incompatibility is also why we had to [finish replacing `mock` with `Mox`](https://github.com/mbta/dotcom/pull/3454) before doing the Elixir language upgrade, since `mock` also pinned `meck` at `0.9.2`.)

*"OTP" in this context is the Erlang version; no relationship to open trip planner.

thecristen (approved):
Seems good to me 👍🏼

PR #3489, opened 2026-09-11
dev(DotcomWeb.Router): show Phoenix's debugger

thecristen:
Phoenix ships with a default debugger view for local development, which we enabled by setting `debug_errors: true` in the Endpoint configuration.

We also have custom error handling set up in `DotcomWeb.Router`, using the `Plug.ErrorHandler` behavior. This is what turns many errors into 500 responses.

However, it turns out that overrides the debugger view. From its docs:

> Note: If this module is used with [Plug.ErrorHandler](https://plug.hexdocs.pm/Plug.ErrorHandler.html), only one of them will effectively handle errors. For this reason, it is recommended that [Plug.Debugger](https://plug.hexdocs.pm/Plug.Debugger.html) is used before [Plug.ErrorHandler](https://plug.hexdocs.pm/Plug.ErrorHandler.html) and only in particular environments, like :dev.

This PR changes our setup a bit to use the debugger on dev, and the current setup elsewhere.

## Screenshots

image

## How to test

Feel free to `raise` whatever errors you want in Elixir to try it out on your own pages locally

### DIFF
--- a/lib/dotcom_web/router.ex
+++ b/lib/dotcom_web/router.ex
@@ -13,26 +13,30 @@ defmodule DotcomWeb.Router do
   """
 
   use DotcomWeb, :router
-  use Plug.ErrorHandler
 
-  alias DotcomWeb.ControllerHelpers
+  if Mix.env() == :dev do
+    use Plug.Debugger, otp_app: :dotcom
+  else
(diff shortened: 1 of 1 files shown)


lvachon1 (approved):
Nice

PR #3491, opened 2026-09-11
feat: show new stop page layout & features

thecristen:
## Scope

**Asana Ticket:** TBA

### Using `live_render/3` to embed LiveViews into a regular page

> [!NOTE]
> This PR first fixes an issue where we couldn't use `live_render/3`. It turned out using different plugs for regular pages vs live view pages was making the root layout kind of confused and the security signature not matching, or something.
> It's fixed by the first commit, which made the search bar break with a 403 error and necessitated the second commit:
- [fix(DotcomWeb.Router): consolidate pipelines](https://github.com/mbta/dotcom/commit/8ad2a986e477d50ff012074141104f593603e7d6)
- [fix(autocomplete): send CSRF token in POST](https://github.com/mbta/dotcom/commit/68e2c0a207f103a1000e481a47d1d19cf2275147)

All other POST requests seem to remain working as expected.

### Kicking off the reworked Stop Page

- Created a feature flag
- Use the feature flag to toggle between old and new layouts

**Changed in both versions**: Some styling tweaks shared between both, mostly the heading sizes

**Added in new layout**
- The basic side-by-side structure, with Upcoming Departures at top in the mobile breakpoint
- Refactored the existing `StopMapLive` LiveView and added it to the new page where the map goes (thanks @joshlarson for the prior art in #2806 !)
- Stubbed out another LiveView for the Upcoming Departures piece

## Screenshots

| Normal | With Feature Flag |
|--------|--------|
| image | image | 

## How to test

Try with and without the feature flag

---
  - https://app.asana.com/0/0/1218398178190666

### DIFF
--- a/lib/dotcom_web/controllers/stop_controller.ex
+++ b/lib/dotcom_web/controllers/stop_controller.ex
@@ -71,6 +71,7 @@ defmodule DotcomWeb.StopController do
           |> Dotcom.StopAmenity.from_stop_facilities()
 
         conn
+        |> assign(:new_stop_page?, Laboratory.enabled?(conn, :new_stop_page))
         |> assign(:breadcrumbs, breadcrumbs(stop, routes_by_stop))
         |> meta_description(stop, routes_by_stop)
         |> assign_alerts()
--- a/lib/dotcom_web/live/stop_map_live.ex
+++ b/lib/dotcom_web/live/stop_map_live.ex
@@ -7,88 +7,62 @@ defmodule DotcomWeb.StopMapLive do
   use DotcomWeb, :live_view
 
   @map_config Application.compile_env(:mbta_metro, :map)
+  @stops_repo Application.compile_env!(:dotcom, :repo_modules)[:stops]
 
-  def(mount(_params, _session, socket)) do
+  def mount(_params, session, socket) do
+    stop_id = session["stop_id"]
+    stop = @stops_repo.get(stop_id)
+
+    child_stops_by_type =
+      stop.child_ids
+      |> Enum.map(&@stops_repo.get/1)
+      |> Enum.group_by(& &1.type)
+
+    entrance_icons =
+      child_stops_by_type
+      |> Map.get(:entrance, [])
+      |> Enum.map(
+        &%{
+          class: "size-5 cursor-pointer",
+          coordinates: [&1.longitude, &1.latitude],
+          name: "door-open",
+          type: "solid"
+        }
+      )
+
+    station_icon = %{
+      anchor: "bottom",
+      class: "size-12 cursor-pointer",
+      coordinates: [stop.longitude, stop.latitude],
+      name: "icon-map-station-marker",
+      type: "icon-svg"
+    }
+
     {
       :ok,
       socket
-      |> assign(:stop_id, "place-boyls")
-      |> assign(:map_config, @map_config)
+      |> assign(:map_config, zoom_to_stop(@map_config, stop))
+      |> assign(:icons, [station_icon | entrance_icons])
     }
   end
 
   def render(assigns) do
-    all_stops =
-      (Stops.Repo.by_route_type(0) ++ Stops.Repo.by_route_type(1))
-      |> Enum.uniq_by(& &1.id)
-      |> Enum.sort_by(& &1.name)
-
-    stop = Stops.Repo.get(assigns.stop_id)
-
-    child_stops = stop.child_ids |> Enum.map(&Stops.Repo.get/1)
-
-    child_stops_by_type = child_stops |> Enum.group_by(& &1.type)
-    entrances = child_stops_by_type |> Map.get(:entrance, [])
-
-    assigns =
-      assigns
-      |> assign(:stop, stop)
-      |> assign(:all_stops, all_stops)
-      |> assign(:child_stops, child_stops)
-      |> assign(:entrances, entrances)
-
     ~H"""
-    <div class="container">
-      <h1>Hello we are your map pin icon</h1>
-
-      <.live_component
-        module={DotcomWeb.Components.Map}
-        id="stop-page-map"
-        class="h-[32rem] w-full"
-        config={@map_config}
-        icons={
-          (@entrances
-           |> Enum.map(
-             &%{
-               class: "size-5 cursor-pointer",
-               coordinates: [&1.longitude, &1.latitude],
-               name: "door-open",
-               type: "solid"
-             }
-           )) ++
-            [
-              %{
-                anchor: "bottom",
-                class: "size-12 cursor-pointer",
-                coordinates: [@stop.longitude, @stop.latitude],
-                name: "icon-map-station-marker",
-                type: "icon-svg"
-              }
-            ]
-        }
-      />
-
-      <div class="flex flex-wrap gap-2 mt-4">
-        <button
-          :for={stop <- @all_stops}
-          class={[
-            "rounded p-sm",
-            stop.id == @stop.id && " bg-brand-primary-lightest"
-          ]}
-          phx-click="switch-stop"
-          phx-value-stop-id={stop.id}
-        >
-          {stop.name}
-        </button>
-      </div>
-    </div>
+    <.live_component
+      module={DotcomWeb.Components.Map}
+      id="stop-page-map"
+      class="h-96 w-full"
+      config={@map_config}
+      icons={@icons}
+    />
     """
   end
 
-  def handle_event("switch-stop", %{"stop-id" => stop_id}, socket) do
-    {
-      :noreply,
-      socket |> assign(:stop_id, stop_id)
-    }
+  defp zoom_to_stop(config, %{latitude: lat, longitude: lon}) do
+    config
+    |> Map.put(:center, [lon, lat])
+    |> Map.put(:zoom, 17)
   end
+
+  defp zoom_to_stop(config, _), do: config
 end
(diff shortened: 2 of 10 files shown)


lib/dotcom_web/live/stop_upcoming_arrivals.ex
```diff
@@ -0,0 +1,22 @@
+defmodule DotcomWeb.Live.StopUpcomingArrivals do
```
joshlarson: **Question:** Didn't we intentionally switch to a naming convention more akin to `DotcomWeb.FooBarLive`, as opposed to `DotcomWeb.Live.FooBar`?

I'm not overly attached to one or the other, but I do think it would be nice to be consistent.

lib/dotcom_web/live/stop_map_live.ex
```diff
@@ -7,88 +7,62 @@ defmodule DotcomWeb.StopMapLive do
...
-      :noreply,
-      socket |> assign(:stop_id, stop_id)
-    }
+  defp zoom_to_stop(config, %{latitude: lat, longitude: lon}) do
+    config
+    |> Map.put(:center, [lon, lat])
+    |> Map.put(:zoom, 17)
   end
+
+  defp zoom_to_stop(config, _), do: config
```
joshlarson: **Issue (blocking-ish):** `zoom_to_stop/2` can never be called with a second parameter that isn't a valid `%Stop{}` struct. While it is true that `@stops_repo.get(stop_id)` could return `nil`, the resulting `stop` is later called like so:
```elixir
coordinates: [stop.longitude, stop.latitude]
```

...which means that when `stop` is `nil`, then we'll crash before we get to the `zoom_to_stop/2` call. 

And Elixir 1.20 is smart enough to figure all that out and warn about it!

So I vote we get rid of this clause in order to get rid of an annoying warning.

lib/dotcom_web/templates/stop/show.html.heex
```diff
@@ -20,86 +20,97 @@
...
-      alerts={alerts_for_amenity(@bike_amenity, @alerts)}
-      date_time={@date_time}
-    />
-    <%= if @stop.station? do %>
-      <h3 class="hidden-md-up">Getting around the station</h3>
-      <.elevator_amenity
+<div class={["container", @new_stop_page? && "lg:flex xl:gap-xl md:gap-lg"]}>
+  <div :if={@new_stop_page?} class="lg:w-2/5 lg:order-last">
+    <h2 class="mt-sm text-xl">{~t"Upcoming Departures"}</h2>
+    {live_render(@conn, DotcomWeb.Live.StopUpcomingArrivals, session: %{"stop_id" => @stop.id})}
```
joshlarson: **Question:** How come you decided to use separate LiveView's here, instead of putting the whole page into a single one? Do you envision combining them at some point, or keeping them separate?

lib/dotcom_web/controllers/stop_controller.ex
```diff
@@ -71,6 +71,7 @@ defmodule DotcomWeb.StopController do
           |> Dotcom.StopAmenity.from_stop_facilities()
 
         conn
+        |> assign(:new_stop_page?, Laboratory.enabled?(conn, :new_stop_page))
```
joshlarson: **Suggestion/Comment (non-blocking):** We have [a nifty plug](https://github.com/mbta/dotcom/blob/2d9ebc7c5a1a19daf66ef8305a5e4ee2eae68097/lib/dotcom_web/plugs/put_flags_in_session.ex) that puts all of the flags into the `assigns` already, so we can just give this flag the atom of `:new_stop_page?` instead of `:new_stop_page` and skip this assign statement.

joshlarson (approved):
This looks great! And I'm excited to build more on top of it!

One of the comments is blocking-ish, but I decided that _ish_ isn't enough to stop me approving... it won't break anything, and is easy enough to address as a follow-up in case you'd rather roll forward without addressing it here.

lib/dotcom_web/live/stop_upcoming_arrivals.ex
thecristen: Updated (in a newly consolidated LiveView)!

lib/dotcom_web/live/stop_map_live.ex
thecristen: Done in [feedback: remove unneeded clause](https://github.com/mbta/dotcom/pull/3491/commits/358e72cd2f809d2d351a0f3cc14910d0a22a2bb7)

lib/dotcom_web/templates/stop/show.html.heex
thecristen: They're now together!

lib/dotcom_web/controllers/stop_controller.ex
thecristen: Yes! Thanks, I'd totally forgotten about that.

lib/dotcom_web/live/stop_information_live.html.heex
```diff
@@ -0,0 +1,106 @@
+<div class="u-bg--gray-bordered-background">
```
joshlarson: **Issue (non-blocking):** I'd love for us to come up with a general pattern for when we use `def render(assigns), do: ~H"<....... />"` versus separate HEEx files. We have some LiveViews that do it one way,  and others that do it the other way, and it feels kind of random?

I personally prefer the inlined approach because that makes it easier to extract pieces out into locally-defined `defp` components. With this approach, I think it's still possible to extract, but then to figure out how something is defined, we have to look at the LiveView file, the separate HEEx template, and possibly an even more separate component file.

lib/dotcom_web/live/stop_information_live.ex
```diff
@@ -0,0 +1,98 @@
...
+      type: "icon-svg"
+    }
+
+    [station_icon]
+  end
+
+  defp zoom_to_stop(config, %{latitude: lat, longitude: lon}) do
+    config
+    |> Map.put(:center, [lon, lat])
+    |> Map.put(:zoom, 17)
```
joshlarson:
```suggestion
    |> Map.put(:zoom, 15)
```

**Issue (non-blocking):** A starting zoom of 17 is super close-in, and makes it hard to see the surrounding area (at DTX for instance, zoom level 17 just shows one intersection). 15 is closer to the current stop page implementation.

Image

lib/dotcom_web/live/stop_map_live.ex
```diff
@@ -7,88 +7,60 @@ defmodule DotcomWeb.StopMapLive do
   use DotcomWeb, :live_view
 
   @map_config Application.compile_env(:mbta_metro, :map)
+  @stops_repo Application.compile_env!(:dotcom, :repo_modules)[:stops]
+
+  def mount(_params, session, socket) do
```
joshlarson: **Issue (non-blocking):** This LiveView is still invoked from the `/preview/stop-map` route, and this change breaks that usage.

That said, I think the actual remedy here is to get rid of this LiveView and its corresponding entry in the router. I only added this LV in order to try out putting the bottom-anchored stop marker onto a MapLibre map - now that we've put the map from this LV onto the `StopInformationLive` LV, I don't think we need this proof-of-concept anymore!

lib/dotcom_web/views/bus_stop_change_view.ex
```diff
@@ -75,12 +75,13 @@ defmodule DotcomWeb.BusStopChangeView do
...
-  def affected_stop_link(_conn, stopname) when is_binary(stopname) do
+  @spec affected_stop_link(Stop.t() | String.t()) :: Phoenix.HTML.Safe.t() | nil
+  def affected_stop_link(stopname) when is_binary(stopname) do
     content_tag(:div, stopname)
   end
 
-  def affected_stop_link(conn, stop) do
-    if(stop, do: link(stop.name, to: stop_path(conn, :show, stop.id), class: "text-primary"))
+  def affected_stop_link(stop) do
+    path = live_path(DotcomWeb.Endpoint, DotcomWeb.StopInformationLive, stop.id)
```
joshlarson: **Question:** How come you're building some of these paths with `"/stops/#{stop_id}"`, and some using the `live_path` function from the Router helper? The result should be the same. 

(FWIW: [The Phoenix Router docs](https://phoenix.hexdocs.pm/Phoenix.Router.html#module-generating-routes) say that `Helpers`, such as `live_path`, are deprecated in favor of invoking verified routes with `~p`... we really should get our `~p` sigil into a good state before `Helpers` is actually removed 🫠)

joshlarson (approved):
Nice! Left a few thoughts and questions, but nothing earth-shattering.

lib/dotcom_web/live/stop_map_live.ex
thecristen: Whoops! Thanks for flagging, I've removed it.

lib/dotcom_web/views/bus_stop_change_view.ex
thecristen: No good reason, just inconsistency. It'll be nice to be able to just use the `~p` :)

test/support/conn_case.ex
```diff
@@ -17,6 +17,7 @@ defmodule DotcomWeb.ConnCase do
 
   def default_conn do
     Phoenix.ConnTest.build_conn()
+    |> Plug.Conn.put_req_header("user-agent", Faker.Internet.UserAgent.user_agent())
```
joshlarson: This is causing [these test failures](https://github.com/mbta/dotcom/actions/runs/35941396732/job/107451497791?pr=3491#step:5:20). [These](https://github.com/mbta/dotcom/blob/25bc0b81906217c1b258b8d970dca1c416458eb1/test/dotcom_web/controllers/app_store_controller_test.exs#L4-L14) two [tests](https://github.com/mbta/dotcom/blob/25bc0b81906217c1b258b8d970dca1c416458eb1/test/dotcom_web/controllers/app_store_controller_test.exs#L59-L69) assume that the browser's user agent will be a desktop, so now that the user agents are sometimes mobile, these tests have become sad.

PR #3496, opened 2026-09-15
cleanup: Remove `redis` dependency from top-level `package.json`

joshlarson:
## Scope

No ticket, but inspired by having no idea what to test for dependabot's #3460, I was curious if we actually needed this.


joshlarson:
[This Splunk query](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22dotcom-dev-green-application%22%20monitor&display.page.search.mode=smart&dispatch.sample_ratio=1&workload_pool=&earliest=-15m&latest=now&display.page.search.tab=events&sid=1789484029.149344) tells me that the monitor is still working, and the success of the various smoke and e2e tests tell me that those are working as well.

I am reasonably certain that this means that we no longer need this dependency.

Screenshot 2026-09-15 at 10 54 41 AM

thecristen:
@joshlarson It wouldn't be reflected in the application deploy because it's used in the monitor (as are most of the NPM dependencies that aren't in `/assets`.

thecristen:
@joshlarson Here's a [log](https://mbta.splunkcloud.com/en-US/app/search/show_source?sid=1789495245.149811&offset=6&latest_time=) of it complaining it can't find `redis` for the Elasticache health check.

joshlarson:
Fair enough - bye bye!

PR #3497, opened 2026-09-15
fix(PageView): avoid caching rendered struct

thecristen:
## Scope

**Asana Ticket:** [BadFunctionError: function #Function<10.85423106/1 in DotcomWeb.PageView> is invalid](https://app.asana.com/1/15492006741476/project/385363666817452/task/1218163392809037?focus=true)
Should fix [this Sentry issue](https://mbtace.sentry.io/issues/7695724338/).

## Implementation

> [!NOTE]
> :robot: I used Copilot for this. Here's its assessment:

### Root cause

`DotcomWeb.PageView.alerts/1` is decorated with `@decorate cacheable(...)` (Nebulex, in-process/ETS-backed cache) and returned the raw result of `render("_alerts.html", ...)`. Since `_alerts.html.heex` is a HEEx template, `render/2` returns a `%Phoenix.LiveView.Rendered{}` struct whose `dynamic` field holds a closure over the compiled module code. That struct got cached and reused across requests.

When a deploy/hot code reload purges the old module version, any later request that reuses the cached struct calls the stale closure via `Phoenix.HTML.Safe.Phoenix.LiveView.Rendered.recur_iodata/1` → `BadFunctionError: ... points to an old version of the code`. The other cached PageView helpers (`important_links`, `top_links`, `tab_list`, `render_news_entries`, `render_upcoming_events`) already avoid this by fully resolving to iodata/strings before caching.

### Fix

Changed `alerts/1` to render to a fully-resolved string (`render_to_string/3` wrapped in `Phoenix.HTML.raw/1`) before it gets cached, matching the existing safe pattern used elsewhere in the same module — eliminating the stale function reference entirely.

Verified: project compiles cleanly and `test/dotcom_web/views/page_view_test.exs` passes (4/4).

## Screenshots

No changes to the alerts tab on the homepage!

## How to test

We can deploy to a staging server.


lvachon1 (approved):
Oops, this was my mistake.  Sorry!

thecristen:
@lvachon1 no worries, I didn't really understand the `render` output working this way would cause this sort of problem myself!

PR #3498, opened 2026-09-15
feat(health): add OTP health check

thecristen:
## Scope

**Asana Ticket:** [🗺️ 🛠️ Add a page to Dotcom that checks OTP's health](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217135535003755?focus=true)

## Implementation

Also updates to use the latest version of `open_trip_planner_client`.

Visit `/_health/open-trip-planner` and get a `200 OK` if Open Trip Planner is ok.

- It feels a little weird to return 503 (since _our_ application is doing fine..) but couldn't think of a better option
- I could update our own monitoring (in `integration/health_checks/otp.js`) to check this new path, instead of pinging OTP itself. Open to doing that in this PR if desired, or in a later PR.

## Screenshots

Nothing to see visually.

## How to test

Could run it with OTP locally.


lib/dotcom_web/controllers/health_controller.ex
```diff
@@ -4,8 +4,32 @@ defmodule DotcomWeb.HealthController do
...
+    conn
+    |> put_resp_header("cache-control", "max-age=60, public")
+    |> send_resp(status, "")
+  end
+
+  defp check_otp do
+    case @cache.get(:otp_health) do
+      nil ->
+        result = %{healthy?: @otp_module.healthy?()}
+        @cache.put(:otp_health, result, ttl: :timer.minutes(1))
```
thecristen, line 28: Open to reducing/removing this, was just trying to avoid potentially hammering OTP here.

joshlarson (approved):
Seems pretty reasonable to me!

PR #3499, opened 2026-09-15
[Line Diagram] [Right Rail]: Fare cards

lvachon1:
## Scope

**Asana Ticket:** [💈➡️Fare cards](https://app.asana.com/1/15492006741476/project/555089885850811/task/1216638931645035?focus=true)

## Implementation

Added logic to determine which fare card to pull from the CMS and show on the right rail.
Adjusted fare card css to use container queries instead of media queries so that our thin right rail uses the thin layout even on wide screens.
Added fare_info entries for the `free_fare` and `rapid_transit_fare` fare classes that do not have fare card CMS content yet.

### Subway Fare Card
Screenshot 2026-09-15 at 2 49 20 PM

### Commuter Rail Fare Card
Screenshot 2026-09-15 at 2 50 19 PM

### Bus Fare Card
Screenshot 2026-09-15 at 2 50 34 PM

### Ferry Fare Card
Screenshot 2026-09-15 at 2 50 52 PM

### Silver Line Rapid Transit Fare Card
Screenshot 2026-09-15 at 2 51 16 PM

### Free Fare Card
Screenshot 2026-09-15 at 2 51 27 PM

## How to test

http://localhost:4001/schedules/Green/line_new - Subway Fare
http://localhost:4001/schedules/71/line_new - Bus Fare
http://localhost:4001/schedules/CR-NewBedford/line_new - Commuter Rail Fare
http://localhost:4001/schedules/Boat-F1/line_new - Ferry Fare
http://localhost:4001/schedules/741/line_new - Silver Line Rapid Transit
http://localhost:4001/schedules/751/line_new - Silver Line Bus
http://localhost:4001/schedules/23/line_new - Free Bus

Confirm that the proper fare card shows up, and that the links work and go to the right pages.  
Confirm that the fare cards on those content pages look right.
Confirm that the fare cards on the new line diagram look like the [figma](https://www.figma.com/design/fXXxB7zL0OsxOItwfWtAxQ/MBTA.com?node-id=9599-37694&m=dev)
Confirm that the fare cards on mode pages (http://localhost:4001/schedules/bus) look right.

### DIFF
--- a/lib/dotcom_web/live/line_diagram_live.ex
+++ b/lib/dotcom_web/live/line_diagram_live.ex
@@ -7,6 +7,7 @@ defmodule DotcomWeb.LineDiagramLive do
   @route_patterns_repo Application.compile_env!(:dotcom, :repo_modules)[:route_patterns]
   @alerts_repo Application.compile_env!(:dotcom, :repo_modules)[:alerts]
   @date_time_module Application.compile_env!(:dotcom, :date_time_module)
+  @routes_repo Application.compile_env!(:dotcom, :repo_modules)[:routes]
 
   alias DotcomWeb.PartialView.{HeaderTab, HeaderTabs}
 
@@ -16,9 +17,11 @@ defmodule DotcomWeb.LineDiagramLive do
     only: [
       header_class: 1,
       route_feature_badge: 1,
-      route_tab_class: 1
+      route_tab_class: 1,
+      single_trip_fares: 1
     ]
 
+  import DotcomWeb.ModeView, only: [mode_fare_card: 1]
   import DotcomWeb.Views.Helpers.AlertHelpers, only: [alert_badge: 1]
 
   on_mount DotcomWeb.Hooks.AssignRoute
@@ -33,6 +36,7 @@ defmodule DotcomWeb.LineDiagramLive do
       params |> Map.get("schedule_direction", %{direction_id: 0}) |> Map.get("direction_id")
 
     tab_params = %{"schedule_direction[direction_id]": direction_id}
+    fare_card = mode_fare_card(route |> Routes.Route.type_atom())
 
     {:ok,
      socket
@@ -41,7 +45,8 @@ defmodule DotcomWeb.LineDiagramLive do
      |> assign(:route_id, route_id)
      |> assign(:route, route)
      |> assign(:tab, "new_line")
-     |> assign(:tab_params, tab_params)}
+     |> assign(:tab_params, tab_params)
+     |> assign(:fare_card, fare_card)}
   end
 
   def make_link(assigns, page, add_params? \\ false) do
@@ -130,8 +135,13 @@ defmodule DotcomWeb.LineDiagramLive do
         >
           ⚠️ Watch Your Step ⚠️
         </marquee>
+        <.fares fare_card={@fare_card} />
       </div>
     </div>
     """
+  end
+
+  def fares(assigns) do
+    assigns.fare_card |> DotcomWeb.PartialView.paragraph(%Plug.Conn{query_params: %{}})
   end
 end


lib/dotcom_web/live/line_diagram_live.ex
```diff
@@ -197,12 +249,41 @@ defmodule DotcomWeb.LineDiagramLive do
           ⚠️ Watch Your Step ⚠️
         </marquee>
         <.route_pdf_sidebar_content route_pdfs={@route_pdfs} date={@date} route={@route} />
+        <div style="container-type: inline-size;" class="w-[380px]">
```
joshlarson: **Issue:** `w-[380px]` is causing this component to get pushed off the edge of the screen between screen widths of 800px and 930px, and also causes it to look too narrow when the screen is below 800px. 

I think setting this to `w-full` has the desired effect (and has the added bonus of allowing the multi-column layout to shine at breakpoints between 620 and 800 px. 

---

I would also recommend changing the `w-[380px]` in the definition of `<.guides />` to `w-full`, although I admit that that does create a bit of an awkward situation where the guide images start to get really big at large "narrow" screen sizes. Exactly how to handle that is probably a design question.

assets/css/_multi-column.scss
```diff
@@ -15,7 +15,10 @@
 
 // Multi-column with either Fare Cards or Descriptive Links
 .c-paragraph--with-cards {
-  @include media-breakpoint-up(sm) {
+  container-type: inline-size;
+
+  @container (width > 544px) {
```
joshlarson:
```suggestion
  @container (width >= 544px) {
```

If you follow my suggestion below to replace `w-[380px]`, then at a browser width of 620px, the container winds up at exactly 544px, which causes this query not to match, and also causes the `width < 544px` query defined in `_fare-card.scss` not to match, which winds up looking like this.

Image

One or the other of the comparisons need to include the "or equal to" case, so that widths of exactly 544px still work properly.

lib/dotcom_web/live/line_diagram_live.ex
```diff
@@ -105,11 +106,62 @@ defmodule DotcomWeb.LineDiagramLive do
...
+    fare_card =
+      cond do
+        fare_class == :free_fare ->
+          :free_fare
+
+        fare_class == :rapid_transit_fare and route.type == 3 ->
+          :rapid_transit_fare
+
+        true ->
+          mode_fare_card(route |> Routes.Route.type_atom())
```
joshlarson, line 128: **Issue:** For ferry in particular, we want to show a route-specific card, because in this context, we know what route we're looking at, and the price range for the ferry mode fare card is entirely because each route has a different price.

It's a bit more of a reach, and maybe out of scope for this ticket, but we could also limit the price range for commuter-rail routes. Needham Line trips can never cost more than $7.00, and the Providence/Stoughton Line is the only one that can rack up the max price of $13.25.

joshlarson (changes requested):
The two changes that I'm requesting for this are on complete opposite sides of the concrete <-> abstract spectrum. 

---

The concrete, tiny change that is blocking for me is the `w-[380px]` thing - that's a bit too rigid a hard-coded width for something that can be viewed at a number of different widths.

---

The high-falutin' abstract one is that I don't think we should be reaching out to the CMS for the fare cards (I'm aware that we did land on that as a "best-guess implementation strategy" when scoping this ticket earlier-on, so I'm sorry about the rollercoaster there! 😅). That said, this PR has exposed a few reasons why I'm reconsidering:
- The code is surprisingly convoluted, and necessarily must be. There's no way to statically examine the code and see what fare card will wind up on a given page, or even _that_ a fare card will wind up on a given page.
- This page will now fail to load if the CMS is down and our cache isn't populated.
- There are a number of fare cards that have no entry in the CMS. We'd either have to get content to add fare cards for us that they're likely never going to use, or we'll have to special-case those.
- Most of the fare card content comes from our code anyway - the CMS just tells us to put the full-price fare on the left(top) and the reduced price fare plus the `Learn more` link on the right(bottom), but the prices and fare media come from Dotcom anyway.
- The fact that we're forced to use `@dialyzer {:nowarn_function, fare_card: 1}` signals that we're invoking a function (in this case `DotcomWeb.PartialView.paragraph/1`) in a way that it wasn't meant to be used.

assets/css/_multi-column.scss
```diff
@@ -15,7 +15,10 @@
 
 // Multi-column with either Fare Cards or Descriptive Links
 .c-paragraph--with-cards {
-  @include media-breakpoint-up(sm) {
+  container-type: inline-size;
+
+  @container (width >= 544px) {
+
```
joshlarson: **Nit:** I think there's a zombie newline here.

assets/css/_multi-column.scss
```diff
@@ -24,9 +27,21 @@
...
+    }
+  }
+
+  .col-sm-6 {
+    width: 100%;
+    float: none;
   }
 }
 
+
```
joshlarson: Zombie newline

lib/dotcom_web/components/fare_card.ex
```diff
@@ -0,0 +1,202 @@
...
+          <a href="/fares/reduced">{~t"Learn more about reduced fares"}</a>
+        </p>
+      </div>
+    </div>
+    <.fare_note route={@route} />
+    """
+  end
+
+  # Stock CMS fare cards already in use on the site
+  def fare_card(%{route: route} = assigns) do
```
joshlarson:
```suggestion
  def fare_card(assigns) do
```

`route` is unused, so this is generating a warning. (`@route` in the HEEx template doesn't count, since that's actually coming from the assigns)

lib/dotcom_web/components/fare_card.ex
```diff
@@ -0,0 +1,202 @@
...
+    """
+  end
+
+  def fare_note(assigns) do
+    ~H"""
+    <div class="text-sm">
+      <div :if={@route.id in ["741", "742", "743", "746"]}>
+        {~t"﹡SL1, SL2, SL3, and SLW are priced as subway fares"}
+      </div>
+      <div class="text-sm">
```
joshlarson: **Nit:** Since the parent element has class `text-sm`, this one doesn't need it.

lib/dotcom_web/components/fare_card.ex
```diff
@@ -0,0 +1,202 @@
...
+    """
+  end
+
+  def fare_link(%{route: %{type: 2}} = assigns) do
+    ~H"""
+    <a href="/fares/commuter-rail-fares">{~t"More Commuter Rail fare options"}</a>
+    """
+  end
+
+  def fare_link(%{route: %{type: 3, id: id}} = assigns) when id in ~w(741 742 743 746) do
```
joshlarson: **Idea/Question/Thought/Speculation (non-blocking):** I wonder if there are other places where we care about silver line rapid bus (as opposed to silver line local bus, or silver line overall). Maybe it's worth creating `silver_line_rapid_bus` analogues to the existing [guard](https://github.com/mbta/dotcom/blob/38bf55d1327655c6e6b2ecea40c3f290aa821c6f/lib/routes/route.ex#L102-L103) and [function](https://github.com/mbta/dotcom/blob/38bf55d1327655c6e6b2ecea40c3f290aa821c6f/lib/routes/route.ex#L254) that more broadly apply to the whole silver line.

lib/dotcom_web/components/fare_card.ex
```diff
@@ -0,0 +1,202 @@
...
+    </div>
+    <.fare_note route={@route} />
+    """
+  end
+
+  # Silver Line rapid transit fare (same as subway)
+  def fare_card(%{route: %{fare_class: :rapid_transit_fare, type: 3}} = assigns) do
+    full_fare =
+      Fares.Repo.for_fare_class(:rapid_transit_fare)
+      |> Fares.Repo.filter(%{duration: :single_trip, includes_media: :charlie_card})
```
joshlarson:
```suggestion
      |> Fares.Repo.filter(duration: :single_trip, includes_media: :charlie_card)
```

[`Fares.Repo.filter/2` expects a keyword list as its second argument](https://github.com/mbta/dotcom/blob/38bf55d1327655c6e6b2ecea40c3f290aa821c6f/lib/fares/repo.ex#L30). It feels like kind of silly requirement, given that the `filter` function immediately converts `opts` to a map, but this type mismatch explains four of the five dialyzer errors that you get if you don't do the `@nowarn_function` thing above.

(The fifth is the fact that dialyzer can't figure out that `%Plug.Conn{query_params: ...}` is actually good enough for the `paragraph` call, and honestly I don't know how to fix that without the `@nowarn_function`, but still might as well clean up the other issues, just in case we do come up with a nice idea)

lib/dotcom_web/components/fare_card.ex
```diff
@@ -0,0 +1,202 @@
...
+        </div>
+        <h3 class="c-fare-card__name">{~t"Fare-Free Bus"}</h3>
+      </div>
+      <div class="c-multi-column__column">
+        <h4 class="mt-0">{~t"Free"}</h4>
+        <p>
+          {~t"Service on this route is free"}<br />
+          <a class="underline text-sm" href="/projects/fare-free-program-routes-23-28-and-29">
+            {~t"Learn more about the Fare-Free Program"}
+          </a>
```
joshlarson: **Issue (blocking):** The formatting of the various `Learn more about...` notes is inconsistent across the different fare cards. The free bus is the only one that's underlined, and notes on the ferry cards are a larger font size than the others.

```suggestion
          <div class="c-fare-card__note">
            <p>
              <a href="/projects/fare-free-program-routes-23-28-and-29">
                {~t"Learn more about the Fare-Free Program"}
              </a>
            </p>
          </div>
```

I would recommend wrapping all of the `Learn more about ...` fare card notes in a `<div class="c-fare-card__note" ><p><a....></p></div>`, and not applying any additional styling to them, so that they come out looking consistent.

Image

Compare ☝️ to 👇 

Image

lib/dotcom_web/components/fare_card.ex
```diff
@@ -0,0 +1,202 @@
...
+    ~H"""
+    <div class="c-fare-card--subway c-fare-card--grouped c-fare-card">
+      <div class="c-fare-card__header">
+        <div class="c-fare-card__icon">
+          <DotcomWeb.Components.RouteSymbols.route_icon
+            route={%Routes.Route{type: 3}}
+            class="c-svg__icon c-svg-icon__rapid-bus"
+          />
+        </div>
+        <h3 class="c-fare-card__name">{~t"Silver Line One-Way"}<sup>*</sup></h3>
```
joshlarson:
```suggestion
        <h3 class="c-fare-card__name">{~t"Silver Line One-Way"}*</h3>
```

**Issue (blocking):** The asterisk is already lifted above the other characters. Superscripting it makes it look _really_ high.

lib/dotcom_web/components/fare_card.ex
```diff
@@ -0,0 +1,202 @@
...
+    |> DotcomWeb.PartialView.paragraph(%Plug.Conn{query_params: %{}})}
+    <.fare_note route={@route} />
+    """
+  end
+
+  def fare_note(assigns) do
+    ~H"""
+    <div class="text-sm">
+      <div :if={@route.id in ["741", "742", "743", "746"]}>
+        {~t"﹡SL1, SL2, SL3, and SLW are priced as subway fares"}
```
joshlarson:
```suggestion
        *{~t"SL1, SL2, SL3, and SLW are priced as subway fares"}
```

**Issue (blocking):** That's a cool centered-asterisk-style character, but the Figma wants a regular asterisk there, and I agree 🙂 

(I'm not totally sure about this, but I don't think we should try translating the `*` as part of the `~t` call. I'm actually not entirely sure that we should be translating the `SL1`/`SL2`, etc either, but we can fiddle with that later.)

lib/dotcom_web/components/fare_card.ex
```diff
@@ -0,0 +1,202 @@
...
+    reduced_fare =
+      Fares.Repo.for_fare_class(:rapid_transit_fare)
+      |> Fares.Repo.filter(%{duration: :single_trip, includes_media: :student_card})
+      |> List.first()
+      |> Map.get(:cents)
+
+    assigns = assigns |> assign(:full_fare, full_fare) |> assign(:reduced_fare, reduced_fare)
+
+    ~H"""
+    <div class="c-fare-card--subway c-fare-card--grouped c-fare-card">
```
joshlarson: **Issue (blocking):** While silver-line-rapid-transit is priced like subway, it should be styled like silver line. Currently, the SL1-3+W line is getting the subway color, which doesn't look quite right.

Image

I think you'll need to add CSS to two places ([one](https://github.com/mbta/dotcom/blob/25bc0b81906217c1b258b8d970dca1c416458eb1/assets/css/_fare-card.scss#L116-L139), [two](https://github.com/mbta/dotcom/blob/25bc0b81906217c1b258b8d970dca1c416458eb1/assets/css/_fare-card.scss#L66-L84)) in order to get all the border colors to come out right.

joshlarson (changes requested):
This looks mostly good, but there are several visual issues that I think need to be addressed before this can be considered done. I would be happy to downgrade all of my comments to non-blocking if this fare card work is blocking anything, but I'm reasonably certain that it's not, hence the blocking feedback.

One additional note, which I did _not_ mark as blocking:

Image

For the fare cards that do come out of the CMS, the `More ... options` link at the bottom is pushed _waaaaay_ down. I do think we'll need to figure out a way to get that to look right, but I'm not going to block approval on that because I have no idea _how_ we could solve this before we come up with a final decision about how this will interact with the CMS.

---

Relatedly, I think the CMS is returning the wrong thing for commuter rail routes. Figma wants a single fancy fare card, but we're currently showing two simple fare cards.

Image

...versus...

Image

...unless I missed a requirements change somewhere?

PR #3501, opened 2026-09-15
fix(Stops.Header): vehicle_type to infer bus stop

thecristen:
## Scope

**Asana Ticket:** [🛑 Add stop ID to stop pages for connecting bus stops](https://app.asana.com/1/15492006741476/project/555089885850811/task/1215333323264097?focus=true)

## Implementation

- Adds retrieving and parsing of a stop's `vehicle_type` field.

## How to test

Check some stops! From the link:
- /stops/22549
- /stops/2548
- and so on...


lvachon1 (approved):
Looks good, though it appears that the data is lacking for many stops outside of dev-green.  I think this might want to wait until that version of the API is deployed before sending this out.

thecristen:
Actually many stops (all rail it seems) do seem to be missing it, but for our use case I think it's sufficient. Thanks!

PR #3502, opened 2026-09-15
chore(deps): Upgrade `phoenix_live_view` to `1.2.11`

joshlarson:
## Scope

**Asana Ticket:** [🛠️ (2026/09) Upgrade Dotcom phoenix_live_view dependency](https://app.asana.com/1/15492006741476/project/555089885850811/task/1218457401577213?focus=true)

## Implementation

This turned out to be slightly non-trivial because of two ways in which PhoenixLiveView appears to have gotten more strict with its validations:
1. It looks like it's validating the absence of stray `}` characters in component invocations. e.g. something like 👇 used to slip past the validator, but now causes a compiler error
```heex
<.foo
  bar={baz}
  }
/>
```

2. It now warns if you pass an `attr` in as part of a `:global` type if it's not one of the attributes that's allowed on all HTML elements (like `id`), and isn't explicitly `included`.

These two validations required an [upgrade to Metro](https://github.com/mbta/mbta_metro/pull/40), because Metro had a dependency on a version of phoenix_storybook that ran afoul of the stray `}` rule, and because Metro didn't explicitly `include` a few attributes that we were passing into `<.icon />` and `<.button />` components.

It also required a fix in the cache_controller, because [it also ran afoul](https://github.com/mbta/dotcom/blob/018f0292c2008fa3004f250c76f807b01c9ed4de/lib/dotcom_web/controllers/cache_controller.ex#L210) of the stray `}` rule.


lvachon1 (approved):
Looks solid

PR #3504, opened 2026-09-16
feat: Add a `/version` endpoint

joshlarson:
## Scope

I don't think we have a ticket for this, but it's something I've been wanting to do for a while. With this change, [the `/version` endpoint](https://dev-blue.mbtace.com/version) will tell us what version of the code is currently running.

## Implementation

This is largely a clone of [when I added a similar feature in Skate](https://github.com/mbta/skate/pull/2549). This change is slightly more involved, because Skate started out with the `SENTRY_RELEASE` environment variable set, and we did not, so we had to add that as well.

> [!NOTE]
> :robot: This entire PR was AI-generated, using the prompt below:
> 
> > There is another project called Skate, located at `~/Projects/skate`. Skate defines an endpoint called `/version` that's based on the `SENTRY_RELEASE` environment variable. Please look at how Skate sets and uses that environment variable, and copy that into Dotcom.

## Screenshots
Screenshot 2026-09-16 at 1 54 14 PM

Screenshot 2026-09-16 at 1 54 27 PM

## How to test

Navigate to the version endpoint in [dev-blue](https://dev-blue.mbtace.com/version) (where this PR is currently deployed), and [local](http://localhost:4001/version).


lvachon1 (approved):
✅

PR #3505, opened 2026-09-16
[News Archive] Add archive notice to old press releases

lvachon1:
## Scope

**Asana Ticket:** [🗞️ 🕸️ Add archive notice to old press releases](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217000162837992?focus=true)

## Implementation
Added age calculation to news_controller, then used that age to determine if a cta should be shown in news_entry/show.html and what message to show.  
Added a little bit of style to match the figma.

## Screenshots

Screenshot 2026-09-17 at 2 02 09 PM

Screenshot 2026-09-17 at 2 02 19 PM

Screenshot 2026-09-17 at 2 02 26 PM

Screenshot 2026-09-17 at 2 02 37 PM

## How to test

Old entries start at around [page 15](http://localhost:4001/news?page=15)
Two years is at around [page 30](http://localhost:4001/news?page=15)
This is our [oldest news entry](http://localhost:4001/news/2005-01-05/mbta-transit-police-name-new-deputy-chief) at [page 231](http://localhost:4001/news?page=231)

Confirm the messages show, and that the year difference calculation is correct.


joshlarson (changes requested):
There are three layout changes that I think this needs.

---

First off, there are some differences between the icon in the Figma versus this branch. I suspect that most of them don't really matter, but the Figma shows the icon top-aligned, as opposed to this branch, which shows the icon centered, and I do think that the top-aligned icon looks better. Also, the spacing doesn't look exactly the same, and while I'm not sure this needs to be pixel-perfect, I will say that subjectively, the current icon-text-distance makes the whole thing look a bit cramped.

**Branch**
Image

**Figma**
Image

I suspect this will require a change to the `<.cta />` component. I don't remember where else that's used, but I bet that any other places it's used will also be enhanced with a top-aligned rather than centered icon.

---

Second, (also visible in ☝️ screenshots) the Figma shows the banner above the "Posted on" date, while this branch has it below.

---

Second off, on wider screens, the archive note has a different width than the content below it, which gives it kind of a disjointed feel. Would it be possible to either shrink the notice or widen the content?

Image

lib/dotcom_web/components/components.ex
```diff
@@ -426,9 +426,15 @@ defmodule DotcomWeb.Components do
...
-      <span class="leading-tight grow">
+    <.cta_wrapper class={"cta-a gap-2 relative" <> @classes} link={@link} {@rest}>
+      <.icon
+        :if={@icon}
+        type={@icon_type}
+        name={@icon}
+        class="size-5 shrink-0 top-4 absolute"
+        aria-hidden
+      />
+      <span class="leading-tight grow ml-[2rem]">
```
joshlarson, line 437: **Issue (blocking):** The `relative`/`absolute`/`top-4` solution here works decently well in the context of the archive notices, but it does break other `<.cta />` consumers. 

- [Trip Planner a11y warning](http://localhost:4001/trip-planner?plan=hsQVX3VudXNlZF9kYXRldGltZV90eXBlxADECGRhdGV0aW1lxCAyMDI2LTA5LTE4VDExOjIwOjIxLjc1NjAyNi0wNDowMMQEZnJvbYTECGxhdGl0dWRly0BFJn3Dej20xAlsb25naXR1ZGXLwFHHSteUpuzEBG5hbWXEDEZvcmVzdCBIaWxsc8QHc3RvcF9pZMQLcGxhY2UtZm9yaGzEBW1vZGVzicQDQlVTxAR0cnVlxAVGRVJSWcQEdHJ1ZcQEUkFJTMQEdHJ1ZcQGU1VCV0FZxAR0cnVlxA5fcGVyc2lzdGVudF9pZMQBMMQLX3VudXNlZF9CVVPEAMQNX3VudXNlZF9GRVJSWcQAxAxfdW51c2VkX1JBSUzEAMQOX3VudXNlZF9TVUJXQVnEAMQCdG-ExAhsYXRpdHVkZctARTKqjrRjScQJbG9uZ2l0dWRly8BRyQrYoRZmxARuYW1lxAdBbGV3aWZlxAdzdG9wX2lkxAtwbGFjZS1hbGZjbMQKd2hlZWxjaGFpcsQEdHJ1ZQ==)
- [Service ended callout](http://localhost:4001/departures?route_id=171&direction_id=0&stop_id=place-nubn) (among others) on departures page
- (Less important, but...) new timetable banner on [timetable pages](http://localhost:4001/schedules/CR-Greenbush/timetable) when [the flag is set](http://localhost:4001/_flags)

There are a few issues causing this to get wonky.

---

Without the space before the `<.cta_wrapper />`'s interpolated `@classes`, any extra classes get smushed into the last item, which effectively erases the last `<.cta_wrapper />`'s default class (`relative` currently), and the consumer's first class.

On the trip planner page linked above, that means that the `<.cta_wrapper />` [gets the class](https://github.com/mbta/dotcom/blob/db4374f693ba34f4f2401875a3e63a47de7f7e35/lib/dotcom_web/live/trip_planner_live.ex#L121) `relativecol-span-full`, which causes the icon's `top-4` to pin it to the nearest `relative` ancestor, which in this case is the whole page instead of the `<.cta_wrapper />`, causing it to float away like a lost balloon.

(...I was at a kid's birthday party last weekend - there were tears 🎈)

Image

(Fun fact: the same thing would happen to the flask icon on the new timetable banner, except that the `<.cta_wrapper />` there is inside of [an element with class `calendar-covered`](https://github.com/mbta/dotcom/blob/62ad346f78e057a1d7e035868d36f015169d7363/lib/dotcom_web/templates/schedule/_timetable.html.heex#L44), which [gives _that_ element a `relative` position](https://github.com/mbta/dotcom/blob/62ad346f78e057a1d7e035868d36f015169d7363/assets/css/_schedule.scss#L137-L140). This is, however, why the text in that banner went from 14px to 16px - the text size class became `relativetext-sm`)

Image

---

Also, while the combination of `absolute` position and `ml-[2rem]` works nicely when there is an icon (assuming that `relative` gets the space it needs), it doesn't work when there is no icon. Most of the upcoming departures callouts use the `<.cta />` component under the hood (via `<.callout />`), and none of them have icons. The result here is that the text gets pushed off-center. See blow with prod compared to this branch:

Image

The off-center-ness is really obvious at narrow browser widths, but it happens at all widths.

---

Here's my suggested rewrite of this chunk:
```suggestion
    <.cta_wrapper class={"cta-a gap-2 " <> @classes} link={@link} {@rest}>
      <div :if={@icon} class="size-5 shrink-0 flex items-center justify-center">
        <.icon type={@icon_type} name={@icon} class="size-4 shrink-0" aria-hidden />
      </div>
      <span class="leading-5 grow">
```

Along with one additional change:
- [In `_callout.scss`](https://github.com/mbta/dotcom/blob/62ad346f78e057a1d7e035868d36f015169d7363/assets/css/_callout.scss#L88), update `.cta-a { align-items }` to be `start` instead of `center`.

This puts the icon back into the regular flex layout, and only allocates space for it when the icon is present, fixing the centering issue, and because the icon is no longer absolutely positioned, it'll stay inside of the `<.cta />`.

The reason I suggested making the icon `size-4` inside of a `size-5` wrapper is because that way, the icon will appear vertically aligned with the first line of the text (because the `-5` of `size-5` matches the `leading-5` of the text), and thus everything will be pretty much exactly in line with the Figma, which shows the icon being 16x16 px.

I'm open to other ways of accomplishing that (or even having the icon be `size-5` and losing the wrapping `<div />`, although subjectively, I think the `size-4` icon looks nicer).

lib/dotcom_web/templates/news_entry/show.html.heex
```diff
@@ -3,6 +3,28 @@
     <div class="c-cms__header">
       <div class="c-cms__title">
         <h1 class="c-cms__title-text">{@news_entry.title}</h1>
+        <div :if={@age > 365} class="c-cms__old_news">
+          <.cta
+            icon="circle-info"
```
joshlarson, line 8: **Issue (blocking):** The Figma says (and subjectively, I agree) that this should be 14px, rather than 16px, which means we should add `classes="text-sm"` here.


```suggestion
            classes="text-sm"
            icon="circle-info"
```

joshlarson (changes requested):
Thanks for addressing the banner width, content ordering, and icon alignment comments!

I left a comment about the icon alignment stuff - the current implementation on this branch borks things for other consumers of the `<.cta />` component, and we need to fix that before merging this.

joshlarson (approved):
Woop!

PR #3507, opened 2026-09-17
fix(cache): Add a TTL to pre-fetched data stored in the cache

joshlarson:
## Scope

No ticket, but I noticed that some cache keys weren't being evicted when I expected they would. Turns out, we were storing a whole bunch of things in the cache with infinite TTL by mistake!

> [!IMPORTANT]
> Since we have a bunch of records in the cache with infinite TTL's, we should clear the cache after releasing this commit in order to wipe those records out.
>
> We should do this in every environment.

## Implementation

We have a few spots where we optimize future lookups by, when we see a list of, say, `%Route{}` structs returned from the `MBTA.Api.Routes.all/1` call, we then cache each one individually so that we can look them up individually without having to make more API calls.

There are two problems with this optimization:
1. By storing the results with an infinite TTL, we prevent Redis from removing them when they go stale. This already causes some problems where we keep stale route data long after new GTFS updates, since we rely on the cache expiring to force us to re-query data from the API. This would also play poorly with https://github.com/mbta/dotcom/pull/3506, because we'd never expire routes that were cached by old versions of the code, even though we'd never query them again, filling up our cache with ancient versions of old data.
2. The optimization in `Routes.Repo.all` that stores `Routes.Repo.cached_get` records works, in the sense that future calls to `Routes.Repo.cached_get` would not re-query the API. That's not true for the `Routes.Repo.get_shape` and `Stops.Repo.stop` records though; for those two invocations, the `args` that we were passing in as the third parameter were a single binary (e.g. `"931_0019"`) instead of a list (e.g. `["931_0019"]`). Since actual `get_shape` and `stop` invocations pass their args to the key generator as a single-entry list, that means that the optimized infinite-TTL records were never actually used, and we re-queried the API anyway.

## How to test

All of the commands below should be run in an `iex -S mix` shell.

To test the `Routes.Repo.cached_get` fix, run the following:

```elixir
Dotcom.Cache.Multilevel.flush
Routes.Repo.all
Dotcom.Cache.Multilevel.ttl("routes.repo|cached_get|WyJPcmFuZ2UiLCBbaW5jbHVkZTogInJvdXRlX3BhdHRlcm5zIl1d")
```

On this branch, you should see a number slightly under 3,600,000 (the number of milliseconds in an hour). On `main`, you should see `:infinity`.

---

To test the `Routes.Repo.get_shapes` fix, the cache keys are different, because the args have changed (see "Implementation" section above).

On this branch, run:
```elixir
Dotcom.Cache.Multilevel.flush
Routes.Repo.get_shapes("Red", direction_id: 0)
Dotcom.Cache.Multilevel.ttl("routes.repo|get_shape|WyI5MzFfMDAxOSJd")
```

You should see a number slightly under 3,600,000.

On `main`, run:

```elixir
Dotcom.Cache.Multilevel.flush
Routes.Repo.get_shapes("Red", direction_id: 0)
Dotcom.Cache.Multilevel.ttl("routes.repo|get_shape|IjkzMV8wMDE5Ig==")
```

You should see `:infinity`. 

If you run the "wrong" thing for the branch, then you should see `nil`.

---

To test the `Stops.Repo.stop` fix, the cache keys are different for the same reason as the `get_shapes` one, so the thing to run on this branch and `main` are different:

On this branch, run:
```elixir
Dotcom.Cache.Multilevel.flush
Stops.Repo.by_route("Blue", 1)
Dotcom.Cache.Multilevel.ttl("stops.repo|stop|WyJwbGFjZS1hcXVjbCJd")
```

You should see a number slightly under 3,600,000.

On `main`, run:
```elixir
Dotcom.Cache.Multilevel.flush
Stops.Repo.by_route("Blue", 1)
Dotcom.Cache.Multilevel.ttl("stops.repo|stop|InBsYWNlLWFxdWNsIg==")
```

You should see `:infinity`. 

Just as above, if you run the "wrong" thing for the branch, then you should see `nil`.

---

If you're curious where I got those hard-coded strings like `"InBsYWNlLWFxdWNsIg=="` and `"WyJwbGFjZS1hcXVjbCJd"` from, I initially got them by calling `Dotcom.Cache.Multilevel.all`, which returns a list of all of the keys, searching for ones that started with `stops.repo`, and fiddling a bit more, e.g. 👇 
```elixir
Dotcom.Cache.Multilevel.all |> Enum.filter(& String.starts_with?(&1, "stops.repo|stop")) |> Enum.map(& {&1, &1 |> String.split("|") |> Enum.at(2) |> Base.decode64!()})
```

The difference between `"InBsYWNlLWFxdWNsIg=="` and `"WyJwbGFjZS1hcXVjbCJd"` in particular can be seen easily by base64-decoding them:

```elixir
iex> Base.decode64!("InBsYWNlLWFxdWNsIg==")
"\"place-aqucl\""

iex> Base.decode64!("WyJwbGFjZS1hcXVjbCJd")
"[\"place-aqucl\"]"
```


thecristen (approved):
Awesome find!!

PR #3509, opened 2026-09-18
fix(schedules): errors from API cause page to 500/show incorrect page state

jlucytan:
## Scope

**Asana Ticket:** [🐞🟢 Fix issue causing GL line diagrams not working on dev-green](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217464268376071)

Additional context in this [Slack thread](https://mbta.slack.com/archives/C06TP48CJ9W/p1789674688549139?thread_ts=1789674683.910489&cid=C06TP48CJ9W)

## Implementation
- When Stops.Repo.by_route receives an error from the API, it returns a result as an {:error, any} tuple 
- Most consumers of `by_route` are expecting a List of Stops instead
- This causes issues in various places
  - single line schedule pages display as if there are no trips scheduled 
  - combined line schedule page (i.e. Green) 500s
- Open implementation questions: 
  -  several calls to `by_route` consumers are used to load the Green line schedule page so this becomes more of a sprawling fix  -- is there a better way to handle the error than changing all consumers to expect the error tuple? 
     - current draft doesn't resolve all of them, maybe I'm close to fixing them all but maybe not 
  - we can probably pass along the error state for the single line schedule and change the text to something along the lines of "error retrieving data" instead of showing the incorrect  
  - however, is it maybe just fine to leave the weirdness in place? Errors currently do not cache on this function, so after API is resolved, all pages return to normal 
  - could resolve partially (let Green 500 but fix single line schedule message to correctly show that it's an issue with data retrieval and not that we have no trips scheduled)


lib/green_line.ex
```diff
@@ -224,7 +224,7 @@ defmodule GreenLine do
   # map of {route_id => [stop_id]} representing all the stops on the route.
   # The {:ok, _} part of the pattern match is due to using Task.async_stream.
   @spec merge_green_line_stops(
-          {:ok, {Route.id_t(), [Stop.t()] | {:error, any}}},
+          {:ok, {Route.id_t(), [Stop.t()] | {:error, any} | {:exit, any}}},
```
thecristen, line 227: I should have updated this when I updated the `Task.async_stream call` in #3393 ! Thanks for adding

thecristen (commented):
Thanks for taking on this investigation! I think you've got a good breakdown of the underlying root cause here.

In response to the open questions:

> several calls to by_route consumers are used to load the Green line schedule page so this becomes more of a sprawling fix -- is there a better way to handle the error than changing all consumers to expect the error tuple?

Options:
1. Add a layer between consumers and by_route that'll see the error and turn it into `[]` in one place instead of having to make that edit everywhere
2. Edit the underlying code so it doesn't return that error. There's actually another layer between `Stops.Repo.by_route/3` and the V3 API already in `Stops.Api.by_route/3` (a legacy pattern which isn't followed elsewhere 😩 ), which is actually what's parsing the response and returning the error tuple (and `Stops.Repo.do_by_route/3` just forwards that along).  

We **_do_** have precedent for papering over these errors with `nil` or `[]` at the repo level (examples: `Routes.Repo.all/0`, `RoutePatterns.Repo.get/1`), so this might be a good option here. Though, this may hinder our ability to detect the error state and choose to render it!

> we can probably pass along the error state for the single line schedule and change the text to something along the lines of "error retrieving data" instead of showing the incorrect
> however, is it maybe just fine to leave the weirdness in place? Errors currently do not cache on this function, so after API is resolved, all pages return to normal
> could resolve partially (let Green 500 but fix single line schedule message to correctly show that it's an issue with data retrieval and not that we have no trips scheduled)

We don't really have a good strategy currently for handling partial missing data in the UI, especially on this old line diagram page. I'm resistant to try to have us tackle that right now, as nice as better messaging would be, especially since these states tend to be transient as you stated.

Do we have a sense as to what it might take to avoid the 500 on the combined Green Line page, even if it does produce a weird UI? I'm thinking that might be the simplest way forward for the moment. I'm guessing it would involve having `GreenLine.stops_on_routes/2` filter out the errors potentially returned by the `GreenLine.merge_green_line_stops/2` calls.

PR #3510, opened 2026-09-18
fix(Stops.Header): improve wrapping behavior

thecristen:
## Scope

**Asana Ticket:** TBA

## Implementation

This was one of the tweaks in the [updated design](https://www.figma.com/design/fXXxB7zL0OsxOItwfWtAxQ/MBTA.com?node-id=11028-22885&m=dev) that I decided to try to tackle.

## Screenshots

(dev on left, this PR on right, in order of decreasing screen width:)

image

image

image

image

image

Hopefully that gives an idea of the new vs. old wrapping behavior!

## How to test

Find some stops with many connections and resize your browser!

---
  - https://app.asana.com/0/0/1218634129185390


joshlarson (approved):
Wow this is so much better!

